PR #124, opened 2015-01-06
Testing utilities

### LINKED ISSUE #31: Testing utilities
Add utilities to ease the testing process.
The goal is to support BDD via feature and unit tests. For feature tests we mean, full stack, browser simulation (like Cucumber does).

This should be part of this PR. We should support the same feature with MiniTest & RSpec

The expected structure is:

```
spec
├── chirp
│   ├── entities
│   ├── repositories
├── support
├── features_helper.rb
├── spec_helper.rb
└── web
    ├── controllers
    ├── features
    └── views
```

- [x] Support MiniTest
- [x] Support RSpec
- [x] Support for feature tests with Capybara
- [x] Code generator must create a directory to for tests (see #29)

There is a [simple setup](https://gist.github.com/jodosha/9830002#file-spec_helper-rb) to use with RSpec to have feature tests. The code is borrowed from `rspec-rails`.

hieuk09:
Closes #31


joneslee85:
:+1:

joneslee85:
one more thing we need to provide is the setup for rack-test for capybara:

We need `spec/support/capybara.rb`:

```
module RSpec
  module FeatureExampleGroup
    def self.included(group)
      group.metadata[:type] = :feature
      Capybara.app = Lotus::Container.new
    end
  end
end
```

and in `spec/features_helper.rb`:

```
# Require this file for feature tests
require 'spec_helper'

require 'capybara'
require 'capybara/rspec'

RSpec.configure do |config|
  config.include RSpec::FeatureExampleGroup

  config.include Capybara::DSL,           feature: true
  config.include Capybara::RSpecMatchers, feature: true
end
```

and empty directory `spec/<app_name>/features`

huydx:
:+1:

hieuk09:
@joneslee85: I updated the PR.

joneslee85:
:+1:

jodosha:
@hieuk09 Thanks for this PR it's really good.

There is only one thing to amend: `lib/lotus/generators/application/container/spec_helper.rb.rspec.tt`.
I haven't added RSpec support at the beginning because historically, `spec_helper.rb` is a moving target. 

Because RSpec now has an `rspec --init` command, I would love to delegate the generation of this file to it. So we don't need to maintain all that complex settings by ourselves.

Thor hints:
* `#run` http://www.rubydoc.info/github/wycats/thor/master/Thor/Actions#run-instance_method
* `#gsub_file` http://www.rubydoc.info/github/wycats/thor/master/Thor/Actions:gsub_file

The last one is for inject custom lines of code. Eg. `config.before(:suite) { Lotus::Application.preload! }`

hieuk09:
@jodosha: In order to call `rspec --init`, you need the rspec gem installed. If the user hasn't install rspec in his global gemset, the command would fail. I suggest we keep the `spec_helper.rb.rspec.tt` file.

joneslee85:
@hieuk09 can't you execute `gem install rspec`?

hieuk09:
I put `rspec` to lotus Gemfile to enable `rspec --init` command, but I'm not sure if it's a good way to do that.
On another hand, the test suit now has some annoying output (generated when running `rspec --init` in `test/command/new_test.rb`) but I haven't thought of any ways to remove it yet.

jodosha:
@hieuk09 You're right, I haven't considered this circular dependency.

As far I understand, your last commit (https://github.com/hieuk09/lotus/commit/045176dfc0ddd438c9c03e947974f336566d9606) requires a pre-installed RSpec gem.

This would fail on developers machines if they haven't that gem yet.

Sorry, but I'm for rolling back to the previous solution and maintain `spec_helper.rb` by ourselves until we find a better solution. Stability comes first.

/cc @joneslee85

jodosha:
If we all agree to revert that commit, I consider this PR ready to be merged. :+1:

hieuk09:
@jodosha: We may add rspec as the dependency of lotus so that when users install lotus, they will install rspec as well. However, it seems to be overkill for a small functionality so I reverted the commit anyway. Thanks for the review :+1:

jodosha:
No, let's stick with the current approach. Users come first.

Luca

joneslee85:
:+1:

PR #128, opened 2015-01-08
Allows to generate the app in the current folder

mengqing:
Cover following cases:

* [x] `lotus new .` to generate new app with app_name = current folder name and path = current path
* [x] `lotus new app_1 --path /my_path` - to generate new app with name app_1 and /my_path

https://github.com/lotus/lotus/issues/127

### DIFF
--- a/lib/lotus/commands/new.rb
+++ b/lib/lotus/commands/new.rb
@@ -14,7 +14,7 @@ module Lotus
         @options  = environment.to_options
         @arch     = @options.fetch(:architecture)
 
-        @target   = Pathname.pwd.join(app_name)
+        @target   = Pathname.pwd.join(@options.fetch(:application_path, ::File.basename(app_name, '*')))
         @source   = Pathname.new(@options.fetch(:source) { ::File.dirname(__FILE__) + '/../generators/application/' }).join(@arch)
 
         @cli      = cli
@@ -33,13 +33,10 @@ module Lotus
         case app_name
         when '.'
           ::File.basename(Dir.getwd)
-        when '/'
-          # If path is root, then use a default app name
-          # most likely won't be used due to file permission
-          'app'
         else
-          # Allows the app name to be a path
-          ::File.basename(app_name)
+          # Returns last component of the app_name
+          # if a path is provided
+          ::File.basename(app_name, '*')
         end
       end
     end
--- a/lib/lotus/commands/new.rb
+++ b/lib/lotus/commands/new.rb
@@ -30,14 +30,10 @@ module Lotus
 
       private
       def _get_real_app_name(app_name)
-        raise ArgumentError.new("Path can not be used for the application name. Please use --path instead ") if app_name.include?('/')
+        raise ArgumentError.new("Invalid application name. If you want to set application path, please use --path option") if app_name.include?('/')
 
-        case app_name
-        when '.'
-          ::File.basename(Dir.getwd)
-        else
-          app_name
-        end
+        app_name = ::File.basename(Dir.getwd) if app_name == '.'
+        app_name
       end
     end
   end
--- a/lib/lotus/commands/new.rb
+++ b/lib/lotus/commands/new.rb
@@ -30,10 +30,11 @@ module Lotus
 
       private
       def _get_real_app_name(app_name)
-        raise ArgumentError.new("Invalid application name. If you want to set application path, please use --path option") if app_name.include?('/')
+        if app_name.include?('/')
+          raise ArgumentError.new("Invalid application name. If you want to set application path, please use --path option")
+        end
 
-        app_name = ::File.basename(Dir.getwd) if app_name == '.'
-        app_name
+        app_name == '.' ? ::File.basename(Dir.getwd) : app_name
       end
     end
   end
--- a/lib/lotus/commands/new.rb
+++ b/lib/lotus/commands/new.rb
@@ -10,11 +10,11 @@ module Lotus
       attr_reader :app_name, :source, :target, :cli, :options
 
       def initialize(app_name, environment, cli)
-        @app_name = app_name
+        @app_name = _get_real_app_name(app_name)
         @options  = environment.to_options
         @arch     = @options.fetch(:architecture)
 
-        @target   = Pathname.pwd.join(@app_name)
+        @target   = Pathname.pwd.join(app_name)
         @source   = Pathname.new(@options.fetch(:source) { ::File.dirname(__FILE__) + '/../generators/application/' }).join(@arch)
 
         @cli      = cli
@@ -26,6 +26,21 @@ module Lotus
 
       def start
         @command.start
+      end
+
+      private
+      def _get_real_app_name(app_name)
+        case app_name
+        when '.'
+          ::File.basename(Dir.getwd)
+        when '/'
+          # If path is root, then use a default app name
+          # most likely won't be used due to file permission
+          'app'
+        else
+          # Allows the app name to be a path
+          ::File.basename(app_name)
+        end
       end
     end
   end
--- a/lib/lotus/commands/new.rb
+++ b/lib/lotus/commands/new.rb
@@ -14,7 +14,7 @@ module Lotus
         @options  = environment.to_options
         @arch     = @options.fetch(:architecture)
 
-        @target   = Pathname.pwd.join(@options.fetch(:application_path, ::File.basename(app_name, '*')))
+        @target   = Pathname.pwd.join(@options.fetch(:path, app_name))
         @source   = Pathname.new(@options.fetch(:source) { ::File.dirname(__FILE__) + '/../generators/application/' }).join(@arch)
 
         @cli      = cli
@@ -30,13 +30,13 @@ module Lotus
 
       private
       def _get_real_app_name(app_name)
+        raise ArgumentError.new("Path can not be used for the application name. Please use --path instead ") if app_name.include?('/')
+
         case app_name
         when '.'
           ::File.basename(Dir.getwd)
         else
-          # Returns last component of the app_name
-          # if a path is provided
-          ::File.basename(app_name, '*')
+          app_name
         end
       end
     end
--- a/test/commands/new_test.rb
+++ b/test/commands/new_test.rb
@@ -490,5 +490,83 @@ describe Lotus::Commands::New do
         end
       end
     end
+
+    describe 'when app_name is .' do
+      let(:app_name) { '.' }
+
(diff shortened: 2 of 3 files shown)


lib/lotus/generators/application/container.rb
```diff
@@ -11,13 +11,16 @@ def initialize(command)
           @slice_generator     = Slice.new(command)
           @lotus_head          = options.fetch(:lotus_head)
           @lotus_model_version = '~> 0.2'
+          @app_name            = app_name
 
           cli.class.source_root(source)
         end
 
         def start
+          @app_name = ::File.basename(Dir.getwd) if @app_name == '.'
```
joneslee85: looks good, now I am wondering if this would work if the current directory is `/`

lib/lotus/generators/application/container.rb
```diff
@@ -11,13 +11,16 @@ def initialize(command)
           @slice_generator     = Slice.new(command)
           @lotus_head          = options.fetch(:lotus_head)
           @lotus_model_version = '~> 0.2'
+          @app_name            = app_name
```
AlfonsoUceda: where does app_name come from?

joneslee85: This change should be done inside the constructor of `Lotus::Commands::New`.

lib/lotus/generators/application/container.rb
mengqing: This won't work if the current directory is /, however I'm not sure if this needs to be supported. Or perhaps we can have a default app name such as 'app' if the current directory is root.

What's your thoughts?

lib/lotus/generators/application/container.rb
mengqing: updated

lib/lotus/commands/new.rb
```diff
@@ -17,6 +17,8 @@ def initialize(app_name, environment, cli)
         @target   = Pathname.pwd.join(@app_name)
         @source   = Pathname.new(@options.fetch(:source) { ::File.dirname(__FILE__) + '/../generators/application/' }).join(@arch)
 
+        @app_name = real_app_name
```
joneslee85: I think you should modify line 13 instead:

```ruby
@app_name = _get_app_name(app_name)
```

lib/lotus/commands/new.rb
```diff
@@ -27,6 +29,16 @@ def initialize(app_name, environment, cli)
       def start
         @command.start
       end
+
+      private
+      def real_app_name
+        if @app_name == '.'
```
joneslee85: what about the case in which `@app_name == '/'`

lib/lotus/commands/new.rb
```diff
@@ -27,6 +29,16 @@ def initialize(app_name, environment, cli)
...
+          ::File.basename(Dir.getwd)
+        else
+          # Allows the app name to be a path
+          ::File.basename(@app_name)
```
joneslee85: I like this :+1: , can you please provide test for this scenario too?

mengqing: tests added

jodosha:
@mengqing Hello and thanks for this PR.

What's this useful for? What's the CLI usage?

joneslee85:
@jodosha the use case is to allow users to parse in a pathname for app_name:

```
lotus new /tmp/my_app
# would create my_app inside /tmp
```

also you could do

```
mkdir /tmp/my_app
cd /tmp/my_app
lotus new .
# it would create my_app in the current folder
```

I myself do not use this feature at all nor am I aware of it. But it does exist in Rails and @mengqing raised it up here, it must be used frequently by some users

mengqing:
Hi @jodosha, @joneslee85 pretty much summerized what it's for and the cli usage of it. I personally always use this syntax for a new rails app, and it works quite well for rvm && rbenv e.g.

```shell
mkdir my_app && cd !$
rvm use --create 2.2.0@my_lotus_app --ruby-version
gem i lotus
lotus new .
```

AlfonsoUceda:
:+1: for support this feature ;)

jodosha:
@mengqing I think this makes sense for Lotus philosophy.

We consider Lotus apps as a deliverability mechanism for "headless" applications. The Container architecture is an example of what I'm saying: `lib/` contains the pure Ruby code (headless) and `apps/` contains web deliverability.

Including this feature may help to easily get onboard gems who need a quick way to expose web facilities. Think of Sidekiq (headless) and the web dashboard that uses Sinatra (`Sidekiq::Web`).

:+1: for me.

lib/lotus/commands/new.rb
```diff
@@ -27,6 +27,21 @@ def initialize(app_name, environment, cli)
       def start
         @command.start
       end
+
+      private
+      def _get_real_app_name(app_name)
+        case app_name
+        when '.'
+          ::File.basename(Dir.getwd)
+        when '/'
```
jodosha: @mengqing This is confusing for me: what we're trying to address here? Does makes sense to have an usage like `lotus new /`?

mengqing: @jodosha, @joneslee85 mentioned about this use case of having '/' as the path, however personally I don't think anyone would be generating the app under root nor is it a good practice (and most likely permission denied). What do you guys think? I can remove it if it doesn't make sense.

jodosha: @mengqing Yes, it doesn't make sense. Please remove it.
We should ensure that `.` is handled as a special case and no other path is allowed.

mengqing: @jodosha should we allow path such as the one @joneslee85 has mentioned eg
```shell
lotus new lib/my_new_lotus_app
# would create my_new_lotus_app inside lib
```
then we can raise exception when app_name is '/'

jodosha: @mengqing @joneslee85 To cover that case, I would prefer to have a `--path` option. This makes both the intent of the CLI and the internals more clear and more protected from bugs.

mengqing: @jodosha makes sense, good idea!

joneslee85: @mengqing I like @jodosha suggestion on having `--path`. :+1: 

And yes, we could remove the `.` after having ^ implemented because it would error out due to permission denied anyway. Much thanks for taking care of this.

lib/lotus/commands/new.rb
```diff
@@ -27,6 +27,18 @@ def initialize(app_name, environment, cli)
       def start
         @command.start
       end
+
+      private
+      def _get_real_app_name(app_name)
+        raise ArgumentError.new("Path can not be used for the application name. Please use --path instead ") if app_name.include?('/')
```
joneslee85: How about: `Invalid application name. If you want to set application path, please use --path option`

lib/lotus/commands/new.rb
```diff
@@ -27,6 +27,18 @@ def initialize(app_name, environment, cli)
...
+
+        case app_name
+        when '.'
```
joneslee85: I'd prefer `if` than `case` here

test/commands/new_test.rb
```diff
@@ -490,5 +490,76 @@ def container_options
...
+      end
+    end
+  end
+
+  describe 'when a path is provided as the app name' do
+    let(:opts)      { Hash[architecture: 'container', application: 'web', application_base_url: '/', lotus_head: false] }
+    let(:app_name)  { 'lib/chirp' }
+
+    it 'raises an ArgumentError' do
+      -> { command.start }.must_raise ArgumentError
```
joneslee85: also worth to check the exception message:

```
exception = -> { command.start }.must_raise ArgumentError
(exception.message).must_equal 'bah blah blah'
```

test/commands/new_test.rb
```diff
@@ -490,5 +490,76 @@ def container_options
...
+
+    before do
+      capture_io { command.start }
+    end
+
+    describe 'when a path is provided' do
+      let(:opts) { container_options.merge(path: 'my_lotus_app') }
+
+      describe 'my_lotus_app' do
+        it 'generates it' do
```
joneslee85: I think it's better to word it as `generates app at specified path`

lib/lotus/commands/new.rb
```diff
@@ -27,6 +27,14 @@ def initialize(app_name, environment, cli)
       def start
         @command.start
       end
+
+      private
+      def _get_real_app_name(app_name)
+        raise ArgumentError.new("Invalid application name. If you want to set application path, please use --path option") if app_name.include?('/')
+
+        app_name = ::File.basename(Dir.getwd) if app_name == '.'
```
joneslee85: I'd prefer this `app_name == '.' ? ::File.basename(Dir.getwd) : app_name`

test/commands/new_test.rb
```diff
@@ -490,5 +490,75 @@ def container_options
...
+    end
+  end
+
+  describe 'when a path is provided as the app name' do
+    let(:opts)      { Hash[architecture: 'container', application: 'web', application_base_url: '/', lotus_head: false] }
+    let(:app_name)  { 'lib/chirp' }
+
+    it 'raises an ArgumentError' do
+      exception = -> { command.start }.must_raise ArgumentError
+      (exception.message).must_equal 'Invalid application name. If you want to set application path, please use --path option'
```
joneslee85: please remove `()`

joneslee85:
except above comments, it is :+1:  from me

lib/lotus/commands/new.rb
```diff
@@ -27,6 +27,14 @@ def initialize(app_name, environment, cli)
       def start
         @command.start
       end
+
+      private
+      def _get_real_app_name(app_name)
+        raise ArgumentError.new("Invalid application name. If you want to set application path, please use --path option") if app_name.include?('/')
```
AlfonsoUceda: I think is better put this because it's a long line

```ruby
if app_name.include?('/')
  raise ArgumentError.new("Invalid application name. If you want to set application path, please use --path option")
end
```

joneslee85: valid point, it'd be easier to read

AlfonsoUceda:
:+1:

mengqing:
:+1:

joneslee85:
:+1:

joneslee85:
@mengqing much thanks for your contribution.

jodosha:
@mengqing This works great, thank you!

mengqing:
@jodosha no problem, glad I helped to improve the app :+1:

PR #131, opened 2015-01-10
Introduce Lotus::Logger

huydx:
Hi @jodosha , @joneslee85 , i've made a minimum prototype of Logger which has below feature:
- Full feature of Ruby std ::Logger
- Always default logdevice to STDOUT
- Has `app_tag` method to inject the identification of current application to the log.

Please review and give me any comments ;)

### DIFF
--- a/lib/lotus/logger.rb
+++ b/lib/lotus/logger.rb
@@ -1,0 +1,94 @@
+require 'logger'
+require 'lotus/utils/string'
+
+module Lotus
+  # Lotus default logger
+  #
+  # Implement with the same interface of ruby std ::Logger
+  # Opts out the logdev parameter and always use STDOUT as outout device
+  #
+  # Lotus logger also has the app tag concept, which used to identify
+  # which application the log come from.
+  #
+  # Lotus Logger default comes with Lotus application
+  # and uses name of highest module namespace as app_tag
+  #
+  # When stands alone, Lotus Logger tries to infer app tag from highest namespace
+  # When has no namespace, Lotus Logger takes [Shared] as default app tag
+  #
+  # @example
+  #   #1 Logger with namespace
+  #   module TestApp
+  #     class AppLogger << Lotus::Logger; end
+  #     def log
+  #       Applogger.new.info('foo')
+  #       #=> output: I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [TestApp] : foo
+  #     end
+  #   end
+  #
+  #   #2 Logger without namespace
+  #   class AppLogger < Lotus::Logger
+  #   end
+  #   Applogger.new.info('foo')
+  #   #=> output: I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [Shared] : foo
+  #
+  #   #3 Logger inside a lotus application
+  #   module LotusModule
+  #     class App < Lotus::Application
+  #       load!
+  #     end
+  #   end
+  #   LotusModule::Logger.info('foo')
+  #   => output: I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [LotusModule] : foo
+  #
+  # @see ::Logger
+  #
+  # @since 0.2.1
+  class Logger < ::Logger
+
+    attr_accessor :app_tag
+
+    # Override Ruby's Logger#initialize
+    #
+    # @param logdev is default to STDOUT
+    #
+    # @since 0.2.1
+    def initialize(app_tag=nil, *args)
+      super(STDOUT, *args)
+      @app_tag = app_tag
+      @formatter = Lotus::Logger::Formatter.new.tap { |f| f.app_tag = self.app_tag }
+    end
+
+    # app_tag is the identification of current app
+    # app_tag is default to use highest namespace if current namespace
+    # if app_tag is blank, lotus use default app_tag 'shared'
+    # @param logdev is default to STDOUT
+    #
+    # @since 0.2.1
+    def app_tag
+      @app_tag || _app_tag_from_namespace || _default_app_tag
+    end
+
+    class Formatter < ::Logger::Formatter
+      attr_accessor :app_tag
+
+      def call(severity, time, progname, msg)
+        time = time.utc
+        progname = "[#{@app_tag}] #{progname}"
+        super(severity, time, progname, msg)
+      end
+    end
+
+    private
+    def _app_tag_from_namespace
+      class_name = self.class.name
+      return nil unless class_name.index('::')
+
+      Utils::String.new(class_name).namespace
+    end
+
+    def _default_app_tag
+      'Shared'
+    end
+  end
+end
--- a/test/logger_test.rb
+++ b/test/logger_test.rb
@@ -1,0 +1,73 @@
+require 'test_helper'
+
+describe Lotus::Logger do
+
+  before do
+    #clear defined class
+    Object.send(:remove_const, :TestLogger) if Object.constants.include?(:TestLogger)
+  end
+
+  it 'like std logger, sets log level to info by default' do
+    class TestLogger < Lotus::Logger; end
+    TestLogger.new.info?.must_equal true
+  end
+
+  it 'always use STDOUT' do
+    output =
+      stub_stdout_constant do
+        class TestLogger < Lotus::Logger; end
+        logger = TestLogger.new
+        logger.info('foo')
+      end
+    output.must_match /foo/
+  end
+
+  it 'has app_tag when log' do
+    output =
+      stub_stdout_constant do
+        module App; class TestLogger < Lotus::Logger; end; end
+        logger = App::TestLogger.new
+        logger.info('foo')
+      end
+    output.must_match /App/
+  end
+
+  it 'has default app tag when not in any namespace' do
+    class TestLogger < Lotus::Logger; end
+    TestLogger.new.send(:app_tag).must_equal 'Shared'
+  end
+
+  it 'infers apptag from namespace' do
+    module App2
+      class TestLogger < Lotus::Logger;end
+      class Bar
+        def hoge
+          TestLogger.new.send(:app_tag).must_equal 'App2'
+        end
+      end
+    end
+    App2::Bar.new.hoge
+  end
+
+  it 'uses custom app_tag from override class' do
+    class TestLogger < Lotus::Logger; def app_tag; 'bar'; end; end
+    output =
+      stub_stdout_constant do
+        TestLogger.new.info('')
+      end
+    output.must_match /bar/
+  end
+
+  it 'has format "#{Severity}, [%Y-%m-%dT%H:%M:%S.%6N #{Pid}] #{Severity} -- [#{app_tag}] : #{message}\n"' do
+    format = "%Y-%m-%dT%H:%M:%S.%6N "
+    stub_time_now do
+      strtime = Time.now.strftime(format)
+      output =
+        stub_stdout_constant do
+          class TestLogger < Lotus::Logger;end
+          TestLogger.new.info('foo')
+        end
+      output.must_equal "I, [1988-09-01T00:00:00.000000 ##{Process.pid}]  INFO -- [Shared] : foo\n"
+    end
+  end
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -105,5 +105,34 @@ class DependenciesReporter
 end
 
 DependenciesReporter.new.run
-require 'fixtures'
 
+def stub_stdout_constant
+  begin_block = <<-BLOCK
+    original_verbosity = $VERBOSE
+    $VERBOSE = nil
+
+    origin_stdout = STDOUT
+    STDOUT = StringIO.new
+  BLOCK
+  TOPLEVEL_BINDING.eval begin_block
+
+  yield
+  return_str = STDOUT.string
+
+  ensure_block = <<-BLOCK
+    STDOUT = origin_stdout
+    $VERBOSE = original_verbosity
+  BLOCK
+  TOPLEVEL_BINDING.eval ensure_block
+
+  return_str
+end
+
+
+def stub_time_now
+  Time.stub :now, Time.utc(1988, 9, 1, 0, 0, 0) do
+    yield
+  end
+end
+
+require 'fixtures'
--- a/lib/lotus.rb
+++ b/lib/lotus.rb
@@ -1,6 +1,7 @@
 require 'lotus/version'
 require 'lotus/application'
 require 'lotus/container'
+require 'lotus/logger'
 
 # A complete web framework for Ruby
 #
--- a/lib/lotus/loader.rb
+++ b/lib/lotus/loader.rb
@@ -2,6 +2,7 @@ require 'lotus/utils/class'
 require 'lotus/utils/kernel'
 require 'lotus/utils/string'
 require 'lotus/routes'
+require 'lotus/logger'
 require 'lotus/routing/default'
 require 'lotus/action/cookies'
 require 'lotus/action/session'
@@ -40,6 +41,7 @@ module Lotus
       _configure_model_framework! if defined?(Lotus::Model)
       _configure_controller_framework!
       _configure_view_framework!
+      _configure_static_variables!
     end
 
     def _configure_controller_framework!
@@ -87,6 +89,16 @@ module Lotus
       end
     end
 
+    def _configure_static_variables!
+      _initialize_logger!
+    end
+
+    def _initialize_logger!
+      unless application_module.const_defined?('Logger', false)
+        logger = Lotus::Logger.new(application_module.to_s)
+        application_module.const_set('Logger', logger)
+      end
+    end
 
     def load_frameworks!
       _load_view_framework!
--- a/test/loader_test.rb
+++ b/test/loader_test.rb
@@ -15,6 +15,7 @@ describe Lotus::Loader do
         assert defined?(CoffeeShop::Entity),     'expected CoffeeShop::Entity'
         assert defined?(CoffeeShop::Repository), 'expected CoffeeShop::Repository'
         assert defined?(CoffeeShop::Presenter),  'expected CoffeeShop::Presenter'
+        assert defined?(CoffeeShop::Logger),     'expected CoffeeShop::Logger'
       end
 
       it 'generates per application classes' do
@@ -72,6 +73,19 @@ describe Lotus::Loader do
       describe 'middleware' do
         it 'preloads the middleware' do
           @application.middleware.must_be_kind_of(Lotus::Middleware)
+        end
+      end
+
+      describe 'logger' do
+        it 'has app module name along with log output' do
+          output =
+            stub_stdout_constant do
+              module CocacolaShop
+                class Application < Lotus::Application; load!; end
+              end
+              CocacolaShop::Logger.info 'foo'
+            end
+          output.must_match /CocacolaShop/
         end
       end
     end


lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
+require 'logger'
+
+module Lotus
+  # Lotus default logger
+  #
+  # Implement with the same interface of ruby std ::Logger
+  # Optout the logdev parameter and always use STDOUT to output the log
```
joneslee85: Opt out not Optout

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+  #
+  # Lotus logger also has the app_tag concept, which used to identify
+  # which application the log come from
+  #
+  # Example Usage:
```
joneslee85: Just `Example:`

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+    #
+    # inject app_tag to logged messages
+    #
+    # @since 0.2.1
+    class LotusFormatter < ::Logger::Formatter
+      attr_accessor :app_tag
+
+      def call(severity, time, progname, msg)
+        raise Exception.new('You must set the app_tag') unless (app_tag && !app_tag.empty?)
+        progname = "[#{@app_tag}] #{progname}"
```
joneslee85: how about renaming `progname` with `app_name`?

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+  # end
+  #
+  # #2 Logger without namespace
+  # class AppLogger < Lotus::Logger
+  # end
+  #
+  # Applogger.new.info('foo')
+  # #=> output: I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [Shared] : foo
+  #
+  # @since 0.1.0
```
joneslee85: `since` is for version of lotus that introduces this change, please leave it as `x.x.x`

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  class Logger < ::Logger
+
+    # Override Ruby's Logger#initialize
+    #
+    # @param logdev is default to STDOUT
+    #
+    # @since 0.2.1
+    def initialize(logdev=nil, *args)
+      super(STDOUT,*args)
```
joneslee85: add a space between `,`

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+  # Lotus default logger
+  #
+  # Implement with the same interface of ruby std ::Logger
+  # Optout the logdev parameter and always use STDOUT to output the log
+  #
+  # Lotus logger also has the app_tag concept, which used to identify
+  # which application the log come from
+  #
+  # Example Usage:
+  # #1 Logger with namespace
```
joneslee85: the convention for multiple examples are:

`Example: Logger with namespace`

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+  # @since 0.1.0
+
+  class Logger < ::Logger
+
+    # Override Ruby's Logger#initialize
+    #
+    # @param logdev is default to STDOUT
+    #
+    # @since 0.2.1
+    def initialize(logdev=nil, *args)
```
joneslee85: why not set `STDOUT` as default value for `logdev`?

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+    def app_tag
+      _app_tag
+    end
+
+    # Lotus Logger default Formatter
+    #
+    # inject app_tag to logged messages
+    #
+    # @since 0.2.1
+    class LotusFormatter < ::Logger::Formatter
```
joneslee85: I'd prefer `Lotus::Logger::Formatter`

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+      def call(severity, time, progname, msg)
+        raise Exception.new('You must set the app_tag') unless (app_tag && !app_tag.empty?)
+        progname = "[#{@app_tag}] #{progname}"
+        super(severity, time, progname, msg)
+      end
+    end
+
+    private
+    def _app_tag
+      namespaces = self.class.name.split('::')
```
joneslee85: this could be a private method

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+      if namespaces.length >= 2
```
joneslee85: why is it 2?

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'test_helper'
+
+describe Lotus::Logger do
+  class IoCapture
+    attr_accessor :stream
+    def write(smt)
+      @stream = smt
+    end
+    def close;end
```
joneslee85: _nitpic_, add space after `,`

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    def flush;end
```
joneslee85: _nitpic_, add space after `,`

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'test_helper'
+
+describe Lotus::Logger do
+  class IoCapture
```
joneslee85: _nitpick_, I think naming `IO` is better`

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    def close;end
+    def flush;end
+  end
+
+  #alias STDOUT to custom class to test output
+  STDOUT = IoCapture.new
+
+  before do
+    #clear defined class
+    Object.constants.include?(:TestLogger)
```
joneslee85: isn't is supposed to be something undef things going in here?

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  it 'always use STDOUT' do
+    class TestLogger2 < Lotus::Logger; end
+    logger = TestLogger2.new
+    logger.info('foo')
+    assert_match(/foo/, STDOUT.stream)
+  end
+
+  describe '.app_tag' do
```
joneslee85: the convention for `.` is to describe class method, in this case, maybe just `app_tag` to denote the fact that we are describing a feature of a function rather than instance/class method

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    it 'has app_tag when log' do
+      module App; class TestLogger3 < Lotus::Logger; end; end
+      logger = App::TestLogger3.new
+      logger.info('foo')
+      assert_match(/App/, STDOUT.stream)
+    end
+
+    it 'raise exception if override logger without app_tag' do
+      class TestLogger4 < Lotus::Logger; def app_tag; end; end
+      assert_raises(Exception) { logger = TestLogger4.new.info('anything') }
```
joneslee85: I think it is a good idea to check what exception has been raised:

```
exception = -> { TestLogger4.new.info('anything') }.must_raise(Exception)
exception.message.must_equal 'something?'
```

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    it 'has default app tag when not in any namespace' do
+      class TestLogger5 < Lotus::Logger; end
+      TestLogger5.new.send(:app_tag).must_equal 'Shared'
```
joneslee85: Is there reason why we could not invoke `app_tag` but having to go through `send`?

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    describe 'has app tag which equal highest namespace' do
+      it 'one nested namespace' do
+        module App
+          class TestLogger6 < Lotus::Logger;end
```
joneslee85: _nitpic_, add space after `,`

test/logger_test.rb
```diff
@@ -0,0 +1,85 @@
+require 'test_helper'
+
+describe Lotus::Logger do
+  class IoCapture
+    attr_accessor :stream
+    def write(smt)
```
simi: What does `smt` stands for?

test/logger_test.rb
simi: *nitpick* after `;`

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  class Logger < ::Logger
+
+    # Override Ruby's Logger#initialize
+    #
+    # @param logdev is default to STDOUT
+    #
+    # @since 0.2.1
+    def initialize(logdev=nil, *args)
+      super($stdout, *args)
```
jodosha: :+1: For @joneslee85 suggestion. Also, we should use `STDOUT` instead of `$stdout`.

lib/lotus/logger.rb
jodosha: :+1:

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+      @formatter = LotusFormatter.new.tap {|f| f.app_tag = self.app_tag}
+    end
+
+    # app_tag is the identification of current app
+    # app_tag is default to use highest namespace if current namespace
+    # if app_tag is blank, lotus use default app_tag 'shared'
+    # @param logdev is default to STDOUT
+    #
+    # @since 0.2.1
+    def app_tag
```
jodosha: If not accessed on the outside, I'd declare it as private method.

lib/lotus/logger.rb
```diff
@@ -0,0 +1,88 @@
...
+    # Lotus Logger default Formatter
+    #
+    # inject app_tag to logged messages
+    #
+    # @since 0.2.1
+    class LotusFormatter < ::Logger::Formatter
+      attr_accessor :app_tag
+
+      def call(severity, time, progname, msg)
+        raise Exception.new('You must set the app_tag') unless (app_tag && !app_tag.empty?)
```
jodosha: `Exception` should never be used.

The reason is that anonymous `rescue` clause won't catch this kind of exceptions.

```ruby
[1] pry(main)> begin
[1] pry(main)*   raise Exception.new('boom')
[1] pry(main)* rescue
[1] pry(main)*   puts 'rescued'
[1] pry(main)* end
Exception: boom
from (pry):2:in `__pry__'
[2] pry(main)>
```

If we don't care to have a custom exception for this case (and I agree), we should use `raise "You must set the app_tag"`. This syntax raises a `RuntimeError`, which is catchable by anonymous `rescue`.

```ruby
[1] pry(main)> begin
[1] pry(main)*   raise "missing app_name"
[1] pry(main)* rescue => e
[1] pry(main)*   puts e.class.name
[1] pry(main)*   puts e.message
[1] pry(main)* end
RuntimeError
missing app_name
=> nil
[2] pry(main)>
```

This pattern should always be used by libraries and frameworks, so developers should never worry to use specific class names with `rescue`.

More at:

* http://robots.thoughtbot.com/rescue-standarderror-not-exception
* http://daniel.fone.net.nz/blog/2013/05/28/why-you-should-never-rescue-exception-in-ruby/
* http://www.mikeperham.com/2012/03/03/the-perils-of-rescue-exception/

lib/lotus/logger.rb
jodosha: Please use `Utils::String#namespace`.

test/logger_test.rb
```diff
@@ -0,0 +1,75 @@
...
+describe Lotus::Logger do
+
+  before do
+    #clear defined class
+    Object.send(:remove_const, :TestLogger) if Object.constants.include?(:TestLogger)
+  end
+
+  it 'like std logger, default log level is info' do
+    class TestLogger < Lotus::Logger; end
+    assert_respond_to(TestLogger.new, :info?)
```
jodosha: This doesn't test if `info?` returns true, it test if it responds to the method.
What about `assert TestLogger.new.info?, "failure message"`.

jodosha:
@huydx Thanks for this PR.

I think tests should assert at least one complete message. I know that this is complex because a timestamp is involved, but you can stub `Time.now` with Minitest in a `before` block.

Timestamps should be UTC.

What is still missing here is the automatic creation of this class when an application is preloaded. In other terms when one of those methods are invoked, Lotus should create `Bookshelf::Logger` class as a `Lotus::Logger` subclass:

  * `Bookshelf::Application.load!`
  * `Bookshelf::Application.load!`
  * `Lotus::Application.preload!`

@joneslee85 Can you please help @huydx with low level details of `Loader`? Thank you!

jodosha:
@huydx @joneslee85 What do you think if we expose class methods for logging.

I find really objects compliant with Lotus philosophy, but it may be a bit noisy to use `Bookshelf::Logger.new.debug "blah"`. What do you think if we have `Bookshelf::Logger.debug "blah"` too?

lib/lotus/logger.rb
huydx: @joneslee85 IMO progname has different meaning, and if you rename progname with app_name, and use both parameter and block like below, what will happen?

```ruby
logger.info('foo') { 'bar' }
#=> output: I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [App] foo : bar
```

lib/lotus/logger.rb
huydx: @joneslee85 IMO the reason i set nil because i think the code like
```ruby
def initialize(logdev=$stdout, *args)
  super($stdout, *args)
```
is not express the idea that we default set the `logdev` param to `$stdout` better as using nil. So what do you think @jodosha @joneslee85

lib/lotus/logger.rb
huydx: :+1: nice catch

lib/lotus/logger.rb
huydx: @jodosha nice suggestion :+1:

lib/lotus/logger.rb
huydx: @jodosha actually i intended to expose the method, just incase use want to check the current app_tag. so what do you thing @jodosha @joneslee85

lib/lotus/logger.rb
huydx: @jodosha :+1: really nice advice, i will follow it

test/logger_test.rb
huydx: @simi it stands for something, but this class is no longer needed so nvm ;)

huydx:
@jodosha class methods for logging seems fine, if we default that `one app, one logger`. 
One more thing, what do you think about replacement of Rack::CommonLogger?

jodosha:
@huydx Yes at the boot time, it should have a class instance initialized eg: `@@logger`.

RE Rack: CommonLogger accepts a logger as second argument. We should pass an instance of `Lotus::Logger.new` so we can have the format that we want.

We cannot use `Bookshelf::Logger` there, because when we start the server, we haven't loaded the applications yet. The only logger available at that time is `Lotus::Logger`.

The place where to inject this functionality is `Lotus::Commands::Server#middleware`. Thank you!

huydx:
@jodosha @joneslee85 i've updated the code with your comments.
There are some points needed to be discussed:
 - I've locate the default log to Lotus::Application::Logger instead of Lotus::Logger. Default to Lotus::Application::Logger is more consistent in finding namespace.
 - I initialize the logger with `inherited` hook in Lotus Application, which is the best place to infer real application name from `inherited(base)`

test/application_test.rb
```diff
@@ -71,4 +71,19 @@ def self.loaded?
...
+      assert_respond_to(CoffeeShop::Application::Logger, :info?)
+    end
+
+    it 'log application namespace' do
+      output =
+        stub_stdout_constant do
+          class Cocacola < Lotus::Application; end
+          Cocacola::Logger.info('')
+        end
+      assert_match(/Cocacola/, output)
```
joneslee85: how about using `must_match` matcher

huydx: @joneslee85 if there is no special reason, i would like to keep using assert style.

joneslee85: @huydx it is just for the sake of spec style consistency

lib/lotus/logger.rb
joneslee85: @huydx I rarely would call this method thus I think it is okay to leave it as private

lib/lotus/logger.rb
```diff
@@ -0,0 +1,83 @@
...
+  #
+  # #2 Logger without namespace
+  # class AppLogger < Lotus::Logger
+  # end
+  #
+  # Applogger.new.info('foo')
+  # #=> output: I, [2015-01-10T21:55:12.727259 #80487]  INFO -- [Shared] : foo
+  #
+  # @since 0.2.1
+
```
joneslee85: _nitpick_ remove this empty line

lib/lotus/logger.rb
```diff
@@ -0,0 +1,83 @@
...
+    # Override Ruby's Logger#initialize
+    #
+    # @param logdev is default to STDOUT
+    #
+    # @since 0.2.1
+    def initialize(app_tag=nil, *args)
+      super(STDOUT, *args)
+      @app_tag = app_tag
+      @formatter = LotusFormatter.new.tap { |f| f.app_tag = self.app_tag }
```
joneslee85: What do you think about `Lotus::Logger::Formatter` ?

test/logger_test.rb
```diff
@@ -0,0 +1,78 @@
...
+describe Lotus::Logger do
+
+  before do
+    #clear defined class
+    Object.send(:remove_const, :TestLogger) if Object.constants.include?(:TestLogger)
+  end
+
+  it 'like std logger, default log level is info' do
+    class TestLogger < Lotus::Logger; end
+    assert_equal(TestLogger.new.info?, true)
```
joneslee85: except above comments, :+1: from me

lib/lotus/logger.rb
huydx: I'm thinking about cases which people want to change app_tag after initialize Logger

test/logger_test.rb
```diff
@@ -0,0 +1,78 @@
...
+        module App; class TestLogger < Lotus::Logger; end; end
+        logger = App::TestLogger.new
+        logger.info('foo')
+      end
+    output.must_match /App/
+  end
+
+  it 'raise exception if override logger without app_tag' do
+    class TestLogger < Lotus::Logger; def app_tag; end; end
+    assert_raises(RuntimeError) { logger = TestLogger.new.info('anything') }
```
joneslee85: How about asserting the exception message:

```
exception = -> { TestLogger.new.info('anything') }.must_raise(RuntimeError)
exception.message.must_equal 'bakh asdlfkjasldkjas'
```

test/application_test.rb
```diff
@@ -71,4 +71,19 @@ def self.loaded?
       @application.name.must_equal @application.class.name
     end
   end
+
+  describe 'Logger' do
+    it 'has logger constant and respond to info method' do
+      assert_respond_to(CoffeeShop::Application::Logger, :info?)
```
joneslee85: can we use `must_respond_to` matcher instead?

lib/lotus/logger.rb
joneslee85: okay, lgtm, leave it  as is

joneslee85: :+1: great work

jdickey:
If I can jump in at this late date…

@jodosha, wrt "Timestamps should be UTC", incredibly enough there are some shops in some benighted corners of the world that prefer all their junk's timestamps to be in local time (I'm in one such). While UTC is the brain-dead-obvious default, It Would Be Very Nice™ if there was an injectable/settable override for that.

(And I probably should pay more attention in the Gitter room...)

huydx:
@jdickey I agree. Btw, current implementation use std logger as parent class, so you can override the behaviour by using different Log Formatter.

test/logger_test.rb
```diff
@@ -0,0 +1,79 @@
...
+    exception = -> { TestLogger.new.info('anything') }.must_raise(RuntimeError)
+    exception.message.must_equal "you must set app_tag at Logger constructor like Lotus::Logger.new('foo')"
+  end
+
+  it 'has default app tag when not in any namespace' do
+    class TestLogger < Lotus::Logger; end
+    TestLogger.new.send(:app_tag).must_equal 'Shared'
+  end
+
+  it 'infer apptag from namespace' do
```
joneslee85: /s/infer/infers

test/logger_test.rb
```diff
@@ -0,0 +1,79 @@
...
+      class Bar
+        def hoge
+          TestLogger.new.send(:app_tag).must_equal 'App2'
+        end
+      end
+    end
+    App2::Bar.new.hoge
+  end
+
+  it 'use custom app_tag from override class' do
```
joneslee85: /s/use/uses

test/logger_test.rb
```diff
@@ -0,0 +1,79 @@
...
+    output =
+      stub_stdout_constant do
+        module App; class TestLogger < Lotus::Logger; end; end
+        logger = App::TestLogger.new
+        logger.info('foo')
+      end
+    output.must_match /App/
+  end
+
+  it 'raise exception if override logger without app_tag' do
```
joneslee85: /s/raise/raises

test/logger_test.rb
```diff
@@ -0,0 +1,79 @@
+require 'test_helper'
+
+describe Lotus::Logger do
+
+  before do
+    #clear defined class
+    Object.send(:remove_const, :TestLogger) if Object.constants.include?(:TestLogger)
+  end
+
+  it 'like std logger, default log level is info' do
```
joneslee85: `sets log level to info by default`

test/application_test.rb
```diff
@@ -71,4 +71,19 @@ def self.loaded?
       @application.name.must_equal @application.class.name
     end
   end
+
+  describe 'Logger' do
+    it 'has logger constant and respond to info method' do
+      CoffeeShop::Application::Logger.must_respond_to :info?
+    end
+
+    it 'log application namespace' do
```
joneslee85: /s/log/logs

test/test_helper.rb
```diff
@@ -105,5 +105,34 @@ def dependencies
...
+  yield
+  return_str = STDOUT.string
+
+  ensure_block = <<-BLOCK
+    STDOUT = origin_stdout
+    $VERBOSE = original_verbosity
+  BLOCK
+  TOPLEVEL_BINDING.eval ensure_block
+
+  return return_str
```
joneslee85: /s/return//

lib/lotus/application.rb
```diff
@@ -37,6 +37,8 @@ def self.inherited(base)
       synchronize do
         applications.add(base)
       end
+
+      define_logger(base)
```
jodosha: @huydx We use to create those facilities in loader: https://github.com/lotus/lotus/blob/master/lib/lotus/loader.rb#L155

For instance, we define a `Bookshelf::Routes` constant which is an object that can be accessed everywhere from the application like we need to to with the logger.

If there isn't any specific need to keep this definition here, I would love to see it together with the other similar objects in `Loader`.

Technical note: Loader creates that classes when we instantiate the application (eg. `Bookshelf::Application.new`). So a logger is accessible when the application starts.

lib/lotus/logger.rb
```diff
@@ -0,0 +1,82 @@
...
+    # @since 0.2.1
+    def app_tag
+      @app_tag || _app_tag_from_namespace || _default_app_tag
+    end
+
+    class Formatter < ::Logger::Formatter
+      attr_accessor :app_tag
+
+      def call(severity, time, progname, msg)
+        raise "you must set app_tag at Logger constructor like Lotus::Logger.new('foo')" unless @app_tag
```
jodosha: @huydx Shouldn't this be moved to `Lotus::Logger#initialize`?

If this is a mandatory information, we want the application fail to start at all. With the current design, it would start but then fail as soon a logging instruction is met.

This suggested change would also eliminate the runtime overhead to check for this information every time we want to log.

lib/lotus/logger.rb
```diff
@@ -0,0 +1,82 @@
...
+        progname = "[#{@app_tag}] #{progname}"
+        super(severity, time, progname, msg)
```
jodosha: Time should be UTC.

test/logger_test.rb
```diff
@@ -0,0 +1,79 @@
...
+  it 'has default time format "%Y-%m-%dT%H:%M:%S.%6N "' do
+    format = "%Y-%m-%dT%H:%M:%S.%6N "
+    stub_time_now do
+      strtime = Time.now.strftime(format)
+      output =
+        stub_stdout_constant do
+          class TestLogger < Lotus::Logger;end
+          TestLogger.new.info('')
+        end
+      output.must_match strtime
```
jodosha: Because we have already stubbed the time in this spec, can you please do `output.must_equal expected`? It's useful in tests to see at least one complete logging output example.

lib/lotus/logger.rb
```diff
@@ -0,0 +1,82 @@
+require 'logger'
+
```
jodosha: `require 'lotus/utils/string'` ;)

lib/lotus/logger.rb
```diff
@@ -0,0 +1,93 @@
...
+  #
+  # Lotus logger also has the app tag concept, which used to identify
+  # which application the log come from.
+  # Lotus Logger default come with Lotus::Application
+  # and take name of class which inherit Lotus::Application as app_tag
+  #
+  # When stand alone, Lotus Logger try to infer app tag from highest namespace
+  # When has no namespace, Lotus Logger take [Shared] as default app tag
+  #
+  # Examples
```
joneslee85: what do you about using `@example` instead of `Example`?

also, I think you should document the way that Lotus app uses this Logger, that is through the singleton way

joneslee85:
:+1: great work

jodosha:
@jdickey I see. For the time being, I prefer to ship a logger without customization. :shipit: and then improve.

Because of this policy, yeah UTC is the "brainless standard" so let's stick with it for now. :wink:

jodosha:
@huydx Thanks for the effort, this looks great! :sparkles: 

@joneslee85 Thanks for helping @huydx with this contribution. I can take care of merge and do a couple of API doc adjustments. :+1:

PR #133, opened 2015-01-14
Enhancement for Lotus App Generator

joneslee85:
What:
Currently, we eager-loading _all_ files in `/lib`, which is not very ideal in the scenario that users want to place some lazy-loading files in this folder, such as the mapping file. So after discussion with @jodosha, it is decided that we:

* [X] Eager-load only  `lib/chirp/**/*`
* [X] Provide an option to load mapping from file, and it will be located at `lib/config/mapping.rb`

Much thanks to @tranvictor for bug report

### DIFF
--- a/lib/lotus/generators/application/container/lib/app_name.rb.tt
+++ b/lib/lotus/generators/application/container/lib/app_name.rb.tt
@@ -19,6 +19,9 @@ Lotus::Model.configure do
   ##
   # Database mapping
   #
+  # Optionally, you can specify mapping file to load with:
+  #   mapping "{__dir__}/config/mapping"
+  #
   mapping do
     # collection :users do
     #   entity     User
--- a/lib/lotus/generators/application/container/lib/app_name.rb.tt
+++ b/lib/lotus/generators/application/container/lib/app_name.rb.tt
@@ -1,5 +1,5 @@
 require 'lotus/model'
-Dir["#{ __dir__ }/**/*.rb"].each { |file| require_relative file }
+Dir["#{ __dir__ }/<%= config[:app_name] %>/**/*.rb"].each { |file| require_relative file }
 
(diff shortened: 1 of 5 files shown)


stevehodgkiss:
What was the bug?

joneslee85:
@stevehodgkiss the bug is that we do eager-loading for all files in `/lib` which cause an issue with lazy-loading file such as *mapping file*. We do not want mapping file to be eager-loaded, we only want to eager-load all files inside `/lib/chirp/`.

lib/lotus/generators/application/container/lib/app_name.rb.tt
```diff
@@ -19,6 +19,9 @@ Lotus::Model.configure do
   ##
   # Database mapping
   #
+  # Optionally, you can specify mapping file to load with:
```
jodosha:
```ruby
# You can specify mapping file to load with:
#
# mapping "{__dir__}/config/mapping"
#
# Alternatively, you can use a block syntax like the following:
```

jodosha:
@joneslee85 Please check my only inline comment. If it's good for you, please amend.
Apart from that LGTM :+1:

stevehodgkiss:
LGTM too :+1:

lib/lotus/generators/slice/templates/application.html.erb
```diff
@@ -1,9 +0,0 @@
-<!doctype HTML>
```
AlfonsoUceda: why remove this?

AlfonsoUceda:
:+1:

lib/lotus/generators/slice/templates/application.html.erb
joneslee85: defunct

PR #135, opened 2015-01-15
Add ApplicationName, fixing dasherized application names and prohibiting use of restricted words.

robyurkowski:
This works for both application names and slice names.

### DIFF
--- a/lib/lotus/application_name.rb
+++ b/lib/lotus/application_name.rb
@@ -1,0 +1,99 @@
+module Lotus
+  # An application name.
+  #
+  # @since x.x.x
+  class ApplicationName
+
+    # A list of words that are prohibited from forming the application name
+    #
+    # @since x.x.x
+    RESERVED_WORDS = %w(lotus).freeze
+
+
+    # Initialize and check against reserved words
+    #
+    # An application name needs to be translated in quite a few ways:
+    # First, it must be checked against a list of reserved words and rejected
+    # if it is invalid. Secondly, assuming it is not invalid, it must be able
+    # to be output roughly as given, but with the following changes:
+    #
+    # 1. downcased,
+    # 2. with surrounding spaces removed,
+    # 3. with internal whitespace rendered as underscores
+    # 4. with underscores de-duplicated
+    #
+    # which is the default output. It must also be transformable into an
+    # environment variable.
+    #
+    # @return [Lotus::ApplicationName] a new instance of the application name
+    #
+    # @since x.x.x
+    def initialize(name)
+      @name = sanitize(name)
+      ensure_validity!
+    end
+
+
+    # Returns the cleaned application name.
+    #
+    # @example
+    #   ApplicationName.new("my-App ").to_s # => "my-app"
+    #
+    # @since x.x.x
+    def to_s
+      @name
+    end
+
+
+    # Returns the application name uppercased with non-alphanumeric characters
+    # as underscores.
+    #
+    # @example
+    #   ApplicationName.new("my-app").to_env_s => "MY_APP"
+    #
+    # @since x.x.x
+    def to_env_s
+      @name.upcase.gsub(/\W/, '_')
+    end
+
+
+    # Returns true if a potential application name matches one of the reserved
+    # words.
+    #
+    # @example
+    #   Lotus::ApplicationName.invalid?("lotus") # => true
+    #
+    # @since x.x.x
+    def self.invalid?(name)
+      RESERVED_WORDS.include?(name)
+    end
+
+
+    private
+
+    # Raises RuntimeError with explanation if the provided name is invalid.
+    #
+    # @api private
+    # @since x.x.x
+    def ensure_validity!
+      if self.class.invalid?(@name)
+        raise RuntimeError,
+          "application name must not be any one of the following: " +
+          RESERVED_WORDS.join(", ")
+      end
+    end
+
+
+    # Cleans a string to be a functioning application name.
+    #
+    # @api private
+    # @since x.x.x
+    def sanitize(name)
+      name
+        .downcase
+        .strip
+        .gsub(/\s/, '_')
+        .gsub(/_{2,}/, '_')
+    end
+  end
+end
--- a/lib/lotus/commands/new.rb
+++ b/lib/lotus/commands/new.rb
@@ -1,4 +1,5 @@
 require 'pathname'
+require 'lotus/application_name'
 require 'lotus/utils/string'
 require 'lotus/utils/class'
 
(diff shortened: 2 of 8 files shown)


lib/lotus/application_name.rb
```diff
@@ -0,0 +1,95 @@
...
+    # 3. with internal whitespace rendered as underscores
+    # 4. with underscores de-duplicated
+    #
+    # which is the default output:
+    #
+    #     ApplicationName.new("my-App ").to_s # => "my-app"
+    #
+    # It must also be transformable into an environment variable:
+    #
+    #     ApplicationName.new("my-app").to_env_s => "MY_APP"
```
AlfonsoUceda: @robyurkowski you can put these examples in the rigth method, e.g

```ruby
@example 
  ApplicationName.new("my-App ").to_s # => "my-app"
def to_s
  ...
end
```

lib/lotus/application_name.rb
```diff
@@ -0,0 +1,95 @@
...
+    # @since x.x.x
+    def to_env_s
+      @name.upcase.gsub(/\W/, '_')
+    end
+
+
+    # Returns true if a potential application name matches one of the reserved
+    # words.
+    #
+    # @since x.x.x
```
AlfonsoUceda: can you put an example here?

lib/lotus/application_name.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    private
+
+    # Raises RuntimeError with explanation if the provided name is invalid.
+    #
+    # @api private
+    # @since x.x.x
+    def ensure_validity!
+      if self.class.invalid?(@name)
+        raise RuntimeError,
```
AlfonsoUceda: I think we should create a custom Exception, e.g:

```ruby
class NameReservedError < ::StandardError
end
```

Thoughts?

AlfonsoUceda:
What do you think to create a test in `test/commands/new_test.rb` with the name `lotus`. I know there is a test in `test/application_name_test.rb` but it's unitary test, in `test/commands/new_test.rb` I think is more for an integration test

lib/lotus/application_name.rb
```diff
@@ -0,0 +1,95 @@
+module Lotus
+  # An application name.
+  #
+  # @since x.x.x
+  class ApplicationName
+
+    # A list of words that are prohibited from forming the application name
+    #
+    # @since x.x.x
+    INVALID_WORDS = %w(lotus).freeze
```
jodosha: Maybe `RESERVED_WORDS` is more appropriate. If you check the next code comment you'll find yourself using the same term:

> Initialize and check against reserved words

lib/lotus/application_name.rb
jodosha: I don't think this is a problem.
This exception isn't meant to be handled by developers, it's only a way to quit the CLI application with a non-successful state.

Indeed it could be optionally expressed as:

```ruby
fail "application name.."
```

lib/lotus/application_name.rb
```diff
@@ -0,0 +1,95 @@
...
+      end
+    end
+
+
+    # Cleans a string to be a functioning application name.
+    #
+    # @api private
+    # @since x.x.x
+    def sanitize(name)
+      name
```
jodosha: Not sure, but it looks like you want to transform the given arg into an "underscored" name. If so, please use `Utils::String#underscore`.

lib/lotus/commands/new.rb
```diff
@@ -10,11 +11,11 @@ class New
       attr_reader :app_name, :source, :target, :cli, :options
 
       def initialize(app_name, environment, cli)
-        @app_name = app_name
+        @app_name = ApplicationName.new(app_name)
         @options  = environment.to_options
         @arch     = @options.fetch(:architecture)
 
-        @target   = Pathname.pwd.join(@app_name)
+        @target   = Pathname.pwd.join(@app_name.to_s)
```
jodosha: I'd say we should decouple string representation (`#to_s`), from file system's manipulations. What do you think if we introduce `ApplicationName#to_path`?

jodosha:
@robyurkowski Thanks for this PR. I left a couple of comments, can you please take care of it?

Apart from those, it looks good to me :+1:

jodosha:
@robyurkowski Before to merge this there are two cleanups to put in place.

1. There is a Git and logical conflict in the app name resolution
2. When I run the specs on my computer, I see an output of a new application created: https://gist.github.com/jodosha/9c377a4f7f9b8b5fea73

Would you like to take care of this? Thank you very much.

lib/lotus/application_name.rb
robyurkowski: Doesn't quite do the same thing:
```
1) Failure:
Lotus::ApplicationName::formats::#to_s#test_0003_renders internal spaces as underscores [/home/rob/Sites/l/lotus/lotus/test/application_name_test.rb:21]:
Expected: "my_app"
  Actual: "my app"


  2) Failure:
Lotus::ApplicationName::formats::#to_s#test_0004_renders with underscores de-duplicated [/home/rob/Sites/l/lotus/lotus/test/application_name_test.rb:26]:
Expected: "my_app"
  Actual: "my _app"


  3) Failure:
Lotus::ApplicationName::formats::#to_s#test_0002_renders trimmed [/home/rob/Sites/l/lotus/lotus/test/application_name_test.rb:16]:
Expected: "my-app"
  Actual: " my-app "
```

The underscoring is part of it, yes, but a big part of it is also sanitization of potential incorrect input. Not opposed to using a `Lotus::Util::String` method or creating one, though.

lib/lotus/commands/new.rb
robyurkowski: :+1: -- even if it's just right now aliased to `#to_s`, that's a good idea in terms of API surface.

robyurkowski:
I think I've fixed most of these but for one thing, and that's the way we provide path for generation. The fact that we can now use `.` as well as `--path` substantially complicates the way we set our `@target`. I think I need some guidance before proceeding:

While we need to be sanitizing our `@app_name` at the very least, should we be sanitizing our path so it's not installed in some crazy location? i.e. should we prevent the user from installing to a path with spaces in it for consistency's sake? If not, I don't see a particular reason why we'd need `ApplicationName#to_path`.

While I'm at it, I'm also going to clean up the shadowed outer variable `app_name`, clarifying its intent as an `app_name_or_path`. The file's getting a bit hard to make sense of because of that.

jodosha:
> should we be sanitizing our path so it's not installed in some crazy location? i.e. should we prevent the user from installing to a path with spaces in it for consistency's sake?

This is another point in favor to decouple application name from path target. In other words, we should sanitize application name as usual, but paths should stay untouched, because that is what the developer wants and we can't take _"smart"_ decisions for them.

robyurkowski:
:+1:. Fix on its way.

jodosha:
@robyurkowski Thank you! :+1:

PR #139, opened 2015-01-17
Adding migration generator

thecatwasnot:
No options currently used

`bundle exec lotus generate migration migration_name` creates `db/migrations/<timestamp>_migration_name.rb`

`bundle exec lotus generate migration slice_name migration_name` creates `apps/slice_name/db/migrations/<timestamp>_migration_name.rb`

### DIFF
--- a/lib/lotus/cli/generate.rb
+++ b/lib/lotus/cli/generate.rb
@@ -6,13 +6,13 @@ module Lotus
   class Generate < Thor
     include Thor::Actions
 
-    desc 'migration NAME', 'generates a model migration'
-    def migration(name = nil)
-      if options[:help] || name.nil?
+    desc 'migration [SLICE] NAME', 'generates a model migration'
+    def migration(*names)
+      if options[:help] || names.empty? || names.length > 2
         invoke :help, ['migration']
       else
         require 'lotus/commands/generate/migration'
-        Lotus::Commands::Generate::Migration.new(name, environment, self).start
+        Lotus::Commands::Generate::Migration.new(names, environment, self).start
       end
     end
 
--- a/lib/lotus/cli/generate.rb
+++ b/lib/lotus/cli/generate.rb
@@ -5,6 +5,7 @@ require 'lotus/environment'
 module Lotus
   class Generate < Thor
     include Thor::Actions
+
     desc 'migration NAME', 'generates a model migration'
     def migration(name = nil)
       if options[:help] || name.nil?
--- a/lib/lotus/cli/generate.rb
+++ b/lib/lotus/cli/generate.rb
@@ -1,0 +1,24 @@
+require 'thor'
+require 'lotus/environment'
+
+
+module Lotus
+  class Generate < Thor
+    include Thor::Actions
+    desc 'migration NAME', 'generates a model migration'
+    def migration(name = nil)
+      if options[:help] || name.nil?
+        invoke :help, ['migration']
+      else
+        require 'lotus/commands/generate/migration'
+        Lotus::Commands::Generate::Migration.new(name, environment, self).start
+      end
+    end
+
+    private
+
+    def environment
+      Lotus::Environment.new(options)
+    end
+  end
+end
--- a/lib/lotus/commands/generate/migration.rb
+++ b/lib/lotus/commands/generate/migration.rb
@@ -9,10 +9,8 @@ module Lotus
         attr_reader :migration_name, :migration_time, :migration_class, :cli, :source, :target
 
         def initialize(name, environment, cli)
-          name = Lotus::Utils::String.new(name)
-          @migration_name = name.underscore
+          set_migration_name(name)
           @migration_time = generate_timestamp
-          @migration_class = name.classify
 
           @target = Pathname.new(environment.root)
           @source = Pathname.new(::File.dirname(__FILE__) + '/../../generators/migration/')
@@ -27,6 +25,12 @@ module Lotus
         end
 
         private
+
+        def set_migration_name(name)
+          name = Lotus::Utils::String.new(name)
+          @migration_name = name.underscore
+          @migration_class = name.classify
+        end
         
         def generate_timestamp
           Time.now.strftime("%Y%m%d%H%M%S")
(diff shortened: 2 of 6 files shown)


lib/lotus/generators/migration.rb
```diff
@@ -0,0 +1,29 @@
...
+      def start
+        opts = {
+          #app_name:         app_name,
+          migration_name:   migration_name,
+          migration_time:   migration_time,
+          migration_class:  migration_class,
+        }
+
+        templates = {
+          'migration.rb.tt' => "db/migrate/#{ opts[:migration_time] }_#{ opts[:migration_name] }.rb"
```
joneslee85: `#{ opts[:migration_time] }_#{ opts[:migration_name] }` can be extracted to a private method

lib/lotus/commands/generate/migration.rb
```diff
@@ -0,0 +1,32 @@
...
+    module Generate
+      class Migration 
+        
+        attr_reader :app_name, :migration_name, :migration_time, :migration_class, :cli, :source, :target
+
+        def initialize(name, environment, cli)
+          name = Lotus::Utils::String.new(name)
+          #@app_name = ""
+          @migration_name = name.underscore
+          @migration_time = Time.now.to_i.to_s
```
joneslee85: Timestamp should be human readable, what do you think about formatting it as `ddmmyyyy`?

test/commands/generate/migration_test.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  before do
+    create_temporary_dir
+  end
+
+  after do
+    chdir_to_root
+  end
+
+  describe 'migration' do
```
joneslee85: how about `describe '#start'`?

lib/lotus/commands/generate/migration.rb
```diff
@@ -0,0 +1,38 @@
...
+        end
+
+        def start
+          @command.start
+        end
+
+        private
+        
+        def generate_timestamp
+          Time.now.strftime("%d%m%Y")
```
AlfonsoUceda: It should be changed, better this for better sorting:

`Time.now.strftime("%Y%m%d")` #=> 20150117

lib/lotus/cli/generate.rb
```diff
@@ -0,0 +1,24 @@
+require 'thor'
+require 'lotus/environment'
+
+
+module Lotus
+  class Generate < Thor
+    include Thor::Actions
+    desc 'migration NAME', 'generates a model migration'
```
AlfonsoUceda: space above please, between this line and include

AlfonsoUceda:
:+1: perfect for me except above comments

lib/lotus/commands/generate/migration.rb
thecatwasnot: Actually it should probably be YmdHs to keep unique.

AlfonsoUceda: perfect ;)

lib/lotus/commands/generate/migration.rb
```diff
@@ -0,0 +1,38 @@
...
+module Lotus
+  module Commands
+    module Generate
+      class Migration 
+        
+        attr_reader :app_name, :migration_name, :migration_time, :migration_class, :cli, :source, :target
+
+        def initialize(name, environment, cli)
+          name = Lotus::Utils::String.new(name)
+          #@app_name = ""
```
joneslee85: this could be removed

lib/lotus/commands/generate/migration.rb
```diff
@@ -0,0 +1,37 @@
...
+module Lotus
+  module Commands
+    module Generate
+      class Migration 
+        
+        attr_reader :migration_name, :migration_time, :migration_class, :cli, :source, :target
+
+        def initialize(name, environment, cli)
+          name = Lotus::Utils::String.new(name)
+          @migration_name = name.underscore
```
joneslee85: What do you think about `Tell, Don't Ask`? I think it'd better to have a private method `generate_name` here.

lib/lotus/generators/migration.rb
```diff
@@ -0,0 +1,34 @@
...
+
+        templates.each do |src, dst|
+          cli.template(source.join(src), target.join(dst), opts)
+        end
+      end
+
+      private
+
+      def migration_file_name(time, name)
+        "#{time}_#{name}"
```
joneslee85: Why don't we access directly to `migration_time` and `migration_name` here?

test/commands/generate/migration_test.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  before do
+    create_temporary_dir
+  end
+
+  after do
+    chdir_to_root
+  end
+
+  describe 'start' do
```
joneslee85: How about following convention `#` prefix to denote a method? That is `#start` instead of `start`

joneslee85:
except above comments and failed build, this is :+1: 

Pending discussion: In order to generate migration for `lib/chirp` and `apps/web`, we need a reliable way to detect the app name. Suggested solution is having app name defined in `.lotusrc` file. I am not really a fan of this approach, I prefer an alternative and implicit way. 

Here is my proposal:

* Enforce naming convention for `lib/chirp`, that is renaming `lib/chirp.rb` to `lib/application.rb`
* Introduce options to set migration path for `Lotus::Model` and `Lotus::Application`


@jodosha thoughts?

thecatwasnot:
My current feeling is that the application db migratioins should live in `db/migrate` and slice specific migrations should live in `apps/web/db/migrate`
IMHO that would be the least surprising option.
I can add the slice migration generator with `bundle exec lotus generate migration web migration_name`

joneslee85:
@thecatwasnot `db/migrate` seems to be poor choice of naming, as suggested, I want to rename it to `db/migrations`, can you do some magic and make it happen? TIA

AlfonsoUceda:
I think we should cover the 2 scenarios:
`bundle exec lotus generate migration migration_name` and `bundle exec lotus generate migration slice_name migration_name`

thecatwasnot:
this is in progress guys, should be done tonight

thecatwasnot:
#256 has more features & is up to date

PR #145, opened 2015-02-03
Update .travis.yml for faster runs

jc00ke:
Use Travis container for faster tests.
Cache Bundler downloads.


jodosha:
@jc00ke Thank you, Sir. :+1:

PR #148, opened 2015-02-04
Full applications preload in console

### LINKED ISSUE #144: Load the app in console
I was poking around the `console` and wanted to check out the routes. @jodosha had mentioned I could access them by using `MyApp::Routes` but I kept getting a missing constant error.

@joneslee85 informed me in chat that I had to actually load the app, which would mount the `Routes`. He suggested I run `MyApp::Container.new` which did make `Routes` available.

It'd be nice if I didn't have to do that when poking around the `console`. It would also be nice to figure out a way to expose routes for routing specs, but that is a separate issue.

jodosha:
Closes #144

### DIFF
--- a/lib/lotus/application.rb
+++ b/lib/lotus/application.rb
@@ -145,7 +145,14 @@ module Lotus
       Lotus::Loader.new(application).load!
     end
 
-    # Preload all the registered applications
+    # Preload all the registered applications, by yielding their configurations
+    # and preparing the frameworks.
+    #
+    # This is useful for testing suites, where we want to make Lotus frameworks
+    # ready, but not preload applications code.
+    #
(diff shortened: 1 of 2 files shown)


lib/lotus/application.rb
```diff
@@ -154,6 +161,24 @@ def self.preload!
...
+    #
+    # This is useful in console where we want all the application code available.
+    #
+    # @return [void]
+    #
+    # @since x.x.x
+    # @api private
+    def self.preload_applications!
+      synchronize do
+        applications.each {|app| app.new }
```
joneslee85: /s/{|/{ |

joneslee85:
:+1:

AlfonsoUceda:
:+1:

PR #152, opened 2015-02-12
Add DBConsole command

Linuus:
As discussed here: https://github.com/lotus/lotus/issues/123

I still haven't gotten around to add the tests yet, but I thought I'd open the PR anyway.

Have I missed any options?

Related: https://github.com/lotus/model/pull/150


joneslee85:
I am good with this PR :+1:

jodosha:
@Linuus @joneslee85 I'm confused. This proposed feature should be used by passing manually CLI args for authentication, database etc..?

joneslee85:
@jodosha you are right, I totally forget about that.

@Linuus the idea is not to provide username/password explicitly. The idea that the CLI would pick up the configuration and handle everything for you.

For example:

`bundle exec lotus dbconsole`
`LOTUS_ENV=test bundle exec lotus dbconsole`

Linuus:
Hmm well, of course! :)




I'll fix it when I'm back home after the weekend.

jodosha:
@Linuus Thank you very much! :+1:

Linuus:
@jodosha, @joneslee85 Do we still want to keep the CLI options so you may pass in overrides? Perhaps that's not necessary.

joneslee85:
@Linuus my vote is to remove them, it should be the job of our configuration file

jodosha:
@Linuus I agree with @joneslee85

Linuus:
@joneslee85 @jodosha I have another question.

Right now this only works with the database configuration set in `lib/my_app.rb`, since I use `Lotus::Model.configuration`. You can pass in a name, like `lotus dbconsole web` but I don't do anything with that parameter at the moment.

How do I get the configuration for the `Web` app in this case?

I think I can do
```ruby
Web::Application.load!
Web::Application.configuration.adapter   #=> {:type=>:sql, :uri=>"postgres://localhost/foo_development"}
```

Is that the way to do it?

Can I get an adapter object somehow? Like when I run `Lotus::Model.configuration.adapter_config`?

joneslee85:
@Linuus very good question, you could access the `adapter_config` (after running `.load!`) by:

```ruby
Web::Model.configuration.adapter_config
```

FYI, the `Lotus::Loader#_configure_model_framework!` duplicate the Model and set it under your app namespace.

It is ideal not to run `.load!` to avoid loading overhead. When you are online tonight, please ping me, I'll see what I could do to help you no this.

joneslee85:
It'd be great if you could test for sqlite uri string escaping

joneslee85:
@Linuus looks good, I am not so sure how are we going to do integration test for this feature, but for now can you please do hand test? Once good, we could merge

Linuus:
Sure. I was unsure myself on how to do it. I tried the other day but nothing good came out of it ;)

I have tested it with sqlite and postgresql (will test mysql too). Only with the default container architecture though.

Should I test something else?

joneslee85:
@Linuus should be good for now.

joneslee85:
ping @Linuus - any update?

Linuus:
Sorry for being slow. I will test mysql tonight.

Linuus:
@joneslee85 I tested with mysql and it seems to work good! :+1:

joneslee85:
Woo! Much thanks

PR #158, opened 2015-02-19
Change SQLite file extension to .sqlite3

mconnell:
As a user of Sublime Text, Switching to the :sql adaptor with sqlite
generates a `<app_name>_development.db` file. Sublime text by default
hides .db extension, and having the original directories generated by
the :file_system adaptor lying around added a little bit of confusion
as to why I couldn't see a database file, even though I could access it
within the REPL.

Although the adaptor is specified within the codebase, the database
file is probably more likely to be stored in another location outwith
the repository when deployed to a server. Having a more useful extension
might be helpful. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ module Bookshelf
       #           uri:  String, 'file:///db/bookshelf'
       #                         'memory://localhost/bookshelf'
       #                         'sqlite:memory:'
-      #                         'sqlite://db/bookshelf.db'
+      #                         'sqlite://db/bookshelf.sqlite3'
       #                         'postgres://localhost/bookshelf'
       #                         'mysql://localhost/bookshelf'
       #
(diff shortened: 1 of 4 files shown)


joneslee85:
@mconnell thanks for filing this PR. I think the file extension naming is a matter of personal choice. IMHO users should mess directly with db file (it might be partly the reason why Sublime Text hide this extension). By having said that, I acknowledge that this might cause confusion for users who uses SQLite3 on production (though I assume the number of production app using SQLite3 is low). I am happy to merge this PR _if_ it is truly a pain point to your workflow.

mconnell:
I'm not terribly attached to the PR, it was mostly I had generated a new lotus application and found it a needless distraction, and someone else might encounter the same situation. I think there's merit in the file extension being more descriptive, but that's only a personal opinion.

jodosha:
@mconnell This makes a lot of sense, `.sqlite3` is the most used extension, `.db` is a minor one. Let's stick to the well adopted. Thank you for this PR :+1:

PR #160, opened 2015-02-23
Set default LOTUS_ENV variable to development

joneslee85:


jodosha:
@joneslee85 Can you please explain the motivations behind it? :smiley_cat: 
`Lotus::Environment` already defaults to `development`.

PR #163, opened 2015-02-26
Init a git repo when generating a new application with container arch

### LINKED ISSUE #132: `lotus new` should initialize a git repository
While I dislike being git-centric, I think this is important behaviour. A plurality of developers are going to be using git, and failing to support that as the default also means that users who quickly add a repository will often miss files that shouldn't be committed, such as `config/.env` et al. which are files that are exceedingly easy to miss.

I realized that I myself contributed my app's session secret earlier as a result. Granted it was the development secret, but if there had been a generated production secret, it would now be in my repository's history.

Linuus:
I read this issue https://github.com/lotus/lotus/issues/132 and thought I'd whip something simple together. Perhaps someone else has already implemented it.

What do you think about this? I guess we should check if the `git` command is available before trying to run it?

Is there a better way to run a command inside the root dir than this?
```ruby
cli.run("git init #{Shellwords.escape(target)}", capture: true)
```

Closes #132

### DIFF
--- a/lib/lotus/generators/application/container.rb
+++ b/lib/lotus/generators/application/container.rb
@@ -1,3 +1,4 @@
+require 'shellwords'
 require 'lotus/generators/abstract'
 require 'lotus/generators/slice'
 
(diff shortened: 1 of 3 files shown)


jodosha:
@Linuus Thank you very much, I've manually merged it!

Linuus:
Oh, that was fast :P

PR #166, opened 2015-02-28
Action generator

### LINKED ISSUE #154: Action generator
## What

Introduce a new command `lotus generate`.

## Why

When working on an application, it's tedious to create several files to make the next spec to pass.

### CLI

```shell
% lotus generate action web books#index
```

  * `action` is the subject of what we want to generate
  * `web` is the name of the application under `apps/`
  * `books#index` is the name of the controller and action
  * Optional argument: `--test=minitest|rspec`. Where `minitest` is the default
  * Optional argument: `--test=erb|haml|slim|etc..`. Where `erb` is the default

### Output

  * `apps/web/controllers/dashboard/index.rb`
  * `spec/web/controllers/dashboard/index_spec.rb`
  * `apps/web/views/dashboard/index.rb`
  * `spec/web/views/dashboard/index_spec.rb`
  * `apps/web/templates/dashboard/index.html.erb` (**empty file**)
  * Amend `apps/web/config/routes.rb` with a new route

### Files

```ruby
# apps/web/controllers/dashboard/index.rb
module Web::Controllers::Dashboard
  class Index
    include Web::Action

    def call(params)
    end
  end
end
```

```ruby
# spec/web/controllers/dashboard/index_spec.rb
# THE CONTENT OF THIS FILE DEPENDS ON THE --test OPTION
require 'spec_helper'

describe Web::Controllers::Dashboard::Index do
  let(:action) { Web::Controllers::Dashboard::Index.new }
  let(:params) { Hash[] }

  it 'is successful' do
    response = action.call(params)
    response[0].must_equal 200
  end
end
```

```ruby
# apps/web/views/dashboard/index.rb
module Web::Views::Dashboard
  class Index
    include Web::View
  end
end
```

```ruby
# spec/web/views/dashboard/index_spec.rb
# THE CONTENT OF THIS FILE DEPENDS ON THE --test OPTION
require 'spec_helper'

describe Web::Views::Dashboard::Index do
  let(:exposures) { Hash[foo: 'bar'] }
  let(:template)  { Lotus::View::Template.new('apps/web/templates/dashboard/index.html.erb') }
  let(:view)      { Web::Views::Dashboard::Index.new(template, exposures) }
 
  it 'exposes #foo' do
    @action.foo.must_equal exposures.fetch(:foo)
  end
end
```

```ruby
# apps/web/config/routes.rb
# OPEN TO DISCUSSION
get '/dashboard', to: 'dashboard#index'
```

### Bonus points

We have in the pipeline two enhancements (#155 and #156) that will serve this feature. All the newly created Lotus apps will have a `.lotusrc` file that store some important informations such as the testing framework and the template engine.

The generator should look at the contents of that file, and generate code accordingly. Eg, create testing files for RSpec or Mintest or generate `.haml` templates.

Here some rules:

  * If `.lotusrc` is missing, and no args are specified, then fallback to the defaults
  * If `.lotusrc` is missing, but one of those args are specified, then follow them
  * If `.lotusrc` is there, and no args are specified, follow the instructions in the file
  * If `.lotusrc` is there, and args are specified, then follow them
  * The command should complain if the testing framework isn't included one of `minitest` or `rspec`
  * The command should allow all the template values (eg. `xyz` is a valid value)

jodosha:
### Feature

See #154 
(it also implements the bonus points listed at the bottom)

### Technical notes

It introduces integration tests for CLI commands.

---

Closes #154


lib/lotus/cli.rb
```diff
@@ -79,6 +79,19 @@ def new(name = nil)
       end
     end
 
+    desc 'generate', 'generates a new action'
+    method_option :path,                desc: 'applications path',        type: :string, default: 'apps'
```
joneslee85: @jodosha what do you think about about removing this option? IMHO, users tend to likely to generate new action at LOTUS_ROOT path

lib/lotus/commands/generate.rb
```diff
@@ -0,0 +1,43 @@
...
+      def initialize(type, app_name, name, env, cli)
+        @cli      = cli
+        @options  = env.to_options.merge(cli.options)
+
+        @app_name = app_name
+        @app      = Utils::String.new(@app_name).classify
+
+        @name     = name
+
+        @source = Pathname.new(::File.dirname(__FILE__) + '/../generators/action/').realpath
```
joneslee85: nitpick, what do you think about aligning this `=` with other lines?

lib/lotus/commands/generate.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module Lotus
+  module Commands
+    class Generate
+      GENERATORS_NAMESPACE = "Lotus::Generators::%s".freeze
+
+      class Error < ::StandardError
+      end
+
+      attr_reader :cli, :source, :target, :app, :app_name, :app_root, :name, :options
```
joneslee85: `app_root` method is already defined below, defining reader for `app_root` seems unnecessary

lib/lotus/commands/generate.rb
```diff
@@ -0,0 +1,43 @@
...
+        @app      = Utils::String.new(@app_name).classify
+
+        @name     = name
+
+        @source = Pathname.new(::File.dirname(__FILE__) + '/../generators/action/').realpath
+        @target = Pathname.pwd.realpath
+
+        require "lotus/generators/#{ type }"
+        generator  = Utils::String.new(type).classify
+        @generator = Utils::Class.load!(GENERATORS_NAMESPACE % generator).new(self)
```
joneslee85: What do you think about extracting this generator class logic into a private `generator_class` method to make the code cleaner?

lib/lotus/generators/action.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      def generate_route
+        path = target.join(_routes_path)
+        path.dirname.mkpath
+
+        FileUtils.touch(path)
+
+        # Insert at the top of the file
+        cli.insert_into_file _routes_path, before: /\A(.*)/ do
+          "get '/#{ @controller }', to: '#{ name }'\n"
```
joneslee85: @jodosha do you think it is a good idea to let user to specify the http method here?

lib/lotus/generators/action.rb
```diff
@@ -0,0 +1,102 @@
...
+      def _routes_path
+        "#{ Pathname.new(app_root).join('config', 'routes') }#{ SUFFIX }"
+      end
+
+      def _action_path
+        "#{ Pathname.new(app_root).join('controllers', @controller, @action) }#{ SUFFIX }"
+      end
+
+      def _view_path
+        "#{ Pathname.new(app_root).join('views', @controller, @action) }#{ SUFFIX }"
```
joneslee85: @jodosha can we DRY it up a bit?

lib/lotus/generators/action.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      def _template_path
+        "#{ Pathname.new(app_root).join('templates', @controller, @action) }#{ TEMPLATE_SUFFIX }#{ options.fetch(:template) { DEFAULT_TEMPLATE }}"
+      end
+
+      def _action_spec_path
+        "#{ Pathname.new('spec').join(app_name, 'controllers', @controller, "#{ @action }_spec") }#{ SUFFIX }"
```
joneslee85: @jodosha can we DRY it up a bit?

joneslee85:
one of the enlisted features is the option to specify the template type for eg haml, it seems to me we do not cater for such cases yet

lib/lotus/commands/generate.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      def start
+        @generator.start
+      rescue Error => e
+        puts e.message
+        exit 1
+      end
+
+      def app_root
+        @app_root ||= [@options[:path], @app_name].join(::File::SEPARATOR)
```
joneslee85: I myself am not a fan of setter/getter all in one, what do you think about renaming it to `set_app_root` and invoke it once in the constructor instead?

lib/lotus/cli.rb
jodosha: @joneslee85 This is to describe where applications are, located. It defaults to `apps/`. But yeah, it doesn't look important.

lib/lotus/generators/action.rb
jodosha: @joneslee85 How so? :smile:

lib/lotus/commands/generate.rb
jodosha: Well, this is a cached value. Do you think we should rename it?

jodosha:
@joneslee85 
> one of the enlisted features is the option to specify the template type for eg haml, it seems to me we do not cater for such cases yet

We look at `.lotusrc` file https://github.com/lotus/lotus/commit/646f15b3cf5e60f18638469cd07c05b08eaccf14

lib/lotus/cli.rb
jodosha: @joneslee85 On the other hand, if we remove that option, we're forced to hardcode that value (`"apps"`) in `Lotus::Commands::Generate` or `Lotus::Generators::Action`. Isn't it better to leave some flexibility to devs?

Please let me know. :smiley_cat:

joneslee85: I'd vote against letting dev to customise it. Once we get to the point where we allow different architecture, we could always go back to this.

lib/lotus/generators/action.rb
joneslee85: must action always be a GET? what about update action which uses PATCH or PUT?

jodosha: This is a facility I expect devs to amend this. To generate the correct information is hard. We have two options: 1. CLI arg ( :scream: ) or 2. Interactive CLI.

I think this is a good compromise for now to let us to ship this monday :shipit:

joneslee85: :shipit: and enhance it later :)

joneslee85:
:+1: ship it!

test/integration/cli/generate_test.rb
```diff
@@ -0,0 +1,103 @@
...
+      @tmp.mkpath
+
+      Dir.chdir(@tmp)
+    end
+
+    def generate_application
+      `bundle exec lotus new #{ @app_name = 'delivery' }#{ options }`
+      Dir.chdir(@root = @tmp.join(@app_name))
+
+      File.open(@root.join('.lotusrc'), 'w') do |f|
```
AlfonsoUceda: @jodosha I think you can use https://github.com/lotus/lotus/blob/master/lib/lotus/lotusrc.rb what do you think?

jodosha: @AlfonsoUceda Because it's doesn't override already existing files. In this test case it already exist because we have just generated a fresh app.

PR #167, opened 2015-03-02
Move dbconsole under namespace db

joneslee85:
Change `lotus dbconsole` to `lotus db console`. This paves way for other DB-related CLI such as `lotus db migrate`.

cc @Linuus


lib/lotus/cli.rb
```diff
@@ -3,6 +3,29 @@
 require 'lotus/version'
 
 module Lotus
+
+  class DBSubcommand < Thor
```
AlfonsoUceda: maybe we should move this class to a file, because there will be more subcommands, what do you think?

lib/lotus/commands/db/console.rb
```diff
@@ -0,0 +1,54 @@
...
+          load_config
+        end
+
+        def start
+          exec connection_string
+        end
+
+        private
+
+        def config
```
vinioliveira: Cool I like it. I have theses methods too. We should extract it to reuse. Since we're going to create an especific class to migrations

Linuus:
:+1:

lib/lotus/commands/db/console.rb
AlfonsoUceda: maybe we can do a parent class or a module for including in these clases

jodosha:
@joneslee85 Great change! :+1:

PR #169, opened 2015-03-02
Adapter (database) config when creating a new application

Linuus:
Fixes https://github.com/lotus/lotus/issues/164

Adds an ~~`adapter`~~ `database` option for the `lotus new` command.
Supports mysql, mysql2, postgresql, sqlite3, memory and filesystem.

Adds the appropriate gem in the Gemfile and url's in .env files.

The templates got kind of messy with a lot of if/else statements. Any ideas on how to clean it up?

Are there more places where the adapter affects the code generation?


lib/lotus/cli.rb
```diff
@@ -73,6 +73,7 @@ def routes
     end
 
     desc 'new', 'generates a new application'
+    method_option :adapter,                       desc: 'adapter config',             type: :string, default: 'filesystem'
```
joneslee85: nitpick, align `:default`

joneslee85: What do you think about rewording to `application persistence adapter`

lib/lotus/generators/application/container/Gemfile.tt
```diff
@@ -16,6 +16,16 @@ gem 'lotusrb',      '<%= Lotus::VERSION %>'
 gem 'lotus-model', '<%= config[:lotus_model_version] %>'
 <%- end -%>
 
+<%- if config[:adapter] == 'mysql' %>
```
joneslee85: What do you think about `case` syntax?

lib/lotus/generators/application/container/config/.env.development.tt
```diff
@@ -1,2 +1,14 @@
 # Define ENV variables for development environment
+<%- if config[:adapter] == 'mysql' %>
+<%= config[:app_name].to_env_s %>_DATABASE_URL="mysql://localhost/<%= config[:app_name] %>_development"
```
joneslee85: What do you think about DRY with `config[:adapter_uri]`? Furthermore, please ensure sqlite3 URI is escaped with Shellword

joneslee85: `config[:adapter_uri][:development]` & `config[:adapter_uri][:test]`

test/commands/new_test.rb
```diff
@@ -225,6 +343,48 @@ def container_options
...
+            content = @root.join('lib/chirp.rb').read
+            content.must_match %(adapter type: :sql, uri: ENV['CHIRP_DATABASE_URL'])
+          end
+        end
+
+        describe 'memory' do
+          let(:opts) { container_options.merge(adapter: 'memory') }
+          it 'generates adapter config for memory' do
+            content = @root.join('lib/chirp.rb').read
+            content.must_match %(adapter type: :memory, uri: ENV['CHIRP_DATABASE_URL'])
```
joneslee85: :clap: I want to kiss you right now. Can you please also look into the way we create `lotusrc` file? Can you please append `adapter=` line in the rc file too? Cheers

lib/lotus/cli.rb
joneslee85: Can we also add alias `-d`?

joneslee85: On second thought, why don't we change the option to `:database`?

Linuus: Sure, perhaps it is more clear to use `:database` than `:adapter`?

lib/lotus/generators/application/container/Gemfile.tt
Linuus: erb has trouble with case syntax. You need to put the `case` and first `when` in the same block. Like this

```ruby
<%- case config[:adapter]
    when 'mysql' %>
gem 'mysql'
<%- when 'mysql2' %>
gem 'mysql2'
<%- when 'postgresql' %>
gem 'pg'
<%- when 'sqlite3' %>
gem 'sqlite3'
<%- end -%>
```

Looks kind of broken to me :)

lib/lotus/cli.rb
Linuus: @jodosha Do you mind if we change it to `:database` instead?

lib/lotus/generators/application/container/Gemfile.tt
jodosha: @Linuus Can you please resolve this in `container.rb`? The pass it to `opts` as `database_gem` and keep the template clean with `gem <%= database_gem %>`.

Linuus: @jodosha Sure, no problems!

lib/lotus/generators/application/container/config/.env.development.tt
jodosha: @Linuus Can you please do the same here? I agree with @joneslee85 : it's better to resolve it in `container.rb`. Also because we'll reuse this code in the future for other generators.

test/commands/new_test.rb
jodosha: @joneslee85 what are the plans for having the adapter there?

Linuus: @joneslee85 Sure. I did that yesterday but that kind of broke the tests. It asks me a bunch of times when running the tests if I want to overwrite the `.lotusrc` file :) I will need to fix that somehow before committing or it will hang the tests on Travis.

lib/lotus/cli.rb
jodosha: @Linuus I think `database` is better :+1: 

The problem with `adapter` is that a generic word and it has no direct link with the database world.

Linuus: `database` it is then!

test/commands/new_test.rb
Linuus: On the other hand, each slice can have different adapters. Should all be added in the rc file in the future?

I'll skip the `.lotusrc` changes for now, since the tests are messing with me anyway. I can add it later if we decide that it is needed.

lib/lotus/generators/application/container/config/.env.development.tt
Linuus: @jodosha Yep! Did it yesterday but haven't pushed yet.

test/commands/new_test.rb
joneslee85: @Linuus please skip `lotusrc` for now. Cheers

Linuus:
@joneslee85 I just pushed my latest changes now. The `container.rb` file is starting to get a bit messy. Perhaps the `database_*` methods should be refactored somewhere else later on.

joneslee85:
:+1: , we could refactor later :)

Linuus:
:+1: Yay!

PR #172, opened 2015-03-05
Cleanup duplicate code

arandaio:


jodosha:
@arandaio Good catch, thank you! :+1:

PR #173, opened 2015-03-06
Set relative template path in generated apps' config

janlelis:
I am suggesting to comment in the templates path in new apps.

I had the following problem (which might still be another issue):
- Lotus by default also looks for templates in `public/**`
- I had a `public/stylesheets/application.css.map` file
- Because it has two file extensions,  Lotus thinks, it has to parse it as a template
- `RuntimeError: No template engine registered for application.css.map` :zap: 

Won't happen if I the relative template path is set.


joneslee85:
:+1:

PR #175, opened 2015-03-06
Integrate with Lotus::Helper

tomkadwill:
Added Lotus::Helper as a dependency and added it to the application generator, as discussed in https://github.com/lotus/lotus/issues/165.

/cc @jodosha

### DIFF
--- a/lib/lotus/generators/application/container.rb
+++ b/lib/lotus/generators/application/container.rb
@@ -9,11 +9,12 @@ module Lotus
         def initialize(command)
           super
 
-          @slice_generator     = Slice.new(command)
-          @lotus_head          = options.fetch(:lotus_head)
-          @test                = options[:test]
-          @database            = options[:database]
-          @lotus_model_version = '~> 0.2'
+          @slice_generator        = Slice.new(command)
+          @lotus_head             = options.fetch(:lotus_head)
+          @test                   = options[:test]
+          @database               = options[:database]
+          @lotus_model_version    = '~> 0.2'
+          @lotus_helpers_version  = '~> 0.0'
 
           cli.class.source_root(source)
         end
@@ -21,12 +22,13 @@ module Lotus
         def start
 
           opts      = {
-            app_name:            app_name,
-            lotus_head:          @lotus_head,
-            test:                @test,
-            database:            @database,
-            database_config:     database_config,
-            lotus_model_version: @lotus_model_version
+            app_name:               app_name,
+            lotus_head:             @lotus_head,
+            test:                   @test,
+            database:               @database,
+            database_config:        database_config,
+            lotus_model_version:    @lotus_model_version,
+            lotus_helpers_version:  @lotus_helpers_version
           }
 
           templates = {
--- a/lib/lotus/generators/application/container.rb
+++ b/lib/lotus/generators/application/container.rb
@@ -9,12 +9,12 @@ module Lotus
         def initialize(command)
           super
 
-          @slice_generator        = Slice.new(command)
-          @lotus_head             = options.fetch(:lotus_head)
-          @test                   = options[:test]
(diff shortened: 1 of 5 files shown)


lib/lotus/generators/application/container/Gemfile.tt
```diff
@@ -12,8 +12,9 @@ gem 'lotus-view',        require: false, github: 'lotus/view'
 gem 'lotus-model',       require: false, github: 'lotus/model'
 gem 'lotusrb',                           github: 'lotus/lotus'
 <%- else -%>
-gem 'lotusrb',      '<%= Lotus::VERSION %>'
-gem 'lotus-model', '<%= config[:lotus_model_version] %>'
+gem 'lotusrb',        '<%= Lotus::VERSION %>'
+gem 'lotus-model',    '<%= config[:lotus_model_version] %>'
+gem 'lotus-helpers',  '<%= config[:lotus_helpers_version] %>'
```
joneslee85: can we remove 1 space in between `gem 'lotus-helpers',` and `'<%= config[:lotus_helpers_version] %>'`?

lib/lotus/generators/application/container.rb
```diff
@@ -21,12 +22,13 @@ def initialize(command)
...
-            database:            @database,
-            database_config:     database_config,
-            lotus_model_version: @lotus_model_version
+            app_name:               app_name,
+            lotus_head:             @lotus_head,
+            test:                   @test,
+            database:               @database,
+            database_config:        database_config,
+            lotus_model_version:    @lotus_model_version,
+            lotus_helpers_version:  @lotus_helpers_version
```
joneslee85: can we remove 1 space in between gem 'lotus-helpers', and '<%= config[:lotus_helpers_version] %>'?

lotusrb.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'lotus-router',     '~> 0.2', '>= 0.2.1'
   spec.add_dependency 'lotus-controller', '~> 0.3', '>= 0.3.2'
   spec.add_dependency 'lotus-view',       '~> 0.3'
+  spec.add_dependency 'lotus-helpers',     '~> 0.0'
```
joneslee85: can you please also indent the versioning too?

test/commands/new_test.rb
```diff
@@ -46,8 +46,9 @@ def container_options
         content = @root.join('Gemfile').read
         content.must_match %(gem 'bundler')
         content.must_match %(gem 'rake')
-        content.must_match %(gem 'lotusrb',      '#{ Lotus::VERSION }')
-        content.must_match %(gem 'lotus-model', '~> 0.2')
+        content.must_match %(gem 'lotusrb',        '#{ Lotus::VERSION }')
+        content.must_match %(gem 'lotus-model',    '~> 0.2')
+        content.must_match %(gem 'lotus-helpers',  '~> 0.0')
```
joneslee85: nitpick, 2 spaces to 1 space

joneslee85:
except comments above, :+1:

AlfonsoUceda:
:+1:

tomkadwill:
Thanks @joneslee85. Updated with you're recommendations.

jodosha:
@joneslee85 @tomkadwill Sorry for being late at the party. First of all, thank you Tom for this integration.

There is one problem, we add both `lotus-helpers` as an hard dependency in the gemspec, but we also add it to the generated `Gemfile` of an app. This can cause conflicts. My idea is to have that gem as an hard dependency (gemspec) and to remove it from the `Gemfile`.

Another issue is that we use `~> 0.0`, it should be `~> 0.1` instead. @tomkadwill Can you please take care of this?

PR #178, opened 2015-03-08
Remove lotus-helpers dependency from Gemfile.tt

tomkadwill:
Related to https://github.com/lotus/lotus/issues/165.

Removed dependency from the Gemfile so that it is only in the gemspec.

Only issue is that the tests will not run at the moment because it cannot find `lotus-helpers (~> 0.1) ruby`

cc/ @jodosha @joneslee85


joneslee85:
:+1:

joneslee85:
@jodosha can you please push 0.1.0 to rubygems.org? cheers

jodosha:
@tomkadwill I think that the content of #182 should be added here, to complete this PR and makes CI green.

PR #180, opened 2015-03-10
Preload application in config/environment file.

smt116:
Hi there,

I had an issue with the `load_paths` and rake tasks / sidekiq workers.

## Links
The background and problem is based on the following code: https://gist.github.com/smt116/db97a1929162f643e073

## Background
Basically I have `services` and `workers` directories in the `apps/streamer`:
```
apps
├── metrics
[...]
└── streamer
    ├── application.rb
    ├── services
    │   └── foo.rb
    ├── tasks
    │   └── streamer.rake
    └── workers
        └── bar.rb
```
I have also `rake streamer:start` rake task which invokes `Streamer::Services::Foo.perform` action. This action streams messages from external, public API. After fetching each message, service passes it to the sidekiq worker which performs some actions on it and saves object into the database (slow `file_system` for test purposes). I have to use worker, because public API won't "wait" for my processing and it have to be as fast as it can.

## Problem
My problem is that `rake streamer:start` **and** `bundle exec sidekiq -C config/sidekiq.yml -r./config/environment.rb` will result in `NameError: uninitialized constant Streamer::Services` exception. I can solve it by adding explicit requires to `config/environment.rb`:
```
require_relative '../apps/streamer/workers/bar'
require_relative '../apps/streamer/services/foo'
```
but the same should be done by `load_paths` from `apps/streamer/application.rb`. I don't want to specify the same thing in two different places each time ;)

## Solution
The solution is to preload application in the `config/environment.rb` file so it will load proper files. I am playing with Lotus framework for 2 days now (and already love it :star2:) so not sure if it won't break something up.

Btw. sidekiq works out-of-the-box:
```
# apps/streamer/workers/foo.rb
require "sidekiq"

module Streamer::Workers
  class Foo
    include Sidekiq::Worker

    def perform(bar)
      ...
    end
  end
end

# apps/streamer/application.rb
module Streamer
    configure do
      root __dir__

      load_paths << [
        'workers',
        'services'
      ]

      handle_exceptions false
    end
  end
end

# bundle exec sidekiq -C config/sidekiq.yml -r./config/environment.rb
```

### DIFF
--- a/lib/lotus/generators/application/container.rb
+++ b/lib/lotus/generators/application/container.rb
@@ -37,6 +37,7 @@ module Lotus
             'Gemfile.tt'               => 'Gemfile',
             'config.ru.tt'             => 'config.ru',
             'config/environment.rb.tt' => 'config/environment.rb',
+            'config/loader.rb.tt'      => 'config/loader.rb',
             'lib/app_name.rb.tt'       => "lib/#{ app_name }.rb",
             'lib/config/mapping.rb.tt' => 'lib/config/mapping.rb',
           }
--- a/lib/lotus/generators/application/container/config/loader.rb.tt
+++ b/lib/lotus/generators/application/container/config/loader.rb.tt
@@ -1,0 +1,3 @@
+require_relative './environment'
+
+Lotus::Application.preload_applications!
--- a/test/commands/new_test.rb
+++ b/test/commands/new_test.rb
@@ -219,6 +219,14 @@ describe Lotus::Commands::New do
       end
     end
 
+    describe 'config/loader.rb' do
+      it 'generates it' do
+        content = @root.join('config/loader.rb').read
+        content.must_match %(require_relative './environment')
+        content.must_match %(Lotus::Application.preload_applications!)
+      end
+    end
+
     describe '.env' do
       it 'generates it' do
         content = @root.join('.env').read
@@ -659,6 +667,14 @@ describe Lotus::Commands::New do
       end
     end
 
+    describe 'config/loader.rb' do
+      it 'generates it' do
+        content = @root.join('config/loader.rb').read
+        content.must_match %(require_relative './environment')
+        content.must_match %(Lotus::Application.preload_applications!)
+      end
+    end
+
     describe '.env.development' do
       it 'patches the file to reference slice env vars' do
         content = @root.join('.env.development').read
@@ -833,6 +849,14 @@ describe Lotus::Commands::New do
         it 'generates it' do
           content = @root.join('config/environment.rb').read
           content.must_match %(require_relative '../lib/new')
+        end
+      end
+
+      describe 'config/loader.rb' do
+        it 'generates it' do
+          content = @root.join('config/loader.rb').read
+          content.must_match %(require_relative './environment')
+          content.must_match %(Lotus::Application.preload_applications!)
         end
       end
 


7even:
I had a similar problem trying to run my controller specs in the container architecture - `uninitialized constant Web::Controllers (NameError)`. [Adding](https://github.com/7even/keeper/commit/581ac314d42ed01e873c9a73f7712d4f9452551f) `Lotus::Application.preload_applications!` to `config/environment.rb` fixes that.

:+1:

jodosha:
@smt116 Thanks for this PR.

That API (`preload_applications!`) loads entirely all the applications and their code. If we accept this change it will **no longer possible to unit test elements in isolation**. Example: unit test an action.

It will also load unnecessary code for commands that don't need it: generators or route inspectors.
This will cause longer wait time for everything in general.

Can you please experiment with a Rake task named `:environment` that does what you have suggested here? Then let your task to depend on it: `task start: :environment`.

I see your pain point and if this Rake solution works we can keep things fast as described above and at the same time solve your problem. I guess it is be shared by other devs too.

What do you think?

jodosha:
@7even Regarding your testing problem, have you seen how things are managed now by code generators?

  * https://github.com/lotus/lotus/blob/master/lib/lotus/generators/application/container/spec_helper.rb.minitest.tt#L7
  * https://github.com/lotus/lotus/blob/master/lib/lotus/generators/application/container/spec_helper.rb.rspec.tt#L28

They use the faster `Lotus::Application.preload!` instead of `.preload_applications!`. The other advantage is that you can still unit test in isolation (see the comment above).

Please let me know if this will fix things for you.

smt116:
Hi @jodosha,

I can do it via `environment` task as a dependency which will preload app, but it will require to add also rake task for running sidekiq worker.

Replacing `.preload_applications!` with `Lotus::Application.preload!` also works. Can we use it or is it still results in unnecessary performance degradation?

7even:
@jodosha unfortunately I already [have this line](https://github.com/7even/keeper/blob/d931ca3509af5523dc5a632cf98703f98b9abd8e/spec/spec_helper.rb#L31) and it doesn't help. I've added a [simple controller spec](https://github.com/7even/keeper/blob/d931ca3509af5523dc5a632cf98703f98b9abd8e/spec/web/controllers/dashboard/index_spec.rb) so the problem can be reproduced - the whole spec suite fails after commenting out [Lotus::Application.preload_applications!](https://github.com/7even/keeper/blob/d931ca3509af5523dc5a632cf98703f98b9abd8e/config/environment.rb#L10).

smt116:
Any updates on this? (check my comment above)

joneslee85:
looks legit to me, I am in favour of the change. 1 vote from me. love to hear from @jodosha

7even:
Loading all the code for every spec doesn't seem nice, it would be cool to avoid that. Unfortunately I can't currently find the time to understand how it works to suggest something particular.

fidothe:
The following gist solves the controller specs chicken-and-egg problem for me (can't require the controller directly via require_relative because `Api::Controllers`, is not defined, cannot load the application in isolation).

https://gist.github.com/fidothe/a4f35b4f0102169f097a

The only really horrid thing I have to do is that `require_relative` call on line 2. That would be solved by `/apps` on the Ruby load path in the way that `/lib` is.

The other problem, that of needing to call `Api::Application.preload! `, could be solved by generating controllers with nested modules, a la this (admittedly pseduocode) https://gist.github.com/fidothe/1dd0ea14a2091d067343

There must be a more explicit way that reduces magic without being too verbose

jodosha:
@fidothe Thanks for the extensive feedback.

### Preload

Starting from `lotusrb-0.3.0` we'll generate [Minitest](https://github.com/lotus/lotus/blob/master/lib/lotus/generators/application/container/spec_helper.rb.minitest.tt#L7) or [RSpec](https://github.com/lotus/lotus/blob/master/lib/lotus/generators/application/container/spec_helper.rb.rspec.tt#L28) `spec_helper.rb` where we do `Lotus::Application.preload!`.

What this [API does](https://github.com/lotus/lotus/blob/master/lib/lotus/application.rb#L160) is to iterate through the applications an to load them. In other words, it's equivalent to your `Api::Application.preload!`.

### Action generator

The boring manual `require_relative` can be solved by the upcoming Action generators #166

smt116:
Okay, but I guess that we are not able to make `sidekiq` (and possible other tools) work out of the box without preloading applications. Or am I wrong?

I am thinking about the case when preloading is unnecessary since it is already approved for specs. From docs I can see that it preloads only the general configuration without application code.

jodosha:
@fidothe https://github.com/lotus/lotus/commit/9ac7a9ed9d66d0ed47e05d170c34415fbae6d092 With this commit we add `require_relative` automatically.

A developer will be able to test an action or a view in isolation. Technically speaking, it will prepare all the applications (subclasses of `Lotus::Application`), but the application code (actions, views) won't be loaded. The `require_relative` directive only loads the isolated action/view. This will guarantee a really fast startup time and a better TDD workflow.

#### Minitest

```ruby
% bundle exec ruby -Ispec spec/web/controllers/dashboard/index_spec.rb
```

#### RSpec

```ruby
% bundle exec rspec spec/web/controllers/dashboard/index_spec.rb
```

jodosha:
@smt116 I recognize your need as something that we have to address. The workflow described above is something really valuable to keep.

Would be a solution to generate another file that does what you need?

```ruby
# config/loader.rb
require_relative './environment'

Lotus::Application.preload_applications!
```

If required, it will load everything: frameworks + application code.

```shell
% bundle exec sidekiq -C config/sidekiq.yml -r./config/loader.rb
```

What do you think? Can you please test and let me know if it works? Thank you!

smt116:
@jodosha It obviously works fine. Can we put it in a framework? should I prepare a pull request with the code?

jodosha:
@smt116 Yes please.

smt116:
Updated.

AlfonsoUceda:
@jodosha @smt116 any news about this?

smt116:
It is ready to merge from my point of view ;)

AlfonsoUceda:
@smt116 can you rebase and resolve conflicts? thanks ;)

smt116:
@AlfonsoUceda done. Sorry for the late response ;)

lib/lotus/generators/application/container.rb
```diff
@@ -37,6 +37,7 @@ def start
             'Gemfile.tt'               => 'Gemfile',
             'config.ru.tt'             => 'config.ru',
             'config/environment.rb.tt' => 'config/environment.rb',
+            'config/loader.rb.tt'      => 'config/loader.rb',
```
brennovich: Maybe would be a good idea to use the same naming as Rails :x

What about `config/boot.rb`?

jodosha:
Closing in favor of Rake tasks support (Ref: #421)

PR #183, opened 2015-03-15
Use modules instead of classes for all repositories

janlelis:
See https://github.com/lotus/model/pull/162


jodosha:
Closing ([explanation](https://github.com/lotus/model/pull/162#issuecomment-81542240)).

Issue #192, opened 2015-03-27
Destroy command


sungwoncho:
Taker: @sungwoncho 

I would like to suggest a destroy command that can undo generation.

It would read something like:
`bundle exec lotus action destroy web home#index`

Any thoughts?

joneslee85:
@sungwoncho we love it, why don't you give it a go?

sungwoncho:
@joneslee85 Okay, I will try to work on it as my schedule permits.

jodosha:
@sungwoncho This makes sense. I agree with @joneslee85 : would you love to take care of it? Thank you very much.

joneslee85:
@sungwoncho ping ping :)

sungwoncho:
@joneslee85 I won't be able to work on this for two weeks tops. Sorry. If anyone wants to take over while I am not working on this, please feel free. I'll update you when I resume my work on this, if no one takes over in two weeks.

joneslee85:
@sungwoncho thanks for the update.

joneslee85:
@sungwoncho ping :)

sungwoncho:
Working on this this weekend!

sungwoncho:
@joneslee85 I have too much on my plate at the moment. Can't promise any progress any time sooner. Feel free to take over.

joneslee85:
@mymai91 @DonDinhIT can you two take care and wrap up Sung's work? Much thanks

mymai91:
yes, we can.

mymai91:
@joneslee85 I see in the Lotus latest version it has two options generator one for application generator, one for model generator. I guess I'll handle destroy for application and model, right? If not help me make it clearer, please.
Thanks.

joneslee85:
@mymai91 this is to destroy action only

mymai91:
@jodosha: What do you think about grouping under 1 sub-command `action`, let’s me give you examples:
```
lotus action generate
lotus action destroy
```

joneslee85:
@jodosha I'd cast a vote to group under subcommand `action`, I kinda like this pattern.

t4deu:
Hi everyone,

This can be done with a more simple way, since Thor is used by the generators, just need to change the Thor behavior before calling the already existing generators. It even removes settings added to existing files(eg: routes.rb). 

I create a PR with the changes for destroy actions, models and migrations.

PR #194, opened 2015-04-03
Add Destroy command

sungwoncho:
This is what I have so far on #192. 

It works, but does not output anything in the console as `generate` command does. Any suggestions?

### DIFF
--- a/lib/lotus/commands/destroy.rb
+++ b/lib/lotus/commands/destroy.rb
@@ -1,0 +1,73 @@
+module Lotus
+  module Commands
+    class Destroy
+
+      ACTION_SEPARATOR = /\/|\#/
+
+      attr_reader :name, :app_name
+
+      def initialize(type, app_name, name, env, cli)
+        @cli      = cli
+        @options  = env.to_options.merge(cli.options)
+
+        @app_name = app_name
+
+        @name     = name
+        @type     = type
+      end
+
+      def start
+        target_file_paths.each do |target_file_path|
+          ::Dir.glob("#{target_file_path}.*").each { |file| ::File.delete(file) }
+        end
+      end
+
+      def app_root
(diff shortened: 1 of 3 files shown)


joneslee85:
@sungwoncho yes, i am happy with it outputing nothing atm. If you like, you could open a new PR to add in output text for generators/destroyers

lib/lotus/commands/destroy.rb
```diff
@@ -0,0 +1,73 @@
...
+        @app_root ||= Pathname.new([@options[:path], app_name].join(::File::SEPARATOR))
+      end
+
+      def spec_root
+        @spec_root ||= Pathname.new('spec')
+      end
+
+      private
+
+        def controller_name
```
joneslee85: our code style is not to indenting private methods

lib/lotus/commands/destroy.rb
```diff
@@ -0,0 +1,73 @@
+module Lotus
+  module Commands
+    class Destroy
+
+      ACTION_SEPARATOR = /\/|\#/
+
+      attr_reader :name, :app_name
+
+      def initialize(type, app_name, name, env, cli)
```
joneslee85: ideally, we would have many versions of destroyers, can we replicate same pattern like those of generators, that is having `destroyers/action.rb`

sungwoncho: Good point. I will improve the PR using your suggestions.

joneslee85:
@sungwoncho ping

jodosha:
@sungwoncho Thanks for this PR. This makes sense, but before to merge it, we need to expand the scope of `destroy`, by including specular logic for `generate`. For instance, remove generated models.

Would you love to work on this? Thank you. :smile:

joneslee85:
@t4deu please create a new PR

t4deu:
Ok, @joneslee85, I will do in this weekend :smile:

PR #197, opened 2015-04-08
Implement Resource generator

nguyenngoc2505:
`lotus generate resource <name_of_resource>` which does similar things like the counter part of Rails

create entity
create repository
create resource route
create all CRUD controller actions
create all view classes
create empty templates for all actions
create tests for actions

For example:
`lotus generate resource posts`

### DIFF
--- a/lib/lotus/commands/generate.rb
+++ b/lib/lotus/commands/generate.rb
@@ -18,11 +18,11 @@ module Lotus
 
       # @since 0.3.0
       # @api private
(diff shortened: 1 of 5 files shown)


joneslee85:
Thanks for your work on this but after discussion with @jodosha, we have decided to drop this generator. The reason behind that is we find it an anti-pattern, for Rails only provides that for your convenience in code generation but it does not solve any usecase. It is recommended for user to use action generator and manually group them into resource by case basis.

PR #203, opened 2015-04-11
Add --skip-view flag to lotus generate action

mvader:
Adding ```--skip-view``` flag when generating a new action prevents the view and the template (and the view spec) from being generated too. This can be useful when developing a JSON api that uses no views or templates.

Example usage:
```bundle exec lotus generate action api users#get --skip-view```

Implements the feature requested in #201

### DIFF
--- a/lib/lotus/cli.rb
+++ b/lib/lotus/cli.rb
@@ -80,13 +80,14 @@ module Lotus
     end
 
     desc 'generate', 'generates a new action'
-    method_option :path,                desc: 'applications path',        type: :string, default: 'apps'
+    method_option :path,                desc: 'applications path',                                         type: :string,  default: 'apps'
+    method_option :skip_view,           desc: 'skip the creation of view and templates (only for action)', type: :boolean, default: false
     method_option :help, aliases: '-h', desc: 'displays the usage method'
 
     # @since 0.3.0
     # @api private
     def generate(type = nil, app_name = nil, name = nil)
-      if options[:help] || (type.nil? && app_name.nil? && name.nil?)
+      if options[:help] || (type.nil? && app_name.nil? && name.nil?) || (options[:skip_view] && type != 'action')
         invoke :help, ['generate']
       else
         require 'lotus/commands/generate'
--- a/lib/lotus/cli.rb
+++ b/lib/lotus/cli.rb
@@ -87,7 +87,7 @@ module Lotus
     # @since 0.3.0
     # @api private
     def generate(type = nil, app_name = nil, name = nil)
-      if options[:help] || (type.nil? && app_name.nil? && name.nil?) || (options[:skip_view] && type != 'action')
+      if options[:help] || (type.nil? && app_name.nil? && name.nil?)
         invoke :help, ['generate']
       else
         require 'lotus/commands/generate'
--- a/lib/lotus/generators/action.rb
+++ b/lib/lotus/generators/action.rb
@@ -48,22 +48,23 @@ module Lotus
           template_path: _template_path,
         }
 
-        templates = {
-          'action.rb.tt' => _action_path,
-          'view.rb.tt'   => _view_path,
-          'template.tt'  => _template_path
-        }
-
         case options[:test]
         when 'rspec'
-          templates.merge!({
-            'action_spec.rspec.tt' => _action_spec_path,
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: lib/lotus/generators/action.rb [hunk 1]
```diff
@@ -48,22 +48,23 @@ def start
...
+          test_type = 'rspec'
+        else
+          test_type = 'minitest'
+        end
+
         templates = {
           'action.rb.tt' => _action_path,
-          'view.rb.tt'   => _view_path,
-          'template.tt'  => _template_path
+          'action_spec.' + test_type + '.tt' => _action_spec_path,
```
joneslee85: What do you think about string interpolation which help prevent exception when `test_type` is nil. that is `"action_spec.#{test_type}.tt"`

joneslee85:
except above comment, it is :+1: from me

lib/lotus/generators/action.rb
joneslee85: my bad, it is impossible not to have `test_type`. please ignore my comment above

lib/lotus/generators/action.rb
```diff
@@ -48,22 +48,23 @@ def start
           template_path: _template_path,
         }
 
+        case options[:test]
```
joneslee85: what do you think about this style instead:

```
test_type = case options[:test]
                   when 'rspec'
                     'rspec'
                   else
                     'minitest'
                   end
```

lib/lotus/generators/action.rb
mvader: I changed it anyway to string interpolation, it's easier to read and more elegant than concatenating the strings in the hash key

lib/lotus/generators/action.rb
mvader: Sure! I didn't even know that this could be done in Ruby.

lib/lotus/generators/action.rb
```diff
@@ -48,22 +48,23 @@ def start
           template_path: _template_path,
         }
 
+        test_type = case options[:test]
+        when 'rspec'
```
joneslee85: please align `when` to be on the same level with `case`

lib/lotus/cli.rb
```diff
@@ -80,13 +80,14 @@ def new(name = nil)
...
-    method_option :path,                desc: 'applications path',        type: :string, default: 'apps'
+    method_option :path,                desc: 'applications path',                                         type: :string,  default: 'apps'
+    method_option :skip_view,           desc: 'skip the creation of view and templates (only for action)', type: :boolean, default: false
     method_option :help, aliases: '-h', desc: 'displays the usage method'
 
     # @since 0.3.0
     # @api private
     def generate(type = nil, app_name = nil, name = nil)
-      if options[:help] || (type.nil? && app_name.nil? && name.nil?)
+      if options[:help] || (type.nil? && app_name.nil? && name.nil?) || (options[:skip_view] && type != 'action')
```
joneslee85: thanks for adding the `type` check here. Come to think of it, there will be lots of generators, and I'd prefer to let `Lotus::Commands::Generate` throws exception if type is invalid and thus can be caught by CLI. For now, please remove it as it is not related to this PR. I welcome another PR if you are keen to take on type check.

mvader:
@joneslee85 done :)

AlfonsoUceda:
@joneslee85 @mvader one thing about this. If you skip the view, then you have to force "self.body = something" in the action because it will produce an error, isn't it?

jodosha:
@AlfonsoUceda Good point. :+1: This will make the dev experience more pleasant by: 1. avoiding a `Rack::Lint` error 2. giving a hint on where to set the body response.

@mvader Can you please enhance with a new PR? Thank you very much.

mvader:
@jodosha @AlfonsoUceda What solution would you prefer: forcing self.body to something in a special action template when there is no view or creating the view and just not creating the template?

I mean, I personally use views to render my JSON API since the json formatting shouldn't be a responsibility of the action but maybe people want to use self.body.

I could implement both if you like it. For example, ```--skip-view``` would use the self.body solution and ````--skip-template``` would just not create the template.

What do you guys think?

jodosha:
@mvader I'd say `self.body = "OK"`. Simple enough.

mvader:
@jodosha cool, I'll make a PR later today with this.

PR #206, opened 2015-04-13
Add self.body to actions generated with --skip-view flag

mvader:
Implemented the changes proposed in #203 after the merge.

### DIFF
--- a/test/commands/generate_test.rb
+++ b/test/commands/generate_test.rb
@@ -59,6 +59,7 @@ describe Lotus::Commands::Generate do
           content.must_match %(  class Index)
           content.must_match %(    include Web::Action)
           content.must_match %(    def call(params))
+          content.wont_match %(      self.body = 'OK')
         end
       end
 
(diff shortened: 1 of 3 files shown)


joneslee85:
:+1:

test/commands/generate_test.rb
```diff
@@ -59,6 +59,7 @@ def default_options
           content.must_match %(  class Index)
           content.must_match %(    include Web::Action)
           content.must_match %(    def call(params))
+          content.wont_match %(      self.body = 'OK')
```
AlfonsoUceda: mmm one thing, in this test `--skip-view` option isn't used, isn't it?

mvader: @AlfonsoUceda nope, that's why I test that self.body is not present in the resultant action. Note the ```wont_match```

AlfonsoUceda: Oh you are rigth, I didn't see 'wont_match' sorry >_<

PR #212, opened 2015-04-20
used attribute accessor instead of reader and writer

ahmetkapikiran:
This pull request contains a little changes on creating getter/setter for class


jodosha:
@ahmetkapikiran Thanks for this PR, but there is a reason why they are separated. The reader is marked as public API, while the writer is private. Merging them with `attr_accessor`, makes hard to distinguish between them.

PR #213, opened 2015-04-21
Introduce Lotus.root

LeoTeam:


test/root_test.rb
```diff
@@ -0,0 +1,10 @@
+require 'test_helper'
+require 'lotus/root'
+
+describe 'Lotus.root' do
```
joneslee85: I think you should do

```
  describe `Lotus` do
    describe `.root` do
    end
  end
```

test/root_test.rb
```diff
@@ -0,0 +1,10 @@
...
+  let(:pathname) {Lotus.root}
+  it 'return root path' do
```
joneslee85: s/return/returns

test/root_test.rb
```diff
@@ -0,0 +1,10 @@
+require 'test_helper'
+require 'lotus/root'
+
+describe 'Lotus.root' do
+  let(:pathname) {Lotus.root}
```
joneslee85: please add empty space after `{`

lib/lotus/root.rb
```diff
@@ -0,0 +1,7 @@
+require 'pathname'
+
+module Lotus
+  def self.root
+    Lotus::Configuration.new.root
```
AlfonsoUceda: IMO, we shouldn't create a new instance of configuration, we have to take the current cofiguration

joneslee85: I don't the above implementation is correct though. What you did above is to retrieve the root value of an app (aka slice), what we wanted here is the root level of the container ( the directory that `config.ru` reside )

What I'd like to suggest is to look into `Lotus::Loader`, there you can work out that we should introduce a root setter for `configuration` perhaps? Let me know if you need any assistant on this ticket.

AlfonsoUceda: Then you have lotus container, https://github.com/lotus/lotus/blob/master/lib/lotus/container.rb inside you have access to configuration

joneslee85:
```
module Lotus
  def self.root
    @root
  end

  def self.root=(path)
    @root = path
  end
end

module Lotus
  class Configuration

    def load!(namespace = nil)
      @namespace = namespace
      evaluate_configurations!

      # set Lotus.root
      if !Lotus.root
        Lotus.root = self.root
      end

      self
    end
  end
end
```

jodosha:
@LeoTeam @mymai91 This approach is wrong :cry:
You use internally `Environment#root`. If you read its API docs it says: 

> It defaults to the current working directory.
   Lotus assumes that all the commands are executed from there.

That means, it delegates to the current directory, **assuming** that all the commands are ran from there.
**The goal of this feature, is to remove that assumption.**

How to manually test it:

```shell
lotus new bookshelf --lotus-head=true
cd bookshelf
```

Edit `Gemfile` with:

```ruby
gem 'lotusrb',                           github: 'LeoTeam/lotus', branch: 'lotus-root'
```

Then

```shell
bundle
cd apps/web
bundle console
```

Test it:

```ruby
irb(main):001:0> Lotus.root
=> #<Pathname:/Users/luca/Desktop/bookshelf/apps/web>
```

It returns `/Users/luca/Desktop/bookshelf/apps/web`, but it should be `/Users/luca/Desktop/bookshelf`.

/cc @joneslee85

jodosha:
@LeoTeam @mymai91 Can you please implement it here? https://github.com/lotus/lotus/blob/master/lib/lotus.rb Thank you!

/cc @joneslee85

jodosha:
@LeoTeam Thank you. Both me and @joneslee85 agree to rely on `Pathname.pwd` for now. Merging this :space_invader:

mymai91:
@jodosha @joneslee85 Good to know. Thanks.

PR #214, opened 2015-04-28
[f] Application generator

### LINKED ISSUE #157: Application generator
**Taker**: @hieuk09 

## What

Introduce a new command `lotus generate`.

## Why

Lotus generator (`lotus new`) generates only one application `apps/web`. If developers want to add other apps, this will be the official way to do that.

### CLI

```shell
% lotus generate app admin
```

  * `app` is the subject of what we want to generate
  * `admin` is the name of the application under `apps/`
  * Optional `--application-base-url`. Defaults to the app name.
  * Optional `--test`. Defaults to the contents of `.lotusrc` or to `minitest`, if that file is missing.

### Output

  * It will generate a Lotus application with the given name: `apps/admin`.
  * Amend `config/environment.rb`

### Files

```ruby
# config/environment.rb
require 'rubygems'
require 'bundler/setup'
require 'lotus/setup'
require_relative '../apps/admin/application' # NEW LINE

Lotus::Container.configure do
  #
  # !!! ATTENTION !!! New applications should be put at the top of this block
  #
  mount Admin::Application, at: '/admin' # NEW LINE
end
```

hieuk09:
[Close #157]

### DIFF
--- a/lib/lotus/commands/generate.rb
+++ b/lib/lotus/commands/generate.rb
@@ -10,6 +10,8 @@ module Lotus
       # @since 0.3.0
       # @api private
       GENERATORS_NAMESPACE = "Lotus::Generators::%s".freeze
+      APP = 'app'.freeze
+      SLICE_TYPE = 'slice'.freeze
 
       # @since 0.3.0
       # @api private
@@ -24,16 +26,15 @@ module Lotus
       # @api private
       def initialize(type, app_name, name, env, cli)
         @cli      = cli
-        @options  = env.to_options.merge(cli.options)
-
-        @app_name = app_name
-        @app      = Utils::String.new(@app_name).classify
-
         @name     = name
-        @type     = type
 
+        @type = sanitize_type(type)
+        @app_name = app_name
         @source   = Pathname.new(::File.dirname(__FILE__) + "/../generators/#{ @type }/").realpath
         @target   = Pathname.pwd.realpath
+
+        @app      = Utils::String.new(@app_name).classify
+        @options  = sanitize_app_name_options(app_name).merge(env.to_options.merge(cli.options))
       end
 
       # @since 0.3.0
@@ -71,6 +72,21 @@ module Lotus
         require "lotus/generators/#{ @type }"
         class_name = Utils::String.new(@type).classify
         Utils::Class.load!(GENERATORS_NAMESPACE % class_name).new(self)
+      end
+
+      # @since x.x.x
+      # @api private
+      def sanitize_app_name_options(app_name)
+        {
+          application: app_name,
+          application_base_url: "/#{app_name}"
+        }
+      end
+
+      # @since x.x.x
+      # @api private
+      def sanitize_type(type)
+        type == APP ? SLICE_TYPE : type
       end
     end
   end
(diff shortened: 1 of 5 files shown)


lib/lotus/commands/generate.rb
```diff
@@ -24,16 +24,21 @@ class Error < ::StandardError
...
         @cli      = cli
-        @options  = env.to_options.merge(cli.options)
-
-        @app_name = app_name
-        @app      = Utils::String.new(@app_name).classify
-
         @name     = name
-        @type     = type
 
+        if type == 'app'
```
nguyenngoc2505: Please remove this

test/integration/cli/generate_test.rb
```diff
@@ -50,20 +55,12 @@ def chdir_to_root
       generate_application
       generate_action
       generate_action_without_view
-      generate_model
```
nguyenngoc2505: why do you remove this test function?

lib/lotus/commands/generate.rb
hieuk09: Fixed, thanks.

lib/lotus/commands/generate.rb
```diff
@@ -72,6 +71,21 @@ def generator
...
+        {
+          application: app_name,
+          application_base_url: "/#{app_name}"
+        }
+      end
+
+      # @since x.x.x
+      # @api private
+      def sanitize_type(type)
+        type == 'app' ? 'slice' : type
```
AlfonsoUceda: can you extract these strings to constants?
```ruby
APP = 'app'.freeze
SLICE_TYPE = 'slice'.freeze
```
something like that

AlfonsoUceda:
:+1:

lib/lotus/commands/generate.rb
hieuk09: Fixed, thanks.

lib/lotus/commands/generate.rb
```diff
@@ -10,6 +10,8 @@ class Generate
       # @since 0.3.0
       # @api private
       GENERATORS_NAMESPACE = "Lotus::Generators::%s".freeze
+      APP = 'app'.freeze
+      SLICE_TYPE = 'slice'.freeze
```
joneslee85: @jodosha are you okay with this term?

joneslee85:
:+1:

lib/lotus/commands/generate.rb
jodosha: @joneslee85 yup :+1:

jodosha:
@hieuk09 Thanks for this PR, we're on the right way.

We should remove `test` CLI option. This decision should be taken by looking at `.lotusrc`. Please note that `Lotus::Environment.new.to_options` will lookup that file for us.

Also, can you please change the name of the generated application from `api` to `admin`? The reason is that in the near future we will be able to do `lotus generate api my_api`, so the name of the fixtures will not be confusing.

Thank you! /cc @joneslee85

hieuk09:
Thanks for your comments, @jodosha. I'll update the PR accordingly.

hieuk09:
@jodosha: I fixed the PR as your comments.

joneslee85:
:+1: nice one

jodosha:
@hieuk09 Thank you very much. :confetti_ball: Merging it! :+1:

PR #216, opened 2015-05-06
issue#211 Add Lotus.env

nguyenngoc2505:
- Implement `Lotus.env`
```ruby
   Lotus.env # => "development"

  Lotus.env?(:development) # => true 
  Lotus.env?(:production)  # => false
```

### DIFF
--- a/lib/lotus.rb
+++ b/lib/lotus.rb
@@ -2,6 +2,7 @@ require 'lotus/version'
 require 'lotus/application'
 require 'lotus/container'
 require 'lotus/logger'
+require 'lotus/environment'
 
 # A complete web framework for Ruby
 #
@@ -9,4 +10,22 @@ require 'lotus/logger'
 #
 # @see http://lotusrb.org
 module Lotus
+
+  # Return the current environment
+  # @since x.x.x
+  #
+  # @example Lotus.env => "development"
+
+  def self.env
+    Environment.new.environment
+  end
+
+  # Check to see if specified environment matches the current environment
+  # @since x.x.x
+  #
+  # @example Lotus.env?(:development) => true
+
+  def self.env?(*names)
+    Environment.new.environment?(*names)
+  end
 end
--- a/lib/lotus/environment.rb
+++ b/lib/lotus/environment.rb
@@ -180,6 +180,13 @@ module Lotus
       @environment ||= ENV[LOTUS_ENV] || ENV[RACK_ENV] || DEFAULT_ENV
     end
 
+    # @since x.x.x
+    #
+    # @see Lotus.env?(name)
+    def environment?(*names)
+      names.map(&:to_s).include?(environment)
+    end
+
     # A set of Bundler groups
     #
     # @return [Array] A set of groups
--- a/test/env_test.rb
+++ b/test/env_test.rb
@@ -1,0 +1,87 @@
+require 'test_helper'
+
+describe 'Lotus' do
+  before do
+    ENV['LOTUS_ENV'] = 'test'
+  end
+
+  describe '.env' do
+    it 'returns environment name' do
+      Lotus.env.must_equal 'test'
+    end
+  end
+
+  describe '.env?' do
+    describe 'when environment is matched' do
+      describe 'when single name' do
+        describe 'when environment var is symbol' do
+          it 'returns true' do
+            Lotus.env?(:test).must_equal true
+          end
+        end
+        describe 'when environment var is string' do
+          it 'returns true' do
+            Lotus.env?("test").must_equal true
+          end
+        end
+      end
+
+      describe 'when multiple names' do
+        describe 'when environment vars are symbol' do
+          it 'returns true' do
+            Lotus.env?(:development, :test, :production).must_equal true
+          end
+        end
+        describe 'when environment vars are string' do
+          it 'returns true' do
+            Lotus.env?("development", "test", "production").must_equal true
+          end
+        end
+
+        describe 'when environment vars include string and symbol' do
+          it 'returns true' do
+            Lotus.env?(:development, "test", "production").must_equal true
+          end
+        end
+      end
+    end
+
+    describe 'when environment is not matched' do
+      before do
+        ENV['LOTUS_ENV'] = 'development'
+      end
+
+      describe 'when single name' do
+        describe 'when environment var is symbol' do
+          it 'returns false' do
+            Lotus.env?(:test).must_equal false
+          end
+        end
+        describe 'when environment var is string' do
+          it 'returns false' do
+            Lotus.env?("test").must_equal false
+          end
+        end
+      end
+
+      describe 'when multiple names' do
+        describe 'when environment vars are symbol' do
+          it 'returns false' do
+            Lotus.env?(:test, :production).must_equal false
+          end
+        end
+        describe 'when environment vars are string' do
+          it 'returns false' do
+            Lotus.env?("test", "production").must_equal false
+          end
+        end
+
+        describe 'when environment vars include string and symbol' do
+          it 'returns false' do
+            Lotus.env?(:test, "production").must_equal false
+          end
+        end
+      end
+    end
+  end
+end
--- a/test/environment_test.rb
+++ b/test/environment_test.rb
@@ -118,6 +118,86 @@ describe Lotus::Environment do
     end
   end
 
+  describe '#environment?' do
+    describe 'when environment is matched' do
+      before do
+        ENV['LOTUS_ENV'] = 'test'
+        @env = Lotus::Environment.new
+      end
+
+      describe 'when single name' do
+        describe 'when environment var is symbol' do
+          it 'returns true' do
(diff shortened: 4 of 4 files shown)


lib/lotus/env.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  # @example env = Lotus.env
+  #   # => #<Lotus::Env:0x000000024e56b8 @string="development">
+  #   # env.to_s
+  #     # "development"
+  #
+  #   # env.development?
+  #     # true
+
+  class Env < Utils::String
```
joneslee85: naming is hard, what do you think about renaming it to `EnvironmentInqurier` instead? I find `Env` vs `Environment` can cause confusion,

nguyenngoc2505: It sound good. Thanks you!

lib/lotus/environment_inqurier.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  # @example env = Lotus.env
+  #   # => #<Lotus::Env:0x000000024e56b8 @string="development">
+  #   # env.to_s
+  #     # "development"
+  #
+  #   # env.development?
+  #     # true
+
+  class EnvironmentInqurier < Utils::String
```
joneslee85: sorry I made a typo, it's supposed to be Inquirer not Inqurier

lib/lotus.rb
```diff
@@ -9,4 +11,9 @@
 #
 # @see http://lotusrb.org
 module Lotus
+  # @since x.x.x
+
+  def self.env
+    @env ||= EnvironmentInqurier.new(Environment.new.environment)
```
jodosha: @nguyenngoc2505 This isn't thread safe. Can you please remove it the `@env ||=` part? Thank you :smile_cat:

lib/lotus/environment_inqurier.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def production?
+      @string == PRODUCTION_ENV
+    end
+
+    def test?
+      @string == TEST_ENV
+    end
+
+    def method_missing(m, *args, &blk)
```
jodosha: binding `*args` and `&blk` isn't necessary as we don't use them.

jodosha:
@nguyenngoc2505 I think string inquirer isn't necessary at all. We can have something like this:

```ruby
Lotus.env?(:production)
```

```ruby
module Lotus
  self.env?(name)
     # NOTE: Lotus::Environment doesn't have #environment? yet
     Lotus::Environment.new.environment?(name)
  end
end
```

nguyenngoc2505:
@jodosha Thanks for your comment, I've updated it.

lib/lotus/environment.rb
```diff
@@ -180,6 +180,13 @@ def environment
       @environment ||= ENV[LOTUS_ENV] || ENV[RACK_ENV] || DEFAULT_ENV
     end
 
+    # @since x.x.x
+    #
+    # @see Lotus.env?(name)
+    def environment?(name)
```
joneslee85: can you write a test?

test/env_test.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe 'Lotus' do
+
+  describe '.env' do
+    it 'returns env object' do
+      Lotus.env.must_be_kind_of String
+    end
+  end
+
+  describe '.env.test?' do
```
joneslee85: I think this should be `env?`

test/env_test.rb
```diff
@@ -0,0 +1,20 @@
...
+    before do
+      ENV['LOTUS_ENV']  = 'test'
+    end
+
+    it 'return true' do
+      Lotus.env?(:test).must_equal true
```
joneslee85: can we also test for false scenario?

test/env_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_helper'
+
+describe 'Lotus' do
+
+  describe '.env' do
+    it 'returns env object' do
+      Lotus.env.must_be_kind_of String
```
joneslee85: can we assert that it returns `test` here?

lib/lotus/environment.rb
nguyenngoc2505: I've written. Thanks!

test/environment_test.rb
```diff
@@ -118,6 +118,32 @@
     end
   end
 
+  describe '#environment?' do
+    describe 'environment?(:test)' do
```
joneslee85: What do you think about removing this line and change the 2 context description line to `when environment is matched` and `when environment is not matched` instead?

test/env_test.rb
```diff
@@ -0,0 +1,33 @@
+require 'test_helper'
+
+describe 'Lotus' do
+  before do
+    ENV['LOTUS_ENV'] = 'test'
+  end
+
+  describe '.env' do
+    it 'returns env object' do
+      Lotus.env.must_be_kind_of String
```
joneslee85: IMO, we do not need this assertion, it does not add any value

test/env_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    it 'return "test"' do
```
joneslee85: s/return/returns

test/env_test.rb
```diff
@@ -0,0 +1,33 @@
...
+      Lotus.env.must_equal 'test'
+    end
+  end
+
+  describe '.env?(:test)' do
```
joneslee85: what do you think about changing it to `.env?` and adding a context line next to it `when environment is matched`?

test/env_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    it 'return true' do
```
joneslee85: s/return/returns

test/env_test.rb
```diff
@@ -0,0 +1,33 @@
...
+      Lotus.env?(:test).must_equal true
+    end
+
+    describe 'when environment is not "test"' do
+      before do
+        ENV['LOTUS_ENV'] = 'development'
+      end
+
+      it 'return false' do
```
joneslee85: s/return/returns

lib/lotus.rb
```diff
@@ -9,4 +10,18 @@
 #
 # @see http://lotusrb.org
 module Lotus
+
+  # @since x.x.x
+  #
```
joneslee85: What do you think about add description for this method, something like `Return the current environment`

lib/lotus.rb
```diff
@@ -9,4 +10,18 @@
...
+  # @example Lotus.env => "development"
+  def self.env
+    Environment.new.environment
+  end
+
+  # @since x.x.x
```
joneslee85: What do you think about adding description for this method? e.g.: `Check to see if specified environment matches the current environment`

lib/lotus.rb
nguyenngoc2505: Added. Thanks!

joneslee85: Should be on top of the since

test/env_test.rb
```diff
@@ -0,0 +1,31 @@
+require 'test_helper'
+
+describe 'Lotus' do
+  before do
+    ENV['LOTUS_ENV'] = 'test'
+  end
+
+  describe '.env' do
+    it 'returns "test"' do
```
joneslee85: What do you think about changing this to "returns environment name"?

lib/lotus.rb
joneslee85: Move the description to top of other lines

test/env_test.rb
nguyenngoc2505: For more detail I think it will be `returns curent environment name`

jodosha:
@nguyenngoc2505 This looks great, thank you :clap: :sparkles: 

May I ask to extra bonus points?

  1. Please test `Lotus.env?` and `Lotus::Environment#environment?` with strings, not just symbols.
  2. What do you think if we let `Lotus.env?` and `Lotus::Environment#environment?` to accept multiple names? It would be handy to check multiple envs at once (see the example below).

### Multiple envs check

Imagine this scenario

```ruby
if Lotus.env?(:staging) || Lotus.env?(:production)
  # do something
end
```

We can provide a shorter version

```
if Lotus.env?(:staging, :production)
  # do something
end
```

Please note that single name should still work.

joneslee85:
@jodosha great suggestion
@nguyenngoc2505 hint: please look into how to use the single splat parameter

nguyenngoc2505:
@jodosha , @joneslee85 Thanks for your suggestion. And #1, Do we accept or reject the `strings` variable? I think we accept the `strings` variable is good.

joneslee85:
@nguyenngoc2505 we would accept String

nguyenngoc2505:
@jodosha, @joneslee85 I've fixed this PR with your suggestion.

joneslee85:
:+1:

PR #219, opened 2015-05-12
Added routing helpers to actions

AlfonsoUceda:
With this PR you can use routing helpers inside actions.

```ruby
module Collaboration::Controllers::RedirectedRoutes
  class Index
    include Collaboration::Action

    def call(params)
      redirect_to routes.legacy_path
    end
  end
end
```

### DIFF
--- a/lib/lotus/loader.rb
+++ b/lib/lotus/loader.rb
@@ -6,6 +6,7 @@ require 'lotus/routing/default'
 require 'lotus/action/cookies'
 require 'lotus/action/session'
 require 'lotus/config/security'
+require 'lotus/action/routing_helpers'
 
 module Lotus
   # Load an application
@@ -60,6 +61,7 @@ module Lotus
             cookies config.cookies.default_options
           end
           prepare { include Lotus::Action::Session } if config.sessions.enabled?
+          prepare { include Lotus::Action::RoutingHelpers }
 
           config.controller.__apply(self)
         end
--- a/test/integration/full_stack_test.rb
+++ b/test/integration/full_stack_test.rb
@@ -233,5 +233,11 @@ describe 'A full stack Lotus application' do
     end
   end
 
+  it 'test route helpers in actions' do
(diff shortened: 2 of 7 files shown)


jodosha:
@AlfonsoUceda Please grep the codebase and update fixtures [like this](https://github.com/lotus/lotus/blob/master/test/fixtures/collaboration/apps/web/app/controllers/books.rb#L58):

```ruby
# from
redirect_to Collaboration::Routes.url(:book, id: @book.id)

# to
redirect_to routes.book_url(id: @book.id)
```

AlfonsoUceda:
@jodosha done!

joneslee85:
@AlfonsoUceda :+1:

test/integration/full_stack_test.rb
```diff
@@ -233,5 +233,11 @@ def response
     end
   end
 
+  it 'test route helpers in actions' do
+    get '/redirected_routes'
+
+    response.status.must_equal 302
+    response.headers["Location"].must_equal "/action_legacy"
```
joneslee85: _ntipick_ why not go with single quote?

lib/lotus/loader.rb
```diff
@@ -60,6 +61,7 @@ def _configure_controller_framework!
             cookies config.cookies.default_options
           end
           prepare { include Lotus::Action::Session } if config.sessions.enabled?
+          prepare { include Lotus::Action::RoutingHelpers }
```
joneslee85: @AlfonsoUceda what do you think about including this helper in view framework too?

AlfonsoUceda: @joneslee85 in this line, routing helpers are loaded for views https://github.com/lotus/lotus/blob/master/lib/lotus/generators/slice/application.rb.tt#L209

jodosha:
@AlfonsoUceda We're doing great! Lotus is getting better and better. Thank you :clap:

Issue #221, opened 2015-05-14
Cookies improvements


jodosha:
We should improve Cookies settings.

  - [ ]  Automatic `:secure` option
  - [ ]  Enhance options syntax

## Automatic :secure option

Alongside with `:httponly` we should improve security, by enabling `:secure` option for cookies.
The value for this option should be set by looking at `Configuration#ssl?`.
If we are in production using HTTPS, `#ssl?` will return `true`, and we set `:secure` on `true`.
If we are in development mode, `#ssl?` will be probably `false`, and then `:secure` will be `false` as well.

## Enhance options syntax

Right now we can enable cookies by passing `true` and setting options, by passing a Hash as second argument:

```ruby
cookies true
# or
cookies true, max_age: 300
```

The last syntax seems a little bit odd, we should improve it like this:

```ruby
cookies max_age: 300
```

#### Tech note

**Right now, this proposed syntax enables cookies, but it doesn't set options.**
The reason is that `Config::Cookies#enabled?` coerces the first argument to a boolean, but will then ignore the second one.

AlfonsoUceda:
@jodosha about enhanced options syntax, then will cookies be enabled only if options are passed?

PR #222, opened 2015-05-16
Add automatic :secure option

AlfonsoUceda:
close task #221

In an application with `https` scheme, `:secure` option for cookies is enabled by default
```ruby
require 'lotus'

module Bookshelf
    configure do
      cookies true

      scheme 'https'
    end
  end
end

Bookshelf::Application.configuration.cookies
# => #<Lotus::Config::Cookies:0x0000000329f880 @enabled=true, @default_options={:httponly=>true, :secure=>true}>
```
In an application with `http` scheme, `:secure` option for cookies is disabled by default
```ruby
require 'lotus'

module Bookshelf
    configure do
      cookies true
    end
  end
end

Bookshelf::Application.configuration.cookies
# => #<Lotus::Config::Cookies:0x0000000329f880 @enabled=true, @default_options={:httponly=>true, :secure=>false}>
```

### DIFF
--- a/lib/lotus/config/cookies.rb
+++ b/lib/lotus/config/cookies.rb
@@ -22,15 +22,16 @@ module Lotus
       #
       # @param enabled [TrueClass, FalseClass] enable cookies
       # @param options [Hash] optional cookies options
+      # @param configuration [Lotus::Configuration] the application configuration
       #
       # @since 0.3.0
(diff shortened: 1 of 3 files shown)


jodosha:
@AlfonsoUceda Another great PR, thank you! :clap:

PR #223, opened 2015-05-17
Fixed action generator, raise exception if action name isn't supplied

AlfonsoUceda:
# Problem
Generate an action without action's name generate wrong files
`lotus generate action admin users`
```
      insert  apps/admin/config/routes.rb
      create  spec/admin/controllers/users/_spec.rb
      create  apps/admin/controllers/users.rb
      create  apps/admin/views/users.rb
      create  apps/admin/templates/users/.html.erb
      create  spec/admin/views/users/_spec.rb
```
```ruby
module Admin::Controllers::Users
  class  # => Note this class without name, break code
    include Admin::Action

    def call(params)
    end
  end
end
```
# Solution

Raise an exception if action's name isn't supplied:
```
lotus generate action admin users
# => Unknown action, please add action's name with this syntax controller_name#action_name
```

### DIFF
--- a/lib/lotus/generators/action.rb
+++ b/lib/lotus/generators/action.rb
@@ -38,6 +38,7 @@ module Lotus
       # @api private
       def start
         assert_existing_app!
+        assert_action!
 
         opts = {
           app:           app,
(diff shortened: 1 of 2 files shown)


joneslee85:
:+1:

jodosha:
@AlfonsoUceda Merged this, thank you! :cat2:

PR #227, opened 2015-05-19
Enhanced cookies syntax options

### LINKED ISSUE #221: Cookies improvements

AlfonsoUceda:
closes #221 

This PR enhances cookies syntax options.

Now cookies are activated with this syntax:
```ruby
cookies true
# or
cookies max_age: 300
```

### DIFF
--- a/lib/lotus/config/cookies.rb
+++ b/lib/lotus/config/cookies.rb
@@ -20,17 +20,37 @@ module Lotus
       # Prevent attackers to steal cookies via JavaScript,
       # Eg. alert(document.cookie) will fail
       #
-      # @param enabled [TrueClass, FalseClass] enable cookies
-      # @param options [Hash] optional cookies options
+      # @param options [Hash, TrueClass, FalseClass] optional cookies options
       #
-      # @since 0.3.0
+      # @since x.x.x
       # @api private
       #
       # @see https://github.com/rack/rack/blob/master/lib/rack/utils.rb #set_cookie_header!
       # @see https://www.owasp.org/index.php/HttpOnly
-      def initialize(enabled = false, options = {})
-        @enabled         = enabled
-        @default_options = { httponly: true }.merge(options)
+      #
+      # @example with boolean option
+      #
+      #  require 'lotus/config/cookies'
+      #
+      #  cookies_config = Lotus::Config::Cookies.new(true)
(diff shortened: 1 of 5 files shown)


lib/lotus/config/cookies.rb
```diff
@@ -20,17 +20,17 @@ class Cookies
...
       #
       # @see https://github.com/rack/rack/blob/master/lib/rack/utils.rb #set_cookie_header!
       # @see https://www.owasp.org/index.php/HttpOnly
-      def initialize(enabled = false, options = {})
-        @enabled         = enabled
-        @default_options = { httponly: true }.merge(options)
+      def initialize(options = {})
+        @options         = options
+        @default_options = { httponly: true }
+        @default_options.merge!(options) if options.is_a? Hash
```
joneslee85: Why do we have to guard this?

AlfonsoUceda: This guard is used because `options` can be true or false. If you know
another approach tell me, I'm not proud with that guard :P

joneslee85: Now I get it, I think it's worth to document that boolean case

AlfonsoUceda: Yes, youare right, I'll update the docs

jodosha:
@AlfonsoUceda This looks great! :clap: Merged. :+1:

PR #230, opened 2015-05-21
Model generator creates wrong RSpec files

nguyenngoc2505:
When model generator create Rspec files it must be:
```ruby
require 'spec_helper'

RSpec.describe User do
  # place your tests here
end
```

### DIFF
--- a/test/commands/generate_test.rb
+++ b/test/commands/generate_test.rb
@@ -248,7 +248,7 @@ describe Lotus::Commands::Generate do
         it 'generates it' do
           content = @root.join('spec/generate/entities/post_spec.rb').read
           content.must_match %(require 'spec_helper')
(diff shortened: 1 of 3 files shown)


test/commands/generate_test.rb
```diff
@@ -248,7 +248,7 @@ def default_options
         it 'generates it' do
           content = @root.join('spec/generate/entities/post_spec.rb').read
           content.must_match %(require 'spec_helper')
-          content.must_match %(describe Post do)
+          content.must_match %(Rspec.describe Post do)
```
jodosha: It's `RSpec`, not `Rspec` :wink:

nguyenngoc2505: Sory, It's my fault. ^^. I've fixed.

PR #238, opened 2015-05-27
Add Platform Control for Code Reloading

Sdogruyol:
Added platform control for code_reloading? environment variable and disable code reloading for JRuby

### DIFF
--- a/lib/lotus/environment.rb
+++ b/lib/lotus/environment.rb
@@ -334,7 +334,13 @@ module Lotus
     # @see Lotus::Commands::Server
     # @see Lotus::Environment::CODE_RELOADING
     def code_reloading?
-      @options.fetch(:code_reloading) { !!CODE_RELOADING[environment] }
+      # JRuby doesn't implement fork that's why shotgun cannot be used.
+      if Utils.jruby?
+        puts "JRuby doesn't support code reloading."
+        false
+      else
+        @options.fetch(:code_reloading) { !!CODE_RELOADING[environment] }
+      end
     end
 
     # Serialize the most relevant settings into a Hash


lib/lotus/environment.rb
```diff
@@ -334,7 +334,13 @@ def env_config
     # @see Lotus::Commands::Server
     # @see Lotus::Environment::CODE_RELOADING
     def code_reloading?
-      @options.fetch(:code_reloading) { !!CODE_RELOADING[environment] }
+      # JRuby doesn't implement fork that's why shotgun cannot be used.
+      if RUBY_PLATFORM == 'java'
```
jodosha: `Utils.jruby?` :wink:

jodosha:
@Sdogruyol Thanks for this PR. Do you think that `lotusrb.gemspec` should be amended accordingly?
Are you aware of Shotgun alternatives that work with JRuby? Thank you.

lib/lotus/environment.rb
Sdogruyol: :+1:

Sdogruyol:
@jodosha you're welcome

Actually there are 2 alternatives.

- There's a tool called Drip for JRuby which is just like Spring for Rails or Shotgun but it has lots of dependencies and maybe hard to put into lotus here's a how to gist for more info on Drip: https://gist.github.com/rwjblue/458291

- The second alternative is that we can check Rails and see how they did implement auto reloading.

jodosha:
@Sdogruyol Lotus itself doesn't implement code reloading and constant autoloading (like Rails). It only leverages on Ruby's `require` instead of monkey patch the core of the language.

To achieve this useful feature in dev mode, we use Shotgun as a _"reloadable"_ container. So it's transparent to Lotus.

The reason is simplicity. Once you remove Shotgun (eg. in production env), you only have plain Ruby mechanisms.

To recap:

  * Drip: We're not looking for a Spring alternative (speedup boot time), but for a Shotgun alternative (code reloading). Does it offers this feature?. That Gist is a 404 for me.

  * Rails code reloading: :-1: :-1: :-1:

Sdogruyol:
@jodosha Here's the working gist: https://gist.github.com/rwjblue/4582914

jodosha:
@Sdogruyol Thank you. It doesn't seem to offer code reloading. Am I right?

Sdogruyol:
@jodosha well it doesn't replace Shotgun directly. 

What about we disable code reloading in JRuby for now? I think firstly we should make Lotus pass all the cases on JRuby 9K.

jodosha:
> What about we disable code reloading in JRuby for now?

:+1:

Sdogruyol:
I'll change it to use ```Utils.jruby?``` and force push :)

AlfonsoUceda:
@Sdogruyol ping

Sdogruyol:
@AlfonsoUceda pong! Sorry for the delay going to push tonight

Sdogruyol:
Done :+1:

AlfonsoUceda:
:+1: cc @jodosha

joneslee85:
To me it is okay :+1:

PR #242, opened 2015-06-03
Generate .env at the root of the app

### LINKED ISSUE #191: Use .env at the root of the application
Quoting a discussion from our chat:

  > @luislavena: any specific reason why .env information is looked up in `config/.env` instead of the root of the repository?
  > @jodosha: Hello, nope, only to keep all the confs under `config/`. If you are thinking about Heroku, please notice that `Lotus::Environment` reads `config/.env` and sets the env vars for you. It uses `Dotenv` internally.
  > @luislavena: thanks, was actually thinking on foreman and tools like azk and others that read repo's root for .env and allow you run one-off commands.

jodosha:
Closes #191 

/cc @lotus/core-team @luislavena


luislavena:
@jodosha great, thank you!

jodosha:
@luislavena Thank you for the good catch :+1:

PR #248, opened 2015-06-04
CSRF protection

### LINKED ISSUE #247: CSRF Protection
The upcoming release of `lotus-helpers` will ship with form helpers. This is the right time to introduce a security defense for one of the most common security threats on the web: [CSRF](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29).

According to OWASP, the strongest mechanism is called [Synchronizer Token Pattern](https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet#General_Recommendation:_Synchronizer_Token_Pattern). It's a secret token stored in session and make it available in the DOM. For each non-GET request that a browser performs, that token should be sent back to the application to compare it with the one in session.

If the match fails, there is a high probability of an attack.

There are a few technical implications for this feature:

  - [x] Enable this feature only if HTTP sessions are enabled
  - [x] Make the token available in DOM (special form hidden field)
  - [x] Offer a standard policy for an attack handling
  - [x] Allow to customize this policy
  - [x] Allow to bypass the security check for an action

jodosha:
### CSRF Protection

This security mechanism is enabled automatically if sessions are turned on.

It stores a "challenge" token in session. For each "state changing request" (eg. `POST`, `PATCH` etc..), we should send a special param: `_csrf_token`.

If the param matches with the challenge token, the flow can continue.
Otherwise the application detects an attack attempt, it reset the session and `Lotus::Action::InvalidCSRFTokenError` is raised.

We can specify a custom handling strategy, by overriding `#handle_invalid_csrf_token`.

Form helper (`#form_for`) automatically sets a hidden field with the correct token. A special view method (`#csrf_token`) is available in case the form markup is manually crafted.

We can disable this check on action basis, by overriding `#verify_csrf_token?`.

### References

  * https://www.owasp.org/index.php/Cross-Site_Request_Forgery_%28CSRF%29
  * https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet

#### Custom Handling

```ruby
module Web::Controllers::Books
  class Create
    include Web::Action

    def call(params)
      # ...
    end

    private

    def handle_invalid_csrf_token
      Web::Logger.warn "CSRF attack: expected #{ session[:_csrf_token] }, was #{ params[:_csrf_token] }"
      # manual handling
    end
  end
end
```

#### Bypass Security Check

```ruby
module Web::Controllers::Books
  class Create
    include Web::Action

    def call(params)
      # ...
    end

    private

    def verify_csrf_token?
      false
    end
  end
end
```

Closes #247

### DIFF
--- a/lib/lotus/action/csrf_protection.rb
+++ b/lib/lotus/action/csrf_protection.rb
@@ -1,0 +1,167 @@
+require 'securerandom'
+
+module Lotus
+  module Action
+    # Invalid CSRF Token
+    #
+    # @since x.x.x
+    class InvalidCSRFTokenError < ::StandardError
+    end
+
+    # CSRF Protection
+    #
+    # This security mechanism is enabled automatically if sessions are turned on.
+    #
+    # It stores a "challenge" token in session. For each "state changing request"
+    # (eg. <tt>POST</tt>, <tt>PATCH</tt> etc..), we should send a special param:
+    # <tt>_csrf_token</tt>.
+    #
+    # If the param matches with the challenge token, the flow can continue.
+    # Otherwise the application detects an attack attempt, it reset the session
+    # and <tt>Lotus::Action::InvalidCSRFTokenError</tt> is raised.
+    #
+    # We can specify a custom handling strategy, by overriding <tt>#handle_invalid_csrf_token</tt>.
+    #
+    # Form helper (<tt>#form_for</tt>) automatically sets a hidden field with the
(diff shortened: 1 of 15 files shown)


AlfonsoUceda:
:tophat: :100: perfect Sir, as far as I know I think it's ok

PR #250, opened 2015-06-06
Added force_ssl option in application and container

### LINKED ISSUE #241: Force SSL
Add a new setting to `Lotus::Configuration` and related generators, to force SSL. See lotus/router#56 for technical details.

```ruby
# apps/web/application.rb
module Web
    configure do
      # ...
      # force_ssl false
    end

    configure :production do
      force_ssl true
    end
  end
end
```

We should allow to specify this option for Container as well:

```ruby
# config/environment.rb
# …
Lotus::Container.configure(force_ssl: ENV['LOTUS_FORCE_SSL']) do
  mount Web::Application, at: ‘/‘
end
```

For backward compatibility, if that option is not specified, this feature must be turned off.

AlfonsoUceda:
Resolves #241

With this PR you can force_ssl in your app or in your container, affecting all apps inside the container.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ end
 
 gem 'lotus-utils',       '~> 0.4', require: false, github: 'lotus/utils',       branch: '0.4.x'
 gem 'lotus-validations', '~> 0.3', require: false, github: 'lotus/validations', branch: '0.3.x'
-gem 'lotus-router',      '~> 0.4', require: false, github: 'lotus/router',      branch: '0.4.x'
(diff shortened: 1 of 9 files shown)


jodosha:
@AlfonsoUceda Thank you for this great work. One last thing is to support `Strict-Transport-Security` header (see https://www.owasp.org/index.php/HTTP_Strict_Transport_Security)

We need a new test in `test/integration/force_ssl_test.rb`: when we receive a request that uses SSL (HTTPS), we should return that header.

That means when we load a `Lotus::Application` via `Loader` and `force_ssl` is true, we need to add a default_header. :wink:

AlfonsoUceda:
@jodosha is this the wanted behaviour?

PR #260, opened 2015-06-16
Removed slice web env database url

AlfonsoUceda:
This PR removes slice database url in .env files.

when you create a new application WEB_DATABASE_URL is no longer generated.


jodosha:
@AlfonsoUceda Thank you :+1:

PR #263, opened 2015-06-19
Fix cannot generate migration error

khaiql:
As you recently introduced the new migration feature, I tried to generate a migration in my test project but unfortunately it got an error:

lotus_error

After a while trying to figure out the cause, I came to a solution as you can see in this PR. I'm not sure whether this is the root cause (cause I know `require`'s parameter doesn't need a file extension), but it works for me anyway. Hope this helps :)

lotus_success

### DIFF
--- a/lib/lotus/commands/generate.rb
+++ b/lib/lotus/commands/generate.rb
@@ -74,7 +74,7 @@ module Lotus
       # @since 0.3.0
       # @api private
       def generator
-        require "lotus/generators/#{ @type }"
+        require "lotus/generators/#{ @type }.rb"
         class_name = Utils::String.new(@type).classify
         Utils::Class.load!(GENERATORS_NAMESPACE % class_name).new(self)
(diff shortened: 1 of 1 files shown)


jodosha:
@khaiql This looks odd to me. I mean, we always require libs without the `.rb` ext. Which OS and Ruby version do you have?

jodosha:
@khaiql I had a look at your stack trace, you're using lotusrb 0.3.2, which doesn't have that feature yet. It will be shipped with 0.4.0 on June 23.

In the meantime, please adjust your `Gemfile` like this:

```ruby
gem 'lotus-model', github: 'lotus/model'
gem 'lotusrb', github: 'lotus/lotus'
```

Closing this.

khaiql:
@jodosha according to my gemfile, I'm using lotusrb 0.4, btw, this error looks odd to me also :)

image

PR #266, opened 2015-06-23
Bump lotus-model dependency

joneslee85:


jodosha:
+1

http://lucaguidi.com

PR #272, opened 2015-06-25
Fixing lotus routes when architecture is application

vyper:
Resolves [lotus/router #60](https://github.com/lotus/router/issues/60)

Command lotus routes don't expect application architecture.


joneslee85:
:+1:

PR #277, opened 2015-06-28
Add dotenv files to .gitignore by default

huynhquancam:
Should we put dotenv files into `.gitignore` by default, as `dotenv` gem is shipped with the framework? 

Sometimes we may absent-mindedly forget to ignore our env files. :smile:

### DIFF
--- a/lib/lotus/generators/application/app/gitignore.tt
+++ b/lib/lotus/generators/application/app/gitignore.tt
@@ -1,2 +1,5 @@
 /db/<%= config[:app_name] %>_development
 /db/<%= config[:app_name] %>_test
+.env
+.env.test
+.env.development


joneslee85:
@huynhquancam thanks for this PR. We do want to commit these dotenv file should users opt to add more variables. Please be noted that these dotenv are loaded on production env.

huynhquancam:
@joneslee85 Thanks. I do think that committing these dotenv files may accidentally expose our sensitive data to the world e.g. S3_SECRET_KEY.

@bkeepers has mentioned about this. https://github.com/bkeepers/dotenv#should-i-commit-my-env-file

Anyway, user is able to optionally add them to `.gitignore` though. :smile:

joneslee85:
@huynhquancam it's true, but you should not use real production key though, and yes, please feel free to add them to ignore file

PR #279, opened 2015-06-29
Fix load container in lotus console

AlfonsoUceda:
If we run `lotus console` with container arch there is a problem generating routes with routes helpers.

An example, we have this container:
```ruby
# environment.rb
Lotus::Container.configure do
  mount Admin::Application, at: '/admin'
  mount CallCenter::Application, at: '/callcenter'
end

# apps/admin/config/routes.rb
get 'home', to: 'home#index', as: :home
```
if we launch `lotus console`
```ruby
# BEFORE
Admin::Routes.path(:home) #=> '/home'

# AFTER THIS FIX
Admin::Routes.path(:home) #=> '/admin/home'
```

### DIFF
--- a/lib/lotus/commands/console.rb
+++ b/lib/lotus/commands/console.rb
@@ -28,8 +28,8 @@ module Lotus
 
         # Add convenience methods to the main:Object binding
         TOPLEVEL_BINDING.eval('self').send(:include, Methods)
-        Lotus::Application.preload_applications!
 
+        load_application
         engine.start
       end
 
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
cc @joneslee85 @jodosha 
Is there a way to test inside console?, like eval code in `Pry` or `irb`

joneslee85:
LGTM :+1: 

@AlfonsoUceda I have no idea how to test inside console yet, anyone?

PR #280, opened 2015-06-30
Fixed random csrf test fail

AlfonsoUceda:
ditto.

This PR solves failing tests with CSRF protection due to csrf protection is disabled in tests


joneslee85:
:+1:

PR #284, opened 2015-07-02
Introduced an extra separator for controller action in action generator

### LINKED ISSUE #282: Support for ZSH (ohmyzsh)
I'm running oh-my-zsh as my default shell, and the lotus commands do not work without prepending `noglob`, otherwise giving the following error:

`zsh: no matches found - controller#action`

I fixed this by adding the alias:
`alias lotus="noglob lotus"`

----

I'm not too sure of the internals, but this might be stressful to people trying to get the generators working while using zsh. Figured I'd at least document it here.

Thanks!

jodosha:
Introduced an extra separator for action generator in order to let devs with a shell that has a glob activated to not run in problems with `#` char.

The following syntax will be alternative:

```shell
lotus generate action web books#show
```

or

```shell
lotus generate action web books/show
```

Closes #282

### DIFF
--- a/lib/lotus/generators/action.rb
+++ b/lib/lotus/generators/action.rb
@@ -10,6 +10,10 @@ module Lotus
       # @api private
       ACTION_SEPARATOR = /\/|\#/
 
+      # @since x.x.x
+      # @api private
+      ROUTE_ENDPOINT_SEPARATOR = '#'.freeze
+
       # @since 0.3.0
       # @api private
       SUFFIX           = '.rb'.freeze
@@ -119,7 +123,7 @@ module Lotus
 
         # Insert at the top of the file
         cli.insert_into_file _routes_path, before: /\A(.*)/ do
-          "get '#{ _route_url }', to: '#{ @name }'\n"
+          "get '#{ _route_url }', to: '#{ _route_endpoint }'\n"
         end
       end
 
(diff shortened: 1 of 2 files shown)


joneslee85:
@jodosha that might be a bit confusing because to me the `/` means the namespace

jodosha:
@joneslee85 How so? :smile: Do you have an example to share? :)

joneslee85:
@jodosha my scenarios is to create a namespace like:

```
lotus generate action web old/books#show
```

that would generate action `Old::Books::Show`

AlfonsoUceda:
:+1:

apux:
I use zsh and it is not possible to use `#` from the command line. I get this error `zsh: no matches found: books#show`.
I tried with `/` and it worked, but I had to change the routes because it was using `/` when it shoul be using `#`.
I think we need an alternative for this. If `/` is not an option, how about a dash `-`, or just an empty space? Something like this:

    lotus generate action web old/books show

joneslee85:
@apux what about `~`?

AlfonsoUceda:
@apux have you tried with this PR in your `Gemfile`?

apux:
@AlfonsoUceda I hadn't. I just did it and it works perfectly.

@joneslee85 Do you think it would be possible to generate more than one action at the same time?
Something like

    lotus generate action web books show index

In that case a simple space would be the best option if we don't want to repeat the resource (more like rails' generator works). Otherwise, a `~` could work too.

mli-max:
I generate with zsh like that `lotus generate action web "dashboard#index"`

apux:
Yes. It is also possible to generate it escaping the `#`

    lotus generate action web dashboard\#index

Both alternatives seem a workaround from the shell perspective more than lotuses, but maybe if they are documented will be enough.

jodosha:
Closing in favor of #300

PR #287, opened 2015-07-03
Add support for local .env files

7even:
This adds support for local dotenv files: `.env.local` for all environments and `.env.#{environment}.local` for a specific one.

These files can be `.gitignore`'d so each developer in the team can set up his own unique environment.


joneslee85:
@7even thanks for the PR, I think this PR is a bit of the edge cases. For majority of the usecases, developers don't commit their dotenv to repo. And there is nothing to stop developer in each to team to modify the existing dotenv for their needs.

Is there any other reason why you want to introduce this change?

7even:
> And there is nothing to stop developer in each to team to modify the existing dotenv for their needs.

Do you mean changing the `.env` which is shared among the team?

I am currently trying to convert an old lotus app to the `0.4` version of the framework; we use our own [settings](https://github.com/roomink/settings) gem which covers all the use cases for us - in particular it keeps the database connection settings. The main settings file is shared between developers but `.local` files are `gitignore`'d so each developer can override it with his own specific settings (postgres username for example).

Now that Lotus uses `dotenv` I would like to switch to using it instead to conform to the Twelve-Factor app but currently dotenv-lotus integration doesn't cover the aforementioned needs. Can you please suggest another option of doing that if this PR is not acceptable?

joneslee85:
> Now that Lotus uses dotenv I would like to switch to using it instead to conform to the Twelve-Factor app but currently dotenv-lotus integration doesn't cover the aforementioned needs. Can you please suggest another option of doing that if this PR is not acceptable?

Firstly, I am hesitate to merge this PR. Because firstly this PR is very customised for edge case, I believe your company might have a shared dev environment in which developers have to inherit those general settings from, whilst the common cases are that developers would likely build and drive their own development env. Secondly, `dotenv` is not supposed to run in production (in fact it is disabled in production env).

I am more than happy to spread the word if you could have this PR as a separate gem for whom who find this useful.

7even:
Well, my point was about an app which reads from a database maintained by another application so in order to avoid confusion we decided on the database name and hard-coded it into the shared settings file in both apps while allowing to override the username in the individual ignored settings files.

I think we can try doing that in the convention form having a `.env.example` file (committed to the repository) explaining how to make your own environment. Closing the PR then.

dikond:
We are facing the same issue. We are using database sharding in our app, and we list DB names and general DB URLs in our `.env` files. Devs may have different URL.

So yeah, we may have `.env.example` in repo and `.env.development` in `.gitignore`, but in this case, dev should check if general settings weren't updated within each pull.

Another variant is like @7even described. General `.env.development` and specific `.env.development.local` which will override some of the configs and in `.gitignore` by default. In this case, we don't have to manually sync from `.env.example`. From my perspective, it is a pretty common thing. Would be nice to support this behaviour.

PR #294, opened 2015-07-06
Some RSpec idioms

benlovell:
I took the opportunity to improve some of the generated RSpec specs. In summary:

* Removed explicit `require_relative` to the SUT file
* Replaced explicit class constants for SUT with `described_class`
* Removed redundant `spec_helper` requires since `.rspec` requires this for us
* Uses single-quoted `it` block strings for consistency

Hopefully these changes are not too opinionated, and thanks - I'm loving lotus! :heart:

### DIFF
--- a/lib/lotus/generators/action/action_spec.rspec.tt
+++ b/lib/lotus/generators/action/action_spec.rspec.tt
@@ -4,7 +4,7 @@ describe <%= config[:app] %>::Controllers::<%= config[:controller] %>::<%= confi
   let(:action) { described_class.new }
   let(:params) { Hash[] }
 
-  it "is successful" do
+  it 'is successful' do
     response = action.call(params)
     expect(response[0]).to eq 200
   end
--- a/lib/lotus/generators/action/action_spec.rspec.tt
+++ b/lib/lotus/generators/action/action_spec.rspec.tt
@@ -1,4 +1,3 @@
-require 'spec_helper'
 require_relative '<%= config[:relative_action_path] %>'
 
 describe <%= config[:app] %>::Controllers::<%= config[:controller] %>::<%= config[:action] %> do
--- a/lib/lotus/generators/action/action_spec.rspec.tt
+++ b/lib/lotus/generators/action/action_spec.rspec.tt
@@ -1,7 +1,7 @@
 require_relative '<%= config[:relative_action_path] %>'
 
 describe <%= config[:app] %>::Controllers::<%= config[:controller] %>::<%= config[:action] %> do
-  let(:action) { <%= config[:app] %>::Controllers::<%= config[:controller] %>::<%= config[:action] %>.new }
+  let(:action) { described_class.new }
   let(:params) { Hash[] }
 
   it "is successful" do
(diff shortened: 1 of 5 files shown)


joneslee85:
@benlovell thanks for submission.

> Removed redundant `spec_helper` requires since `.rspec` requires this for us

Personally I am not sold with this idiom much, to me it adds more *surprise factor*. Thoughts? @jodosha @AlfonsoUceda

benlovell:
@joneslee85 this one does sometimes cause some friction, although permit me to point out: https://github.com/lotus/lotus/blob/f498682b6b8ec3916c32863ce9ed1bee55135794/lib/lotus/generators/application/app/spec_helper.rb.rspec.tt#L12 :wink:

mli-max:
@joneslee85 We don't require `spec_helper.rb` in rails projects. It requires in `.rspec`. Usually we require `rails_helper.rb` in each spec files :smile:

jodosha:
Hey @benlovell Thanks for this PR. The "long" `require_relative` is needed for testing an action/view in isolation. Eg. `bundle exec rspec path/to/action_spec.rb`.

By default Lotus loads the bare minimum amount of code: gem dependencies, the framework, and application configurations. Application code (such as actions) isn't loaded, unless you trigger certain methods (eg. `Lotus::Container.new`). The goal is to have a fast start time in order to make TDD easy.

So, when you run `rspec` like described above, you need to explicitly require the object under test. We prefer `require_relative` over `require`, because it's faster.

Can you please revert that change? Thank you!

benlovell:
@jodosha I did investigate this but it wasn't the behaviour I observed. That said, I may well be missing something obvious...

```ruby
describe Web::Views::Books::Index do
  let(:exposures) { Hash[foo: 'bar'] }
  let(:template)  { Lotus::View::Template.new('apps/web/templates/books/index.html.erb') }
  let(:view)      { described_class.new(template, exposures) }
  let(:rendered)  { view.render }

  it 'exposes #foo' do
    expect(view.foo).to eq exposures.fetch(:foo)
  end
end

```

```
hurrdurr~/code/bookshelf % bundle exec rspec spec/web/views/books/index_spec.rb
.

Finished in 0.00127 seconds (files took 0.54657 seconds to load)
1 example, 0 failures
```

And:

```ruby
describe Web::Controllers::Books::Index do
  let(:action) { described_class.new }
  let(:params) { Hash[] }
 
  it 'is successful' do
    response = action.call(params)
    expect(response[0]).to eq 200
  end
end
```

```
hurrdurr~/code/bookshelf % bundle exec rspec spec/web/controllers/books/index_spec.rb
.
 
Finished in 0.00112 seconds (files took 0.55421 seconds to load)
1 example, 0 failures
```

WDYT?

jodosha:
@benlovell The behavior that you observed isn't the intended one :) We need to understand what triggers the code loading. We'll get back to you soon. Thanks!

benlovell:
@jodosha thanks! Happy to remove 8113e08 if you want to get this merged as-is for now.

joneslee85:
@benlovell please restore the `require_relative` then it's good to me

benlovell:
@joneslee85 OK, 8113e08 is gone now, thanks! <3

joneslee85:
@benlovell please rebase it

benlovell:
@joneslee85 done.

PR #305, opened 2015-07-23
Disable welcome page in the test environment

andrii:
The first high-level feature test should not fail because the text cannot be found on the welcome page, but because the route is undefined.

**EDIT**: According to the Getting Started guide, in test env, a Page Not found would be returned instead


joneslee85:
@andrii thanks for the PR, I am bit puzzled here, firstly, can you please point me to the failed integration test? I do have other question, please get in touch with me on chat when you are online

PR #310, opened 2015-07-29
+ Support for different HTTP methods.

pascalbetz:
* Added `method` option for action generator
* Added tests

### DIFF
--- a/lib/lotus/cli.rb
+++ b/lib/lotus/cli.rb
@@ -83,6 +83,7 @@ module Lotus
     method_option :application_base_url, desc: 'application base url',                                      type: :string
     method_option :path,                 desc: 'applications path',                                         type: :string
     method_option :url,                  desc: 'relative URL for action',                                   type: :string
+    method_option :method,               desc: "HTTP method for action. Upper/lower case is ignored. Must be one of GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS, TRACE.", type: :string, default: 'GET'
     method_option :skip_view,            desc: 'skip the creation of view and templates (only for action)', type: :boolean, default: false
     method_option :help, aliases: '-h',  desc: 'displays the usage method'
 
--- a/lib/lotus/generators/action.rb
+++ b/lib/lotus/generators/action.rb
@@ -1,5 +1,6 @@
 require 'lotus/generators/abstract'
 require 'lotus/utils/string'
+require 'lotus/routing/route'
 
 module Lotus
   module Generators
@@ -117,8 +118,8 @@ module Lotus
       # @since x.x.x
       # @api private
       def assert_http_method!
-        if !HttpRouter::Route::VALID_HTTP_VERBS.include?(_http_method.upcase)
-          raise Lotus::Commands::Generate::Error.new("Unknown HTTP method '#{_http_method}', please use one of #{HttpRouter::Route::VALID_HTTP_VERBS.join(', ')}.")
+        if !Lotus::Routing::Route::VALID_HTTP_VERBS.include?(_http_method.upcase)
+          raise Lotus::Commands::Generate::Error.new("Unknown HTTP method '#{_http_method}', please use one of #{Lotus::Routing::Route::VALID_HTTP_VERBS.join(', ')}.")
         end
       end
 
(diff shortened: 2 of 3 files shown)


pascalbetz:
@joneslee85 please review.

lib/lotus/cli.rb
```diff
@@ -83,6 +83,7 @@ def new(name = nil)
     method_option :application_base_url, desc: 'application base url',                                      type: :string
     method_option :path,                 desc: 'applications path',                                         type: :string
     method_option :url,                  desc: 'relative URL for action',                                   type: :string
+    method_option :method,               desc: 'HTTP method for action',                                    type: :string, default: 'get'
```
joneslee85: please mention all supported http verbs in description

lib/lotus/generators/action.rb
```diff
@@ -30,6 +30,16 @@ class Action < Abstract
...
 
+      # Default HTTP method used when generating an action.
+      # @since x.x.x
+      # @api private
+      DEFAULT_HTTP_METHOD = 'get'.freeze
+
+      # Supported HTTP methods when generating an action.
+      # @since x.x.x
+      # @api private
+      AVAILABLE_HTTP_METHODS = %w{get post put patch delete trace}.freeze
```
joneslee85: can we use the verbs supported by our router https://github.com/lotus/router/blob/master/lib/lotus/routing/http_router.rb#L10 ?

pascalbetz:
@joneslee85 : thanks for feedback. Ammended my changes.

lib/lotus/generators/action.rb
```diff
@@ -108,6 +114,14 @@ def assert_action!
         end
       end
 
+      # @since x.x.x
+      # @api private
+      def assert_http_method!
+        if !HttpRouter::Route::VALID_HTTP_VERBS.include?(_http_method.upcase)
```
joneslee85: IMHO `HttpRouter` is low-level, there should be a public interface in `Lotus::Router` that we should use, please check

lib/lotus/cli.rb
```diff
@@ -83,6 +83,7 @@ def new(name = nil)
     method_option :application_base_url, desc: 'application base url',                                      type: :string
     method_option :path,                 desc: 'applications path',                                         type: :string
     method_option :url,                  desc: 'relative URL for action',                                   type: :string
+    method_option :method,               desc: "HTTP method for action. Upper/lower case is ignored. Must be one of GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS, TRACE.", type: :string, default: 'GET'
```
joneslee85: should you retrieve the methods from the verbs constant instead of hardcoding here?

lib/lotus/generators/action.rb
pascalbetz: There does not seem another list of supported methods. So we:

* duplicate it in the CLI
* add methods to Lotus::Router (```Lotus::Router.supported_http_methods``` and/or ```Lotus::Router.http_method_supported?('bla')```)
* move the array to Lotus::Router
* Leave as is

lib/lotus/cli.rb
pascalbetz: It would force me to require the class. Thought that might be not wanted for a description (it's loaded all the time then, even for other commands). Should be easier/cleaner once these commands have been refactored like the db commands.
What do you think? Require it seems fine to me.

joneslee85: well, if so, i'd prefer not to, hardcode as it is

lib/lotus/generators/action.rb
joneslee85: :+1: for `Lotus::Router.valid_http_verbs` PR

pascalbetz: Lotus::Routes currently defines the DSL. Not sure if it should go there.

I'm unsure why those methods are defined at all. Can not find anything that references that list of methods at all (besides my PR). What is the purpose?

joneslee85: please use `Lotus::Routing::Route::VALID_HTTP_VERBS`

pascalbetz: Done.  Not sure if it is better now, had to require it. Left it as a single commit so far, can squash if you like it.

Now, since apparently the existing code uses _verb_ and _method_ (but more _verb_) i was wondering if i should rename everything in my PR from method to verb?

PR #315, opened 2015-08-08
Fix test

joneslee85:
The `Backend::Application` and `Frontend::Application` is no-where
to be found in `test/fixtures.rb`. I use `CoffeeShop::Application`
and `Reviews::Application` instead, because they are both defined
in `test/fixtures.rb`


AlfonsoUceda:
:+1:

PR #324, opened 2015-08-20
Refactor the CLI for better maintainability

pascalbetz:
The current implementation reached its limits, adding new features, switches, options was painful and error prone experience.

Goals:

* create one class per subcommand
* remove direct dependency from Thor, use Thor only for parsing CLI and processing of templates
* extract Thor config to one file per command which is then included into the main Thor file
* improve testability and test coverage
* compare generated output against fixture files not strings within the test

Relevant Issue: https://github.com/lotus/lotus/issues/297

Check list by @joneslee85:
Ensure every commands works:

* [X] lotus new
* [X] lotus version
* [X] lotus routes
* [X] lotus server
* [X] lotus help
* [X] lotus console
* [X] lotus generate action
* [X] lotus generate app
* [X] lotus generate help
* [X] lotus generate migration
* [X] lotus generate model

PR #325, opened 2015-08-22
Refactor generators database config

brennovich:
- Extract common behaviour between generators (app, container)
- Add unit specs covering generators'  database config logic
- Add multi-platform support (mri, jruby)
- Normalize behaviour of database configuration between App and Container generators (add `.sqlite` into database name when SQLite engine)

I'm aware of #324, but I'm taking a different approach, instead dumping all database config logic into `Abstract` class (where I believe we should only define interface, not logic) I extracted to a new class `DatabaseConfig`. It think this should be incorporated in #324

### DIFF
--- a/lib/lotus/generators/database_config.rb
+++ b/lib/lotus/generators/database_config.rb
@@ -1,0 +1,88 @@
+module Lotus
+  module Generators
+    class DatabaseConfig
+      SUPPORTED_ENGINES = {
+        'mysql'      => { type: :sql,         mri: 'mysql',   jruby: 'jdbc-mysql'    },
+        'mysql2'     => { type: :sql,         mri: 'mysql2',  jruby: 'jdbc-mysql'    },
+        'postgresql' => { type: :sql,         mri: 'pg',      jruby: 'jdbc-postgres' },
+        'postgres'   => { type: :sql,         mri: 'pg',      jruby: 'jdbc-postgres' },
+        'sqlite'     => { type: :sql,         mri: 'sqlite3', jruby: 'jdbc-sqlite3'  },
+        'sqlite3'    => { type: :sql,         mri: 'sqlite3', jruby: 'jdbc-sqlite3'  },
+        'filesystem' => { type: :file_system, mri: nil,       jruby: nil             },
+        'memory'     => { type: :memory,      mri: nil,       jruby: nil             }
+      }.freeze
+
+      attr_reader :engine, :name
+
+      def initialize(engine, name)
+        @engine = engine
+        @name = name
+
+        SUPPORTED_ENGINES.key?(engine) or fail "\"#{ engine }\" is not a valid database type"
+      end
+
+      def to_hash
+        {
+          gem: gem,
+          uri: uri,
+          type: type
+        }
+      end
+
+      def type
+        SUPPORTED_ENGINES[engine][:type]
+      end
+
+      def sql?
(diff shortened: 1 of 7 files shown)


joneslee85:
@brennovich I love this approach, it makes senses as it is much easier to work with other platforms variations. Whenever you have time, ping me on chat, I want to pair with you to review this.

lib/lotus/generators/database_config.rb
```diff
@@ -0,0 +1,88 @@
+module Lotus
+  module Generators
+    class DatabaseConfig
+      SUPPORTED_ENGINES = {
+        'mysql'      => { type: :sql,         mri: 'mysql',   jruby: 'jdbc-mysql'    },
```
joneslee85: I don't think we should support the old `mysql` driver anymore, can you please change it to `mysql2` for MRI?

joneslee85: no, on second look, it seems to me we do support that, please ignore my comment above

test/integration/cli/generate_test.rb
```diff
@@ -13,7 +14,7 @@
 
     def create_temporary_dir
       @tmp = Pathname.new(@pwd = Dir.pwd).join('tmp/integration/cli/generate')
-      @tmp.rmtree if @tmp.exist?
+      FileUtils.rm_rf(@tmp) if @tmp.exist?
```
joneslee85: is there value of checking `@tmp.exist?` anymore?

lib/lotus/generators/database_config.rb
brennovich: Maybe we should drop support, I believe that not even Rails supports this

brennovich:
@joneslee85 sure! Any preferred service (hangouts, Skype, irc, etc)?

joneslee85:
@brennovich use our gitter.im chat room, http://chat.lotusrb.org

test/integration/cli/generate_test.rb
brennovich: You're right :)

Dropped 0af5573

carloslopes:
Awesome! :heart:

joneslee85:
Looks good, please squash into 1

cc @lotus/core

lib/lotus/generators/database_config.rb
joneslee85: @brennovich please do it in a separate PR instead

lib/lotus/generators/database_config.rb
```diff
@@ -0,0 +1,88 @@
...
+        when 'mysql2'
+          "mysql2://localhost/#{ name }"
+        when 'postgresql', 'postgres'
+          "postgres://localhost/#{ name }"
+        when 'sqlite', 'sqlite3'
+          "sqlite://db/#{ Shellwords.escape(name) }"
+        when 'memory'
+          "memory://localhost/#{ name }"
+        when 'filesystem'
+          "file:///db/#{ name }"
```
joneslee85: please also do `Shellwords.escape` here

brennovich: Done https://github.com/brennovich/lotus/commit/0dd955929d9d0f3d49ccc2f38ecfdbf48ef1c2e7#diff-38afc003617f0593d1a62c002f6a8f42R74

brennovich:
Squashed :)

PR #327, opened 2015-08-25
Add logger option on Lotus configuration

rpanachi:
I've added a logger method on Lotus Configuration to make life easier :)

This allow to define a different logger instance by environment, avoiding to redefine logger class by hand, like this:

```ruby
configure :development do
  logger Logger.new(STDOUT)
end
```

```ruby
configure :production do
  # production configs here
  logger Logger.new("/var/log/app.log")
end
```

To use on application:

```ruby
module Bookshelf::Controllers::Books
  class Index
    include Bookshelf::Action
    def call(params)
      Bookshelf::Logger.info "Lotus rules!"
    end
  end
end
```

### DIFF
--- a/lib/lotus/configuration.rb
+++ b/lib/lotus/configuration.rb
@@ -1642,11 +1642,38 @@ module Lotus
       @view ||= Config::FrameworkConfiguration.new
     end
 
-    # This options is used to set/get the application_module logger instance
+    # Defines a logger instance to the configuration
+    #
+    # This logger instance will be used to set the logger available on application module
+    #
+    # If no logger instance is defined, a Lotus::Logger will be set by default
     #
     # @return [Logger, NilClass] logger instance
     #
-    # @since 0.4.0
+    # @since x.x.x
+    #
+    # @example Define a logger
+    #   require 'lotus'
+    #
+    #   module Bookshelf
+    #     class Application < Lotus::Application
+    #       configure do
+    #         logger Logger.new(STDOUT)
+    #       end
+    #       load!
+    #     end
+    #
+    #     module Controllers::Error
+    #       include Bookshelf::Controller
+    #
+    #       action 'Index' do
+    #         def call(params)
+    #           Bookshelf::Logger.info "Logging to STDOUT"
+    #         end
+    #       end
+    #     end
+    #   end
+    #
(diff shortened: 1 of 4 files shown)


manuwell:
that is nice! 

i think is more explicit than setting the constant Logger on container.

lib/lotus/configuration.rb
```diff
@@ -1642,6 +1642,19 @@ def view
       @view ||= Config::FrameworkConfiguration.new
     end
 
+    # This options is used to set/get the application_module logger instance
+    #
+    # @return [Logger, NilClass] logger instance
```
joneslee85: can you please add example here? Please follow the documentation convention :)

lib/lotus/configuration.rb
```diff
@@ -1642,6 +1642,19 @@ def view
...
+    #
+    # @since 0.4.0
```
joneslee85: s/0.4.0/x.x.x

test/configuration_test.rb
```diff
@@ -825,4 +825,29 @@ class FileSystem
...
   end
+
+  describe '#logger' do
+    describe "when not previously set" do
+      before do
+        @configuration = Lotus::Configuration.new
+      end
+
+      it 'defaults to nil' do
+        @configuration.logger.must_be :nil?
```
joneslee85: what do you think about `must_be_nil` ?

rpanachi:
@joneslee85 I've applied your suggestions and added a test for Loader. Please let me know if it's good to merge. Thanks

joneslee85:
:+1:

test/loader_test.rb
```diff
@@ -99,6 +99,19 @@ class Application < Lotus::Application; load!; end
 
           output.must_match(/BeerShop/)
         end
+
+        it 'supports configured custom logger' do
+          class MyLogger < Logger; end
```
joneslee85: @rpanachi i'd prefer that you declare the new logger class in `before` and killing it in `after`. The same goes for the `DrinkShop:: Application`

rpanachi:
@joneslee85 I refactored the tests related to logger on loader_test. What do you think? Thanks

PR #329, opened 2015-08-31
Removed multiple CI version

ferdinandrosario:
Hi, as of now we are running CI for 16 various Ruby versions. shall we remove each specific version and test with only the latest stable release?
```
rvm:
  2.0
  2.1
  2.2
  rbx-2
  jruby-head
  ruby-head
```
This will ensure only the latest patch as well we can help travis by reducing the CI count....

If there is a specific reason then kindly Ignore this PR...
This is one of my suggestion...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,18 +4,8 @@ cache: bundler
 script: 'bundle exec rake test:coverage'
 rvm:
   - 2.0.0
-  - 2.1.0
-  - 2.1.1
-  - 2.1.2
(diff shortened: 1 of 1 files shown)


joneslee85:
@ferdinandrosario thanks for the PR, we want to ensure that there is no regression whatsoever with every minor version of major series. Furthermore, there have been few intermittent failtures on specific minor versions occured in the past.

ferdinandrosario:
Ok :+1:

PR #341, opened 2015-09-15
introduces request and response format and deprecates default_format

theocodes:
This is based on https://github.com/lotus/controller/pull/125.

It introduces those same changes to application configuration and adds the example for `default_response_format` to the generated files.

Please check the notes over at the other PR.

thoughts?


Gemfile
```diff
@@ -9,7 +9,8 @@ end
 gem 'lotus-utils',       '~> 0.5', require: false, github: 'lotus/utils',       branch: '0.5.x'
 gem 'lotus-validations', '~> 0.3', require: false, github: 'lotus/validations', branch: '0.3.x'
 gem 'lotus-router',      '~> 0.4', require: false, github: 'lotus/router',      branch: '0.4.x'
-gem 'lotus-controller',  '~> 0.4', require: false, github: 'lotus/controller',  branch: '0.4.x'
+gem 'lotus-controller',  '~> 0.4', require: false, github: 'theocodes/controller', branch: 'response-format-improvements'
```
mvader: is this on purpose?

test/configuration_test.rb
```diff
@@ -605,25 +605,47 @@ class FileSystem
...
+    it 'raises an error if the given format cannot be coerced into symbol' do
+      -> { @configuration.default_request_format(23) }.must_raise TypeError
+    end
+  end
+
+  describe '#default_response_format' do
     describe "when not previously set" do
       it 'returns nil' do
-        @configuration.default_format.must_equal :html
+        @configuration.default_response_format.must_equal nil
```
jodosha: @theocodes HTML must be honored. Lotus::Controller is designed for HTTP endpoints, so it uses a generic content type (`application/octet-stream`, Rack default).

Lotus full stack app are designed to be web apps, where HTML must be the default.

lib/lotus/configuration.rb
```diff
@@ -987,20 +1039,33 @@ def adapter(options = {})
...
     end
 
+    # Set a format as default fallback for all the requests without a strict
+    # requirement for the mime type.
+    #
+    # @since 0.1.0
+    #
+    # @deprecated Use {#default_request_format} instead.
+    def default_format(format = nil)
+      Lotus::Utils::Deprecation.new('default_format is deprecated, please use default_request_format')
```
jodosha: @theocodes We should test this deprecation.

Gemfile
theocodes: @mvader yes. This requires changes that have been applied to 'lotus/controller' also.

test/configuration_test.rb
theocodes: @jodosha I see your point. But bare in mind that if this value is nil, the framework will guess the content type by looking at the request content type.  (https://github.com/lotus/controller/blob/master/lib/lotus/action/mime.rb#L155)

The issue with setting `:html` to be the default there is that the framework will then never try to detect the request type because default_response_format will always have a value.

In other words, this logic hasn't changed. The only difference is that if `default_response_format` is set, it will use that. Otherwise it will use the same logic as before.

On the other hand, if you wish to change the logic that is currently in place and have the default_response_format always be `:html` by default, thats fine too. As long as we understand that slight change in semantics :)

please me know your thoughts

theocodes: @jodosha I have addressed everything here apart from `default_response_format` having a value by default.  let me know if you indeed want me to change that.

jodosha:
@theocodes I just merged this, thank you! :+1:

theocodes:
Happy to help :)

PR #342, opened 2015-09-17
Merge assets sources

### LINKED ISSUE #333: Asset directories overwrite earlier directories
Given this structure:
- apps
  - web
    - assets
      - fonts
      - images
      - javascripts
        - application.js
      - stylesheets
        - application.css
    - ...
    - vendor
      - fonts
      - images
      - javascripts
        - jquery.js
      - stylesheets
        - basecss.css
     - ...
- ...

Given the configuration:
````Ruby
module Web
  class Appliction > Lotus::Aplication
    configure do
      # ...
      
      assets << [
        'assets',
        'vendor',
      ]
      
      #...
      serve_assets true #for the current env.
      #...
  end
end
````

Expected the following to be served under the mounted path
- `javascripts/application.js`,
- `javascripts/jquery.js`, 
- `stylesheets/application.css` and 
- `stylesheets/basecss.css`

Actually these are served under the mounted path:
- `stylesheets/basecss.css`
- `javascripts/jquery.js`, 

**Workaround**:
- Give all the sub-directories unique names or flatten the vendor folder.

Why is this not a good workaround?
- Paths should be merged, not overwritten. 
- Earlier entries (assets) should survive later entries (vendors) so you can overwrite files.
- Vendor files may depend on fixed paths which makes the workaround cumbersome

jodosha:
Closes #333


SleeplessByte:
This seems right!

PR #343, opened 2015-09-17
Remove debugging leftover from assets.rb.

simi:


simi:
:( Did a broke something?

PR #346, opened 2015-09-22
Generate nested actions correctly

caius:
Hi!

Whilst trying to generate a nested action earlier I noticed the generator doesn't consume the cli arguments correctly.

Given the command `bundle exec lotus generate action "admin/dashboard#create"`:

Currently in master this will pick out the controller constant as being `Admin` and the action constant as being `Dashboard`, which then generates the module `Admin::Dashboard` in `app/:name/controllers/admin/dashboard.rb`.

This PR fixes this up by only splitting that last cli argument by `#` instead of `#` and `/`, which leads to the controller being `Admin::Dashboard` & the action being `Create`. Which in turn (correctly) generates `Admin::Dashboard::Create` in `app/:name/controllers/admin/dashboard/create.rb`.

Test fails against master & passes against this branch.

Thanks!

### DIFF
--- a/test/commands/generate/action_test.rb
+++ b/test/commands/generate/action_test.rb
@@ -42,6 +42,17 @@ describe Lotus::Commands::Generate::Action do
       end
     end
 
+    it "handles nested routes correctly" do
+      with_temp_dir do |original_wd|
+        setup_container_app
+        command = Lotus::Commands::Generate::Action.new({}, 'web', 'admin/books#index')
+        capture_io { command.start }
+
+        assert_file_exists 'apps/web/templates/admin/books/index.html.erb'
+        assert_file_includes('apps/web/config/routes.rb', "get '/admin/books', to: 'admin/books#index'")
+      end
+    end
+
     describe 'route options' do
       it 'uses the --method option to specify HTTP method in the route' do
         with_temp_dir do |original_wd|
--- a/lib/lotus/commands/generate/action.rb
+++ b/lib/lotus/commands/generate/action.rb
@@ -33,11 +33,14 @@ module Lotus
 
           controller_and_action_name = Utils::String.new(controller_and_action_name).underscore.gsub(QUOTED_NAME, '')
 
-          controller_name, action_name = controller_and_action_name.split(ACTION_SEPARATOR)
+          controller_name, action_name = controller_and_action_name.split(ROUTE_ENDPOINT_SEPARATOR)
 
           @application_name = Utils::String.new(application_name).classify
           @controller_name = Utils::String.new(controller_name).classify
           @action_name = Utils::String.new(action_name).classify
+
+          # Handle nested actions correctly
+          @controller_pathname = Utils::String.new(@controller_name).underscore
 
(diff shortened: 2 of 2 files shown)


joneslee85:
@caius thanks for the PR. This brings to my attention that we should facilitate the conversation on the guideline of having / or not having nested routes. Because IMHO nested action could be extracted to a separate app of its own. Having said that, I don't think we should impose such strict rule but rather recommending users the best practices.

Please rebase and get the test green then I'll merge this

caius:
@joneslee85 ah interesting, hadn't thought to pull the action into a separate app to do nested routes. This is the first lotus app I've built though, coming from a rails/sinatra background so I'm not be as enlightened to the lotus way as I could be :grin: 

Have rebased & fixed up the action generator a bit more with recent changes - it was trying to generate `templates/admin::books/index.rb` with just my original fix rebased :laughing:.

test/commands/generate/action_test.rb
```diff
@@ -40,6 +40,20 @@
       end
     end
 
+    it "handles nested routes correctly" do
+      with_temp_dir do |original_wd|
+        setup_container_app
+        command = Lotus::Commands::Generate::Action.new({}, 'web', 'admin/books#index')
+        capture_io { command.start }
+
+        assert File.exist?('apps/web/templates/admin/books/index.html.erb')
```
joneslee85: there is a helper for this, please check spec/support folder

test/commands/generate/action_test.rb
```diff
@@ -40,6 +40,20 @@
...
+
+        expected_route = "get '/admin/books', to: 'admin/books#index'"
+        actual = File.read('apps/web/config/routes.rb')
+        assert actual.include?(expected_route)
```
joneslee85: what do you think about using must_include instead?

caius:
@joneslee85 thanks for the comments, had cargoculted from elsewhere in the file and didn't spot the custom assertions. Have fixed up the test to use `must_include` and the file existence assert instead. Could you take another look please? :smile:

joneslee85:
:+1: 

cc @AlfonsoUceda @jodosha for another round of review

test/commands/generate/action_test.rb
```diff
@@ -42,6 +42,20 @@
...
+      with_temp_dir do |original_wd|
+        setup_container_app
+        command = Lotus::Commands::Generate::Action.new({}, 'web', 'admin/books#index')
+        capture_io { command.start }
+
+        assert_file_exists 'apps/web/templates/admin/books/index.html.erb'
+
+        expected_route = "get '/admin/books', to: 'admin/books#index'"
+        actual = File.read('apps/web/config/routes.rb')
+        actual.must_include(expected_route)
```
t4deu: You can use the `assert_file_includes` here too, like this:
```ruby
assert_file_includes("apps/web/config/routes.rb", "get '/admin/books', to: 'admin/books#index'")
```
This custom assertion can compare multiple strings or regex

caius: Oh cool, missed that. :blush: Fixed up & pushed now.

PR #349, opened 2015-09-30
Generate lib/app_name/mailers/templates on app creation

joneslee85:


jodosha:
@joneslee85 Great quick response! :clap: 

Why do we need these new `.gitkeep` files? `lib/lotus/generators/application/container/lib/chirp/mailers/.gitkeep`

I mean, `chirp` is part of a fixture, right?

joneslee85:
@jodosha yup

jodosha:
@joneslee85 I mean, should we remove them?

joneslee85:
@jodosha we should, another PR

Issue #348, opened 2015-09-29
Stop users from generating model with names that is not a valid ruby classname


joneslee85:
```
bundle exec lotus generate 1235
```

As you can the naming is not a valid ruby classname and thus would yield error when you run the app.

We want to block users from having naming that is not a valid inflected classname

jasoncharnes:
This would be something I'd be willing to take a stab at.

joneslee85:
@jasoncharnes please give it a go

jasoncharnes:
@joneslee85 will do! should be able to work on it this weekend. thanks!

jasoncharnes:
@joneslee85 could you explain, or provide an example, of a non-valid inflected class name? Sorry, just want to make sure I'm covering all the bases.

joneslee85:
We inflect snake case name to valid Ruby class name. A valid class name can contain number but could not start with a number. Solution for this problem can be:

* regex to detect format then throw error
* regex to detect format then append a non-number prefix, eg `Migration123CreateMyTable`

joneslee85:
I like th first approach

PR #353, opened 2015-10-03
Adds validation for model name

jasoncharnes:
#348

### DIFF
--- a/lib/lotus/commands/generate/model.rb
+++ b/lib/lotus/commands/generate/model.rb
@@ -15,6 +15,7 @@ module Lotus
           @generator = Lotus::Generators::Generator.new(template_source_path, base_path)
 
           assert_model_name!
+          assert_model_has_valid_name!
         end
 
         def start
@@ -32,6 +33,14 @@ module Lotus
         def assert_model_name!
           if model_name.nil? || model_name.strip.empty?
             raise ArgumentError.new('Model name nil or empty.')
+          end
+        end
+
+        # @since x.x.x
+        # @api private
+        def assert_model_has_valid_name!
+          unless model_name.match(/^[a-z]/i)
+            raise ArgumentError.new("Invalid model name")
           end
         end
 
--- a/lib/lotus/commands/generate/model.rb
+++ b/lib/lotus/commands/generate/model.rb
@@ -36,11 +36,12 @@ module Lotus
           end
         end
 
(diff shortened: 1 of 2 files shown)


test/commands/generate/model_test.rb
```diff
@@ -9,6 +9,10 @@
       -> { Lotus::Commands::Generate::Model.new({}, '') }.must_raise ArgumentError
       -> { Lotus::Commands::Generate::Model.new({}, '   ') }.must_raise ArgumentError
     end
+
+    it 'validates model name' do
+      -> { Lotus::Commands::Generate::Model.new({}, '123') }.must_raise ArgumentError
```
joneslee85: what do you think about asserting the exception message here too?

```
exception = -> { Lotus::Commands::Generate::Model.new({}, '123') }.must_raise ArgumentError
exception.message.must_equal 'The error message'
```

lib/lotus/commands/generate/model.rb
```diff
@@ -35,6 +36,14 @@ def assert_model_name!
           end
         end
 
+        # @since x.x.x
+        # @api private
+        def assert_model_has_valid_name!
+          unless model_name.match(/^[a-z]/i)
+            raise ArgumentError.new("Invalid model name")
```
joneslee85: What do you think about providing extra details on what a valid naming is?

joneslee85: also can you please add doc comment for `assert_model_name!`?

jasoncharnes:
@joneslee85 appreciate the feedback, and I agree. I pushed an updated commit with the changes.

jodosha:
@jasoncharnes Thanks for this fix.
Do you and @joneslee85 think it's a good idea to merge `assert_model_name!` with `assert_model_has_valid_name!`? as they both verify if the given name meets a set of rules: non-empty, no numbers.

jasoncharnes:
@jodosha I was actually thinking about that while I adding `assert_model_has_valid_name!`. I'd be glad to update it if @joneslee85 agrees.

joneslee85:
@jasoncharnes please do

jasoncharnes:
@jodosha @joneslee85 let me know if this is/isn't along the lines of what your were thinking. Really digging Lotus! Thx for letting me get involved.

PR #370, opened 2015-10-15
A catch all exception handler.

manuwell:
By adding a catch all exception handler on the configuration file, the application can handle the exception and do whatever it wants.

This is a propose of how I might the feature would be. 

With kind of configuration, the application can notify exceptions, log to a centralized log, etc. 

What do you think?

In the `application.rb` configure we will have something like:

``` #!ruby
on_exception do |exception, env|
  [500, {}, "Something went wrong =\. #{exception.message}"]
end
```


AlfonsoUceda:
@manuwell I think you can use a middleware for this.

cc @lotus/core-team

manuwell:
I couldn't catch the exception on middleware once the lotus exception catcher returns a rack response without setting the exception.

Maybe I did it wrong. 

But, in this way I can handle the exception as I like and render a specific view, save into DB. I gonna have all the lotus application power. 

Thks

AlfonsoUceda:
@manuwell how did you make the middleware? can you give me a gist of your example?

manuwell:
@AlfonsoUceda I created [this repo](https://github.com/manuwell/lotus-exeception-example) to ilustrate what I am trying to do with a middleware to catch a template exception. 

as you can see, the middleware coudn't capture the expcetion.

in [this branch](https://github.com/manuwell/lotus-exeception-example/tree/with-exception-handler) I applyied this PR, and as you can see the exception was handled.

jodosha:
@manuwell Hello and thanks for this idea.

What's the semantic of that handler? It's configured at the level of the application, but it has a returning 
value that is a Rack response, which is a concern of the actions. This is confusing for me. :cry: 

How does it works with action's handlers? Which of them takes higher priority? How does it plays with [custom error pages](http://lotusrb.org/guides/views/custom-error-pages/)?

What happens in development mode? What about production? Is the exception bubbled in order to let NewRelic or similar tools to record it?

Can you please explain it? Thanks :smile:

jodosha:
@manuwell This is a gentle PING :smile_cat:

brennovich:
I have a few concerns about this. Exception handling is very expensive in MRI and specially in implementations such as JRuby. If we apply this global handler we can possibly be adding a vulnerability point into user application making DoS attack easier to be successful.

> What happens in development mode? What about production?

Good question, IMO it should only be available in development, it'll enable users to build things like `better_errors` library

> Is the exception bubbled in order to let NewRelic or similar tools to record it?

Nice catch, we definitely shouldn't interferer in profiling tools life-cycle.

joneslee85:
@manuwell can you please rebase this PR? TIA

joneslee85:
@jodosha can you please share your spike on the solution you came up with that fall to rack exception here?

jodosha:
Here it is: https://github.com/hanami/hanami/commit/4c1b6e1024756f29d3c9c9207a99fd55a3c41bd0

The idea is to let exception handler products to do their job. They need the exception to be stored at `env['rack.exception']` and let the exception to not be caught. As a reference, please read https://github.com/hanami/controller/issues/133

---

This proposed solution eliminates the need for developers to worry how to handle exceptions by themselves.

joneslee85:
ping @manuwell

jodosha:
Closing this in favor of #528 

Thank you all! :clap:

PR #371, opened 2015-10-16
Totally refactored work with lotusrc

### LINKED ISSUE #369: .lotusrc double creation
Command `bundle exec lotus new blog` creates two `.lotusrc` files – in current and `./blog` directories. Does it correct?

aderyabin:
__It's very urgent!!!__

Currently, every `lotus` command line generates `.lotusrc` file, even `lotus help`. The reason is https://github.com/lotus/lotus/blob/master/lib/lotus/setup.rb#L3.
 
More complex problem is that `Lotus::Lotusrc.new` generates this file ([prooflink](https://github.com/lotus/lotus/commit/cf6695e6a66bdd8e8f2fa5e83b11de432954819c#diff-00c8fc8ba6a42495f2a5f805f67e19faR76)) , although creation should be executed only on generation new application ([like this](https://github.com/lotus/lotus/commit/f80de2e7ae615a8348be7fc830f07ee907699860#diff-bad57f74cb2621a0a3d2d6e390ac7aa9R41)).

I propose to detect `lotusrc` existance, if no then use [default options](https://github.com/lotus/lotus/compare/lotus:master...aderyabin:double_lotusrc?expand=1#diff-00c8fc8ba6a42495f2a5f805f67e19faR72)

It should resolve #369

### DIFF
--- a/lib/lotus/environment.rb
+++ b/lib/lotus/environment.rb
@@ -173,7 +173,7 @@ module Lotus
     #   # the one defined in the parent (eg `FOO` is overwritten). All the
     #   # other settings (eg `XYZ`) will be left untouched.
     def initialize(options = {})
-      @options = Lotus::Lotusrc.new(root, options).read
+      @options = Lotus::Lotusrc.new(root).options
       @options.merge! Utils::Hash.new(options.clone).symbolize!
       @mutex   = Mutex.new
       @mutex.synchronize { set_env_vars! }
--- a/lib/lotus/lotusrc.rb
+++ b/lib/lotus/lotusrc.rb
@@ -2,7 +2,7 @@ require 'pathname'
 require 'dotenv/parser'
 
 module Lotus
-  # Create and read the .lotusrc file in the root of the application
+  # Read the .lotusrc file in the root of the application
   #
   # @since 0.3.0
   # @api private
@@ -20,7 +20,7 @@ module Lotus
     # @since 0.3.0
     # @api private
     #
-    # @see Lotus::Lotusrc#read
+    # @see Lotus::Lotusrc#options
     DEFAULT_ARCHITECTURE = 'container'.freeze
 
     # Architecture key for writing the lotusrc file
@@ -28,7 +28,7 @@ module Lotus
     # @since 0.3.0
     # @api private
     #
-    # @see Lotus::Lotusrc#read
+    # @see Lotus::Lotusrc::DEFAULT_OPTIONS
     ARCHITECTURE_KEY = 'architecture'.freeze
 
     # Test suite default value
@@ -36,7 +36,7 @@ module Lotus
     # @since 0.3.0
     # @api private
     #
-    # @see Lotus::Lotusrc#read
+    # @see Lotus::Lotusrc::DEFAULT_OPTIONS
     DEFAULT_TEST_SUITE = 'minitest'.freeze
 
     # Test suite key for writing the lotusrc file
@@ -44,7 +44,7 @@ module Lotus
     # @since 0.3.0
     # @api private
     #
-    # @see Lotus::Lotusrc#read
+    # @see Lotus::Lotusrc::DEFAULT_OPTIONS
     TEST_KEY = 'test'.freeze
 
     # Template default value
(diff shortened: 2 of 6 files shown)


jodosha:
@aderyabin Hello, how that would happen? Currently (`v0.5.0`) there is a [guard](https://github.com/lotus/lotus/blob/master/lib/lotus/lotusrc.rb#L116) that creates the file only if missing.

Now, that being said, I'm keen to remove that auto-creation, because we assume that now all the apps created with a version minor of `0.3.0` have now a valid `.lotusrc` file. Note, `v0.3.0` is the version that [introduced](https://github.com/lotus/lotus/blob/master/CHANGELOG.md#v030---2015-03-23) this feature. :+1: 

There are a couple of comments that I left. Can you please address them? Thank you :+1:

lib/lotus/lotusrc.rb
```diff
@@ -72,54 +84,23 @@ class Lotusrc
...
-    # @see Lotus::Lotusrc::TEMPLATE_KEY
-    def create
-      unless exists?
-        rcfile = File.new(path_file, "w")
-        rcfile.puts "#{ ARCHITECTURE_KEY }=#{ @options.fetch(:architecture, DEFAULT_ARCHITECTURE) }"
-        rcfile.puts "#{ TEST_KEY }=#{ @options.fetch(:test, DEFAULT_TEST_SUITE) }"
-        rcfile.puts "#{ TEMPLATE_KEY }=#{ @options.fetch(:template, DEFAULT_TEMPLATE) }"
-        rcfile.close
-      end
+    def options
```
jodosha: @aderyabin Can you please get back to `#read` instead of `#options`? The verb indicates the intent of reading the file.

lib/lotus/lotusrc.rb
```diff
@@ -132,6 +113,16 @@ def exists?
...
     end
 
+    private
+
+    def symbolize(hash)
+      Utils::Hash.new(hash).symbolize!
+    end
+
+    def file_options
+      symbolize(exists? ? Dotenv::Parser.call(content) : {})
```
jodosha: @aderyabin We assume that this now exists for all Lotus apps, no need for this check here. That would make `#exists?` unnecessary at this point.

lib/lotus/environment.rb
```diff
@@ -173,7 +173,7 @@ class Environment
     #   # the one defined in the parent (eg `FOO` is overwritten). All the
     #   # other settings (eg `XYZ`) will be left untouched.
     def initialize(options = {})
-      @options = Lotus::Lotusrc.new(root, options).read
+      @options = Lotus::Lotusrc.new(root, options).options
```
jodosha: @aderyabin Do we still need to pass `options` as second argument to the constructor? Can you please check if we can get rid of it? Thank you.

aderyabin:
@jodosha 
Each command line usage Lotus, for example `lotus help` in current implementation calls `Environment.new`. `Environment.new` calls `Lotusrc.new`. Last should check `.lotusrc` existence because command `lotus help` could be called from any path (I mean not project root path). Also this command should not generate `.lotusrc` file but use default values.

```ruby
DEFAULT_OPTIONS = Utils::Hash.new({
  ARCHITECTURE_KEY => DEFAULT_ARCHITECTURE,
  TEST_KEY         => DEFAULT_TEST_SUITE,
  TEMPLATE_KEY     => DEFAULT_TEMPLATE
}).symbolize!.freeze
```

aderyabin:
As for renaming `#read ` to `#options`. If `.lotusrc` doesn't exists `read` is not a good verb.

jodosha:
> As for renaming #read to #options. If .lotusrc doesn't exists read is not a good verb.

It's the same `File.new('...').read` semantic. :smile:

aderyabin:
Yep, if file exists, but if file not exists method returns _default options_.

aderyabin:
@jodosha What is your latest decision?

jodosha:
@aderyabin It's fine :smile: 

Before to merge, can you please verify this? https://github.com/lotus/lotus/pull/371/files#r42261568 Thank you!

jodosha:
@aderyabin Thanks, it looks good! :+1:

PR #375, opened 2015-10-21
Implemented Chunked Streaming

berfarah:
From ticket https://github.com/lotus/lotus/issues/303

Thanks to @pascalbetz who figured out why the test wasn't working and helped validate!

@jodosha What's the history behind removing Rack::Chunked [here](https://github.com/lotus/lotus/blob/master/lib/lotus/commands/server.rb#L36)? Is it a bug only in WEBrick?

### DIFF
--- a/test/fixtures/streaming/application.rb
+++ b/test/fixtures/streaming/application.rb
@@ -22,11 +22,7 @@ module StreamingApp
       def call(params)
         self.format = :text
         self.body = Enumerator.new do |y|
-          y << "one"
-          sleep 0.5
-          y << "two"
-          sleep 0.5
-          y << "three"
+          %w(one two three).each { |s| y << s }
         end
       end
     end
--- a/test/fixtures/streaming/application.rb
+++ b/test/fixtures/streaming/application.rb
@@ -1,0 +1,34 @@
+require 'rack/chunked'
+
+module StreamingApp
+  class Application < Lotus::Application
+    configure do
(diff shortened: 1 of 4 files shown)


pascalbetz:
We should add notes:

* how to enable streaming
* that streaming does not work with shotgun enabled

Not sure where that should go. Perhaps in lotus-controller gem in action.rb?

test/fixtures/streaming/application.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  module Controllers::Streaming
+    class Get
+      include StreamingApp::Action
+
+      def call(params)
+        self.format = :text
+        self.body = Enumerator.new do |y|
+          y << "one"
+          sleep 0.5
```
pascalbetz: can you remove the ```sleep``` calls?

berfarah: Yep, right, forgot those. Will do.

berfarah:
I think maybe putting it in the [README](https://github.com/lotus/controller/blob/master/README.md) for controller? That's where cookies and sessions are outlined.

pascalbetz:
sounds good!

jodosha:
@berfarah @pascalbetz Excellent, you rock!! :metal: 

I think the explanation goes just before this section https://github.com/lotus/controller/blob/master/README.md#no-rendering-please

PR #377, opened 2015-10-22
RSpec disable_monkey_patching! mode compatibility

jodosha:
RSpec has introduced with 3.0 a mode to avoid to monkey-patching. It's a configuration disabled by default (see https://www.relishapp.com/rspec/rspec-core/v/3-0/docs/configuration/zero-monkey-patching-mode).

This PR makes our generated specs for actions and views to play well with it.

Actual:

```ruby
describe ... do
end
```

Wanted:

```ruby
RSpec.describe do
end
```

/cc @lotus/core-team


AlfonsoUceda:
:+1:

Issue #155, opened 2015-02-13
App generator --template arg


jodosha:
When a new application is generated via `lotus new`, we should allow to pass a new argument: `--template`. This indicates the default template used by the application. If not specified, it defaults to `erb`.

This serves as a base for: [Action generator](https://github.com/lotus/lotus/issues/154) #154.

AlfonsoUceda:
@jodosha and me we was talking about this feature. For now, we can't do this feature, we have detected when we pass the template option, e.g. `--template=haml` we need to replace the default `erb` files with `haml` but we can't support in lotus every template.

We need to create a template generator system for supporting this feature. Like rails does with, e.g. `haml-rails` in the future we'll have similar gems, `lotus-haml`

We have to pospone this feature.

jodosha:
Closing this for now.

PR #379, opened 2015-10-22
support template option on new application generation

aderyabin:
I've added support `--template=slim` option on new app generation, `erb` value by default
```bash
  lotus new bookshelf --template=slim
```

### DIFF
--- a/lib/lotus/cli.rb
+++ b/lib/lotus/cli.rb
@@ -72,11 +72,12 @@ module Lotus
 
       $ > lotus new fancy_app --lotus_head
     EOS
-    method_option :database, aliases: ['-d', '--db'], desc: "application database (#{Lotus::Generators::DatabaseConfig::SUPPORTED_ENGINES.keys.join('/')})", default: Lotus::Generators::DatabaseConfig::DEFAULT_ENGINE
-    method_option :architecture, aliases: ['-a', '--arch'], desc: 'application architecture (container/app)', default: Lotus::Commands::New::Abstract::DEFAULT_ARCHITECTURE
+    method_option :database, aliases: ['-d', '--db'], desc: "application database (#{Lotus::Generators::DatabaseConfig::SUPPORTED_ENGINES.keys.join('/')})", default: Lotus::DEFAULT_DATABASE_ENGINE
+    method_option :architecture, aliases: ['-a', '--arch'], desc: 'application architecture (container/app)', default: Lotus::DEFAULT_ARCHITECTURE
     method_option :application_name, desc: 'application name, only for container', default: Lotus::Commands::New::Container::DEFAULT_APPLICATION_NAME
-    method_option :application_base_url, desc: 'application base url', default: Lotus::Commands::New::Abstract::DEFAULT_APPLICATION_BASE_URL
-    method_option :test, desc: "application test framework (#{Lotus::Generators::TestFramework::VALID_FRAMEWORKS.join('/')})", default: Lotus::Generators::TestFramework::DEFAULT_FRAMEWORK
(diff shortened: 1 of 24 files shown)


joneslee85:
looks good, I like it, though I think `template` is a bit confusing, I would take it as the generator skeleton templates, what  do you think about renaming it `template_engine` instead? I know it is longer but it does express its intent clearly

lib/lotus/cli.rb
```diff
@@ -75,6 +75,7 @@ def console
     method_option :database, aliases: ['-d', '--db'], desc: "application database (#{Lotus::Generators::DatabaseConfig::SUPPORTED_ENGINES.keys.join('/')})", default: Lotus::Generators::DatabaseConfig::DEFAULT_ENGINE
     method_option :architecture, aliases: ['-a', '--arch'], desc: 'application architecture (container/app)', default: Lotus::Commands::New::Abstract::DEFAULT_ARCHITECTURE
     method_option :application_name, desc: 'application name, only for container', default: Lotus::Commands::New::Container::DEFAULT_APPLICATION_NAME
+    method_option :template, desc: 'template engine', default: 'erb'
```
joneslee85: can we update the description with the list of available option? I really want to make it easy for users, please see line 75 for inspiration

aderyabin: So, let's be consistent and replace `template` to `template_engine` in `.lotusrc` too

joneslee85: well, that one might break the compatibility, I'd vote for replacing it, if we agree to do so, it's worth to mention it in the next release notes

cc @jodosha

jodosha:
@joneslee85 @aderyabin please have a look at https://github.com/lotus/lotus/issues/155

lib/lotus/cli.rb
jodosha: @joneslee85 If we decide to proceed with this feature, we must handle `.lotusrc` as well.

aderyabin:
Thanks for reference.

On Thu, Oct 22, 2015 at 4:26 PM, Luca Guidi <notifications@github.com>
wrote:

> @joneslee85 <https://github.com/joneslee85> @aderyabin
> <https://github.com/aderyabin> please have a look at #155
> <https://github.com/lotus/lotus/issues/155>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/lotus/lotus/pull/379#issuecomment-150222983>.
>



-- 
-------------------------
Andrey Deryabin

aderyabin:
@jodosha @AlfonsoUceda Are you still want to create a template engine gems like `lotus-haml` ?

aderyabin:
This PR affects `Lotusrc`. It's better to accept #371 before current.

jodosha:
> Are you still want to create a template engine gems like lotus-haml ?

@aderyabin We actually closed #155 because we do **NOT** want to maintain `lotus-haml` et. co

aderyabin:
I think that support template engines is not so difficult:
– add templates (index.html.xxx and application.html.xxx)
– add gem to Gemfile.

@jodosha  Am I correct?

AlfonsoUceda:
@aderyabin officially we don't want support templates but people can create gems for supporting lotus with those templates ;)

jodosha:
@aderyabin 

The main problem is to maintain `application.html.xxx` for all that template engines: https://github.com/lotus/view#engines

The templates created by action generator are empty, so it's just a matter to specify the right extension. We take it from `.lotusrc`. If you manually edit it `template=haml` and try to generate a new action, you'll see it works fine.

Adding the gem to `Gemfile` is easy too.

jodosha:
I'm gonna close this, because I'm afraid it can lead to inconsistencies and/or request for support for other template engines. This is something that should be eventually handled by third party gems.

Issue #381, opened 2015-10-27
Could not generate new action for app which name contains underscore


joneslee85:
**Step to reproduce**:
1) Generate new app with HEAD `lotus new --application=api_v1 demo --lotus-head`
2) Change into this app dir: `cd demo` and `bundle install`
3) Generate new action: `bundle exec lotus generate action api_v1 books#index`
4) Expected: No errors, Actual: Got errors

Seems to me a regression.

jasoncharnes:
I'd be glad to give this a go tonight.

joneslee85:
@jasoncharnes it's yours

PR #383, opened 2015-10-28
Converts camel case app name into snake case when looking for applica…

jasoncharnes:
…tion during action generation. Also found an issue with those of use who experience an issue when generating actions with zsh and have to use '/' instead of '#' between controller and action.

#381

### DIFF
--- a/lib/lotus/commands/generate/action.rb
+++ b/lib/lotus/commands/generate/action.rb
@@ -8,7 +8,7 @@ module Lotus
 
         # @since x.x.x
         # @api private
-        ACTION_SEPARATOR = /\/|\#/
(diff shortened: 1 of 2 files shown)


joneslee85:
is it possible to have test ?

jasoncharnes:
Absolutely, don't know where my head was at last night. Just pushed the tests.

joneslee85:
:+1:

jodosha:
@jasoncharnes Thank you for this fix :blue_heart:

jasoncharnes:
@jodosha glad to do it! I appreciate the opportunity to contribute.

PR #388, opened 2015-11-16
Lotus assets integration

jodosha:
This feature introduces integration with the new `lotus-assets` gem.

## Default Assets Source

Until now the default source for assets was a public directory of an application (eg. `apps/web/public`), where we used to have a few default directories (eg `javascripts`, `stylesheets`).

The problem with this approach was the misleading name `public` in these apps, as they weren't configured with the production web server. The common deployment layout is to use one [`root` directory in Nginx configuration](http://nginx.org/en/docs/beginners_guide.html).

My deployment strategy **before** this integration was implemented, was to manually create a `public/` directory at the root of the application, copy the assets via a script and let Nginx to serve them statically from there.

```
server {
    listen 80;
    server_name localhost;

    root /home/deploy/appname/public;

    # ...
}
```

### Breaking Changes

  1. New applications will be generated with `apps/web/assets/{images,stylesheets,javascripts}` directories.
  2. The old `apps/web/public` will **NOT** be used anymore as source for assets.

## Public Directory

New Lotus projects will be generated with a `public/` directory at the root. This should be configured as `root` directory for Nginx.

All the files under it, SHOULD be ignored by Git, and not put under version control.

## New Configuration Syntax

As of v0.5 applications were generated with the following configuration syntax to add new asset sources. The public directory was implicitly added: `apps/web/public`.

```ruby
module Web
    configure do
      # ...
      # assets << [
      #   'vendor/javascripts'
      # ]
    end
  end
end
```

Because now assets can have advanced configuration (see https://github.com/lotus/assets/pull/20), we changed `assets` into a block.

New apps will be generated like this:

```ruby
module Web
    configure do
      # ...
      # assets do
      #   sources << [
      #     'vendor/assets'
      #   ]
      # end
    end
  end
end
```

### Breaking Changes

  1. Default directory will be `apps/web/assets` instead of `apps/web/public`.
  2. The `assets` method now only accepts a block, and sources must be passed inside that block (see above).

/cc @lotus/core-team


lib/lotus/config/assets.rb
```diff
@@ -48,11 +48,33 @@ def initialize(path)
         end
       end
 
+      class Sources
```
jodosha: @AlfonsoUceda This class is transitory. Assets sources will be handled at the level of Container, not single app. There is also a `method_missing` below that I accepted only with this transition in mind.

jodosha: @lotus/core-team If you will, I can implement the next feature here as well.

lib/lotus/loader.rb
```diff
@@ -98,6 +99,25 @@ def _configure_view_framework!
       end
     end
 
+    def _configure_assets_framework!
+      configuration.assets_destination = Lotus.root.join('public', 'assets')
```
jodosha: @AlfonsoUceda Sadly yes, we need that temp variable `config`. If we try to access `configuration` from `duplicate`, it will reference `Lotus::Assets.configuration`, rather than the accessor that we have in `Loader`.

In other words:
```ruby
Lotus::Assets.duplicate(namespace) do
  puts configuration # => Lotus::Assets.configuration WRONG
  puts config # => Lotus::Configuration RIGHT
end
```

PR #392, opened 2015-11-20
Fix bug where lotus port can not be set from .env

kenglxn:
This PR contains two commits, one introducing a test that demonstrates the bug, and the commit fixing this bug.


jodosha:
@kenglxn Hey, thank you. :+1:

kenglxn:
No problem. Thank you for this awesome framework 😊

PR #393, opened 2015-11-23
Timing

oreoshake:
While network timing attacks are not trivial, using secure_compare will ensure they can't happen.

See https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/metal/request_forgery_protection.rb#L318 for prior art.

No tests since this change should be transparent. Creating a timing attack PoC for a test seems like a terrible idea. Existing tests pass.


jodosha:
@oreoshake Hey, thank you very much for checking this. Merging the PR. :+1:

PR #402, opened 2015-11-27
Change hardcoded 2014 to current year

cllns:
:house_with_garden:

### DIFF
--- a/lib/lotus/templates/welcome.html.erb
+++ b/lib/lotus/templates/welcome.html.erb
@@ -44,7 +44,7 @@
       </div>
     </div>
     <div id="footer">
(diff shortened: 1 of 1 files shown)


jodosha:
@cllns @joneslee85 Sorry, but you have to fix this.

As of `v0.5.0` it prints:

```html
2014 - present
```

With this merged PR, it will be:

```html
2015 - present
```

Which it's redundant.

If we want to use a dynamic time in welcome page, it should print:

```html
2014 - 2015
```

Which can be achieved via

```erb
2014 - <%= Time.now.year %>
```

Can you please fix it?

cllns:
Oops! Great find! See #406

PR #403, opened 2015-11-27
Change welcome template h1 to base64 encoded logo

cllns:
I used [TinyPNG](https://tinypng.com/) to shrink [lotus-dark@2x.png](https://github.com/lotus/lotus.github.io/blob/build/source/images/lotus-dark%402x.png) from 6KB to 2KB, then used [base64css](http://www.base64css.com) to convert it to base64.

Here's how it looks:
screen shot 2015-11-27 at 16 15 23

Compared to:
screen shot 2015-11-27 at 16 31 04

It makes the markup considerably uglier, but is arguably simpler than dealing with loading an asset in the template for a welcome screen. base64 encoding the image is [the approach Rails takes](https://github.com/rails/rails/blob/master/railties/lib/rails/templates/rails/welcome/index.html.erb#L63) (though they do it in CSS, which I'm not opposed to doing).


jodosha:
@cllns Great, thank you :+1:

PR #404, opened 2015-11-28
Automatically use correct HTTP verbs for resourceful routes

cllns:
In the the router, we have [RESTful Resources](http://lotusrb.org/guides/routing/restful-resources/).

This PR updates the action generator to automatically set the proper HTTP methods for actions named `create`, `update` and `destroy` (i.e. `POST`, `PATCH` and `DELETE`). I don't suggest we try to guess any others, but since these are used in `resources` they're okay in my opinion.

(A user provided `--method` still overrides this)

If this PR is accepted, I'll update the [`getting-started.md` guide](https://github.com/lotus/lotus.github.io/search?p=2&q=default&type=Code&utf8=%E2%9C%93) as well.

### DIFF
--- a/lib/lotus/commands/generate/action.rb
+++ b/lib/lotus/commands/generate/action.rb
@@ -23,6 +23,15 @@ module Lotus
         # @api private
         DEFAULT_HTTP_METHOD = 'GET'.freeze
 
+        # HTTP methods used when generating resourceful actions.
+        # @since x.x.x
+        # @api private
+        RESOURCEFUL_HTTP_METHODS =  {
+          "Create" => "POST",
+          "Update" => "PATCH",
+          "Destroy" => "DELETE"
+        }.freeze
+
         def initialize(options, application_name, controller_and_action_name)
           super(options)
           if !environment.container?
@@ -92,7 +101,13 @@ module Lotus
         # @since x.x.x
         # @api private
         def http_method
-          options.fetch(:method, DEFAULT_HTTP_METHOD).downcase
+          options.fetch(:method, resourceful_http_method).downcase
+        end
+
+        # @since x.x.x
+        # @api private
+        def resourceful_http_method
+          RESOURCEFUL_HTTP_METHODS.fetch(@action_name, DEFAULT_HTTP_METHOD)
         end
 
         # @since x.x.x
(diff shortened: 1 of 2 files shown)


joneslee85:
I really like this feature, to me :+1: cc @lotus/core-team

AlfonsoUceda:
:+1:

cllns:
Terrific!

On a similar note (but as a separate PR), how would you feel about me "improving" the action generator more, to automatically guess the proper URL as well, again based on [RESTful Resources](http://lotusrb.org/guides/routing/restful-resources/)? 

For example, `lotus generate action web books#edit` would add:

```
get '/books/:id/edit', to: 'books#edit'
```

instead of: 

```
get '/books', to: 'books#edit'
```

joneslee85:
@cllns I like that idea, cc @lotus/core-team

lib/lotus/commands/generate/action.rb
```diff
@@ -92,7 +92,17 @@ def skip_view?
...
         # @api private
         def http_method
-          options.fetch(:method, DEFAULT_HTTP_METHOD).downcase
+          options.fetch(:method, resourceful_http_method).downcase
+        end
+
+        # @since x.x.x
+        # @api private
+        def resourceful_http_method
+          {
```
jodosha: @cllns Can you please extract this mapping into a constant as well? Thank you very much.

jodosha:
@cllns I love this idea. Thank you for implementing it. :+1: 
I left one comment, if you can address it, the we can merge this PR.

> On a similar note (but as a separate PR), how would you feel about me "improving" the action generator more, to automatically guess the proper URL as well, again based on RESTful Resources?

:+1:

cllns:
Great! All set.

I'm not sure if `resourceful_http_method` is the best method name anymore. Do you think `default_http_method` (or something else) would be better? Now that we have the hash named as a constant, it helps document what we're doing a bit more. Now we have two things named 'resourceful http method(s)': the constant and the method.

Also let me know if you prefer commits squashed and I'll take care of that :)

jodosha:
@cllns Thanks for the quick update. :+1: 

> Also let me know if you prefer commits squashed and I'll take care of that :)

Yes please.

jodosha:
@lotus/core-team FYI - I'm merging this, even if CI isn't passing, because of the SimpleCov breaking change, now fixed in master.

PR #405, opened 2015-11-30
Use SVG badges

olivierlacan:
Much crisper, nicer, and more scalable than the PNG versions. :-)


jodosha:
@olivierlacan Hey, thank you for this PR!

PS @lotus/core-team I'm merging this, even if CI isn't passing, but there is a breaking change with `SimpleCov` that I'm gonna fix right now in master.

PR #406, opened 2015-11-30
Fix footer to say 2014–*current year*

cllns:
Fixes #402


jodosha:
@cllns Thank you! :+1:

PR #414, opened 2015-12-07
Create resourceful routes for generated actions

cllns:
Expands on #404 to generate Resourceful routes for generated actions.

So

`lotus generate action web books#new`

will generate:

`get '/books/new', to: 'books#new'`

rather than 

`get '/books', to: 'books#new'`
etc.

Based on [the conventions](http://lotusrb.org/guides/routing/restful-resources/).

Let me know if you want me to take a different approach. I could make the route `resource :books, only: [:new]` instead, though that'd get messy after a few generated actions for the same resource, since we just pre-pend to routes, we don't read it for current state.

Do I need to do anything with the `named_route`, for the URL helpers?

### DIFF
--- a/lib/lotus/commands/generate/action.rb
+++ b/lib/lotus/commands/generate/action.rb
@@ -32,6 +32,17 @@ module Lotus
           "Destroy" => "DELETE"
         }.freeze
 
+        # For resourceful actions, what to add to the end of the base URL
+        # @since x.x.x
+        # @api private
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@cllns it looks good ;) I think the single route is a good approach instead use resource or resources because it can be more difficult to handle that.

To be conservative, I wouldn't add more, if the user wants to use a named_helper, it should add it.

cc @lotus/core-team

jodosha:
@cllns Thanks I've just merged this PR! :+1:

PR #418, opened 2015-12-12
Added attributes option when generate a model

AlfonsoUceda:
This PR introduces a new option when you generate a model
```
lotus generate model car --attributes=brand,model
```
This command creates a file like this:
```ruby
class Car
  include Lotus::entity
 
  attributes :brand, :model
end

### DIFF
--- a/lib/lotus/cli_sub_commands/generate.rb
+++ b/lib/lotus/cli_sub_commands/generate.rb
@@ -67,8 +67,11 @@ module Lotus
 
         > $ lotus generate model car
 
+        > $ lotus generate model car --attributes=brand,model
+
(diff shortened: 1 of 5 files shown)


jodosha:
@AlfonsoUceda I'm not sure about this change for a few reasons:

1. Even if we leave Lotus::Model as it is, the potential scope of this enhancement is larger that the actual. This opens to requests like typed attributes `lotus generate model car brand_id:integer name:string`.
2. We are holding our decisions about data (Lotus::Model and Lotus::Validations) until 0.7

jodosha:
@AlfonsoUceda Closing this for now.

PR #421, opened 2015-12-17
Rake tasks

### LINKED ISSUE #210: Default Rake task: environment
### Description

Add support for a default Rake task `environment`. This should serve as a base for user defined rake tasks.

Real world example:

```ruby
namespace :storage do
  desc 'Clear local storage'
  task clear: :environment do
    # Here I want to have access to the objects of my application
    # Example:
    BookRepository.clear
  end
end
```

### Implementation hints

```ruby
task :environment do
  require 'lotus/environment'
  require Lotus::Environment.new.env_config
end
```

This article can be helpful for testing this feature: https://robots.thoughtbot.com/test-rake-tasks-like-a-boss

jodosha:
This change introduces four default new Rake tasks:

  * `preload` - It **only** preloads the project configuration and env variables.
  * `environment` - It preloads the **entire** project code, including models, actions, views, etc..
  * `db:migrate` - For ecosystem compat. It's equivalent to `lotus db migrate`
  * `assets:precompile` - For ecosystem compat. It's equivalent to `lotus assets precompile`

## When to use `preload`

This Rake task is a **fast** way to preload:

  1. Gem dependencies in `Gemfile`
  2. Lotus
  3. Project configurations such as env variables and application configurations (eg. `apps/web/application.rb`

Use this as a Rake task prerequisite when you **DO NOT** need to access to project code (eg. entites, actions, views, etc..)

### Example

```ruby
# Rakefile

task foo: :preload do
  puts ENV['LOTUS_ENV']              # => "development"
  puts ENV['BOOKSHELF_DATABASE_URL'] # => "postgres://localhost/bookshelf_development"
  puts defined?(User)                # => nil
end
```

## When to use `environment`

This Rake task does:

  1. All the steps that `preload` does
  2. Load all the project code

Use this as a Rake task prerequisite when you **DO** need to access to project code (eg. entites, actions, views, etc..)

### Example

```ruby
# Rakefile

task foo: :environment do
  puts ENV['LOTUS_ENV']              # => "development"
  puts ENV['BOOKSHELF_DATABASE_URL'] # => "postgres://localhost/bookshelf_development"

  UserRepository.clear
end
```

## When to use the other Rake tasks

You **SHOULD NOT** use them directly. They are there for compatibility with Ruby ecosystem. Automated tasks from SaaS (eg. Heroku) are modeled after Rails. They expect to find Rails Rake tasks (eg. `db:migrate`).

See https://github.com/heroku/heroku-buildpack-ruby/issues/442

If you are in control of your deployment, you **SHOULD** use `lotus` commands (eg. `lotus db migrate`).

## New Projects

As of `0.6` new projects will include the following line in `Rakefile`:

```ruby
require 'lotus/rake_tasks'
```

### Upgrade notes

If projects generated before `0.6` want to take advantage of this new Rake support, they just need to add that line in their `Rakefile`.

---
Closes #210


lib/lotus/rake_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+      # use:
+      #
+      #   * lotus db migrate
+      #   * lotus assets precompile
+      #
+      # This is the preferred way to run Lotus command line tasks.
+      # Please use them when you're in control of your deployment environment.
+      namespace :db do
+        task :migrate do
+          system "bundle exec lotus db migrate"
```
thhermansen: Sorry for being a bit late; but do we need something like this to ensure if migration fail rake exits with something other than `0` as exit value?

```ruby
system("bin/hanami db migrate") || exit(1)
```

jodosha: @thhermansen yes, you're right. I created a ticket for this: https://github.com/hanami/hanami/issues/549 Thanks!

PR #424, opened 2015-12-24
Delete unnecessary Utils::Hash#symbolize! method in lotusrc

davydovanton:
Hello!

I found that `ARCHITECTURE_KEY`, `TEST_KEY` and `TEMPLATE_KEY` constants use in `DEFAULT_OPTIONS` only. Therefore I think that we can remove this constants and unnecessary `Utils::Hash#symbolize!` call frome this code.

Also I think that we need add description to all this keys to `DEFAULT_OPTIONS` documentation but I don't know how I can do it :disappointed:.

### DIFF
--- a/lib/lotus/lotusrc.rb
+++ b/lib/lotus/lotusrc.rb
@@ -29,6 +29,8 @@ module Lotus
     # @api private
     #
     # @see Lotus::Lotusrc::DEFAULT_OPTIONS
+    #
+    # @deprecated in 0.6.0
     ARCHITECTURE_KEY = 'architecture'.freeze
 
     # Test suite default value
@@ -45,6 +47,8 @@ module Lotus
     # @api private
     #
     # @see Lotus::Lotusrc::DEFAULT_OPTIONS
+    #
+    # @deprecated in 0.6.0
     TEST_KEY = 'test'.freeze
 
     # Template default value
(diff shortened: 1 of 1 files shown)


jodosha:
@davydovanton Hey, thanks for this PR! You let me to discover an interesting issue.

These constants were introduced to reduce code distance between `Lotus::Lotusrc` and the code generator that creates `.lotusrc` for new projects. I suspect that the recent CLI refactoring (#324) changed this internal safety trick.

Let me explain.

These two components (`Lotus::Lotusrc` and the related generator) can easily get out of sync. Think of a typo in one of the two, then projects will be loaded with wrong values.

---

I think we should tight again `.lotusrc` generation to `Lotus::Lotusrc` details.
The following files are templates used to generate new `.lotusrc`. They have hardcoded keys:

  * `lib/lotus/generators/application/app/lotusrc.tt`
  * `lib/lotus/generators/application/container/lotusrc.tt`

Would you like to open a different PR that interpolates keys (eg `architecture`) with the constants from `Lotus::Lotusrc` (eg. `ARCHITECTURE_KEY`)?

---

That being said, I'm closing this PR, because the suggested change, still needs these constants.

davydovanton:
okay, I think I understand your idea :)
and yes I want to open different PR with this feature

PR #427, opened 2015-12-28
[WIP] Remove app naming prefix for DATABASE_URL unix env var

joneslee85:
Why?

It would make the unix var much more friendly with Heroku.

cc @lotus/core-team

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
 # Lotus
 A complete web framework for Ruby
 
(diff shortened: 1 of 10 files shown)


davydovanton:
good idea :+1:

jodosha:
@joneslee85 This is a good idea. :+1: 

FYI - I've removed the `v0.6.0` milestone, this should be part of `v0.7.0`.

PR #428, opened 2015-12-28
Move shellwords require to lib file instead of spec file

joneslee85:


jodosha:
@joneslee85 Really good catch :+1:

PR #429, opened 2015-12-29
Fix hardcoded keys and values in .lotusrc template

davydovanton:
Reference #424

/cc @jodosha


jodosha:
@davydovanton Hey, thanks for the quick action! Merging it! :+1:

davydovanton:
nice, thanks @jodosha!

PR #433, opened 2015-12-30
Translate RACK_ENV=deployment into LOTUS_ENV=production

jodosha:
This proposal enhances our compatibility with Rack servers.
For instance, Unicorn only understands `"development"` and `"deployment"` environments. This is due to Rack basic implementation of environments.

On servers where `LOTUS_ENV` isn't set, we can fallback to this translation.


joneslee85:
:+1:

PR #435, opened 2015-12-31
Remove app prefix for generated DATABASE_URL

joneslee85:
Why?

This make Lotus more friendly with Heroku because Heroku ruby-buildpack is now using `DATABASE_URL`

### DIFF
--- a/lib/lotus/generators/application/app/.env.development.tt
+++ b/lib/lotus/generators/application/app/.env.development.tt
@@ -1,4 +1,4 @@
 # Define ENV variables for development environment
-<%= config[:upcase_app_name] %>_DATABASE_URL="<%= config[:database_config][:uri][:development] %>"
+DATABASE_URL="<%= config[:database_config][:uri][:development] %>"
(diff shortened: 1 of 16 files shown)


AlfonsoUceda:
@joneslee85 you have to change it here: https://github.com/lotus/lotus/blob/master/lib/lotus/generators/application/app/lib/app_name.rb.tt#L11-L18

joneslee85:
@AlfonsoUceda yes, it's not finished, thanks for pointing that out

AlfonsoUceda:
ah sorry I didn't see the WIP >_<

joneslee85:
@AlfonsoUceda you can review it now

jodosha:
@joneslee85 It looks good to me :+1: 

Please hold on before to merge it. We want to make the 0.6 release first.

jodosha:
@joneslee85 I reverted this because this is a breaking change and we can't block master with it. There are minor fixes that needs to be merged and released first.

PR #436, opened 2015-12-31
Set correct definition for application and project words

davydovanton:
/cc @jodosha

### DIFF
--- a/lib/lotus/cli.rb
+++ b/lib/lotus/cli.rb
@@ -18,7 +18,7 @@ module Lotus
 
     desc 'server', 'starts a lotus server'
     long_desc <<-EOS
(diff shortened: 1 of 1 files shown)


joneslee85:
I kinda like the wording project than application as it might be less confusing for users, but if we are to accept this PR, we need to update our API docs thoroughly too

davydovanton:
@joneslee85 no problem. I can send PR with this changes to documentation too :blush:

jodosha:
@joneslee85 The next version of the guides https://github.com/lotus/lotus.github.io/pull/114 is ready with the new terminology. I asked @davydovanton to open this PR.

davydovanton:
thanks for merge, @jodosha! :+1:

PR #442, opened 2016-01-02
Explicit control on assets sources for each application.

jodosha:
**This PR changes the current implementation in `master`. That means this isn't a breaking change with 0.5, but with `master`. This because assets are going to be introduced soon with 0.6. So it's just a change that developers shouldn't notice unless their using `HEAD` of this repository.**

---

Before this PR, we use to add the **implicit** following assets sources:

  * For Container it was `apps/web/assets`
  * For Application it was `app/assets`

There are two problems with this approach:

  1. It isn't consistent with other settings. For instance, for loading paths, we don't add them implicitly, but we add them to `application.rb`. (See _Example 1_).

  2. This isn't flexible. If an application doesn't have assets (eg. HTTP API app), it's forced to keep around the default directory (eg. `apps/api/assets`) or the `lotus assets precompile` command will fail and so the deploy.

---

This proposal, solves both the problems above:

  1. `Lotus::Loader` doesn't add the implicit source anymore.
  2. New projects are generated with this setting in `application.rb` (see _Example 2_).

**In this way, we have a consistent, and flexible design that doesn't require any manual intervention from developers.**

If an API app doesn't need `apps/api/assets`, they can remove the directory and comment the relative setting in `apps/api/application.rb`.

---

### Example 1

We use to generate new project with explicit load paths:

```ruby
require 'lotus/helpers'
require 'lotus/assets'

module Web
    configure do
      load_paths << [
        'controllers',
        'views'
      ]

      # ...
    end
  end
end
```

### Example 2

We now generate new projects by adding that default source for assets as an explicit setting:

```ruby
require 'lotus/helpers'
require 'lotus/assets'

module Web
    configure do
      assets do
        sources << [
          'assets'
        ]
      end

      # ....
    end
  end
end
```

### DIFF
--- a/lib/lotus/generators/app/application.rb.tt
+++ b/lib/lotus/generators/app/application.rb.tt
@@ -119,13 +119,12 @@ module <%= config[:classified_app_name] %>
       #
 
       # Specify sources for assets
-      # The directory `<%= config[:app_base_path] %>/assets` is added by default
       #
-      # assets do
-      #   sources << [
-      #     'vendor/assets'
-      #   ]
-      # end
+      assets do
+        sources << [
+          'assets'
+        ]
+      end
 
       ##
       # SECURITY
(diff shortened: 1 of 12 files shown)


AlfonsoUceda:
:+1: @jodosha great work

joneslee85:
@jodosha Nice one

PR #446, opened 2016-01-06
Update test file for Lotusrc class

davydovanton:
* Use existing `@lotusrc` instead `Lotus::Lotusrc.new(@root)`;
* Add missing test for checking returned options keys;

### DIFF
--- a/test/lotusrc_test.rb
+++ b/test/lotusrc_test.rb
@@ -2,22 +2,24 @@ require 'test_helper'
 
 describe Lotus::Lotusrc do
   describe '#options' do
-    describe 'file exists' do
-      before do
-        Dir.chdir($pwd)
-        @old_pwd = Dir.pwd
-        Dir.chdir 'test/fixtures/lotusrc/exists'
-        @root = Pathname.new(Dir.pwd)
-        @lotusrc = Lotus::Lotusrc.new(@root)
-      end
+    before do
+      Dir.chdir($pwd)
+      @old_pwd = Dir.pwd
+      Dir.chdir lotusrc_path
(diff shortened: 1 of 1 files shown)


joneslee85:
@davydovanton can you please update the description what the refactoring is?

davydovanton:
@joneslee85 yes, of course

AlfonsoUceda:
@davydovanton because you are refactoring this test, can you extract to a before block below https://github.com/lotus/lotus/pull/446/files#diff-b83ba0c0465de991a609fb85be6dd33dR4 the common things except the `Dir.chdir 'path'` line? to avoid duplicated code. what do you think?

davydovanton:
@AlfonsoUceda done. I replaced all `before` block but for each case I created different path to lotusrc file. What do you think? This is better for you?

AlfonsoUceda:
@davydovanton could you do the same for after block? then :+1: to me, thank you so much ;)

davydovanton:
oh, sorry I did not notice `after` block :disappointed: 
Done too

davydovanton:
@AlfonsoUceda could you stop current travis build? (https://travis-ci.org/lotus/lotus/builds/100646008) this will save our time :smiley:

AlfonsoUceda:
haha don't worry ;) stopped

davydovanton:
nice, thank you very much @AlfonsoUceda! :tada:

AlfonsoUceda:
@davydovanton merging, because this is only a refactor in a test. thanks!!
cc @lotus/core-team

davydovanton:
Thanks for merge!

PR #447, opened 2016-01-06
Introducing lotus runner command

AlfonsoUceda:
### Introducing lotus runner command

We are adding new command for running your scripts with the Lotus environment preloaded, because you can need to deploy a script in your production app to do tasks.

The command is `lotus runner`, you need to pass a file or an expression:
```ruby
lotus runner 'puts Lotus.env' # => development
```
or
```ruby
# my_file.rb
puts Lotus.env
```
```ruby
lotus runner ./my_file.rb # => development
```


lib/lotus/commands/runner.rb
```diff
@@ -0,0 +1,53 @@
...
+      # @param options [Hash] Environment's options
+      # @param expression_or_file [String] Expression or file to evaluate
+      #
+      # @since x.x.x
+      # @see Lotus::Environment#initialize
+      # @see Lotus::Environment#require_application_environment
+      # @see Lotus::Application.preload!
+      def initialize(options, expression_or_file)
+        @expression_or_file = expression_or_file
+        raise NoArgumentGivenException.new if @expression_or_file.nil?
```
jodosha: @AlfonsoUceda What about `ArgumentError`? It's a common Ruby pattern for this kind of situations :smile:

AlfonsoUceda: You are right and it's built-in, I'll change it. ;)

lib/lotus/commands/runner.rb
```diff
@@ -0,0 +1,48 @@
...
+      # @param options [Hash] Environment's options
+      # @param expression_or_file [String] Expression or file to evaluate
+      #
+      # @since x.x.x
+      # @see Lotus::Environment#initialize
+      # @see Lotus::Environment#require_application_environment
+      # @see Lotus::Application.preload!
+      def initialize(options, expression_or_file)
+        @expression_or_file = expression_or_file
+        raise ArgumentError.new("Provide file or expression to execute") if @expression_or_file.nil?
```
davydovanton: I think it's unnecessary string. Because if `expression_or_file` is nil in `run` command  Thor throw exception. What do you think?

jodosha: @davydovanton It will throw a `NoMethorError`, which isn't user friendly, and it will perceived as a bug in Lotus. It's better a message that explains what's the problem, like we do in this case.

davydovanton: okay, I agree. `NoMethodError ` is terrible name for error in this place. :+1:

PR #448, opened 2016-01-07
Configurable assets compressors

jodosha:
Generate new projects and applications with the following assets settings:

```ruby
# apps/web/application.rb
module Web
    configure do
      # ...

      ##
      # ASSETS
      #
      assets do
        # JavaScript compressor
        #
        # Supported engines:
        #
        #   * :builtin
        #   * :uglifier
        #   * :yui
        #   * :closure
        #
        # See: http://lotusrb.org/guides/assets/compressors
        #
        # In order to skip JavaScript compression comment the following line
        javascript_compressor :builtin

        # Stylesheet compressor
        #
        # Supported engines:
        #
        #   * :builtin
        #   * :yui
        #   * :sass
        #
        # See: http://lotusrb.org/guides/assets/compressors
        #
        # In order to skip stylesheet compression comment the following line
        stylesheet_compressor :builtin

        # Specify sources for assets
        #
        sources << [
          'assets'
        ]
      end
    end
  end
end
```

---

The default compressor for both javascripts and stylesheets is `:builtin`. **It's not an efficient compressor, but it's a good starting point, it's pure Ruby, it's fast and it doesn't require any external dependency.**

Developers can choose their own favorite engine and use it.

---

/cc @lotus/core-team

### DIFF
--- a/lib/lotus/commands/assets/precompile.rb
+++ b/lib/lotus/commands/assets/precompile.rb
@@ -18,7 +18,12 @@ module Lotus
 
         def preload_applications
           @environment.require_application_environment
-          Lotus::Application.preload!
+
+          if @environment.container?
+            Lotus::Container.new
+          else
+            Lotus::Application.preload!
+          end
         end
 
         def precompile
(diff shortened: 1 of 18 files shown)


davydovanton:
:+1:

PR #449, opened 2016-01-08
Fix problem with dash char in app name

davydovanton:
If you create new app with app architecture and name which include dash char (`-`) lotus generator created invalid names for modules. I.e this command:
```
$ lotus new test-app --architecture=app
```

Will create `Test-app` module:
``` ruby
module Test-app
    # ...
  end
end
```

However the rails solved this problem. When you create `test-app` application rails generator set `TestApp` to all partials.

Therefore this commit change all `-` chars to `_` in `app_name` before calling `classify` method. However this changes increase allocation (because `tr` method call through `method_missing`). But I hope that it's
not critical issue on this part.

### DIFF
--- a/lib/lotus/commands/new/app.rb
+++ b/lib/lotus/commands/new/app.rb
@@ -104,7 +104,7 @@ module Lotus
         end
 
         def classified_app_name
-          Utils::String.new(app_name).classify
+          Utils::String.new(app_name).classify.tr('::', '')
         end
 
         # def application_base_path
--- a/test/commands/generate/app_test.rb
+++ b/test/commands/generate/app_test.rb
@@ -45,6 +45,11 @@ describe Lotus::Commands::Generate::App do
       end
     end
 
+    it 'returns valid classified app name' do
+      command = Lotus::Commands::Generate::App.new({ architecture: 'container' }, 'awesome-test-app')
+      command.template_options[:classified_app_name].must_equal 'AwesomeTestApp'
+    end
+
     it 'create files' do
       with_temp_dir do |original_wd|
         setup_container_app(original_wd)
--- a/lib/lotus/commands/generate/app.rb
+++ b/lib/lotus/commands/generate/app.rb
@@ -96,7 +96,7 @@ module Lotus
         end
 
         def classified_app_name
-          Utils::String.new(app_name).classify
+          Utils::String.new(app_name).classify.tr('::', '')
         end
 
         def assert_application_name!(value)
--- a/test/commands/new/app_test.rb
+++ b/test/commands/new/app_test.rb
@@ -43,15 +43,15 @@ describe Lotus::Commands::New::App do
         end
       end
     end
-  end
 
-  describe 'rspec' do
-    it 'creates files' do
-      with_temp_dir do |original_wd|
-        command = Lotus::Commands::New::App.new({'test' => 'rspec'}, 'new_app')
-        capture_io { command.start }
+    describe 'rspec' do
+      it 'creates files' do
+        with_temp_dir do |original_wd|
+          command = Lotus::Commands::New::App.new({'test' => 'rspec'}, 'new_app')
+          capture_io { command.start }
 
-        assert_generated_app('rspec', original_wd)
+          assert_generated_app('rspec', original_wd)
+        end
       end
     end
   end
--- a/test/commands/new/app_test.rb
+++ b/test/commands/new/app_test.rb
@@ -54,6 +54,11 @@ describe Lotus::Commands::New::App do
         end
       end
     end
+
+    it 'returns valid classified app name' do
+      command = Lotus::Commands::New::App.new({}, 'awesome-test-app')
+      command.template_options[:classified_app_name].must_equal 'AwesomeTestApp'
+    end
   end
 
   def assert_generated_app(test_framework, original_wd)


davydovanton:
Also I think we need to check what will happen when you try to create new app with `test-app` name in container architecture. What do you think?

joneslee85:
Yes please

For now I think it is okay as is but moving forward we should think dash is a separator for namespace, but till then

lib/lotus/commands/new/app.rb
```diff
@@ -104,7 +104,7 @@ def upcase_app_name
         end
 
         def classified_app_name
-          Utils::String.new(app_name).classify
+          Utils::String.new(app_name).tr('-', '_').classify
```
AlfonsoUceda: this should be fixed in lotus/utils repo

https://github.com/lotus/utils/blob/master/lib/lotus/utils/string.rb#L73
I think we have to add https://github.com/lotus/utils/blob/master/lib/lotus/utils/string.rb#L67

davydovanton: Тo, I don't agree with you. If you open this link you see that `test-app` should be equal `Test::App`.
http://guides.rubygems.org/name-your-gem/

davydovanton:
However, when I try to create `test-app` in container architecture I have similar problem. :disappointed:

davydovanton:
@joneslee85 I fixed behavior for `lotus generate app test-app` command too. But I don't like this solution because I repeat code in `classified app_name` method. What do you think? May be is it better to replace this method to `Utils::String` or other place?

davydovanton:
and also I have a strange [failed test](https://travis-ci.org/lotus/lotus/jobs/101341838#L256) for `Static` class.

davydovanton:
@joneslee85 can I do anything to speed up merge of this PR? :)

AlfonsoUceda:
@davydovanton one thing about this, when you do `lotus new test-app --architecture=app` do you expect `Test::App` or `TestApp`

davydovanton:
I expect `TestApp` but in `app` architecture I will get `Web` :)

AlfonsoUceda:
@davydovanton then this fix should be in lotus-utils, not here

davydovanton:
I don't think so.
This changes fixed only `lotus new` command. `Utils::String.new('test-app').classify` should return `Test::App` but this behaviourus will breake lotus app. Because you can't create something like this:
``` ruby
module Test::App
    # ...
  end
end
```

AlfonsoUceda:
@davydovanton nop, that doesn't work like that.
```ruby
Lotus::Utils::String.new("Lotus::Utils").classify.to_s
 => "Lotus::Utils" #GOOD
Lotus::Utils::String.new("test-app").classify.to_s
 => "Test-app" # BAD
Lotus::Utils::String.new("test-app").classify.to_s
 => "TestApp" # EXPECTED
Lotus::Utils::String.new("test_app").classify.to_s
 => "TestApp" #GOOD
```

davydovanton:
@AlfonsoUceda yes, you're right. But if you open this page - http://guides.rubygems.org/name-your-gem/ you can see that `'test-app'` string should be equals `Test::App`.

lib/lotus/commands/new/app.rb
jodosha: @davydovanton, @AlfonsoUceda is right `#classify` doesn't work properly and needs to be fixed.

```ruby
➜ bundle console
Resolving dependencies...
irb(main):001:0> require 'lotus/utils/string'
=> true
irb(main):002:0> Lotus::Utils::String.new('app-name').classify.to_s
=> "App-name"
```

It returns `"App-name"`, which is clearly an invalid Ruby class name. We expect `"App::Name"` as result.

jodosha:
@davydovanton @AlfonsoUceda Unfortunately Lotus doesn't support deep naming. We have a convention that the top level module must have a depth of one. So `test-app` for now must be translated as `TestApp`, otherwise the world falls apart.

---

I don't like this convention, but we're forced to keep it because of `Configuration.for`:

  * https://github.com/lotus/controller/blob/master/lib/lotus/controller/configuration.rb#L90
  * https://github.com/lotus/view/blob/master/lib/lotus/view/configuration.rb#L90
  * https://github.com/lotus/assets/blob/master/lib/lotus/assets/configuration.rb#L74

This method is invoked when we _"duplicate"_ the frameworks (eg. from `Lotus::Action` we get `Web::Action`. With that method we can understand that an action that includes `Admin::Action` belongs to the `Admin` application.

---

Probably with 2.0 we'll remove this convention, but for now it has to stay. It isn't a good time to make radical changes like this.

davydovanton:
@jodosha may be is better to replace `self.for` to utils module? If you agree, I can create PR.

and what about this PR? is this changes good for you?

jodosha:
@davydovanton No, thanks but `Configuration.for` should stay as it is. That should be revisited later.

The `String#classify` fix works well :+1: - I'm gonna merge it now.

---

**We still need to respect `"app-name" => "AppName"` convention**. That means we should apply a transformation to the output of `#classify`: `Utils::String.new('app-name').gsub('::', '')`.

@davydovanton Can you please amend this PR? TY!

davydovanton:
@jodosha done.

davydovanton:
I have one failed build in travis (https://travis-ci.org/lotus/lotus/jobs/102324284). But this build failed because of tests cann't find `Assets` class.

jodosha:
@davydovanton Thank you! :+1: I've restarted that failing CI Job. Waiting for it.

jodosha:
@davydovanton Where are the tests for this? Can you please point to me where they are located? Thank you! :+1:

davydovanton:
@jodosha sorry, I forgot this :disappointed: 
done :smiley:

davydovanton:
@jodosha I don't know why but some builds failed with this error:
```
  1) Error:
Lotus::Commands::Generate::App::with valid arguments#test_0003_returns valid classified app name:
ArgumentError: App generator is only available for container architecture.
    /home/travis/build/lotus/lotus/lib/lotus/commands/generate/app.rb:110:in `assert_architecture!'
```
Do you have any ideas?

test/commands/generate/app_test.rb
```diff
@@ -45,6 +45,11 @@
       end
     end
 
+    it 'returns valid classified app name' do
+      command = Lotus::Commands::Generate::App.new({}, 'test-app')
```
jodosha: @davydovanton Can we use `"awesome-test-app"` instead?

I want to make sure that multiple dashes are handled well in the name. One day we could change `String#tr` with `String#sub` and the test still pass with `"test-app"`, but the feature would be broken.

lib/lotus/commands/new/app.rb
```diff
@@ -104,7 +104,7 @@ def upcase_app_name
         end
 
         def classified_app_name
-          Utils::String.new(app_name).classify
+          Utils::String.new(app_name).classify.tr('::', '')
```
jodosha: @davydovanton Is there any way to DRY this with `Commands::Generate::App#classified_app_name`?

davydovanton: I thought about this. But we need to replace this method to utils repository for escape DRY. Also I thought about `Lotus::Generators::Generatable` but it incorrect place for this method too.

test/commands/generate/app_test.rb
davydovanton: good idea :+1:

lib/lotus/commands/new/app.rb
jodosha: @davydovanton Fine with that for now.

jodosha:
@davydovanton This looks good now, merging :+1:

davydovanton:
@joneslee85, @AlfonsoUceda, @jodosha thanks for help :+1:

PR #459, opened 2016-01-13
Introducing hanami runner command

AlfonsoUceda:
### Introducing hanami runner command

We are adding new command for running your scripts with the Hanami environment preloaded, because you can need to deploy a script in your production app to do tasks.

The command is `hanami runner`, you need to pass a file or an expression:
```ruby
hanami runner 'puts Hanami.env' # => development
```
or
```ruby
# my_file.rb
puts Hanami.env
```
```ruby
hanami runner ./my_file.rb # => development
```


davydovanton:
@AlfonsoUceda any updates?

jodosha:
Closing this stale PR because: there isn't a shared real use case for this and because @AlfonsoUceda is taking time off from OSS.

PR #461, opened 2016-01-18
Remove default console engine from cli.rb

sergikon:
If default console engine is specified for Thor, then 'engine_lookup' in console.rb never gets to work: `load_engine options.fetch(:engine) { engine_lookup }`. As a result it's not enough to just bundle 'pry' but also have to specify it with --engine=pry.

### DIFF
--- a/lib/lotus/cli.rb
+++ b/lib/lotus/cli.rb
@@ -51,7 +51,7 @@ module Lotus
     $ > lotus console --engine=pry
     EOS
     method_option :environment, desc: 'path to environment configuration (config/environment.rb)'
-    method_option :engine, desc: "choose a specific console engine: (#{Lotus::Commands::Console::ENGINES.keys.join('/')})", default: Lotus::Commands::Console::DEFAULT_ENGINE
+    method_option :engine, desc: "choose a specific console engine: (#{Lotus::Commands::Console::ENGINES.keys.join('/')})"
     method_option :help, desc: 'displays the usage method'
     def console
       if options[:help]


joneslee85:
@sergikon It is in fact intentional design to specify `--engine`. Because imagine if you have multiple engines in your `Gemfile`, how would you be able to tell which to load?

I agree with you that adding `--engine` option explicitly every time is a pain in the neck, what I would like to suggest is default option for engine in `.lotusrc` file.

cc @jodosha @AlfonsoUceda

sergikon:
@joneslee85 I'm not questioning the option itself, but I think that specifying a default for Thor is a bug.

`commands/console.rb` has code to choose the engine if it's not specified:
~~~
def engine_lookup
  (ENGINES.find { |_, klass| Object.const_defined?(klass) } || DEFAULT_ENGINE).first
end

def engine
  load_engine options.fetch(:engine) { engine_lookup }
end
~~~

and tests reveal the intention:

~~~
describe '#engine' do
  describe 'when all the supported engines are loaded' do
...
    it 'prefers Pry' do
~~~

but all that doesn't work if `options.fetch(:engine)` returns the default option because we have specified it for Thor.

Specifying engine in `.lotusrc` was the first thing I thought, it just seems natural, so it should probably be implemented too. But it's just a further enhancement, while the above is just a bug.

joneslee85:
@sergikon i see. lgtm

joneslee85:
I can merge this now. But can you please do me favour by creating a new issue suggesting adding an engine option for `.lotusrc`?

jodosha:
@sergikon Thanks for catching this bug, I can confirm it and this PR fixes the problem.
I invite you to add tests to prevent this regression in the future. Thank you very much.

/cc @joneslee85

PR #464, opened 2016-01-19
Add `--version' and `-v' as aliases to `version'

cllns:
Just upgraded to 0.6.1, and tried to run:
```
$ lotus -v
Could not find command "_v".
```
And then:
```
$ lotus --version
Could not find command "__version".
```
Finally I typed `lotus` to see that the correct command is `lotus version`. 

This patch allows either of those alternate forms, and adds a test for `lotus version` as well.
I wouldn't be against adding `-V` too, but I don't think it's necessary.

### DIFF
--- a/test/cli_test.rb
+++ b/test/cli_test.rb
@@ -194,6 +194,35 @@ describe Hanami::Cli do
     end
   end
 
+  describe 'version' do
+    describe 'when `version` command' do
+      it 'prints Hanami version' do
+        assert_output("v#{Hanami::VERSION}\n") do
+          ARGV.replace(%w{version})
+          Hanami::Cli.start
+        end
+      end
+    end
+
+    describe 'when passing --version to hanami command, with no subcommand' do
+      it 'prints Hanami version' do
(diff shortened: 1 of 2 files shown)


cllns:
Rebased this PR from `master`, to reflect the name change from Lotus to Hanami

test/cli_test.rb
```diff
@@ -194,6 +194,29 @@
     end
   end
 
+  describe 'version' do
+    it '`version` command' do
```
joneslee85: what do you think about changing it to `prints Hanami version` instead?

cllns: Great points all around!

cllns:
@joneslee85 just took care of your comments, to make the tests better :) 
Let me know if you want them to be different at all!

jodosha:
@cllns Thank you, this can be useful :+1: 

@joneslee85 Please take care of the rest. https://github.com/hanami/hanami.github.io/pull/147

joneslee85:
@jodosha okay doky

PR #467, opened 2016-01-22
Dry generators + standardize error messages

accessd:


lib/lotus/generators/generatable.rb
```diff
@@ -46,6 +46,11 @@ def template_source_path
       def target_path
         raise NotImplementedError, "must implement the target_path method"
       end
+
+      def argument_blank?(value)
+        value.nil? || value.strip.empty?
```
jodosha: @accessd Can you please replace it with `Lotus::Utils::String.new(value).blank?`? This method has a better support for UTF-8 chars. Thank you :+1:

accessd: @jodosha There's no `blank?` method. Should I add it?

something like:
```
def blank?
  @string.nil? || /\A[[:space:]]*\z/ === @string
end
```

accessd:
rebased

accessd:
How about to add `blank?` to `Hanami::Utils::String`? https://github.com/accessd/utils/commit/ec2928384294c5941c9e429cd7a81362139630b7

lib/hanami/generators/generatable.rb
```diff
@@ -46,6 +46,11 @@ def template_source_path
       def target_path
         raise NotImplementedError, "must implement the target_path method"
       end
+
+      def argument_blank?(value)
+        value.nil? || value.strip.empty?
```
jodosha: @accessd This doesn't respect UTF-8 chars.

accessd:
updated. using the new method `Hanami::Utils::Blank.blank?`

jodosha:
@accessd Thank you. Just merged this! :+1:

PR #468, opened 2016-01-22
Console engine fallback to DEFAULT_ENGINE

railsme:
Fallback to ```DEFAULT_ENGINE``` should return array, not string ```irb```.
Calling ```#first``` will return ```i``` from DEFAULT_ENGINE as first letter in string.

### DIFF
--- a/lib/hanami/commands/console.rb
+++ b/lib/hanami/commands/console.rb
@@ -22,7 +22,7 @@ module Hanami
         'irb'  => 'IRB'
       }.freeze
 
-      DEFAULT_ENGINE = 'irb'.freeze
+      DEFAULT_ENGINE = ['irb'].freeze
 
       # @since 0.1.0
       attr_reader :options
--- a/test/commands/console_test.rb
+++ b/test/commands/console_test.rb
@@ -82,8 +82,20 @@ describe Hanami::Commands::Console do
     end
 
     describe 'when nothing is loaded' do
+      module ConsoleLoadEngineStub
+        private
+
+        def load_engine(engine)
+          Object.const_set(
+            Hanami::Commands::Console::ENGINES[engine],
+            Module.new { def self.start; end }
+          )
+        end
+      end
+
       before do
-        stub_engine 'IRB'
+        console.extend ConsoleLoadEngineStub
+        @remove_const = true
       end
 
       after do


lib/hanami/commands/console.rb
```diff
@@ -60,7 +60,7 @@ def engine
       # @since 0.1.0
       # @api private
       def engine_lookup
-        (ENGINES.find { |_, klass| Object.const_defined?(klass) } || DEFAULT_ENGINE).first
+        (ENGINES.find { |_, klass| Object.const_defined?(klass) } || [DEFAULT_ENGINE]).first
```
davydovanton: What do you think, may be is better to set `DEFAULT_ENGINE` constant to array of engine? Like this:
``` ruby
DEFAULT_ENGINE = ['irb'].freeze
```

railsme: sure, I can fix it :)

davydovanton: it'll be great :+1:

railsme: done :)

lib/hanami/commands/console.rb
```diff
@@ -22,7 +22,7 @@ def reload!
         'irb'  => 'IRB'
       }.freeze
 
-      DEFAULT_ENGINE = 'irb'.freeze
+      DEFAULT_ENGINE = ['irb'].freeze
```
joneslee85: What do you think about leaving this constant as is and change the caller so that it wraps this constant to an array instead?

accuser: Why is this failing though? IRB is included in the engines hash, but is obviously not being found. Whilst changing the way DEFAULT_ENGINE is defined / wrapped, is this still not going to fail? If IRB can't be found, it can't be found.

railsme: irb is not required by default in code, that's why we have no 'IRB' const.
After falling back to default engine ```#load_engine``` will require it and const will be defined.

joneslee85: @railsme what I meant is to change the lookup code instead:

```
      def engine_lookup
        if engine = ENGINES.find { |_, klass| Object.const_defined?(klass) }
          engine.first
        else
          DEFAULT_ENGINE
        end
      end
```

railsme: what about this solution?
```ruby
      def engine_lookup
        ENGINES.keys.find(-> { DEFAULT_ENGINE }) do |engine|
          Object.const_defined?(ENGINES[engine])
        end
      end
```

railsme: Ha, looks better, but works slower :laughing:

joneslee85: @railsme I'd go for your array solution because of the perf gain over the others. I'll wait for another round of review from @jodosha and @AlfonsoUceda

railsme: @joneslee85 your solution is the same performance. Don't know which to use.
Found another trick
```ruby
      ENGINES = {
        'pry'  => :Pry,
        'ripl' => :Ripl'
        'irb'  => :IRB
      }.freeze
```
prevents from unneeded type casts inside ruby's `Object.const_defined?` and `Object.const_get`

joneslee85: @railsme I'd leave that hash as is. Because using `if` is just slightly slower than `array` way but with the gain of readability, I'll wait for @jodosha to weigh in his opinion

jodosha:
This is probably a regression introduced by https://github.com/hanami/hanami/pull/461

jodosha:
@railsme Thank you very much for this fix :+1:

PR #471, opened 2016-01-22
Change Lotus logo to Hanami logo

cllns:
Unfortunately the new image is much larger than the old one. I used 64 colors on the PNG to convert it
then did the same process as #403 

The image is 500 pixels wide, but with the image tag I change the width to 350, so we could reduce the size of the image. The down-side is that it'd look more blurry on Retina screens (it's going to look a little blurry anyway).

screen shot 2016-01-22 at 11 39 05

@jodosha


cllns:
Size after TinyPNG, before Base64: 11233 bytes
Base64 size: 15002 bytes

I think the old one was like 2KB to 2.5KB

davydovanton:
looks cool! :+1:

artofhuman:
:+1:

jodosha:
@cllns Thank you! :+1:

PR #473, opened 2016-01-22
Fix template generation when template options is not erb

davydovanton:
If `template` option differs from `erb` in `.hanamirc` file the `hanami g app` command create `application_layout.html.erb` file.

This patch fix this problem for slim and haml templates. Also this patch fix `#hanamirc` method in `Generate::App`.

### DIFF
--- a/lib/hanami/commands/generate/app.rb
+++ b/lib/hanami/commands/generate/app.rb
@@ -7,11 +7,14 @@ module Hanami
     class Generate
       class App < Abstract
 
+        DEFAULT_TEMPLATE_FORMATS = 'erb'.freeze
+
+        APPLICATION_TEMPLATE_FORMATS = %w[erb slim haml].freeze
+
         attr_reader :base_path
 
         def initialize(options, application_name)
           super(options)
-
           assert_application_name!(application_name)
           assert_architecture!
 
@@ -23,7 +26,7 @@ module Hanami
           add_mapping('application.rb.tt', 'application.rb')
           add_mapping('config/routes.rb.tt', 'config/routes.rb')
           add_mapping('views/application_layout.rb.tt', 'views/application_layout.rb')
-          add_mapping('templates/application.html.erb.tt', 'templates/application.html.erb')
+          add_mapping("templates/application.html.#{ template_engine }.tt", "templates/application.html.#{ template_engine }")
           add_mapping('favicon.ico', 'assets/favicon.ico')
 
           add_mapping('.gitkeep', 'controllers/.gitkeep')
@@ -53,6 +56,11 @@ module Hanami
 
         private
 
+        def template_engine
+          engine = super
+          APPLICATION_TEMPLATE_FORMATS.find { |t| t == engine } || DEFAULT_TEMPLATE_FORMATS
+        end
+
         def application_base_url
           options.fetch(:application_base_url, "/#{app_name}")
         end
@@ -77,6 +85,10 @@ module Hanami
               %(#{ upcase_app_name }_SESSIONS_SECRET="#{ SecureRandom.hex(32) }"\n)
             end
           end
+        end
+
+        def hanamirc
+          @hanamirc ||= Hanamirc.new(base_path)
         end
 
         def target_path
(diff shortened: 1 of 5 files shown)


lib/hanami/commands/generate/app.rb
```diff
@@ -7,11 +7,14 @@ module Commands
     class Generate
       class App < Abstract
 
+        DEFAULT_TEMPLATE_FORMATS = 'erb'.freeze
+
+        APPLICATION_TEMPLATE_FORMATS = %w[erb slim haml].freeze
```
joneslee85: We tend to support what Tilt could. It seems to be too strict to hardcode all engines here

davydovanton: Yes, you're right. But I can create layout files for each templates.

joneslee85: I think the code should be dynamic but we could have tests to cover all supported template engines

davydovanton: Do you have any ideas how I can create any template dynamically? For example, I have `template.erb`:
``` erb
<div><%= text %></div>
```

And I need to get template.slim like this:
``` slim
div
  = text
```

Or `template.haml`:
``` haml
%div
  = text
```

jodosha:
By looking at the list of [Tilt supported engines](https://github.com/rtomayko/tilt#tilt---), these are clearly the only engines that we want to support.

There could be other engines (eg `.str`) that are not widely used.

---

We do something similar for Model commands. We support all the databases that Sequel supports, but only for a few of them (PG, MySQL and Sqlite(3)) we offer database advanced features.

---

I'd say to expand this with support for `hanami new`, by adding `--template=erb|haml|slim` argument. What do you think @hanami/core-team?

---

Ref https://github.com/hanami/hanami/issues/155
Ref https://github.com/hanami/hanami/pull/379
Ref https://github.com/hanami/hanami/issues/155

joneslee85:
@jodosha i am with you on this, we can start with a limited number of 1st-class supported engines just like we did with test frameworks and console engines.

joneslee85:
@davydovanton sorry for late reply.

* I am questioning if we should gracefully fallback to `erb`, should we fail fast and warn user to check their `.hanamirc` setting for `template` instead?
* Can you please add test for the case hanamirc with template=erb?
* Can you please add test for the case where hananirc has no template=?

Cheers

lib/hanami/generators/app/templates/application.html.slim.tt
```diff
@@ -0,0 +1,8 @@
+doctype html
+html
+  head
+    title
+      | <%= config[:classified_app_name] %>
+    - == favicon
```
jodosha: @davydovanton This should be changed into:

```slim
= favicon
```

With the actual line I got a Slim syntax error:

screen shot 2016-02-09 at 11 28 55

lib/hanami/generators/app/templates/application.html.slim.tt
```diff
@@ -0,0 +1,8 @@
...
+  body
+    - == yield
```
jodosha: Same of https://github.com/hanami/hanami/pull/473/files#r52290293

jodosha:
@joneslee85 If you omit `template` in `.hanamirc`, both with a blank value `template=` or with entirely removing the line, a new application is generated with ERb.

That being said, we can add a warning into `Hanamirc` each time we try to access a value, but it's missing in the file. I'd add it as a separated PR.

jodosha:
@davydovanton @joneslee85 Apart from the Slim issue described above, this looks good for me.

Here's the specs for the second part of this enhancement: https://github.com/hanami/hanami/issues/507

davydovanton:
@jodosha I fixed slim template

joneslee85:
:+1:

davydovanton:
thanks for merge :tada:

jodosha:
@davydovanton @joneslee85 That "fix" in master https://github.com/davydovanton/lotus/commit/35229102cf930e00c63a3882baeaaff5bf06d65a still causes problem. See https://github.com/hanami/hanami/pull/515#discussion-diff-52923862

Issue #462, opened 2016-01-19
Add console engine option to `.lotusrb`.


sergikon:
There is `--engine=[ pry | ripl | irb ]` option to `lotus console` command, which allows to choose an engine if several are installed and if built-in engine choosing algo suggests a different one.

Adding this option to `.lotusrb` can save some typing. Perhaps, `console=irb`?

jodosha:
@sergikon Thanks for starting this discussion.

With your last fix (#461) When you add `pry` to `Gemfile` it's auto-detected. That means `bundle exec lotus console` starts a Pry session.

Can you explain a bit the reason behind your suggestion? Thank you! :smile:

sergikon:
@jodosha Frankly, I don't see much of a necessity for this, but I can imagine a situation when it would help: suppose we have all the engines bundled and while one developer prefers Pry, another is using IRB. The options for IRB user are:

1. Remove Pry (noot good).
2. Specify `--engine=irb` every time.
3. Add `console=irb` to his own `.lotusrc`.

The third options seems natural, but brings another question, should `.lotusrc` be stored in repo, or rather ignored so every dev can have his own file? Or should there be another `.lotusrc.dev` as with `.env`?

So, it's a low priority enhancement suggestion, that I think can wait till someone really needs it and willing to work on.

jodosha:
@sergikon Thanks for the quick reply.

> suppose we have all the engines bundled

We [support only](https://github.com/lotus/lotus/blob/master/lib/lotus/commands/console.rb#L19) IRb, Pry and Ripl. The first one is always present because part of stdlib. The other two must be intentionally bundled by the team, but there is an overlap between them, and the team should decide which one to add and what to take out.

> while one developer prefers Pry, another is using IRB

This is a case that we don't want to support because it would involve to:

  * Change the format of `.lotusrc`, and decide what to do if `console` is missing.
  * Because the main `.lotusrc` MUST be added to SCM, so we need to add a fallback. For instance `.lotusrc.local` or `.lotusrc.dev`, but this is too much IMO

Closing this, thanks again for your analysis :+1:

PR #478, opened 2016-01-24
Add :console option for hanami `new` CLI

joneslee85:
Introducing a new `:console` option for `hanami new` command. This allows user to choose from 3 supported console engines, that is `irb` (default), `pry` and `ripl`.

Depending on the architecture, the new project will be generated with `console` key in `.hanamirc` and gem appended to `Gemfile`

cc @hanami/core-team


jodosha:
@joneslee85 What's the point of this feature? We're able to pick the wanted engine by looking at `Gemfile`.

Please have a look at https://github.com/hanami/hanami/issues/462

joneslee85:
@jodosha this allow user to choose engine when they first create the project. 

It is useful for users who want to create reusable project template.

I made a mistake with introducing `console` entry for `.hanamirc`. That part can be removed.

PR #486, opened 2016-01-27
fix Hanami::Static for public subdirectories

railsme:
To reproduce the issue run these commands:
```
hanami new reprod
cd reprod
hanami generate action web home#index --url=/
hanami generate app admin
hanami generate action admin dashboard#index --url=/
hanami s
```

First request will respond as it should.
The second request will be 404 _File not found: /admin_

It depends on reimplemented ```Hanami::Static#can_serve```.
Variable ```destination``` returns false positive result for directories that are not possible to serve.

### DIFF
--- a/lib/hanami/static.rb
+++ b/lib/hanami/static.rb
@@ -32,7 +32,7 @@ module Hanami
 
     def can_serve(path, original = nil)
       destination = Dir[PUBLIC_DIRECTORY].find do |file|
-        file.index(path).to_i > 0
+        file.index(path).to_i > 0 && file.end_with?(path) && File.file?(file)
       end
 
       (super(path) || !!destination) && _fresh?(original, destination)


davydovanton:
@railsme thanks! :+1: 
Could you add test for this case?

railsme:
@davydovanton could you help a bit, I cannot imagine how to test this private method in rack middleware :smile_cat:

davydovanton:
@railsme sorry, I didn't notice that it's a private method :disappointed:

railsme:
@davydovanton that wasn't sarcasm. I really want to know how to test such behaviour in rack middleware🤘

davydovanton:
@railsme so, if you're wondering how I would do:
I not found any tests for `Static` class that's why I'd create new test file for this class.
And also I'd create tests for public `#call` method and add special cases which check behaviour for `#can_serve` method.

railsme:
@davydovanton, I think the best place to get some info - Rack::Static tests. Thank you for advice and let's make hanami better! :smile_cat:

jodosha:
@railsme Thank you for this PR, but I'm closing this in favor of https://github.com/hanami/hanami/pull/506 which includes tests too.

PR #491, opened 2016-01-30
Fix model generation for camelcased app name

### LINKED ISSUE #490: Different in behavior of 'hanami new' and 'hanami generate model'
I found difference in behavior of `hanami new` and `hanami generate model` when appname includes capital letters.

For example,  `MyProject/lib` directory is as following after I ran `hanami new MyProject`

```
MyProject/lib/
  - config/
  - myproject/
  - myproject.rb
```

And it is as following after I ran `hanami generate model model`

```
MyProject/lib/
  - config/
  - myproject/
  - MyProject/
      - repositories/model.rb
      - entities/model.rb
  - myproject.rb
```

There is `Dir["#{ __dir__ }/investgate/**/*.rb"].each { |file| require_relative file }` in `myproject.rb`. So, Model won't loaded.

davydovanton:
Fixed #490

### DIFF
--- a/lib/hanami/commands/generate/model.rb
+++ b/lib/hanami/commands/generate/model.rb
@@ -57,7 +57,7 @@ module Hanami
         end
 
         def model_root
-          Pathname.new('lib').join(::File.basename(Dir.getwd))
+          Pathname.new('lib').join(app_name)
         end
 
         # @since 0.5.0
@@ -75,13 +75,13 @@ module Hanami
         # @since 0.5.0
         # @api private
         def entity_spec_path
-          target_path.join('spec', ::File.basename(Dir.getwd), 'entities', "#{ model_name_underscored }_spec.rb")
+          target_path.join('spec', app_name, 'entities', "#{ model_name_underscored }_spec.rb")
         end
 
         # @since 0.5.0
         # @api private
         def repository_spec_path
-          target_path.join('spec', ::File.basename(Dir.getwd), 'repositories',
+          target_path.join('spec', app_name, 'repositories',
             "#{ model_name_underscored }_repository_spec.rb")
         end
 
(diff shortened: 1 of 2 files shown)


zakuro9715:
Why don't we use ApplicationName?

`hanami new` command uses ApplicationName.
https://github.com/hanami/hanami/blob/master/lib/hanami/commands/new/abstract.rb#L67-L73

I think we might want to use ApplicationName like `hanami new` for model generation.

lib/hanami/commands/generate/model.rb
```diff
@@ -57,7 +57,7 @@ def model_name_valid
         end
 
         def model_root
-          Pathname.new('lib').join(::File.basename(Dir.getwd))
+          Pathname.new('lib').join(app_name_downcased)
```
joneslee85: what do you think about renaming it to `app_name` instead?

test/commands/generate/model_test.rb
```diff
@@ -33,7 +33,16 @@
...
       end
     end
+  end
 
+  describe 'sanitizes application name' do
+    it 'downcases it' do
+      with_temp_dir('TestApp') do |original_wd|
+        command = Hanami::Commands::Generate::Model.new({}, 'car')
+        capture_io { command.start }
+        assert_generated_file(original_wd.join('test/fixtures/commands/generate/model/car.rb'), 'lib/testapp/entities/car.rb')
```
joneslee85: I don't think `testapp` is the right application name for `TestApp`, I'd prefer it follows the same naming convention of Ruby, that is `test_app`.

davydovanton: I don't know because hanami generate `testapp.rb` db config file. @jodosha what do you think?

lib/hanami/commands/generate/model.rb
davydovanton: yep, I like it :+1:

test/commands/generate/model_test.rb
jodosha: :+1: For @joneslee85 suggestion.

PR #492, opened 2016-01-31
Add missing file adapter description to generators

davydovanton:
I found that this information [described in docs](http://hanamirb.org/guides/models/overview/) but not described in generated files.


joneslee85:
:+1:

davydovanton:
thanks :star2:

jodosha:
This is good :+1:

Issue #495, opened 2016-02-06
Error on deploy "bookshelf" with rspec to heroku


vyper:
I used the commands based in main page:

```bash
gem install hanami # installed 0.7.1
hanami new bookshelf --test=rspec
cd bookshelf && bundle
# added this file
echo 'body { background-color: #ccc; }' >  apps/web/assets/stylesheets/application.css
git add . && git commit -m "Initial commit"
heroku apps:create
heroku config:add BOOKSHELF_DATABASE_URL="file:///db/bookshelf"
heroku config:add HANAMI_ENV=production
git push heroku master # here, the error
```

# Error
```
[...]
remote: -----> Writing config/database.yml to read from DATABASE_URL
remote:        Could not detect rake tasks
remote:        ensure you can run `$ bundle exec rake -P` against your app with no environment variables present
remote:        and using the production group of your Gemfile.
remote:        rake aborted!
remote:        LoadError: cannot load such file -- rspec/core/rake_task
remote:        /tmp/build_ac7604013813153e87aa6e06f7e07536/Rakefile:3:in `require'
remote:        /tmp/build_ac7604013813153e87aa6e06f7e07536/Rakefile:3:in `<top (required)>'
remote:        /tmp/build_ac7604013813153e87aa6e06f7e07536/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/rake_module.rb:28:in `load'
remote:        /tmp/build_ac7604013813153e87aa6e06f7e07536/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/rake_module.rb:28:in `load_rakefile'
remote:        /tmp/build_ac7604013813153e87aa6e06f7e07536/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/application.rb:689:in `raw_load_rakefile'
remote:        /tmp/build_ac7604013813153e87aa6e06f7e07536/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/application.rb:94:in `block in load_rakefile'
remote:        /tmp/build_ac7604013813153e87aa6e06f7e07536/vendor/bundle/ruby/2.2.0/gems/rake-10.5.0/lib/rake/application.rb:176:in `standard_exception_handling'
[...]

PR #496, opened 2016-02-07
Fix Rakefile generation when using RSpec

vyper:
During the deploy (probably production environment) have not rspec, then,
there is an exception because of `require 'rspec/core/rake_task'`

Fix for: #495


joneslee85:
thx

jodosha:
@vyper @joneslee85 thank you all :+1:

Issue #466, opened 2016-01-21
`config/routes.rb` routes file as DSL or Router initialization


cllns:
Right now, setting your `app/web/config/routes.rb` to:
```
Router.new do
  resources :books
end
```
won't work. Instead, it should just be
```
resources :books
```
That is, the `config/routes.rb` should just be the DSL.

We should do one of the following:

1. Allow a `Lotus::Router` from that file, instead of assuming it's the only ever the DSL block. [This is the file](https://github.com/lotus/lotus/blob/master/lib/lotus/config/routes.rb) that would need to be edited, I think. Alternately, it could be done in the superclass, but I'm not sure if that'll ever be an issue for any Mapping other than Routes.

2. Throw an error if the block yields a `Lotus::Router` and instruct the user that the `config/routes.rb` file should just be the DSL, not the entiry `Lotus::Router` initialization.

3. Modify the documentation to mention that `config/routes.rb` should be the DSL only. A big warning should go in `lotus/router`'s README, since that assumes you want the `Lotus::Router` object. People who are using `lotus/lotus` need to know to only provide the DSL.

I like **1)** more, since it's more flexible, and someone can look at the documentation for `Lotus::Router` and use it immediately, without having to know that `config/routes.rb` should only be the DSL.

Meanwhile **2)** is probably just as much code, since you have to get the Router from the file anyway, but is better if we want enforce that everyone's routes file looks the same. That is, it might be confusing if the `Lotus::Router.new do` / `end` block is optional.

(Thanks to @fespinoza for bringing up this issue in the chat.)

jodosha:
@cllns @fespinoza Thanks for starting this discussion.

Right now we generate a file like this:

```ruby
# Configure your routes here
# See: http://www.rubydoc.info/gems/lotus-router/#Usage
```

We probably want to generate this:

```ruby
# Configure your routes here
# See: http://lotusrb.org/guides/routing/overview/
```

Do you think that pointing to the Lotus guides, instead of the API docs can change the perception? The first documents full stack apps, while the latter is for standalone usage of the router.

---

FYI - As of `lotus-router-0.5.0`, you can wrap that DSL with `Lotus:Router.define`. [See the PR](https://github.com/lotus/router/pull/74).

fespinoza:
that would be a start, as using lotus for the first time, that confused me a lot, maybe it's possible to give an example even as in:

```ruby
# Configure your routes here
# See: http://lotusrb.org/guides/routing/overview/
#
# Example
[...]

PR #497, opened 2016-02-07
Improve reference for the template of routes.rb

fespinoza:
Fixes https://github.com/hanami/hanami/issues/466

Now, when using the generator, the link in the comment will point to
hanami's routing overview which is the context on which the file will be
used, that including an example.


davydovanton:
LGFM :shipit:

jodosha:
@fespinoza @joneslee85 Thank you :+1:

Issue #499, opened 2016-02-08
Introduce new convention for project/app naming


joneslee85:
We want to offer a standard and consistent way of generating new project/application name:
  * Adopt same gem naming / file naming convention as of [How to name your gem](http://guides.rubygems.org/name-your-gem/)
  * CLI is able to takes in both snakecase/camelcase and convert them to snakecase naming

## Actual

```
hanami new TestApp
# => testapp would be generated
```

## Proposed

```
hanami new TestApp
# => test_app
```

```
hanami new Test_app
# => test_app
```

```
hanami new Namespace::TestApp
# => create folder namespace/test_app
```

davydovanton:
I like this proposal :+1:

jodosha:
@joneslee85 Can you please explain which Ruby constant names would be generated by this proposal? Eg. What `Test_app` should produce?

What's the idea behind this?

joneslee85:
@jodosha 

`Test_app` would produce `test_app` project folder.

The need behind is the consistency for naming project and generated code.

jodosha:
This is fixed now in master.

To try it, please follow these instructions:
https://github.com/hanami/hanami#how-to-use-hanami-head

But use this last command instead:

```shell
➜ bundle exec hanami new Namespace::TestApp --hanami-head=true
      create  .hanamirc
      create  .env.development
      create  .env.test
      create  Gemfile
      create  config.ru
      create  config/environment.rb
      create  lib/namespace/test_app.rb
      create  public/.gitkeep
      create  config/initializers/.gitkeep
      create  lib/namespace/test_app/entities/.gitkeep # THIS LOOKS RIGHT
      create  lib/namespace/test_app/repositories/.gitkeep
      create  lib/namespace/test_app/mailers/.gitkeep
      create  lib/namespace/test_app/mailers/templates/.gitkeep
      create  spec/namespace/test_app/entities/.gitkeep
      create  spec/namespace/test_app/repositories/.gitkeep
      create  spec/namespace/test_app/mailers/.gitkeep
      create  spec/support/.gitkeep
      create  db/.gitkeep
      create  Rakefile
      create  spec/spec_helper.rb
      create  spec/features_helper.rb
      create  .gitignore
         run  git init . from "."
      create  apps/web/application.rb
      create  apps/web/config/routes.rb
      create  apps/web/views/application_layout.rb
      create  apps/web/templates/application.html.erb
      create  apps/web/assets/favicon.ico
[...]

PR #502, opened 2016-02-08
Introduce new convention for project/app naming

davydovanton:
Reference #499

What will happen after calling `hanami new MyApp`:
screenshot 2016-02-08 20 05 40

### DIFF
--- a/lib/hanami/application_name.rb
+++ b/lib/hanami/application_name.rb
@@ -1,3 +1,5 @@
+require 'hanami/utils/string'
+
 module Hanami
   # An application name.
   #
@@ -28,7 +30,7 @@ module Hanami
     #
     # @since 0.2.1
     def initialize(name)
-      @name = sanitize(name)
+      @name = sanitize(Hanami::Utils::String.new(name).underscore)
       ensure_validity!
     end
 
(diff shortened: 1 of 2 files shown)


lib/hanami/application_name.rb
```diff
@@ -28,7 +30,7 @@ class ApplicationName
     #
     # @since 0.2.1
     def initialize(name)
-      @name = sanitize(name)
+      @name = sanitize(Hanami::Utils::String.new(name).underscore)
```
davydovanton: what do you think, maybe 'Utils' module is better place for `#sanitize` method?

joneslee85:
@davydovanton let's hold off the implementation until we reach a consensus

davydovanton:
@joneslee85 okay. I agree with you :+1:

jodosha:
@joneslee85 @davydovanton The more I see these PRs fixing bugs with naming conventions, the more I think we should extract this logic into an internal object like `Hanami::ComponentName` (sorry for the poor name :sweat_smile:) and use it for all the code generators.

davydovanton:
@jodosha what do you think may be is better to close this PR?

jodosha:
@davydovanton Yes, I think this is fixed now. Thanks! :+1:

Issue #501, opened 2016-02-08
Relaxed Hanami dependency for new projects


jodosha:
## Actual

We generate new projects with a strict dependency on `hanami` gem.

```ruby
# Gemfile
gem 'hanami', '0.8.0'
```

## Proposed

Because we follow SemVer, we can relax the dependency. In case of a patch release, developers can update their apps via a simple `bundle update hanami`, **instead of edit their `Gemfile`**.

New generated projects should have this `Gemfile`.

```ruby
# Gemfile
gem 'hanami', '~> 0.8'
```

---

/cc @hanami/contributors WDYT?

pascalbetz:
+1

accuser:
Better to make use of semantic versioning and relax the dependency. If anyone wants to freeze the gem dependency, they will know what they are doing and why.

:+1:

joneslee85:
Me like. Me plus one

hlegius:
:+1:

jodosha:
Solved by #503

PR #503, opened 2016-02-08
Generate less strict hanami version on Gemfile

davydovanton:
Reference #501

/cc @jodosha

### DIFF
--- a/lib/hanami/commands/new/abstract.rb
+++ b/lib/hanami/commands/new/abstract.rb
@@ -15,7 +15,8 @@ module Hanami
         DEFAULT_ARCHITECTURE = 'container'.freeze
         DEFAULT_APPLICATION_BASE_URL = '/'.freeze
 
-        attr_reader :options, :target_path, :database_config, :test_framework
+        attr_reader :options, :target_path, :database_config,
+          :test_framework, :hanami_model_version
 
         def initialize(options, name)
           @options = Hanami::Utils::Hash.new(options).symbolize!
@@ -86,8 +87,8 @@ module Hanami
           File.directory?(target.join('.git'))
         end
 
-        def hanami_model_version
-          @hanami_model_version
+        def hanami_version
+          "~> #{Hanami::VERSION.scan(/\A\d{1,2}\.\d{1,2}/).first}"
         end
 
         def hanami_head?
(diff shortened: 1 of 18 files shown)


joneslee85:
+1

lib/hanami/generators/application/app/Gemfile.tt
```diff
@@ -15,7 +15,7 @@ gem 'hanami-mailer',      require: false, github: 'hanami/mailer'
 gem 'hanami-assets',      require: false, github: 'hanami/assets'
 gem 'hanami',                             github: 'hanami/hanami'
 <%- else -%>
-gem 'hanami',       '<%= Hanami::VERSION %>'
+gem 'hanami',       '~> <%= Hanami::VERSION_WITHOUT_TINY %>'
```
jodosha: @davydovanton We should unify with `hanami-model` definition and use `<%= config[:hanami_version] %>`.
The logic for defining that version should take care of `Hanami::VERSION`, but **NOT** to define a new constant.

Example:

```ruby
version = Hanami::VERSION.scan(/\A\d{1,2}\.\d{1,2}/).first
```

AlfonsoUceda: @davydovanton could you take care of this? thanks ;)

davydovanton: @AlfonsoUceda yes, of course :smiley:

davydovanton:
@AlfonsoUceda @jodosha I created `hanami_version` option.

lib/hanami/commands/new/abstract.rb
```diff
@@ -90,6 +90,10 @@ def hanami_model_version
           @hanami_model_version
         end
 
+        def hanami_version
+          Hanami::VERSION.scan(/\A\d{1,2}\.\d{1,2}/).first
```
jodosha: @davydovanton Please have a look at https://github.com/hanami/hanami/pull/503/files#diff-b6e68c896b787eed14d8d92616b64789R18, there is a discrepancy between `hanami` and `hanami-model` versions. I think we should probably let `#hanami_version` to include `~>` as prefix of the version. Eg. `"~> 0.8"`.

In this way the Ruby code is in complete control of the version number and the template is logic-less, as it only interpolates the value.

davydovanton: Sounds reasonably :+1:

davydovanton:
@jodosha I fixed you comment and also I replaced getter `hanami_model_version` method to `attr_reader` construction.

jodosha:
@davydovanton This looks good :+1: Merging it, thank you :+1:

davydovanton:
@jodosha thanks for review :+1:

PR #504, opened 2016-02-09
Implement full CSP to enhance security

joneslee85:
[Content Security Policy](http://www.w3.org/TR/CSP) or CSP is a set of standardised HTTP headers that whitelists valid sources that are allowed. This reduces of risks being attacked with XSS. 

We want to offer the top-notch security practices out of the box for the framework. Currently Hanami has implemented 75% of supported directives. This PR is to complete the full support of directives as specified in CSP Level 3 spec.

## Browser support status

* [CSP 1.0](http://caniuse.com/#feat=contentsecuritypolicy)
* [CSP Level 2](http://caniuse.com/#feat=contentsecuritypolicy2)

## TODO:

* [X] object-src
* [X] child-src
* [X] media-src
* [X] base-uri
* [X] plugin-types
* [X] manifest-src
* [X] frame-ancestors
* [x] reflected-xss
* [x] referrer
* [X] form-action

### DIFF
--- a/lib/hanami/config/security.rb
+++ b/lib/hanami/config/security.rb
@@ -50,7 +50,7 @@ module Hanami
         if value.nil?
           @content_security_policy
         else
-          @content_security_policy = value
+          @content_security_policy = value.split(';').map(&:strip).join('; ')
         end
       end
     end
--- a/lib/hanami/generators/app/application.rb.tt
+++ b/lib/hanami/generators/app/application.rb.tt
@@ -200,7 +200,28 @@ module <%= config[:classified_app_name] %>
       #  * http://content-security-policy.com/
       #  * https://developer.mozilla.org/en-US/docs/Web/Security/CSP/Using_Content_Security_Policy
       #
-      security.content_security_policy "default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self';"
+      # Content Security Policy references:
+      #
+      #  * https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives
+      #
+      security.content_security_policy %{
+        form-action 'self';
+        referrer origin-when-cross-origin;
+        reflected-xss block;
+        frame-ancestors 'self';
+        base-uri 'self';
+        default-src 'none';
+        script-src 'self';
+        connect-src 'self';
+        img-src 'self';
+        style-src 'self';
+        font-src 'self';
+        object-src 'self';
+        plugin-types application/pdf;
+        child-src 'self';
+        frame-src 'self';
+        media-src 'self'
+      }
 
       ##
       # FRAMEWORKS
--- a/test/config/security_test.rb
+++ b/test/config/security_test.rb
@@ -1,0 +1,55 @@
+require 'test_helper'
+
+describe Hanami::Config::Security do
+  let(:security) { Hanami::Config::Security.new }
+
+  describe '#content_security_policy' do
+    describe 'when the argument is nil' do
+      before do
+        security.content_security_policy('img-src localhost')
+      end
+
+      it 'returns the value' do
+        security.content_security_policy.must_equal('img-src localhost')
+        security.content_security_policy(nil).must_equal('img-src localhost')
+      end
+    end
+
+    describe 'when the argument is not single line string' do
+      it 'assigns new value' do
+        security.content_security_policy('script-src example')
+        security.content_security_policy.must_equal('script-src example')
+      end
+    end
+
+    describe 'when the argument is multiline string' do
+      it 'concatenates all lines into one and assigns new value' do
+        security.content_security_policy(%{
+          script-src example;
+          img-src localhost
+        })
+        security.content_security_policy.must_equal('script-src example; img-src localhost')
+      end
+    end
+  end
+
+  describe '#x_frame_options' do
+    describe 'when the argument is nil' do
+      before do
+        security.x_frame_options('ALLOW ALL')
+      end
+
+      it 'returns the value' do
+        security.x_frame_options.must_equal('ALLOW ALL')
+        security.x_frame_options(nil).must_equal('ALLOW ALL')
+      end
+    end
+
+    describe 'when the argument is not nil' do
+      it 'assigns new value' do
+        security.x_frame_options('DENY')
+        security.x_frame_options.must_equal('DENY')
+      end
+    end
+  end
+end
--- a/test/fixtures.rb
+++ b/test/fixtures.rb
@@ -40,6 +40,25 @@ module CoffeeShop
 
       default_response_format :html
 
+      security.x_frame_options "DENY"
+      security.content_security_policy %{
+        form-action 'self';
+        referrer origin-when-cross-origin;
+        reflected-xss block;
+        frame-ancestors 'self';
+        base-uri 'self';
+        default-src 'none';
+        connect-src 'self';
+        img-src 'self';
+        style-src 'self';
+        font-src 'self';
+        object-src 'self';
+        plugin-types application/pdf;
+        child-src 'self';
+        frame-src 'self';
+        media-src 'self'
+      }
+
       scheme 'https'
       host   'hanami-coffeeshop.org'
 
--- a/test/loader_test.rb
+++ b/test/loader_test.rb
@@ -45,6 +45,14 @@ describe Hanami::Loader do
         CoffeeShop::Controller.configuration.default_response_format.must_equal(:html)
       end
 
+      it 'configures controller to use default headers' do
+        CoffeeShop::Controller.configuration.default_headers.
+          must_equal({
+            "X-Frame-Options" => "DENY",
+            "Content-Security-Policy" => "form-action 'self'; referrer origin-when-cross-origin; reflected-xss block; frame-ancestors 'self'; base-uri 'self'; default-src 'none'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self'; object-src 'self'; plugin-types application/pdf; child-src 'self'; frame-src 'self'; media-src 'self'",
+          })
+      end
+
       it 'generates controllers namespace' do
         assert defined?(CoffeeShop::Controllers), 'expected CoffeeShop::Controllers'
       end
--- a/lib/hanami/generators/application/app/config/application.rb.tt
+++ b/lib/hanami/generators/application/app/config/application.rb.tt
@@ -197,7 +197,28 @@ module <%= config[:classified_app_name] %>
       #  * http://content-security-policy.com/
       #  * https://developer.mozilla.org/en-US/docs/Web/Security/CSP/Using_Content_Security_Policy
       #
-      security.content_security_policy "default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self';"
+      # Content Security Policy references:
+      #
+      #  * https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives
+      #
+      security.content_security_policy %{
+        form-action 'self';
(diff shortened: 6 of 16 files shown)


jodosha:
@joneslee85 This is :100: - Why still WIP?

jodosha:
@joneslee85 Can we think of another PR for partial [CSP 2.0](http://caniuse.com/#feat=contentsecuritypolicy2) support? I'm thinking of [`form-action`](https://www.w3.org/TR/CSP/#directive-form-action) directive that is another defense against CSRF attacks.

Thank you :+1:

joneslee85:
@jodosha I want to support up to Level 3 of CSP. I am doing one by one with manual testing to make sure things works correctly. Sorry, it would take few more days.

For CSP 2.0, I want to do in a separate PR for the sake of review-ability.

AlfonsoUceda:
@joneslee85 perfect :+1:

joneslee85:
@jodosha @AlfonsoUceda it is now ready for review

jodosha:
@joneslee85 I merged master and pushed on this branch, this should fix the build.

jodosha:
@joneslee85 It didn't worked, BTW. See https://travis-ci.org/hanami/hanami/builds/110629855

aderyabin:
@joneslee85 :trophy:

lib/hanami/generators/app/application.rb.tt
```diff
@@ -200,7 +200,11 @@ module <%= config[:classified_app_name] %>
       #  * http://content-security-policy.com/
       #  * https://developer.mozilla.org/en-US/docs/Web/Security/CSP/Using_Content_Security_Policy
       #
-      security.content_security_policy "default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self';"
+      # Content Security Policy references:
+      #
+      #  * https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives
+      #
+      security.content_security_policy "form-action 'self'; referrer origin-when-cross-origin; reflected-xss block; manifest-src 'self'; frame-ancestors 'self'; base-uri 'self'; default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self'; object-src 'self'; plugin-types application/pdf application/x-shockwave-flash application/x-java-applet; child-src 'self'; media-src 'self'"
```
jodosha: @joneslee85 Can we make `#content_security_policy` to accept multiline strings [as we agreed](https://github.com/hanami/hanami/issues/510#issuecomment-183095185) on #510 ?

test/fixtures.rb
```diff
@@ -40,6 +40,10 @@ class Application < Hanami::Application
 
       default_response_format :html
 
+      security.x_frame_options "DENY"
+      security.content_security_policy "form-action 'self'; referrer origin-when-cross-origin; reflected-xss block; manifest-src 'self'; frame-ancestors 'self'; base-uri 'self'; default-src 'none'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self'; object-src 'self'; plugin-types application/pdf application/x-shockwave-flash application/x-java-applet; child-src 'self'; media-src 'self'"
```
jodosha: @joneslee85 What happens when both CSP and CSPRO are enabled? We should send only one header, not both.

test/loader_test.rb
```diff
@@ -45,6 +45,15 @@
         CoffeeShop::Controller.configuration.default_response_format.must_equal(:html)
       end
 
+      it 'configures controller to use default headers' do
+        CoffeeShop::Controller.configuration.default_headers.
+          must_equal({
+            "X-Frame-Options" => "DENY",
+            "Content-Security-Policy" => "form-action 'self'; referrer origin-when-cross-origin; reflected-xss block; manifest-src 'self'; frame-ancestors 'self'; base-uri 'self'; default-src 'none'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self'; object-src 'self'; plugin-types application/pdf application/x-shockwave-flash application/x-java-applet; child-src 'self'; media-src 'self'",
```
jodosha: @joneslee85 Same of https://github.com/hanami/hanami/pull/504/files#r53554599 :wink:

jodosha:
@joneslee85 I fixed the database failing test, now the build fails because of the changes of this PR. It should be trivial to fix them. Have a look at https://travis-ci.org/hanami/hanami/builds/110638517 :wink:

test/fixtures.rb
joneslee85: Very good question. 
According to the spec, it is totally _okay_ to have both defined (that is having 2 headers) as both will be honored. As long as they follow one simple rule:

* Each resource should be _ONLY_ defined either in one type of header

joneslee85: You see, I've been wondering if we should introduce helper to help our use guard against the wrongly usage of CSP header because not many would go through from start to ending of W3C spec. There are certain policy that can be only used in Report Only mode.

test/loader_test.rb
joneslee85: I am working that out myself, the header does not work if it contains new line and empty spaces in newlines.

test/fixtures.rb
jodosha: @joneslee85 Every time I used CSPRO was for projects already in production since years. That header is useful to collect data for a long period of time and then plan the switch off with CSP.

That represent a soft transition from an in browser warning that you can collect to a browser exception that stops the page to work as usual.

Do we have this legacy problem with Hanami?

The idea of introducing CSP in the early versions of the framework, was to avoid that painful transition. I think we can safely remove CSPRO support.

In case someone needs it, they can always use: 1. Default headers 2. a shared module for each action (hint `controller.prepare`) and always send that report-only header.

What do you think?

test/loader_test.rb
jodosha: @joneslee85 I was referring to CSP vs CSPRO debate. :wink: 

For multiline strings my idea is to transform the given string into a CSP compatible one. That means to change how [`Hanami::Config::Security#content_security_policy`](https://github.com/hanami/hanami/blob/master/lib/hanami/config/security.rb#L49) works

```ruby
  def content_security_policy(value = nil)
        if value.nil?
          @content_security_policy
        else
          @content_security_policy = value # TRANSFORM THE INPUT HERE
        end
      end
```

test/fixtures.rb
joneslee85: > Do we have this legacy problem with Hanami?
No, we don't. 

> I think we can safely remove CSPRO support.
Sure, we can do that.

> In case someone needs it, they can always use: 1. Default headers 2. a shared module for each action (hint controller.prepare) and always send that report-only header.
:+1:

joneslee85:
@jodosha please help review

test/config/security_test.rb
```diff
@@ -0,0 +1,45 @@
+require 'test_helper'
+
+describe Hanami::Config::Security do
+  let(:security) { Hanami::Config::Security.new }
+
+  describe '#content_security_policy' do
+    describe 'when the argument is nil' do
+      before do
+        security.content_security_policy('img-src localhost')
```
jodosha: @joneslee85 Shall we try with multiline strings here?

joneslee85: @jodosha test added

lib/hanami/generators/app/application.rb.tt
```diff
@@ -200,7 +200,28 @@ module <%= config[:classified_app_name] %>
...
-      security.content_security_policy "default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self';"
+      # Content Security Policy references:
+      #
+      #  * https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives
+      #
+      security.content_security_policy %{
+        form-action 'self';
+        referrer origin-when-cross-origin;
+        reflected-xss block;
+        manifest-src 'self';
```
jodosha: @joneslee85 According to [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#manifest-src):

> The `manifest-src` directive specifies which manifest can be applied to the resource.
> **Note**: If this directive is absent, the user agent will look for the `default-src` directive.

This should be a ["source list"](https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives) (aka URL), not a keyword like `'self'`.

Because we don't support [Web App Manifest](https://www.w3.org/TR/appmanifest/), we should remove this entry at all, and let the browser to fallback to `default-src`, as described by Mozilla docs above.

lib/hanami/generators/app/application.rb.tt
```diff
@@ -200,7 +200,28 @@ module <%= config[:classified_app_name] %>
...
+        frame-ancestors 'self';
+        base-uri 'self';
+        default-src 'none';
+        script-src 'self';
+        connect-src 'self';
+        img-src 'self';
+        style-src 'self';
+        font-src 'self';
+        object-src 'self';
+        plugin-types application/pdf application/x-shockwave-flash application/x-java-applet;
```
jodosha: @joneslee85 What do you think if we leave only PDF here? The latter two plugins were made famous by exploits that can harm users browsers. I'd leave developers to enable this themselves.

joneslee85: @jodosha I like that idea :+1:

jacobbednarz:
What about an option for reporting the violations via `report-uri`?

lib/hanami/generators/app/application.rb.tt
```diff
@@ -200,7 +200,27 @@ module <%= config[:classified_app_name] %>
...
+        base-uri 'self';
+        default-src 'none';
+        script-src 'self';
+        connect-src 'self';
+        img-src 'self';
+        style-src 'self';
+        font-src 'self';
+        object-src 'self';
+        plugin-types application/pdf;
+        child-src 'self';
```
jacobbednarz: Are you aiming for CSP 2 support only? If not, you'll need `frame-src` in here too.

lib/hanami/config/security.rb
```diff
@@ -50,7 +50,7 @@ def content_security_policy(value = nil)
         if value.nil?
           @content_security_policy
         else
-          @content_security_policy = value
+          @content_security_policy = value.split(';').map(&:strip).join('; ')
```
jacobbednarz: Is there any validation on what directives you're going to allow? I'm not 100% of your implementation idea so this may not actually be a concern :stuck_out_tongue:

lib/hanami/generators/app/application.rb.tt
joneslee85: Good point. for the sake of Safari support, I add it in later commit

lib/hanami/config/security.rb
joneslee85: This is for the multiline input such as `%{ multilines }` that makes the config value look easy for reading.

jacobbednarz: Sorry - I probably haven't explained my point very well here. What I intended to convey was is there any validation that the user cannot input an invalid directive or option here? For example, consider the following CSP.

```
Content-Security-Policy: default *; img-src yolocopter; frame-src 'none';
```

This would cause the browser to error since `img-src: yolocopter` isn't a valid source list or keyword. This in turn renders the CSP useless as it cannot be enforced by the browser. However, from my initial comment, I'm not sure if this is something you want to support or leave it up to the developer to implement it correctly.

joneslee85: @jacobbednarz I was proposing a proper helper but we agreed not to and leave it up to the developer.

jacobbednarz: :+1:

jodosha:
@joneslee85 I've just merged this :sparkles: :100: :sparkles:

jacobbednarz:
:tada:

PR #517, opened 2016-02-16
Introduce new logger configuration

### LINKED ISSUE #481: Logger Stream
`Hanami::Logger` supports only stdout as output stream. This has [several advantages](http://12factor.net/logs) such as leaving the freedom to ops team to redirect the stream like they want, or to not accumulate endless (and useless) log files during development.

We should keep stdout as default and preferred stream to log informations.

However, we want to make this more flexible and let developers to **use their own logger**, or **write to file**. In order to make this possible, we want to introduce a new setting:

```ruby
# apps/web/config/application.rb
configure do
  logger.stream 'path/to/file'
end
```

The argument can be:

  * `String`, `Pathname` with relative or absolute path to file
  * A stream like instances of `IO`, `StringIO` like `$stdout` or `STDOUT`

**It MUST default to `STDOUT`.**

### Generating Apps

Projects with default generated app (`Web`) and new generated app (`hanami generate app`), MUST include that setting in the **main** `configure` block **as a commented line of code**.

### Upgrading Existing Projects

Projects that will upgrade to this new version will still use stdout, as it will remain the default output.

davydovanton:
### TODO list

- [X] Create new `Config::Logger` class
- [X] Create `stream` and `build` methods
- [x] Returns `Config::Logger` instance in `logger` config option and fix `_configure_logger!` logic
- [x] Update configuration file
- [ ] Add informations about logger to guides

/cc @hanami/core-team 

---

Closes #481

### DIFF
--- a/lib/hanami/config/logger.rb
+++ b/lib/hanami/config/logger.rb
@@ -9,7 +9,7 @@ module Hanami
       attr_reader :application_module
 
       def initialize
-        @log_device = STDOUT
+        @device = STDOUT
       end
 
       # Log device option value. STDOUT by default
@@ -26,9 +26,9 @@ module Hanami
       # @api private
       def stream(value = nil)
         if value.nil?
-          @log_device
+          @device
         else
-          @log_device = value
+          @device = value
         end
       end
 
@@ -39,7 +39,7 @@ module Hanami
       # @since x.x.x
       # @api private
       def build
-        ::Hanami::Logger.new(log_device: @log_device)
+        ::Hanami::Logger.new(device: @device)
       end
     end
   end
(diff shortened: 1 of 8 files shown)


lib/hanami/config/logger.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      # Returns new Hanami::Logger instance with all options
+      #
+      # @return [Hanami::Logger] the new logger instance
+      #
+      # @since x.x.x
+      # @api private
+      def build
+        ::Hanami::Logger.new(log_device: @log_device)
```
jodosha: @davydovanton Now it's `device:`, not `log_device:` :wink:

davydovanton:
@jodosha could you check this?

joneslee85:
@davydovanton looks :+1:

jodosha:
@davydovanton It's really good. Thank you. Merging it! :+1:

PR #518, opened 2016-02-16
Make Dotenv a soft dependency

### LINKED ISSUE #516: Make Dotenv a soft dependency for development
## Actual

`dotenv` is a runtime dependency for `hanami`. We have included it, because hosting vendors use to provide configurations via env variables.

It happens that devs get confused about the main `.env` and they think it's for production purposes. This assumption is WRONG. Storing informations in that file is not a good security practice because that means to add them inside the Git repository.

Production env vars must be setup according to the hosting instructions. Heroku for instance, uses `heroku config:add`.

## Proposed

* Remove `dotenv` from runtime dependencies
* Generate new projects with `dotenv` in `:development`, `:test` groups in `Gemfile`
* Don't generate `.env` anymore
* Don't try to load `.env` anymore, but only `.env.development` or `.env.test`

Please note that `Hanami::Hanamirc` [uses `Dotenv` internally to parse `.hanamirc`](https://github.com/hanami/hanami/blob/master/lib/hanami/hanamirc.rb#L127). We must replace that code with a parser made by ourselves.

---

/cc @hanami/contributors WDYT?

aderyabin:
Closes #516. I'm open for discussion.
// cc @jodosha

### DIFF
--- a/lib/hanami/environment.rb
+++ b/lib/hanami/environment.rb
@@ -1,9 +1,12 @@
 require 'thread'
 require 'pathname'
-require 'dotenv'
 require 'hanami/utils'
 require 'hanami/utils/hash'
 require 'hanami/hanamirc'
+begin
+  require 'dotenv'
+rescue LoadError
+end
 
 module Hanami
   # Define and expose information about the Hanami environment.
@@ -40,12 +43,6 @@ module Hanami
     # @since 0.6.0
     # @api private
     RACK_ENV_DEPLOYMENT = 'deployment'.freeze
-
-    # Default `.env` file name
-    #
-    # @since 0.2.0
-    # @api private
-    DEFAULT_DOTENV = '.env'.freeze
 
     # Default `.env` per environment file name
     #
@@ -461,7 +458,7 @@ module Hanami
     # @since 0.2.0
     # @api private
     def set_application_env_vars!
-      Dotenv.load     root.join(DEFAULT_DOTENV)
+      return unless defined?(Dotenv)
       Dotenv.overload root.join(DEFAULT_DOTENV_ENV % environment)
     end
 
--- a/lib/hanami/hanamirc.rb
+++ b/lib/hanami/hanamirc.rb
@@ -1,5 +1,4 @@
 require 'pathname'
-require 'dotenv/parser'
 require 'hanami/utils/hash'
 
 module Hanami
@@ -82,6 +81,12 @@ module Hanami
       TEMPLATE_KEY     => DEFAULT_TEMPLATE
     }).symbolize!.freeze
 
+    # Key/value separator in hanamirc file
+    #
+    # @since 0.8.0
+    # @api private
+    SEPARATOR = '='.freeze
+
     # Initialize Hanamirc class with application's root and environment options.
     #
     # @param root [Pathname] Application's root
@@ -124,7 +129,22 @@ module Hanami
     end
 
     def file_options
-      symbolize(exists? ? Dotenv::Parser.call(content) : {})
+      symbolize(exists? ? parse_file(path_file) : {})
+    end
+
+    # Read hanamirc file and parse it's values
+    #
+    # @since 0.8.0
+    # @api private
+    #
+    # @return [Hash] hanamirc parsed values
+    def parse_file(path)
+      {}.tap do |hash|
+        File.readlines(path).each do |line|
+          key, value = line.split(SEPARATOR)
+          hash[key] = value.strip
+        end
+      end
     end
 
     # Return the hanamirc file's path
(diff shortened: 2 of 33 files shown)


lib/hanami/environment.rb
```diff
@@ -461,7 +459,7 @@ def set_hanami_env_vars!
     # @since 0.2.0
     # @api private
     def set_application_env_vars!
-      Dotenv.load     root.join(DEFAULT_DOTENV)
+      return unless defined?(Dotenv)
       Dotenv.overload root.join(DEFAULT_DOTENV_ENV % environment)
```
jodosha: @aderyabin Because now we're calling Dotenv just once, shall we use `#load` instead?

lib/hanami/hanamirc.rb
```diff
@@ -119,12 +118,21 @@ def exists?
 
     private
 
+    def parse(str)
+      {}.tap do |hash|
+        str.split("\n").each do |line|
```
jodosha: Instead of this split, we should change how `#content` works and rely on how Ruby iterates for each line of a file. See `File#readlines`.

lib/hanami/hanamirc.rb
```diff
@@ -119,12 +118,21 @@ def exists?
...
+          key, value = line.split('=')
```
jodosha: @aderyabin Can we please export `=` as a string constant and freeze it? Thanks!

test/environment_test.rb
```diff
@@ -23,17 +23,19 @@
...
-      end
+      describe 'ignores .env' do
+        it 'does not sets env vars from .env' do
+          ENV['FOO'].must_be_nil
+        end
 
-      it 'sets port from .env' do
-        @env.port.must_equal 42
-      end
+        it 'does not sets port from .env' do
```
jodosha: @aderyabin The fact that we're ignoring `.env` doesn't mean we want to ignore if a port is specified there. Can you please restore this test and maybe moving that port value into another `.env.*` fixture? Thanks!

lib/hanami/hanamirc.rb
aderyabin: I working on Parser module, because `.hanamirc` can stores comments, empty lines and so on

lib/hanami/environment.rb
aderyabin: @jodosha 
`#load` and `#overload` has only one difference `#load` uses `#apply` and `#overload` uses `#apply!` [proof link](https://github.com/bkeepers/dotenv/blob/c44af19abeb12240995358fb350c1dab42d3c0fe/lib/dotenv/environment.rb#L20-L26)

```
    def apply
      each { |k, v| ENV[k] ||= v }
    end

    def apply!
      each { |k, v| ENV[k] = v }
    end
```

So have to use `#overload` because if the variable exists it should be overrided.

jodosha: @aderyabin Perfect :+1:

lib/hanami/hanamirc/parser.rb
```diff
@@ -0,0 +1,63 @@
+module Hanami
+  class Hanamirc
+    module Parser
```
jodosha: @aderyabin I don't think this is necessary. This is the `Dotenv::Parser`, which needs to handle edge cases because of human editing of `.env` files.

Our `.hanamirc` is stricter and I think that the previous, simpler implementaion (`split("=")`) was enough. This because our file isn't meant to be edited **and** it doesn't have complex values that include double quotes or UNIX `export` keyword.

test/environment_test.rb
```diff
@@ -23,17 +23,30 @@
...
-      end
+      describe 'ignores .env' do
+        it 'does not sets env vars' do
+          ENV['FOO'].must_be_nil
+        end
 
-      it 'sets port from .env' do
-        @env.port.must_equal 42
+        it 'does not sets port' do
+          @env.port.must_equal 2300
```
jodosha: @aderyabin Please set the port in a `.env.test` fixture and verify that this still works. Please have a look at https://github.com/hanami/hanami/pull/392

jodosha: @aderyabin I just noticed the test is placed below in the file. Sorry.

aderyabin: No problem. I have to check failed tests.

jodosha:
@aderyabin I manually tried all the combinations and reviewed the code. It works great, thank you!

As side note, [I added a guard](https://github.com/hanami/hanami/commit/5170c04c0ecc297170d78b95d5a2292d92d84fca) that don't tries to invoke `Dotenv.overload` in case a `.env.*` file is missing. For now it works because `.overload` [ignores non existing files](https://github.com/bkeepers/dotenv/blob/e13bf8a877e2237b1c91c66c775f88bbcaafe314/lib/dotenv.rb#L32). If they change that behavior, then our applications would have fallen apart, but with this guard we're protected. :wink:

PR #519, opened 2016-02-18
Use spec formatter [DO NOT MERGE!]

joneslee85:
Recently we got so many intermittent tests, this PR is to use Spec formatter so I could tell which tests are running in which order for troubleshooting with TravisCI

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,5 +26,6 @@ platforms :jruby do
   gem 'jdbc-sqlite3'
 end
 
+gem 'minitest-reporters'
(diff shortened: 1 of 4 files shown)


jodosha:
@joneslee85 Great, can you please also verify why Coveralls reports such a low coverage rate? Thank you very much :+1:

joneslee85:
You have fixed it in https://github.com/hanami/hanami/commit/1e2047fa710df384df498e382b08a212d0a27e31, so I am closing this for now

PR #522, opened 2016-02-18
Digest url modified to correct address

### LINKED ISSUE #521: Application.rb Template has incorrect digest URL
When you generate a new application with Hanami, the application.rb contains the following snippet of comment:

```
# Use digest file name for asset paths
#
# See: http://hanamirb.org/guides/assets/digest
 digest  true
```

The request URL for the digest does not exist and I assume the correct location is
 http://hanamirb.org/guides/assets/overview

If this is correct I can go ahead and make the change?

daneb:
Closes #521


jodosha:
@daneb Thank you very much! :+1:

Issue #249, opened 2015-06-05
Experimental code reloading


jodosha:
### Overview

**Lotus itself doesn't implement code reloading**.

We use a reloadable context only for development mode via `lotus server` command. The reason is simplicity. Once we bypass that reloadable context, Lotus works like any other Ruby library, without messing with constant autoloading tricks.

The library we use for reloading is: Shotgun. It's written in Ruby and uses `fork(2)` to spawn a new process when a new request comes it. It has portability issues: it works only on MRI (see https://github.com/lotus/lotus/issues/237).

### Experimental Solution

We want to keep this architecture: Lotus shouldn't care about code reloading. We should have this feature implemented on the outside, still with `lotus server`.

There is a POSIX compatible solution that provides code reloading for generic processes: `entr(1)`.
I did some manual tests and seems to work well. The idea is to replace Shotgun with Entr in the long term. Before to do that, we should check if it works properly for Lotus.

#### Pros

  * It removes a Rubygem runtime dependency for `lotusrb`
  * It's probably faster, as it's written in C, instead of Ruby
  * It can improve portability with other Ruby implementations

#### Cons

  * It requires dev machines to install `entr`. Note that it has both `apt-get` and `brew` installers.

### Details

We need to add a CLI arg for `lotus server`: `--experimental-reloading`.
When used, it should prefer `entr` over `shotgun`.
If the binary can't be found, it should raise an exception.

The internal implementation should resemble this command:

```shell
find . | entr -r ...
```

http://entrproject.org/

pascalbetz:
@jodosha 
What needs to be done for entr to work? Can you try the streaming sample with entr (if you think that it might be shotgun which prevents streaming from working) or explain how you used entr.

Details:
https://github.com/lotus/lotus/issues/303

thecatwasnot:
@pascalbetz I played with using entr on a lotus project inside of docker, basically running entr on host machine, to restart containers.  Dropped it in a Makefile:
```
.PHONY: watch
watch:
        docker-compose kill dev
        while sleep 1; do find apps/ config/ lib/ config.ru Gemfile.lock ! -name '*.sw*' | entr -dr docker-compose up dev; done

.PHONY: watch-bundler
watch-bundler:
        ls Gemfile | entr docker-compose run dev bundle install

.PHONY: watch-spec
watch-spec:
        while sleep 1; do find apps/ lib/ spec/ ! -name '*.sw*' | entr -d docker-compose run dev bundle exec rspec; done
```
[...]

PR #523, opened 2016-02-18
Code reloading with entr

aderyabin:
Functionality fully described [here](https://github.com/hanami/hanami/issues/249#issuecomment-181797020):

1. Remove `shotgun` as hard dependency from `hanami.gemspec`
2. Generate new projects with this gem in `Gemfile` (see below)
3. For the server: detect if `Shotgun` is present and activate code reloading via `Shotgun` (which is the actual implementation)
4. Otherwise try to fallback to `entr`
5. When none of them isn't available, it should not try to activate code reloading.

// cc @jodosha 

**NOTE**: There are no link to Hanami Guides.

### DIFF
--- a/lib/hanami/commands/server.rb
+++ b/lib/hanami/commands/server.rb
@@ -1,78 +1,144 @@
 require 'rack'
+require 'hanami/server'
 
 module Hanami
   module Commands
-    # Rack compatible server.
-    #
-    # It is run with:
-    #
-    #   `bundle exec hanami server`
-    #
-    # It runs the application, by using the server specified in your `Gemfile`
-    # (eg. Puma or Unicorn).
-    #
-    # It enables code reloading by default.
-    # This feature is implemented via process fork and requires `shotgun` gem.
-    #
-    # @since 0.1.0
-    # @api private
-    class Server < ::Rack::Server
-      attr_reader :options
+    class Server
 
-      # @param options [Hash] Environment's options
+      # Message text when Shotgun enabled but interpreter does not support `fork`
       #
-      # @since 0.1.0
-      # @see Hanami::Environment#initialize
+      # @since 0.8.0
+      # @api private
+      WARNING_MESSAGE = 'Your platform doesn\'t support code reloading.'.freeze
+
+      ENTR_EXECUTE_COMMAND = 'find ./**/*.rb | entr -r bundle exec hanami rackserver %s'.freeze
+
+      attr_reader :server
+
       def initialize(options)
-        @_env    = Hanami::Environment.new(options)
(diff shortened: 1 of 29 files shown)


lib/hanami/commands/server.rb
```diff
@@ -24,12 +27,15 @@ class Server < ::Rack::Server
       # @since 0.1.0
       # @see Hanami::Environment#initialize
       def initialize(options)
+        @entr_options =  options.merge(code_reloading: false).inject([]) {|res, (k, v)| res << "--#{k}=#{v}" ; res}.join(" ")
```
jodosha: @aderyabin Does this means we're open to accept any kind of CLI args and pass down to `entr`?

lib/hanami/commands/server.rb
```diff
@@ -24,12 +27,15 @@ class Server < ::Rack::Server
...
         @_env    = Hanami::Environment.new(options)
         @options = _extract_options(@_env)
-
         if code_reloading?
+          begin
```
jodosha: @aderyabin Can we please declutter `#initialize`, by extracting a method like `prepare_code_reloading!`? Thank you! :+1: This should include `@entr_options` line too. :smile:

lib/hanami/commands/server.rb
aderyabin: yes, all passed CLI args will be forwarded to `entr` but with disabled code reloading

lib/hanami/commands/server.rb
```diff
@@ -17,6 +17,9 @@ module Commands
     # @since 0.1.0
     # @api private
     class Server < ::Rack::Server
+      ENTR_EXECUTE_COMMAND = 'find . | entr -r bundle exec hanami server %s'.freeze
+      ENTR_CHECK_COMMAND = 'which entr'.freeze
```
jodosha: @aderyabin Using `which` makes this code non portable on Windows. Please check if this snippet can work for us: http://stackoverflow.com/a/5471032/498386

lib/hanami/commands/server.rb
jodosha: @aderyabin What's the need? I'm not criticizing the choice, but trying to understand the reasons behind :smile:

jodosha:
@aderyabin This is a great implementation. Before to merge I want to try it for a development session. :+1:

jodosha:
@aderyabin I've finally merged this. Thank you :+1:

rbishop:
I'm just here to clap my hands in excitement to see a project using finally `entr` instead re-inventing file system polling and bringing in all sorts of crazy dependencies. :clap: :clap:

PR #524, opened 2016-02-18
do not show warning message if code reloading is disabled

aderyabin:
Warning message is always displaying even in tests where code reloading is disabled. I think that is not correct.

### DIFF
--- a/lib/hanami/environment.rb
+++ b/lib/hanami/environment.rb
@@ -382,8 +382,8 @@ module Hanami
     def code_reloading?
       if code_reloading = @options.fetch(:code_reloading) { !!CODE_RELOADING[environment] }
         # JRuby doesn't implement fork that's why shotgun cannot be used.
-        if Utils.jruby?
-          puts "JRuby doesn't support code reloading."
+        unless Kernel.respond_to?(:fork)
+          puts "Your platform doesn't support code reloading."
           return false
         end
       end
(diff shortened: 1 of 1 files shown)


joneslee85:
:+1:

jodosha:
@aderyabin Thank you, this enhancement makes sense.

What if we take the chance to make it less narrowed? Windows platform doesn't implement `fork` as well, and it isn't available on MRI. We should check for `Kernel.fork`.

```shell
~:jruby-9.0.5.0 ➜ irb
irb(main):001:0> Kernel.respond_to?(:fork)
=> false
irb(main):002:0> quit
 ~:jruby-9.0.5.0 ➜ chruby 2.3.0
 ~:ruby-2.3.0 ➜ irb
irb(main):001:0> Kernel.respond_to?(:fork)
=> true
irb(main):002:0>
```

And then we should change the warning message too: `"Your platform doesn't support code reloading."`.

aderyabin:
@jodosha Done

jodosha:
@aderyabin Thank you for this quick change!

The impact of this change can affect #523 . In `master` we had that jruby check, which we want to make more generic to report similar errors on platforms that don't support `fork`.

But with the new strategy (#523), code reloading don't necessarily depends on `fork` when using `entr`. `fork` is needed only for Shotgun. WDYT?

aderyabin:
Yes, yes you are right, but it's a problem #523 and should be solved there.

PR #530, opened 2016-02-24
Add logger level option

### LINKED ISSUE #482: Logger Levels
## Log Levels

Right now there isn't any management for log levels. We want to introduce default levels for each environment as described below:

  * `development`: `debug`
  * `test`: `error`
  * `production`: `info`

### Customized Error Levels

A developer can set a different log level with the following syntax:

```ruby
# apps/web/config/application.rb
configure do
  logger.level :debug
end
```

The following values MUST be accepted

* A symbol that express one of the default `Logger` levels (eg `:debug`)
* A string that express one of the default `Logger` levels (eg `"DEBUG"`)
* A constant from Hanami (eg. `Hanami::Logger::DEBUG`)
* A constant from Logger (eg. `Logger::DEBUG`)

Note, `Logger` from Ruby stdlib already accepts all these arguments, we only need to define constants for `Hanami::Logger`.

**This should take the precedence over the default value of the current env.**

### Generating Apps

Projects with default generated app (`Web`) and new generated app (`hanami generate app`), MUST include that setting in the **main** `configure` block **as a commented line of code**.

### Upgrading Existing Apps

There is nothing that breaks for existing apps, by adding default levels.

davydovanton:
Closes #482

### DIFF
--- a/lib/hanami/config/logger.rb
+++ b/lib/hanami/config/logger.rb
@@ -8,20 +8,8 @@ module Hanami
     class Logger
       attr_reader :application_module
 
-      # @since x.x.x
-      # @api private
-      LEVEL_HASH = {
-        warn:    Hanami::Logger::WARN,
-        info:    Hanami::Logger::INFO,
-        fatal:   Hanami::Logger::FATAL,
-        error:   Hanami::Logger::ERROR,
-        debug:   Hanami::Logger::DEBUG,
-        unknown: Hanami::Logger::UNKNOWN
-      }.freeze
-
       def initialize
         @device = STDOUT
-        @level  = Hanami::Logger::DEBUG
       end
 
       # Log device.
@@ -86,7 +74,7 @@ module Hanami
         if value.nil?
           @level
         else
-          @level = detect_level(value)
+          @level = value
         end
       end
 
@@ -120,16 +108,8 @@ module Hanami
       # @see Hanami::Config::Logger#stream
       # @see Hanami::Config::Logger#engine
       def build
-        logger = ::Hanami::Logger.new(@app_name, device: @device)
-        logger.level = @level
-
-        @engine || logger
-      end
-
-    private
-
-      def detect_level(level)
-        level.is_a?(Integer) ? level : LEVEL_HASH[level.to_sym.downcase]
+        @engine ||
+          ::Hanami::Logger.new(@app_name, device: @device, level: @level)
       end
     end
   end
--- a/lib/hanami/config/logger.rb
+++ b/lib/hanami/config/logger.rb
@@ -8,8 +8,20 @@ module Hanami
     class Logger
       attr_reader :application_module
 
+      # @since x.x.x
+      # @api private
+      LEVEL_HASH = {
+        warn:    Hanami::Logger::WARN,
+        info:    Hanami::Logger::INFO,
+        fatal:   Hanami::Logger::FATAL,
+        error:   Hanami::Logger::ERROR,
+        debug:   Hanami::Logger::DEBUG,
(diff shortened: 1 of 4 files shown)


davydovanton:
@hanami/core-team could you check this? Thanks! :purple_heart:

lib/hanami/config/logger.rb
```diff
@@ -89,8 +120,16 @@ def app_name(value = nil)
       # @see Hanami::Config::Logger#stream
       # @see Hanami::Config::Logger#engine
       def build
-        @engine ||
-          ::Hanami::Logger.new(@app_name, device: @device)
+        logger = ::Hanami::Logger.new(@app_name, device: @device)
```
jodosha: @davydovanton Do you think it's the case we're gonna modify the initializer so we can achieve the following code?

```ruby
@engine || ::Hanami::Logger.new(@app_name, device: @device, level: @level)
```

More elegant, doesn't instantiate an object that can theoreticallly not be used, and avoids mutations of the logger after it's being instantiated. It looks like a win-win situation.

lib/hanami/config/logger.rb
```diff
@@ -8,8 +8,20 @@ module Config
...
+      # @since x.x.x
+      # @api private
+      LEVEL_HASH = {
+        warn:    Hanami::Logger::WARN,
+        info:    Hanami::Logger::INFO,
+        fatal:   Hanami::Logger::FATAL,
+        error:   Hanami::Logger::ERROR,
+        debug:   Hanami::Logger::DEBUG,
+        unknown: Hanami::Logger::UNKNOWN
+      }
```
joneslee85: freeze it

test/config/logger_test.rb
```diff
@@ -35,6 +35,32 @@
     end
   end
 
+  describe '#level' do
+    it 'takes logger level as a symbol' do
```
joneslee85: can we cover scenario in which lvl is integer?

lib/hanami/config/logger.rb
```diff
@@ -89,8 +120,16 @@ def app_name(value = nil)
...
+        logger = ::Hanami::Logger.new(@app_name, device: @device)
+        logger.level = @level
+
+        @engine || logger
+      end
+
+    private
+
+      def detect_level(level)
+        level.is_a?(Integer) ? level : LEVEL_HASH[level.to_sym.downcase]
```
jodosha: @davydovanton What if we let `Hanami::Logger` to check this? If you think, we're using too much knowledge of its implementation. We analyize integer and symbols (see `LEVEL_HASH`) on its behalf. This change would make `Hanami::Logger` capable to take the right decision by itself. :wink:

jodosha: @davydovanton What happens if I pass an integer that can't be recognized as level? Eg `999`.

joneslee85: Plus one

test/config/logger_test.rb
```diff
@@ -60,5 +86,14 @@
         builded_logger.must_be_instance_of ::Logger
       end
     end
+
+    describe 'when user set level' do
```
joneslee85: s/set/sets

test/config/logger_test.rb
davydovanton: yes, of course

davydovanton:
Pleaze, don't merge this before https://github.com/hanami/utils/issues/121

jodosha:
@davydovanton https://github.com/hanami/utils/pull/121 it's now merged in master and available at `0.8.x`.

PR #532, opened 2016-02-26
Fix configuration file for app architecture

### LINKED ISSUE #531: Logger: Support Application Architecture
We introduced new logger features with #517 and #530 .
These PR were merged, but they only ship with generators that support Container Architecture. Application Architecture was left behind.

We should address this disparity of features.

davydovanton:
Closed #531


jodosha:
@davydovanton Merged, thank you! :+1:

PR #533, opened 2016-02-29
Clean up CLI descriptions

jbodah:
This PR include several minor changes to the CLI descriptions:

1. start descriptions with capitals (this is Thor's default and it's kind of jarring to see a mix)

2. always refer to hanami as "hanami", not as a mix of that and "Hanami"

3. clean up trailing text in descriptions and unify styles on "(this, that, etc)"


jodosha:
@jbodah Thank you, this is the kind of improvement that only fresh eyes can see. :+1:

PR #536, opened 2016-02-29
Stop generating lib/config/mapping.rb file

### LINKED ISSUE #535: Remove lib/config/mapping.rb, make it the default, or add comments
I lost about an hour because of this (and some confusion with another bug which had been resolved in 0.8.x). I would recommend either removing this file, making it the default, or adding comments saying that the file is meant to be loaded from `lib/my_project.rb` to avoid confusion. Personally I prefer either of the first two options to avoid having a dead file in the project

joneslee85:
[Fix #535]


jodosha:
@joneslee85 Thanks for this quick PR :+1: 

There is still one spot left: `lib/bookshelf.rb`

```ruby
  # You can specify mapping file to load with:
  #
  # mapping "#{__dir__}/config/mapping"
  #
  # Alternatively, you can use a block syntax like the following:
  #
```

davydovanton:
:+1:

PR #537, opened 2016-03-01
Prepend routes after leading comments

cllns:
Going through the getting started guide, once you run `hanami generate action web books#index`, your routes file looks like this:

```
get '/books', to: 'books#index'
# Configure your routes here
# See: http://www.rubydoc.info/gems/hanami-router/#Usage
get '/', to: 'home#index'
```

When what you'd expect is it to look like:

```
# Configure your routes here
# See: http://www.rubydoc.info/gems/hanami-router/#Usage
get '/books', to: 'books#index'
get '/', to: 'home#index'
```

- The other option is to put the route at the bottom of the file, but then it might not match, which is confusing)
- In the regex, I opted for `\s` instead of `\n` because I'm not sure if `\n` matches newlines on Windows. I tried using 'multiline' mode, but that was more trouble than it was worth.
- We could just override Thor's `prepend_to_file` to have this behavior for all strings that are not comments.

### DIFF
--- a/lib/hanami/commands/generate/action.rb
+++ b/lib/hanami/commands/generate/action.rb
@@ -105,7 +105,7 @@ module Hanami
           routes_path.dirname.mkpath
           FileUtils.touch(routes_path)
 
-          generator.prepend_to_file(routes_path, "#{ http_method } '#{ route_url }', to: '#{ route_endpoint }'\n")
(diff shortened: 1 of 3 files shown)


davydovanton:
amazing :+1:

joneslee85:
:+1:

jodosha:
@cllns Awesome :heart_eyes: Merged it :+1:

PR #538, opened 2016-03-01
Welcome page: Remove trailing period, to make copy-and-pasting easier

cllns:
When I double click the `<code>` block to highlight the whole thing, it also includes this trailing period. It doesn't really add anything, and is a little annoying in this situation, so I think we can get rid of it.


AlfonsoUceda:
@cllns thanks!

Issue #514, opened 2016-02-14
Follow ruby code style convensions


aderyabin:
As an idea, I suggest to follow Ruby code style fully described by Bozhidar ([link](https://github.com/bbatsov/ruby-style-guide)) and integrate Rubocop. 

I saw the discussion about Rubocop [here](https://github.com/hanami/hanami/pull/171) but let's back to the question again. Yes, I know that default Rubocop config is angry, but we can modify it. 

From my side I'm ready to start working on Rubocop integration, but I'm interested in thoughts by @jodosha, @jeremyf and @joneslee85 before starting.

jodosha:
@aderyabin Hello and thanks for this discussion. Are you talking about the internals of the framework or about the generated code (aka Hanami projects)?

aderyabin:
@jodosha Both.
as the example, look at this lines https://github.com/hanami/hanami/blame/master/lib/hanami/commands/generate/app.rb#L64-L73.
Ten lines of code and two different styles `#{app_name}` and `#{ app_name }`.

Or another example. Hanami codebase has both `raise` and `fail`. 

Yes, there are a lot more important tasks, but Hanami is enjoying the beauty, and code should be beautiful too, I believe.

jdickey:
I've been using RuboCop on all my projects for years but, in the last year or so, I've had the sense they're going deep into the fringe. They've flip-flopped on a couple of style conventions as time has gone on.

I've locked my current projects to RuboCop 0.35.1, with a simple [configuration file](https://gist.github.com/jdickey/970d70a9b58fe5519d50) to override exactly two defaults:

<script src="https://gist.github.com/jdickey/970d70a9b58fe5519d50.js"></script>

Lately, I've been relying far more on Tim Roessner's [Reek](https://github.com/troessner/reek) for most of my style enforcement; you may want to take a look at it, too.,

jodosha:
@aderyabin Absolutely not for the internals of the framework. Sorry, but we can't afford this change.

For the Hanami projects it depends on how much work it is for us and how much demand it's out there. I never used RuboCop because I feel it too strict. I'm with @jdickey on this.

joneslee85:
@aderyabin i am too a big fan of strict code linting but I do find it is annoying at times, especially for contributors. But I am all for small clean-up commits (submitted along with big PRs) to make the code more consistent.

jbodah:
[...]

PR #540, opened 2016-03-03
Change x_frame_options DENY to single quotes

cllns:
From #514.

Two code changes, ten fixtures changes, to make tests pass.


joneslee85:
Personally I prefer single quote over soviet quote but I think we should also be mindful to others moving forward. There will be pending PRs that touch many files and trivial PRs to fix code style would make it hard for other to rebase their works.

cllns:
I actually prefer double quotes :dancer: 
But I prefer consistency above that, and the code `hanami` generates for an app is all single quotes (except for this).

I did `hanami new test-hanami-app`, this was the only case of double quotes when they weren't needed.

jodosha:
@cllns 
> But I prefer consistency

Agreed. Merging this :wink:

PR #541, opened 2016-03-08
Remove default HTTP method from thor's method_option

### LINKED ISSUE #539: Resourceful routes - always GET
#414 isn't working as desired, with version 0.7.2.

Running `hanami generate action web books/destroy` should generate this route:
```
delete '/books/:id', to: 'books#destroy'
```

but it generates this instead:
```
get '/books/:id', to: 'books#destroy'
```

So the URL suffix is working properly, but the HTTP action (verb) is not. It just uses GET for everything (unless you pass in `--method`).

This is true for all of the commands, so this is an issue for `create` and `update` too, since they're the other non-GET resourceful routes.

It's weird because there's [a test for this](https://github.com/hanami/hanami/blob/master/test/commands/generate/action_test.rb#L162). I'm not sure why that test is not catching this.

cllns:
Closes #539 

The 'default' isn't always 'GET' anymore, since it changes for resourceful routes.

### DIFF
--- a/lib/hanami/cli_sub_commands/generate.rb
+++ b/lib/hanami/cli_sub_commands/generate.rb
@@ -31,7 +31,7 @@ module Hanami
 
         > $ hanami generate action web cars#create --method=post
       EOS
-      method_option :method, desc: "The HTTP method to be used for the generated route. Must be one of (#{Hanami::Routing::Route::VALID_HTTP_VERBS.join('/')})", default: Hanami::Commands::Generate::Action::DEFAULT_HTTP_METHOD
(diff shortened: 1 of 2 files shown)


jodosha:
@cllns Hey, thanks for this PR. There are failing tests that you may want to fix :wink: 

> The 'default' isn't always 'GET' anymore, since it changes for resourceful routes.

What about non-RESTful routes? Eg.

```shell
% bundle exec hanami generate action web books#sales --url=/books/sales
```

cllns:
Fixed the test and added a note in the `desc:` about the default being GET!

jodosha:
@cllns Where is the note?

cllns:
@jodosha Forgot to force push :sweat_smile:. Just did so! Sorry about that!

jodosha:
I manually tested it and it fixes the problem. Thanks @cllns !

PR #543, opened 2016-03-09
Fixes #542 - relative path for nested actions

### LINKED ISSUE #542: Generated relative path inaccurate for nested actions
If I do

```shell
hanami g action app foo/bar#baz
```

All the files generate in the right places, but the `require_relative` is inaccurate. Because of the additional directory nesting, it needs one more `../`

What I get:
```ruby
# apps/foo/controllers/bar/baz.rb
module Foo::Controllers::Bar::Baz
end

# spec/foo/controllers/bar/baz_spec.rb
require_relative '../../../../apps/foo/controllers/bar/baz'

# should be
require_relative '../../../../../apps/foo/controllers/bar/baz'
```

The code is [here](https://github.com/hanami/hanami/blob/ec6411c77b38b87e1e162e1c5bb73ade9da4642f/lib/hanami/commands/generate/action.rb#L236). I believe the @controller_pathname could be scanned for slashes and then additional `..` could be added based on that.

Here's a quick and dirty way of fixing it:
```ruby
def relative_action_path
  nestings = @controller_pathname.count('/') * ['..']
  path = Pathname.new('.').join('..', '..', '..', *nestings)
  path = path.join('..') if environment.container?
  path.join(application_path, 'controllers', @controller_pathname, @action_name.downcase)
end
```

Thoughts?

berfarah:
For reference: https://github.com/hanami/hanami/issues/542

@cllns Do you feel like these tests go overboard? I figured the alternative (and shorter) would've been doing the below, but at the cost of not making sure that it actually references the right file, leaving the room open for typos.

```ruby
assert_file_includes('spec/web/controllers/admin/books/index_spec.rb', '../../../../../apps/web/controllers/admin/books/index.rb')
```

```ruby
assert_file_includes('spec/web/views/admin/books/index_spec.rb', '../../../../../apps/web/views/admin/books/index.rb')
```

### DIFF
--- a/lib/hanami/commands/generate/action.rb
+++ b/lib/hanami/commands/generate/action.rb
@@ -60,6 +60,7 @@ module Hanami
           @controller_name = Utils::String.new(controller_name.join("/")).classify
           @action_name = Utils::String.new(action_name).classify
           @controller_pathname = Utils::String.new(@controller_name).underscore
+          @controller_patharray = @controller_pathname.split("/")
 
           assert_application_name!
           assert_controller_name!
@@ -210,23 +211,23 @@ module Hanami
         end
 
         def view_path
-          application_path.join('views', @controller_pathname, "#{@action_name.downcase}.rb")
+          application_path.join('views', *@controller_patharray, "#{@action_name.downcase}.rb")
         end
 
         def view_spec_path
-          spec_root.join('views', @controller_pathname, "#{@action_name.downcase}_spec.rb")
+          spec_root.join('views', *@controller_patharray, "#{@action_name.downcase}_spec.rb")
         end
 
         def template_path
-          application_path.join('templates', @controller_pathname, "#{@action_name.downcase}.html.#{template_engine.name}")
+          application_path.join('templates', *@controller_patharray, "#{@action_name.downcase}.html.#{template_engine.name}")
         end
 
         def action_path
-          application_path.join('controllers', @controller_pathname, "#{@action_name.downcase}.rb")
+          application_path.join('controllers', *@controller_patharray, "#{@action_name.downcase}.rb")
         end
 
         def action_spec_path
-          spec_root.join('controllers', @controller_pathname, "#{ @action_name.downcase}_spec.rb")
+          spec_root.join('controllers', *@controller_patharray, "#{ @action_name.downcase}_spec.rb")
         end
 
         def spec_root
@@ -234,15 +235,17 @@ module Hanami
         end
 
         def relative_action_path
-          path = Pathname.new('.').join('..', '..', '..')
-          path = path.join('..') if environment.container?
-          path.join(application_path, 'controllers', @controller_pathname, @action_name.downcase)
+          relative_base_path.join(application_path, 'controllers', *@controller_patharray, @action_name.downcase)
         end
 
         def relative_view_path
-          path = Pathname.new('.').join('..', '..', '..')
-          path = path.join('..') if environment.container?
-          path.join(application_path, 'views', @controller_pathname, @action_name.downcase)
+          relative_base_path.join(application_path, 'views', *@controller_patharray, @action_name.downcase)
+        end
+
+        def relative_base_path
+          nestings = ['..'] * @controller_pathname.count(Utils::String::UNDERSCORE_SEPARATOR)
+          nestings << '..' if environment.container?
+          path = Pathname.new('.').join('..', '..', '..', *nestings)
         end
 
         def app_base_dir
--- a/test/commands/generate/action_test.rb
+++ b/test/commands/generate/action_test.rb
@@ -71,6 +71,11 @@ describe Hanami::Commands::Generate::Action do
 
         assert_file_exists 'apps/web/templates/admin/books/index.html.erb'
         assert_file_includes('apps/web/config/routes.rb', "get '/admin/books', to: 'admin/books#index'")
+
+        relative_action_path = command.template_options[:relative_action_path]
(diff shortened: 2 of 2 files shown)


cllns:
Awesome! 

Personally, I think the shorter version is better, since you actually `assert_file_includes` rather than `assert_file_exists`and the full string path is more explicit. Sure a typo is easier, but the test shouldn't be changing much anyway. What do others think?

lib/hanami/commands/generate/action.rb
```diff
@@ -234,14 +234,16 @@ def spec_root
         end
 
         def relative_action_path
-          path = Pathname.new('.').join('..', '..', '..')
-          path = path.join('..') if environment.container?
+          nestings = ['..'] * @controller_pathname.count('/')
```
jodosha: @berfarah We should replace hardcoded file separator `'/'` with `::File::SEPARATOR`.

jodosha: @berfarah the first part of `relative_action_path` is identical to `relative_view_path`. Shall we consider to extract a method to be used by both of these consumers?

berfarah: Thanks! Absolutely :)

berfarah:
@jodosha Another thing - in `action.rb` at the [top](https://github.com/hanami/hanami/blob/master/lib/hanami/commands/generate/action.rb#L11):

should
```ruby
ACTION_SEPARATOR = /[\/,#]/
```

be 

```ruby
ACTION_SEPARATOR = /[#{File::SEPARATOR},#]/
```
?

Assuming the user can use either `foo/bar#baz`, `foo/bar/baz` or `foo\bar\baz` to create an action.

Also, how do you feel about the tests asserting the relative file exists vs checking the string?

cllns:
Should we add a test for deeper nested actions and views? The code seems to support arbitrary nesting, so it might be good to add a test for that :) Maybe just 3 or 4 levels deep?

lib/hanami/commands/generate/action.rb
berfarah: @jodosha I've looked into it, and I think using `File::SEPARATOR` would break functionality on Windows here.

At the top of `action.rb`, the `@controller_pathname` variable is assigned by using `Hanami::Utils::String().underscore`.

`#underscore` will only replace class divisions with `/` ([see here](https://github.com/hanami/utils/blob/master/lib/hanami/utils/string.rb#L43)).

If we wanted to use `File::SEPARATOR` here we'd have to adjust `#underscore` to use it as well.

It would work for `hanami g action foo\bar#baz` but it would break for `hanami g action Foo::Bar::Baz` - also, as mentioned below, it won't work for `foo\bar\baz`

jodosha:
@berfarah RE `ACTION_SEPARATOR`. That is a convention to separate controller and action name. The suggested char is `#` (eg. `home#index`), and `/` is an arbitrary fallback that has nothing to do with file separators.

We could've been used another char instead (eg `home`:rabbit:`index`). :wink:

lib/hanami/commands/generate/action.rb
jodosha: @berfarah 
> At the top of action.rb, the @controller_pathname variable is assigned by using Hanami::Utils::String().underscore.

Ouch, this should be addressed. The usage of `underscore` is useful to transform `Foo::Bar::Baz` into predictable chunks. But then the mistake was to use the output string directly without to normalize via `::File::Separator`.

jodosha:
@berfarah To recap: we want the following syntaxes to work across the OSes:

```shell
bundle exec hanami generate action web books#index
bundle exec hanami generate action web books/index
bundle exec hanami generate action web sales/books#index
bundle exec hanami generate action web sales/books/index
```

Am I missing something?

berfarah:
Sounds good - I think I'll be `split` ing the underscored value and passing it in as an array wherever the path is required. I'll update my PR on Monday or Tuesday. Have a nice weekend!

berfarah:
@jodosha Fixed. I've added an instance variable `@controller_patharray` which splits on `/`. By using `*@controller_patharray` wherever a `pathname.join` is used, we automatically will get `File::SEPARATOR` used. `#underscore` still needs to be used for routing, so I left it at that and we use both instance variables instead.

While I like @cllns's idea, I've left it at the current nesting rather than rewriting the existing nesting test and didn't want to create a new one to prevent from slowing down the tests further. If you want one, I can add one no problem. I've also refactored as suggested.

lib/hanami/commands/generate/action.rb
```diff
@@ -234,15 +235,17 @@ def spec_root
...
 
         def relative_view_path
-          path = Pathname.new('.').join('..', '..', '..')
-          path = path.join('..') if environment.container?
-          path.join(application_path, 'views', @controller_pathname, @action_name.downcase)
+          relative_base_path.join(application_path, 'views', *@controller_patharray, @action_name.downcase)
+        end
+
+        def relative_base_path
+          nestings = ['..'] * @controller_pathname.count("/")
```
jodosha: @berfarah Shouldn't this be still `File::SEPARATOR` instead of `/`?

berfarah: @jodosha In this case, it's counting the `/`s in `@controller_pathname` - which can only have `/` and creating an array with that many nestings.

jodosha: @berfarah Thanks for the clarification. Having these slashes around is confusing, that's why we need to name them via constants.

To avoid further ping/pong, I'm gonna merge this and cleanup by myself.

Issue #520, opened 2016-02-18
Initializers for container architecture


tylerhunt:
I created a new app with the default (container) architecture, and the generator produced the directory `config/initializers` in the project root, but this directory is never loaded. Instead, when the app loads, it's attempting to look for initializers using the glob `apps/web/config/initializers/**/*.rb`.

Are initializers supposed to be available at the container level, only at the app level within the container, or in both places?

jodosha:
@tylerhunt Hey sorry for the confusion. The expected behavior is to have initializers at the project (aka container) level (`config/initializers`), not on application basis (`apps/web/config/initializers`).

The intent here is to let you to setup **project wide** dependencies. For instance Sidekiq's `Sidekiq.configure_server` can be an example and you would add it at `config/initializers/sidekiq.rb`.

tylerhunt:
The [guide][] also indicates the opposite:
> For a given application named `Web`, they MUST be placed under `apps/web/config/initializers`.

[guide]: http://hanamirb.org/guides/applications/initializers/

jodosha:
@tylerhunt I think this is a huge mistake of mine. I apologize for this confusion. We should fix this.

tylerhunt:
@jodosha No problem. Just pointed it out because I came across it, so it can be updated when the code is changed.

lucasallan:
@joneslee85 @jodosha can I work on this?

jodosha:
@joneslee85 are you already working on this? Otherwise let @lucasallan to take care of it.

joneslee85:
@jodosha I've started it. Please leave it to me

lucasallan:
Let me know if you need help @joneslee85

jodosha:
Unfortunately delegating the responsibility to `Hanami::Container` can fix the problem of having one centralized loaders from `config/initializers` (instead of `apps/web/config/initializers`), but it doesn't guarantee to be always executed **and** the order of loading.

---

First of all, there are some circumstances where `Hanami::Container` is not instantiated. For instance, when we run a single unit test on a project, in order to keep the start time short, we don't load the full project code.

That means basically to **avoid to instantiate** `Hanami::Container`. Now imagine a unit test relying on an initializer: if we delegate to `Hanami::Container` the role to load that initializer, but at the same time we don't instantiate `Hanami::Container` at all, that unit test will blow up.

---

[...]

PR #544, opened 2016-03-09
[WIP] Fix container initializer loading

joneslee85:
### What this is about?

Please refer to https://github.com/hanami/hanami/issues/520 for more details.

The expected behaviour is that the project should look into the global `config/initializers` instead of the local app `config/initializers`/


jodosha:
@joneslee85 For new fixtures you add, if they are projects, don't forget to add the `.hanamirc` file for each of them.

jodosha:
Unfortunately delegating the responsibility to `Hanami::Container` can fix the problem of having one centralized loaders from `config/initializers` (instead of `apps/web/config/initializers`), but it doesn't guarantee to be always executed **and** the order of loading.

---

First of all, there are some circumstances where `Hanami::Container` is not instantiated. For instance, when we run a single unit test on a project, in order to keep the start time short, we don't load the full project code.

That means basically to **avoid to instantiate** `Hanami::Container`. Now imagine a unit test relying on an initializer: if we delegate to `Hanami::Container` the role to load that initializer, but at the same time we don't instantiate `Hanami::Container` at all, that unit test will blow up.

---

Right now, we know that `Hanami::Container` is instantiated at the end of the load process, by the generated project code. This loading sequence OK for now: we aim for initializers to be loaded as last thing as we do now. But again, by delegating this role to `Hanami::Container` works by accident.

---

To solve these issues, we need to improve how a project is loaded first, and then apply a new solution for the initializers.

PR #551, opened 2016-03-19
Exit with status code if rake command was raised

### LINKED ISSUE #549: Exit in case a Rake tasks fails
## Background

Each Hanami project ships with a couple of "standard" Rake tasks. These tasks are `db:migrate` and `assets:precompile`. We support them, because of Rails influence in SaaS world. Heroku runs them automatically for all the Ruby (not Rails) apps.

Read more at: http://hanamirb.org/guides/applications/rake/

## Enhancement

We should exit the parent process in case the Rake task fails.

---

Thanks to @thhermansen for suggesting this.

davydovanton:
Fixed #549

/cc @hanami/core-team

### DIFF
--- a/test/rake_helper_test.rb
+++ b/test/rake_helper_test.rb
@@ -2,13 +2,13 @@ require 'test_helper'
 require 'hanami/rake_helper'
 
 describe Hanami::RakeHelper do
+  before do
+    Hanami::RakeHelper.install_tasks
+  end
+
+  let(:app) { Rake.application }
+
   describe '.install_tasks' do
-    before do
-      Hanami::RakeHelper.install_tasks
-    end
-
-    let(:app) { Rake.application }
-
     it 'defines "preload"' do
       task = rake_task "preload"
       task.prerequisites.must_equal []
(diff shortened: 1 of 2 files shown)


test/rake_helper_test.rb
```diff
@@ -30,6 +30,20 @@
     end
   end
 
+  describe 'when "db:migrate" task failed' do
+    it 'exits with failde status code' do
```
joneslee85: s/failde/failed

jodosha:
I manually tested it and it works. Thanks @davydovanton :+1: 

Steps to reproduce:

```shell
➜ hanami new bookshelf --hanami-head=true
➜ cd bookshelf
➜ vim Gemfile # edit: `gem 'hanami', github: 'davydovanton/lotus', branch: 'rake-exit'
➜ bundle
➜ HANAMI_ENV=production bundle exec rake db:migrate # this should raise an error
➜ echo $? # => 1
```

davydovanton:
thanks for merge :+1:

Issue #494, opened 2016-02-05
Container: remove per application Rack::MethodOverride in favor of one at the level of Container middleware stack


jodosha:
## Actual

If we have two applications `web` and `admin`, they both use `Rack::MethodOverride` in their internal middleware stack.

## Proposed

As of v0.6.0 `Hanami::Container` has its own middleware stack which wraps all the applications within a project. We can mount `Rack::MethodOverride` here instead.

klebervirgilio:
What tests would need to be done in order to make sure that's happing?

PR #552, opened 2016-03-22
Fix for issue 494 - Move Rack::MethodOverride middleware from application to container

stevehook:
This is an attempt to fix https://github.com/hanami/hanami/issues/494

I've moved `Rack::MethodOverride` from the application middleware stack to the container, but only when using the container architecture.

If the application is running the application architecture then there is no container middleware stack so I think we need to keep `Rack::MethodOverride` in the per application stack. To detect this I'm checking `Hanami.environment.container?` - is this the correct way to determine the architecture?

There was no obvious way to test the contents of the container middleware stack directly after making the change, I may be missing something here? I've added an integration test that sets up a container app and checks that `Rack::MethodOverride` is present by invoking a PATCH action (it fails without the change).

### DIFF
--- a/lib/hanami/container.rb
+++ b/lib/hanami/container.rb
@@ -52,6 +52,7 @@ module Hanami
         require 'hanami/static'
         @builder.use Hanami::Static
       end
+      @builder.use Rack::MethodOverride
 
       @builder.run @routes
     end
--- a/lib/hanami/middleware.rb
+++ b/lib/hanami/middleware.rb
@@ -100,7 +100,7 @@ module Hanami
         _load_assets_middleware
         _load_session_middleware
         _load_default_welcome_page_for(application)
-        use Rack::MethodOverride
+        _load_method_override_middleware
 
         true
       end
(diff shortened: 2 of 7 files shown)


stevehook:
There is an intermittent spec failure, can be reproduced with `rake test SEED=47157`. Will investigate...

joneslee85:
@stevehook thanks for the PR. At this moment I'm also working on fixing an issue with initialiser at container. You might have known that the container does not handle much logic, it simply do router and app mounting. My PR is to introduce container config like that of the application level. I just want to let you know in adv that my work and your work night conflict with each other. If possible, please get in touch with me on Gitter. We could discuss of possibility making a separate PR for container config so both of our work can leverage.

jodosha:
@stevehook Thanks for this PR. It looks good :+1: 

@joneslee85 I'm merging this because it's a change we want to introduce and doesn't affect too much `Container` (only one line). We're not sure which design your re-arranging work will have. We can't hold this PR for so long. Better to merge it.

PR #553, opened 2016-03-23
Destroy generated route, regardless of url param

### LINKED ISSUE #451: Destroy command: action route doesn't get removed
It affects `v0.6.0`

Steps to reproduce:

```shell
➜ lotus new bookshelf && cd bookshelf && bundle
➜ bundle exec lotus generate action web home#index --url=/
➜ bundle exec lotus destroy action web home#index
➜ cat apps/web/config/routes.rb
get '/', to: 'home#index' # THIS SUPPOSED TO BE REMOVED
# ...
```

---
Ref: #357 
/cc @t4deu

cllns:
Closes #451 .

The code that normally removes the route basically runs the generate command backwards (a.k.a. [`revoke`](https://github.com/erikhuda/thor/blob/f0c2166534e122636f5ce04d61885736ef605617/lib/thor/actions/inject_into_file.rb#L62)). In this case, we want to be more permissive, removing the route regardless of the URL.

Rather than monkey patching Thor, we let it do its thing, then we run a `gsub_file` to remove the offending line. 

I'm not sure we want to add the `@controller_and_action_name` instance variable, but it seems weird to split it to get `@controller_name` and `@action_name`, then join them back together.
Maybe we want to only have`@controller_and_action_name`, and convert `controller_name` and `action_name` into methods. Happy to do that work, if it's the direction we want to go.

### DIFF
--- a/lib/hanami/commands/generate/action.rb
+++ b/lib/hanami/commands/generate/action.rb
@@ -64,9 +64,9 @@ module Hanami
             application_name = File.basename(Dir.pwd)
           end
 
-          controller_and_action_name = Utils::String.new(controller_and_action_name).underscore.gsub(QUOTED_NAME, '')
+          @controller_and_action_name = Utils::String.new(controller_and_action_name).underscore.gsub(QUOTED_NAME, '')
 
-          *controller_name, @action_name = controller_and_action_name.split(ACTION_SEPARATOR_MATCHER)
+          *controller_name, @action_name = @controller_and_action_name.split(ACTION_SEPARATOR_MATCHER)
 
           @application_name = Utils::String.new(application_name).classify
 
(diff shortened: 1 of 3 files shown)


cllns:
I just manually checked the test case from #451 and this fixes that.

Only downside is that when you run `destroy`, stdout shows shows two different things happening for  `apps/web/config.routes`:

```bash
        gsub  apps/web/config/routes.rb
      remove  spec/web/controllers/home/index_spec.rb
      remove  apps/web/controllers/home/index.rb
      remove  apps/web/views/home/index.rb
      remove  apps/web/templates/home/index.html.erb
      remove  spec/web/views/home/index_spec.rb
    subtract  apps/web/config/routes.rb
```
(the `remove`s are all red, while `subtract` and `gsub` are green)

cllns:
Fixed the issue where it said both `gsub` and `subtract`. Now the output of destroying the action is:
```
      remove  spec/web/controllers/book/indx_spec.rb
      remove  apps/web/controllers/book/index.rb
      remove  apps/web/views/book/index.rb
      remove  apps/web/templates/book/index.html.erb
      remove  spec/web/views/book/index_spec.rb
    subtract  apps/web/config/routes.rb
```
which is the same as it is without this PR.

jodosha:
This works @cllns Thank you! 👍

PR #554, opened 2016-03-24
Fix spec description

ariejan:
Bit of a lame pull request, but herer it is. 

I was looking into why my server does code reloading and encountered an erroneous description in the specs. Fixed it.


jodosha:
@ariejan Hey, thank you very much! All the PRs that improve the code are always well welcome. Merging! :+1:

PR #555, opened 2016-03-24
Remove simplecov

jodosha:


Rakefile
```diff
@@ -9,7 +9,7 @@ end
 
 namespace :test do
   task :coverage do
-    ENV['COVERAGE'] = 'true'
+    ENV['COVERALL'] = 'true'
```
cllns: Would `COVERALLS` be better than `COVERALL`?

PR #557, opened 2016-03-26
Namespace container

joneslee85:
## What is this about?
To namespace container project.

## Why are we doing this?
There is a needs for a placeholder to contain global project config/logic codes. It makes sense to place them in container level, that is `Hanami::Container`.

By having a class that extends `Hanami::Container`, we could allow user to customise this layer without monkey-patching this class.

before:

```
Hanami::Container.configure do; end;
```

after:

```
module YourApp
  class Container < Hanami::Container
    configure do
    end

    #extra logic
  end
end
```

## Notes
The Release Notes should document about this change to avoid confusion for users.

cc @hanami/core-team

### DIFF
--- a/lib/hanami/commands/generate/app.rb
+++ b/lib/hanami/commands/generate/app.rb
@@ -65,8 +65,8 @@ module Hanami
         end
 
         def add_mount_app
-          generator.inject_into_file base_path.join('config/environment.rb'), after: /Hanami::Container.configure do/ do |match|
-            "\n  mount #{ classified_app_name }::Application, at: '#{ application_base_url }'"
+          generator.inject_into_file base_path.join('config/environment.rb'), after: /configure do/ do |match|
+            "\n      mount #{ classified_app_name }::Application, at: '#{ application_base_url }'"
           end
         end
 
--- a/lib/hanami/commands/new/container.rb
+++ b/lib/hanami/commands/new/container.rb
@@ -18,14 +18,15 @@ module Hanami
 
         def template_options
           {
-            app_name:             app_name,
-            hanami_head:          hanami_head?,
-            test:                 test_framework.framework,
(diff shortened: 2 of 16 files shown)


jodosha:
@joneslee85 Thanks for this improvement! :smile:

`Hanami::Container` internally uses class variables (eg. `@@options`). This is thread safe if it's a singleton within the Ruby process. But now that we're inheriting from it, it's better to use _class level instance variables_ (eg. `@options`). See: http://www.railstips.org/blog/archives/2006/11/18/class-and-instance-variables-in-ruby/

jodosha:
@joneslee85 If a few case, we're using `Hanami::Container.new` as mechanism to preload the framework and **all** the user project code.

It's the case of [`hanami console`](https://github.com/hanami/hanami/blob/b003f10606ce11d40843cf74d1dc087992d00c4f/lib/hanami/commands/console.rb#L83).

---

With this proposed change, the subclass will hold the configuration, so `Hanami::Container.new` will not work anymore as now. How we can programmatically guess the subclass name? To make it work we should do `Bookshelf::Container.new`.

Rails has a similar problem: `Blog` is a subclass of `Rails::Application`. To safely access that subclass, they introduced `Rails.application`, which is a singleton. But if we get back to the singleton (`Hanami.container` ?), we haven't solved the problem that this PR aims to fix, because we already have `Hanami::Container` as a singleton.

jodosha:
@joneslee85 We need to amend [Capybara generator](https://github.com/hanami/hanami/blob/b003f10606ce11d40843cf74d1dc087992d00c4f/lib/hanami/generators/application/container/capybara.rb.rspec.tt#L5), which still refer to `Hanami::Container`.

jodosha:
Closing this PR as stale. If we think this will be needed, please bring the discussion back.

PR #565, opened 2016-05-02
Update ruby version to 2.3.1 & 2.2.5 on travis

vyper:
- https://www.ruby-lang.org/en/news/2016/04/26/ruby-2-3-1-released/
- https://www.ruby-lang.org/en/news/2016/04/26/ruby-2-2-5-released/


jodosha:
@vyper Thank you! 👍

PR #566, opened 2016-05-04
Require gems in bin/hanami file

davydovanton:
## Why this patch is important?
Now I working on integration the [rodauth][rodauth] gem and hanami application. And [I'm creating gem][gem] wrapper for this. On this gem I want to call custom `auth` generator like this:

```
$ bundle exec hanami generate auth
```

For this I need to require my gem to `bin/hanami` file. And with this patch I'll be have a simple way to creating custom actions or generators.

For example, if you call `hanami g` with my gem you see this output:

screenshot 2016-05-04 04 01 07

/cc @hanami/contributors and @hanami/core-team 

[gem]: https://github.com/davydovanton/hanami-rodauth
[rodauth]: http://rodauth.jeremyevans.net

### DIFF
--- a/bin/hanami
+++ b/bin/hanami
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
-
 require 'bundler'
 require 'hanami/cli'
+require 'hanami/setup'
+
 Hanami::Cli.start
--- a/lib/hanami/cli_base.rb
+++ b/lib/hanami/cli_base.rb
@@ -1,0 +1,30 @@
+require 'thor'
+
+module Hanami
+  class CliBase < Thor
+    # Add new custom CLI command to special CLI class
+    #
+    # @since x.x.x
+    #
+    # @example Usage with Cli class
+    #   require 'hanami'
+    #   require 'hanami/cli'
+    #
+    #   Hanami::Cli.custom_commands do
+    #     desc 'custom', 'Generate a something'
+    #     long_desc <<-EOS
+    #       long description for your custom command
+    #     EOS
+    #     def custom
+    #       if options[:help]
+    #         invoke :help, ['auth']
+    #       else
+    #         # ...
+    #       end
+    #     end
+    #   end
+    def self.custom_commands(&blk)
+      class_eval(&blk) if block_given?
+    end
+  end
+end
--- a/lib/hanami/cli_base.rb
+++ b/lib/hanami/cli_base.rb
@@ -1,7 +1,5 @@
-require 'thor'
-
 module Hanami
-  class CliBase < Thor
+  module CliBase
     # Add new custom CLI command to special CLI class
     #
     # @since x.x.x
@@ -10,7 +8,7 @@ module Hanami
     #   require 'hanami'
     #   require 'hanami/cli'
     #
-    #   Hanami::Cli.custom_commands do
+    #   Hanami::Cli.define_commands do
     #     desc 'custom', 'Generate a something'
     #     long_desc <<-EOS
     #       long description for your custom command
@@ -23,7 +21,7 @@ module Hanami
     #       end
     #     end
     #   end
-    def self.custom_commands(&blk)
+    def define_commands(&blk)
       class_eval(&blk) if block_given?
     end
   end
--- a/lib/hanami/cli_base.rb
+++ b/lib/hanami/cli_base.rb
@@ -1,8 +1,11 @@
 module Hanami
   module CliBase
-    # Add new custom CLI command to special CLI class
+    # Add new custom CLI command to special CLI class.
+    # Please be careful. This is a private method that
+    # can be deleted soon.
     #
     # @since x.x.x
+    # @api private
     #
     # @example Usage with Cli class
     #   require 'hanami'
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -1,11 +1,13 @@
+require 'thor'
 require 'hanami/cli_base'
 require 'hanami/commands/console'
 require 'hanami/commands/new/app'
 require 'hanami/commands/new/container'
 
 module Hanami
-  class Cli < CliBase
+  class Cli < Thor
     # include Thor::Actions
+    extend CliBase
 
     desc 'version', 'Prints hanami version'
     long_desc <<-EOS
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -1,10 +1,10 @@
-require 'thor'
+require 'hanami/cli_base'
 require 'hanami/commands/console'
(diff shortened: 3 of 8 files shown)


jodosha:
@davydovanton Hey thanks for working on this integration. That means a lot for us and it's definitely the right direction since now on. Bravo! 👏 

We need to support your effort within the framework. But before I have a question: did you tried by requiring `hanami/setup` from your code? It seems to do the same thing that you're looking for. https://github.com/hanami/hanami/blob/b4fbed780a3df64d9b7697d66ee56f932c0f57d3/lib/hanami/setup.rb

Please let me know.

davydovanton:
@jodosha thanks for review 👍 

Yes, I tried to require `hanami/setup` to my code but it didn't work😞 Also I changed `Bundler.require` call to required `hanami/setup`.

jodosha:
@davydovanton Does the last change at https://github.com/hanami/hanami/pull/566/commits/9a8db0445fec53046c2b65e96987569a2bf1c005 work for your app?

davydovanton:
@jodosha yep, all work fine :data:

jodosha:
@davydovanton One last thing before to merge this: how did you managed to hook into `hanami generate` command? I've seen that you're able to do `hanami g auth`. That seems great!

davydovanton:
@jodosha it was hard, but I founded solution: I reopen `Module#included` method and call `class_eval` for included class. Code here: https://github.com/davydovanton/hanami-rodauth/blob/master/lib/hanami/auth.rb
and here:
https://github.com/davydovanton/hanami-rodauth/blob/master/lib/hanami.rb#L8

jodosha:
@davydovanton It's hackish.. 😿  Let's put this way: what you would do if you had to build third party integration from scratch? For instance: how to register commands and subcommands?

I don't want to be picky, I just think that this is a good chance to implement something useful for the Community, and give you the chance to implement it. WDYT?

davydovanton:
@jodosha yep, you're right it's hachish 😞 

I want to write blog post about it. Also I think I can add manual to guides too. What do you think?

_**P.S.:** I created [gist](https://gist.github.com/davydovanton/9450519b0f645e041eb3ccc807444dc6) with simple example_

davydovanton:
Also I think we can to create special method or logic for implement custom CLI commands. Smth like this:

``` ruby
require 'hanami/cli'

module Hanami
  class CustomCommand
    add_cli_command do
      desc 'custom_command', 'Desc'
      long_desc <<-EOS
      EOS
      def auth
        Hanami::Commands::CustomCommand.new(options).start
      end
    end
  end
end
```

davydovanton:
@jodosha I created special helper for adding new CLI commands. Now custom commands creates like this:
``` ruby
Hanami::Cli.add_cli_commands do
  desc 'custom', 'Empty command'
  def custom
    puts 'custom command'
  end
end
```

bin/hanami
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env ruby
-
 require 'bundler'
 require 'hanami/cli'
+require 'hanami/setup'
```
jodosha: @davydovanton Is this still needed?

davydovanton: yep, I need to load "plugins" for calling `#custom_commands` in this gems

lib/hanami/cli_base.rb
```diff
@@ -0,0 +1,30 @@
...
+    #     EOS
+    #     def custom
+    #       if options[:help]
+    #         invoke :help, ['auth']
+    #       else
+    #         # ...
+    #       end
+    #     end
+    #   end
+    def self.custom_commands(&blk)
```
jodosha: @davydovanton what about `define_command`?

davydovanton: @jodosha looks better, but in this block you can define more than one command. What about `define_commands`?

lib/hanami/cli_base.rb
```diff
@@ -0,0 +1,30 @@
+require 'thor'
+
+module Hanami
+  class CliBase < Thor
```
jodosha: @davydovanton It looks like we're using inheritance to share code. For this purpose, a module to include it's better.

What if we turn this into into a module and let the classes to include it?

davydovanton: @jodosha I think that base class is more idiomatically for different commands. But also I can agree with you too.

jodosha: @davydovanton I think composition is better for roles, while inheritance for behavior specialization.

In this case, all the subclasses of `CliBase` exist to share the code (which is a role in this case) and they don't specialize base class behavior. So composition (via module inclusion) is preferrable here. 😄

lib/hanami/cli_base.rb
jodosha: @davydovanton Ok, I'm fine with the plural as well.

davydovanton:
@jodosha done 🎉

jodosha:
@davydovanton How I can add my own generator?

Whould this be possible?

```shell
hanami generate resource book
```

davydovanton:
@jodosha you need call `define_commands` class method for CLI class and put your command to block. I.e.:
``` ruby
Hanami::Cli.define_commands do
  desc 'custom', 'Generate a something'
  long_desc <<-EOS
    long description for your custom command
  EOS
  def custom
    if options[:help]
      invoke :help, ['auth']
    else
      # ...
    end
  end
end
```

Yes, of course, you can create any thor cli command.

jodosha:
@davydovanton Sorry for the misunderstanding. What I really mean is: can we **decorate existing commands?**.

We already provide `generate`, if we merge this, can we add a subcommand to it? From your reply it looks like only new commands can be added: `hanami foo`, but not `hanami generate foo`.

davydovanton:
@jodosha oh, I understand. Yes, we can add subcommand for any command. For this you need to call method for special class (`Hanami::CliSubCommands::Generate` for generators or `Hanami::CliSubCommands::DB` for `hanami db` commands). I.e.:

``` ruby
Hanami::CliSubCommands::Generate.define_commands do
  desc 'custom', 'Generate a something'
  long_desc <<-EOS
    long description for your custom command
  EOS
  def foo
    if options[:help]
      invoke :help, ['foo']
    else
      # ...
    end
  end
end
```

This code will create `hanami g foo` command.

davydovanton:
@jodosha I'll add test for subcommands. Give me 1 minute 😊

davydovanton:
@jodosha done. You can see worked example in [here](https://github.com/hanami/hanami/pull/566/commits/72058fbb120369de596a9fad746f726fe33a3d13)

davydovanton:
Thanks for merge @jodosha 👍

PR #567, opened 2016-05-07
Fix usage of Mutex

flash-gordon:
Hi,

I found that Hanami uses mutexes improperly by calling `synchronize` alongside with `Mutex.new`. It is not how synchronization meant to work because you must create `Mutex` object in advance and `synchronize` your code later using the object. Otherwise there would be no synchronization at all o_O. Please take a look at my snippets:
```ruby
# Use Mutex.new every time we run #synchronize
jruby-head :001 > def synchronize
jruby-head :002?>   Mutex.new.synchronize do
jruby-head :003 >     yield
jruby-head :004?>   end
jruby-head :005?> end
 => :synchronize
jruby-head :006 >
jruby-head :007 > $gvar = 0
 => 0
jruby-head :008 >
jruby-head :009 > 5.times { Thread.new { 10_000.times { synchronize { $gvar += 1 } } } }
 => 5
jruby-head :010 > $gvar
 => 33393
```
```ruby
# Use shared Mutex value
 jruby-head :001 > $mutex = Mutex.new
 => #<Mutex:0x27ae2fd0>
jruby-head :002 >
jruby-head :003 > def synchronize
jruby-head :004?>   $mutex.synchronize do
jruby-head :005 >     yield
jruby-head :006?>   end
jruby-head :007?> end
 => :synchronize
jruby-head :008 >
jruby-head :009 > $gvar = 0
 => 0
jruby-head :010 >
jruby-head :011 > 5.times { Thread.new { 10_000.times { synchronize { $gvar += 1 } } } }
 => 5
jruby-head :012 > $gvar
 => 50000
```

I replaced all `Mutex.new` calls with class variables so everything should be fine now. I also had to assign a value to `@@applications` explicitly due to a deadlock on lazy assignment.

### DIFF
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -122,6 +122,13 @@ module Hanami
     end
 
     class << self
+      # @since 0.8.0
+      # @api private
+      @@mutex = Mutex.new
+
+      # @since 0.2.0
+      # @api private
+      @@applications = Set.new
 
       # Registry of Hanami applications in the current Ruby process
       #
@@ -130,9 +137,7 @@ module Hanami
       # @since 0.2.0
       # @api private
       def applications
-        synchronize do
-          @@applications ||= Set.new
-        end
+        @@applications
       end
 
       # Configure the application.
@@ -238,7 +243,7 @@ module Hanami
       # @since 0.2.0
       # @api private
       def synchronize
-        Mutex.new.synchronize do
+        @@mutex.synchronize do
           yield
         end
       end
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -40,6 +40,10 @@ module Hanami
       end
     end
 
+    # @since 0.8.0
+    # @api private
+    LOCK = Mutex.new
+
     # Return the routes for this application
     #
     # @return [Hanami::Router] a route set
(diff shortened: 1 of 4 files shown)


Erol:
Good catch! 👍

davydovanton:
What do you think, can we use constant instead class variable?

flash-gordon:
@davydovanton yeah, I don't like class variables either! I even hate them for their really weird behavior. Although I was careful enough to avoid rough edges I would be happy to replace all `@@mutex` with something like `_LOCK`. I'm still not sure about `@@applications` because the value is supposed to change during startup time and declaring it as a constant is confusing a bit in my opinion. It can be handled with something like this (if you wish of course)
```ruby
@applications = Set.new

def applications
  if self. equal?(Hanami::Application)
    @applications
  else
    Hanami::Application.applications
  end
end
```

davydovanton:
@flash-gordon yes, I agree with `@@application`. It's better to use class variable here. But I'll be happy if you replace mutex object on constant :)

flash-gordon:
OK, done!

jodosha:
@flash-gordon Ouch, I wrote that code, dunno how it was possible.. Thanks for the good catch! 👍

PR #569, opened 2016-05-09
Add notice to about SERVE_STATIC_ASSETS

### LINKED ISSUE #562: SERVE_STATIC_ASSETS for production
Right now, SERVE_STATIC_ASSETS = "true" for development and test, but not production.

This makes sense when deploying to a server like Nginx.

But for Heroku, [SERVE_STATIC_ASSETS must be "true"](https://devcenter.heroku.com/articles/rails-4-asset-pipeline#serve-assets).

I ran across this issue when deploying to Heroku (says something like `/public/assets.json` not found) and so did @jxxcarlson.

The [Assets guide 'overview' covers this](https://github.com/hanami/hanami.github.io/blob/build/source/guides/assets/overview.md#serve-static-assets) but it's not clear everyone will read that before deploying to Heroku.

A few options I see, in order of most helpful fixing this problem to least:

1. Consider changing the default for production to SERVE_STATIC_ASSETS="true", and have the guides instruct users to set it to false if using Nginx, etc.? The [Hanami homepage](http://hanamirb.org/) has instructions for how to deploy to Heroku. The upside here is that "it just works" for all hosting environments. The downside is that, for Nginx you don't want your assets going through Ruby.

2. Is it possible to add a message when deploying a Hanami app to Heroku and `SERVE_STATIC_ASSETS="false"`, alerting them that they need to set SERVE_STATIC_ASSETS to "true"?

3. At the very least, can we add the specific error message to the Assets 'Overview' guide, to facilitate searching for the error? (or create a 'Deploying' Assets guide)

cllns:
Resolves #562.

Since Heroku is the only one who should be running `rake assets:precompile` (other should use `hanami assets precompile`, this is a good place to tell users about `SERVE_STATIC_ASSETS=true`.

I happily welcome suggestions for improvements to the message :)

### DIFF
--- a/lib/hanami/rake_helper.rb
+++ b/lib/hanami/rake_helper.rb
@@ -60,6 +60,11 @@ module Hanami
 
       namespace :assets do
         task :precompile do
+          puts "=============================================================="
+          puts "NOTE: In order to serve static assets on Heroku (and others), "
+          puts "the environment variable SERVE_STATIC_ASSETS must equal 'true'"
+          puts "To do this, run `heroku config:set SERVE_STATIC_ASSETS=true`  "
+          puts "=============================================================="
           system("bundle exec hanami assets precompile") || exit($?.exitstatus)
         end
       end
--- a/test/integration/rake/test_application.rb
+++ b/test/integration/rake/test_application.rb
@@ -114,10 +114,24 @@ describe 'Rake tasks (Application)' do
     end
 
     it "precompile assets" do
-      success = system "bundle exec rake assets:precompile"
-      success.must_equal true
+      capture_subprocess_io {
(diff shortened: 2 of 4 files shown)


davydovanton:
@cllns nice 👍
But I have a idea for discussion. What do you think if we colorized this message to red or yellow color?
``` ruby
# you can use this for red color message
puts "\e[31m RED \e[0m"

# and this for yellow
puts "\e[33m YELLOW \e[0m"
```

jodosha:
@cllns Thank you, but I'm closing this because Heroku is just one of the vendors. See the complete explanation here: https://github.com/hanami/hanami/issues/562#issuecomment-222083389

Issue #572, opened 2016-05-19
Hanami assets slow in development


pascalbetz:
Reported by Nick Sutterer.

Issues are served slow in development mode:

127.0.0.1 - - [17/May/2016:08:31:19 +1000] "GET /assets/bootstrap.min.css HTTP/1.1" 304 - 1.5921
127.0.0.1 - - [17/May/2016:08:31:19 +1000] "GET /assets/offcanvas.css HTTP/1.1" 304 - 1.6118
127.0.0.1 - - [17/May/2016:08:31:19 +1000] "GET /assets/offcanvas.js HTTP/1.1" 304 - 1.7339
127.0.0.1 - - [17/May/2016:08:31:19 +1000] "GET /assets/bootstrap.min.js HTTP/1.1" 304 - 1.7420

PR #573, opened 2016-05-19
Serve static assets via Shotgun::Static

pascalbetz:
This fixes https://github.com/hanami/hanami/issues/572 as discussed on gitter.

Before:

bildschirmfoto 2016-05-19 um 19 48 02

After:
bildschirmfoto 2016-05-19 um 19 59 40

(Tested with https://github.com/apotonick/gemgem-hanami)

### DIFF
--- a/lib/hanami/server.rb
+++ b/lib/hanami/server.rb
@@ -49,10 +49,34 @@ module Hanami
       mw = Hash.new { |e, m| e[m] = [] }
       mw["deployment"].concat([::Rack::ContentLength, ::Rack::CommonLogger])
       mw["development"].concat(mw["deployment"] + [::Rack::ShowExceptions, ::Rack::Lint])
+      mw["development"].push(::Shotgun::Static) if enable_shotgun_static?
       mw
     end
 
     private
+
+    # @since x.x.x
+    # @api private
+    def enable_shotgun_static?
+      code_reloading? && shotgun_available?
+    end
+
+    # @since x.x.x
+    # @api private
+    def shotgun_available?
+      begin
+        require 'shotgun'
+        true
+      rescue LoadError
+        false
+      end
+    end
+
+    # @since x.x.x
+    # @api private
+    def code_reloading?
+      @_env.code_reloading?
+    end
 
     # @since 0.8.0
     # @api private
--- a/lib/hanami/server.rb
+++ b/lib/hanami/server.rb
@@ -64,12 +64,10 @@ module Hanami
     # @since x.x.x
     # @api private
     def shotgun_available?
-      begin
-        require 'shotgun'
-        true
-      rescue LoadError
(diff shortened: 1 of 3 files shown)


lib/hanami/server.rb
```diff
@@ -49,11 +49,35 @@ def middleware
...
+    def enable_shotgun_static?
+      code_reloading? && shotgun_available?
+    end
+
+    # @since x.x.x
+    # @api private
+    def shotgun_available?
+      begin
+      require 'shotgun'
+      true
```
cllns: indent 68 & 69?

pascalbetz:
@cllns fixed formatting.

test/server_test.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'test_helper'
```
cllns: Most other tests use `require 'test_helper'` instead of `require_relative`

test/server_test.rb
```diff
@@ -0,0 +1,17 @@
...
+    middlewares = Hanami::Server.new(code_reloading: true).middleware['development'].map(&:name)
+    middlewares.must_include('Shotgun::Static')
+  end
+
+  it 'does not add Shotgun::Static when code reloading is disabled' do
+    middlewares = Hanami::Server.new(code_reloading: false).middleware['development'].map(&:name)
+    middlewares.wont_include('Shotgun::Static')
+  end
+
+
```
cllns: remove empty lines

test/server_test.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'test_helper'
+require 'hanami/server'
+
```
cllns: remove empty line

cllns:
@pascalbetz Awesome! Just found a couple other things. Sorry I didn't batch them together 😅

pascalbetz:
@cllns updated. will squash when everything is fine.

lib/hanami/server.rb
```diff
@@ -49,11 +49,35 @@ def middleware
...
+
+    # @since x.x.x
+    # @api private
+    def shotgun_available?
+      begin
+        require 'shotgun'
+        true
+      rescue LoadError
+        false
+      end
```
davydovanton: May be is better to remove `begin` and use only `rescue`? What do you think? I.e.:
``` ruby
def shotgun_available?
  require 'shotgun'
  true
rescue LoadError
  false
end
```

pascalbetz: I don't like implicit begin blocks (because we always will need explicit begin blocks as well and i don't like two styles for the same thing) but the ruby styleguide prefers them. Whatever is more common and preferred in Hanami.

jodosha:
@pascalbetz Thank you! 👍

Issue #571, opened 2016-05-16
routes helper ignores mount point when testing


mereghost:
Apparently the routes helper ignores the mount point of the application when running tests.

The following code should be enough to reproduce the error:

apps/web/controllers/stuffs/index.rb
```ruby
module Web::Controllers::Stuffs
  class Index
    include Web::Action
    def call(params)
      self.headers.merge!({'Location' => String.new(routes.stuff_url(1)))
      self.body = 'Irrelevant'
    end
  end
end
```
apps/web/config/routes.rb
```ruby
resources :stuffs
```

config/environment.rb
```ruby
# ...
Hanami::Container.configure { mount Web::Application, at: '/my/path' }
```

spec/web/features/return_a_weird_location_header_spec.rb
```ruby
require 'features_helper'

describe 'return a weird location header' do
  it 'respects the mount path' do
    visit '/my/path/stuffs'
    expect(page.driver.response.headers['Location']).to match /my\/path/
  end
end
```

I have tested it with Capybara (and later with Rack::Test only) both on stable and head. It seems that, while the router recognizes the mount point, as the request is successful, it does not pass this knowledge down to the controller.

Let me know if any further info is necessary.

Erol:
Hi @jodosha, would you mind if I take a stab at this? I was able to trace why this is happening and can open a WIP PR where we can discuss how it can be resolved.

jodosha:
@mereghost Hi Marcello, can you please check if #579 fixes your problem? Thanks!

mereghost:
@jodosha Hi Luca, it works. Thanks for taking a stab at it @Erol and @jodosha  =)

jodosha:
Fixed by #579

PR #579, opened 2016-05-26
Fix route mounting on testing environments

Erol:
This fix attempts to address #571 by initializing `Hanami::Container` before preloading all Hanami applications on testing environments.

Without an initialized `Hanami::Container`, the Hanami application loader fails to see the mounting configuration for the application and assumes a `prefix` of `nil`.

I also observed that on testing environments, `Hanami::Loader#_assign_routes_to_application_module!` gets called twice: the first one with a route prefix of `nil` and the second one with the correct mount prefix. An alternative is to modify https://github.com/hanami/hanami/blob/master/lib/hanami/loader.rb#L217-L222 to drop the Routes factory constant so it can be re-assigned again:

```ruby
def _assign_routes_to_application_module!
  application_module.send(:remove_const, 'Routes') if application_module.const_defined?('Routes')
  routes = Hanami::Routes.new(application_routes)
  application_module.const_set('Routes', routes)
end
```

What do you guys think?

### DIFF
--- a/lib/hanami/container.rb
+++ b/lib/hanami/container.rb
@@ -4,16 +4,22 @@ require 'hanami/router'
 
 module Hanami
   class Container
-    class Router < ::Hanami::Router
+    class Mounting
+      def self.load!(&blk)
+        new.instance_exec(&blk)
+      end
+
       def mount(app, options)
-        app = app.new(path_prefix: options.fetch(:at)) if hanami_app?(app)
-        super(app, options)
+        app.configuration.path_prefix options.fetch(:at) if hanami_app?(app)
       end
 
       private
 
       def hanami_app?(app)
         app.ancestors.include? Hanami::Application
+      end
+
+      def method_missing(m, *args)
       end
     end
 
@@ -23,6 +29,8 @@ module Hanami
       Mutex.new.synchronize do
         @@options       = options
         @@configuration = blk
+
+        Mounting.load!(&blk)
       end
     end
 
--- a/lib/hanami/generators/application/container/spec_helper.rb.minitest.tt
+++ b/lib/hanami/generators/application/container/spec_helper.rb.minitest.tt
@@ -4,6 +4,4 @@ ENV['HANAMI_ENV'] ||= 'test'
 require_relative '../config/environment'
 require 'minitest/autorun'
 
-$container = Hanami::Container.new
-
 Hanami::Application.preload!
--- a/lib/hanami/generators/application/container/spec_helper.rb.minitest.tt
+++ b/lib/hanami/generators/application/container/spec_helper.rb.minitest.tt
@@ -4,4 +4,6 @@ ENV['HANAMI_ENV'] ||= 'test'
 require_relative '../config/environment'
 require 'minitest/autorun'
 
+$container = Hanami::Container.new
+
 Hanami::Application.preload!
--- a/test/integration/container_test.rb
+++ b/test/integration/container_test.rb
@@ -10,9 +10,10 @@ describe Hanami::Container do
       mount Back::Application,  at: '/back'
     end
 
-    @container = Hanami::Container.new
     Front::Application.load!
     Back::Application.load!
+
+    @container = Hanami::Container.new
   end
 
   def app
--- a/test/integration/container_test.rb
+++ b/test/integration/container_test.rb
@@ -45,7 +45,7 @@ describe Hanami::Container do
       'GET, HEAD  /back/users                    Back::Controllers::Users::Index'
     ]
     matches.each do |match|
-      @container.routes.inspector.to_s.must_match match
+      app.routes.inspector.to_s.must_match match
     end
   end
 
--- a/lib/hanami/generators/application/container/features_helper.rb.minitest.tt
+++ b/lib/hanami/generators/application/container/features_helper.rb.minitest.tt
@@ -4,7 +4,7 @@ require_relative './spec_helper'
 require 'capybara'
 require 'capybara/dsl'
 
-Capybara.app = Hanami::Container.new
+Capybara.app = $container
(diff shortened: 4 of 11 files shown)


jodosha:
@Erol Hey thanks for this PR, but can you please explain it? I'm confused.

In the description you mentioned some changes to `Hanami::Loader`, but the current proposal only changes our fixtures and how the application will be generated.

lib/hanami/generators/application/container/spec_helper.rb.minitest.tt
```diff
@@ -4,4 +4,6 @@ ENV['HANAMI_ENV'] ||= 'test'
 require_relative '../config/environment'
 require 'minitest/autorun'
 
+$container = Hanami::Container.new
```
jodosha: @Erol This can't be initialized here.

If we initialize `Hanami::Container` here that means that the entire framework **and all the project code** (apps, controllers, models, views) will be loaded. This is fine for _feature tests_, where we want to test the entire stack, but for single unit tests isn't the case.

In other words, when you run this unit test.

```ruby
bundle exec rspec spec/web/controllers/home/index_spec.rb
```

Hanami only loads the framework and app configurations (`Hanami::Application.preload!`). Then that unit test has a `require_relative '../../../../apps/web/controllers/home/index'` instruction to load only that object in isolation. This makes TDD possible with Hanami.

---

I understand what you're trying to do here, but instead of initialize it, we should rely on a different strategy like `Hanami::Container.preload!`. This should only look at the mounted applications and **without to initialize them**, invoke `.preload!` on them by passing the mounting path.

Erol:
Sorry for not being clear. I found two approaches to address the route mounting issue.

---

The first approach is what's currently commited in this PR. It involves modifying the generators to initialize `Hanami::Container` before calling `Hanami::Application.preload!`. Without an initialized `Hanami::Container`, the application loader fails to see the mounting configuration for the application and assumes a prefix of `nil`.

This does not work:

```ruby
Hanami::Container.configure do
  mount Web::Application, at: '/my/path'
end

Hanami::Application.preload!
```

This works and will pass the correct mounting prefix:

```ruby
Hanami::Container.configure do
  mount Web::Application, at: '/my/path'
end

Hanami::Container.new
Hanami::Application.preload!
```

---

The second approach does not touch any of the test fixture generators. Instead, we can modify `Hanami::Loader` to allow changing the `Routes` factory even if it has already been defined:

```ruby
def _assign_routes_to_application_module!
  application_module.send(:remove_const, 'Routes') if application_module.const_defined?('Routes')
  routes = Hanami::Routes.new(application_routes)
  application_module.const_set('Routes', routes)
end
```

I observed that on testing environments, the above method gets called twice. The first call will have a mount prefix of `nil` but the second one has the correct mount prefix. By allowing `Routes` to be redefined, the correct routes factory is defined.

I'll also commit this second approach so the changes are clear.

lib/hanami/generators/application/container/spec_helper.rb.minitest.tt
Erol: Maybe we can just go with the second approach then? I just committed the code for it. We only need to allow redefining the `Routes` application module constant which stores the routes.

jodosha:
@Erol 

> I observed that on testing environments, the above method gets called twice.

This is an implicit behavior, that I even discovered right now. We can't write a patch on the assumption that it will be called twice forever.

Please investigate this approach here: https://github.com/hanami/hanami/pull/579#discussion_r64870758

lib/hanami/generators/application/container/spec_helper.rb.minitest.tt
jodosha: Please have a look at: https://github.com/hanami/hanami/pull/579#issuecomment-222100561

Erol:
Sure. I'll update the PR once I come up with something 👍

Erol:
Hi @jodosha, I'd like your feedback on this new fix.

Instead of waiting for `Hanami::Container` to be initialized to load application prefixes, they are now evaluated and assigned on `Hanami::Container.configure` via a `Hanami::Container::Mounting` class.

I also dropped `Hanami::Container::Router` since the custom `mount` code to configure prefixes on it is no longer required.

WDYT?

test/integration/container_test.rb
```diff
@@ -10,9 +10,10 @@
       mount Back::Application,  at: '/back'
     end
 
-    @container = Hanami::Container.new
     Front::Application.load!
     Back::Application.load!
+
+    @container = Hanami::Container.new
```
Erol: I delayed the creation of a `Hanami::Container` instance to simulate what happens on generated test suites. We can also just opt to initialize the container instance on tests that require it.

lib/hanami/container.rb
```diff
@@ -4,17 +4,23 @@
...
+        app.configuration.path_prefix options.fetch(:at) if hanami_app?(app)
       end
 
       private
 
       def hanami_app?(app)
         app.ancestors.include? Hanami::Application
       end
+
+      def method_missing(m, *args)
```
jodosha: @Erol Why this is needed?

Erol: @jodosha I wasn't sure what directives are allowed inside `Hanami::Container.configure`. If `mount` is the only directive we allow, this is not needed.

If there are other directives, I think I can also stub them using actual methods instead of `method_missing`.

jodosha: @Erol I see, maybe this should be documented, or I will forget soon 😂 .

Erol: @jodosha You are right, will document that. I'll update this later!

jodosha:
@Erol I just merged this, thanks. I checked with an app of mine.

PR #581, opened 2016-05-27
Fix various doc typos

Erol:


jodosha:
@Erol Thank you very much, merging this 👍

PR #582, opened 2016-05-27
Optimize route helpers by defining actual methods after being called once

Erol:
I'd like to get feedback if this is a worthy optimization.

This update _caches_ route helper methods by defining an actual method whenever a dynamic route helper is called. This way, multiple successive calls to `method_missing` to call a particular route helper is unnecessary when running on the same process.

Example:

```ruby
routes.books_path #=> First call is slow. Evaluated via #method_missing
routes.books_path #=> Second call is fast. Calls actual #books_path method
```

### DIFF
--- a/lib/hanami/routes.rb
+++ b/lib/hanami/routes.rb
@@ -139,10 +139,18 @@ module Hanami
     # @since 0.3.0
     # @api private
     def method_missing(m, *args)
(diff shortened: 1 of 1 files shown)


jodosha:
@Erol Thanks for this PR, I don't like runtime changes to objects and it isn't thread safe. 👎

Erol:
Got it. Thanks for the feedback!

PR #583, opened 2016-05-27
Create format option for logger

davydovanton:
Reference to https://github.com/hanami/utils/pull/122

/cc @jodosha

### DIFF
--- a/lib/hanami/config/logger.rb
+++ b/lib/hanami/config/logger.rb
@@ -10,6 +10,7 @@ module Hanami
 
       def initialize
         @stream = STDOUT
+        @format = ::Hanami::Logger::Formatter.new
       end
 
       # Log stream.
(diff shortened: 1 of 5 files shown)


jodosha:
@davydovanton Thanks for this PR, but we shouldn't hardcode that logic. We probably want a per env setting in configuration blocks.

```ruby
module Web
    configure do
      logging.formatter :json
    end
  end
end
```

davydovanton:
@jodosha good point, I like it 👍

davydovanton:
@jodosha done :tada

joneslee85:
👍 @davydovanton

jodosha:
@davydovanton This is almost done 😉 .

My suggestion is to generate [new applications](https://github.com/hanami/hanami/blob/fc4abbe5232df59cf20dd6608920434b57bc1dfd/lib/hanami/generators/app/application.rb.tt#L295) with this setting.

```ruby
configure :development do
  # Logger format. It defaults to DEFAULT
  # logger.format :default
end

configure :production do
  logger.format :json
end
```

davydovanton:
yep, I liked it 👍

jodosha:
@davydovanton Thank you! Just merged.

PR #587, opened 2016-05-28
Disable Rake::TestTask warnings in generated apps/containers

ericfreese:
Proposal to address hanami/utils#123 by generating Rakefiles with `Rake::TestTask` configured to disable warnings as proposed in https://github.com/hanami/utils/issues/123#issuecomment-219198446.

I haven't looked too deeply into whether this is a problem with RSpec, but I'd guess it's not, since RSpec appears [not to use `Rake::TestTask`](https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/rake_task.rb).

Thoughts?


joneslee85:
Thanks for the PR but I do not think it is a good idea t disable warning as it hides the warnings that are supposed to get addressed. Due to the big scope of the warning, I would love to see many PRs to address all warnings .

ericfreese:
I'm all for showing warnings in my own code, but the problem seems to come when I get warnings for all the gems I depend on (and gems they depend on, and so on). It doesn't seem realistic to expect warnings to be fixed in every gem.

My guess would be that most Hanami users are not interested in seeing these warnings every time they run their tests (I could absolutely be wrong here, since I'm new to the project), but gem developers and others who are interested in addressing these warnings can always enable them by deleting this line.

My vote would go for disabling the warnings by default, but I'd love to hear more opinions.

jodosha:
@ericfreese 
> It doesn't seem realistic to expect warnings to be fixed in every gem.

It's true.

---

We're receiving a lot of complains from new people. The result is a poor user experience and a perception of a shabby framework.

As previously stated we're doing as much as possible to fix these warnings in our gems, and helping direct (`http_router`, `mail`) and indirect (`haml`) `hanami` dependencies fix their code. But a warning-free ecosystem is really hard to achieve in the immediate. Ruby is such a liberal language that let a project to run well despite these hiccups.

---

My vote is to keep warnings while we develop Hanami, but to suppress them for final projects. This PR should be expanded with RSpec support and then merged.

RSpec's `spec_helper.rb` has a `config.warnings = true` that should be set to `false`. See https://github.com/hanami/hanami/blob/9d41d06d90c7ccba720e37f7b9bfe3cbd544617c/lib/hanami/generators/application/app/spec_helper.rb.rspec.tt#L75

@ericfreese Can you please take care of the rest? Thanks

ericfreese:
Updated to address RSpec as well.

ericfreese:
Actually, I'm not quite sure what's going on here. It looks like that section you linked is commented out.

Did I make the change you were intending?

ericfreese:
Running a couple tests, it seems RSpec does not enable warnings by default, so no code change is necessary to ignore warnings in RSpec tests.

@jodosha The line of code referenced is inside a comment block of RSpec settings "suggested to provide a good initial experience". IMO, we could keep that line as is, recommending that people enable warnings in their project, but having warnings disabled by default.

jodosha:
> Actually, I'm not quite sure what's going on here. It looks like that section you linked is commented out.

Yes, it's commented by default. That is a file generated by `rspec --init`. Almost all the default `spec_helper.rb` is commented out.

However, if a developer remove the comments, they will find the `config.warning` set to `false`. In that case they still won't see the warnings. Only if they take an **explicit** action to turn on that setting with `true`.. well, they will see the warnings.

This is equivalent to Minitest's Rake task: if a developer will remove `t.warning = false` or set it to `true`, they know what they're doing.

jodosha:
@ericfreese Thanks for this PR, merging. 👍

PR #588, opened 2016-05-30
Make simplecov return the correct coverage

beatrichartz:
Currently the coverage for this repo is reported as 44% - good news, Hanami actually has 93% coverage 🎉

### DIFF
--- a/test/support/coverage.rb
+++ b/test/support/coverage.rb
@@ -8,6 +8,12 @@ module Coverage
     configure_simplecov!
   end
 
+  module_function def cover_as!(suite_name)
+    return unless enabled?
+
+    SimpleCov.command_name(suite_name)
+  end
+
   private
 
   module_function def travis?
--- a/test/support/coverage.rb
+++ b/test/support/coverage.rb
@@ -1,0 +1,43 @@
+module Coverage
+  module_function def configure!
+    return unless enabled?
+
+    require 'simplecov'
+    require 'coveralls'
+
+    configure_simplecov!
+  end
+
+  private
+
+  module_function def travis?
+    !!ENV['TRAVIS']
+  end
+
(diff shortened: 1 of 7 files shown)


test/support/coverage.rb
```diff
@@ -0,0 +1,43 @@
+module Coverage
```
davydovanton: What do you think, may be is better to use class instead module? :)

davydovanton:
good job, I liked it 👍

test/support/coverage.rb
beatrichartz: I quite like the `module` - `module_function` approach - I can change it to a class with class functions if that suits the style of the repo better, do you want me to?

davydovanton: no, it was only question. For me class is more clear than `module_function`, that's why I asked.

beatrichartz: I don't mind changing it either way :) Just let me know. Thanks for taking the time to review this!

jodosha:
@beatrichartz Wonderful! I was looking at this problem today. Probably the coverage is even higher, but there are some tests that aren't counted because ran in a separated process. See `script/ci`.

Do you think it's possible to merge the coverage for multiple processes? See https://github.com/colszowka/simplecov/issues/147#issuecomment-6603303

beatrichartz:
@jodosha good catch, added the integration tests in, simplecov merges them automatically. Also made jRuby run in debug mode so the correct coverage can be provided from the Travis build.

test/support/coverage.rb
joneslee85: TIL `module_function`, IMO I quite like it

jodosha:
@beatrichartz Thank you! Just merged this PR! 🎉

PR #590, opened 2016-06-01
Remove require 'rubygems'

sakuro:
Hanami supports MRI 2+ and the gem system is built-in since 1.9.
Do we still need loading it explicitly?

All tests are passing without the lines (except `S`s)


jodosha:
@sakuro Thank you. Merging this. 👍

hungcoi6699:
hello Toàn Cầu

PR #599, opened 2016-06-24
Allow views to render any HTTP status code.

### LINKED ISSUE #509: Modify rendering policy to permit template rendering for HTTP status other than 200 & 201
# Actual

The current rendering policy will only permit the template associated with an action to be rendered if the action is deemed successful:

```
#hanami/lib/hanami/rendering_policy.rb
SUCCESSFUL_STATUSES = (200..201).freeze
    
def _render_action(action, response)
  if successful?(response)
    view_for(action, response).render(action.exposures)
  end
end

def successful?(response)
  SUCCESSFUL_STATUSES.include?(response[STATUS])
end
```

When returning an HTTP status other than 200 or 201, a [custom error page](http://hanamirb.org/guides/views/custom-error-pages/) is used in its place, but is specific to the app and not the current action. These templates represent more of an unrecoverable error situation, similar to an exception. Thus the custom error pages can be likened to default exception handlers.

# Proposed

Allow the template associated with the action to be rendered and respond with the HTTP status code set by the action. For example:

```
module Web::Controllers::Session
  class Create
    include Hanami::Action

  params do
    param :email, presence: true
    param :password, presence: true
  end

  def call(params)
    if params.invalid?
      self.status = 400 # or 422
    elsif self.current_user = AuthenticateUser.call(params) 
      redirect_to routes.user_path(id: current_user.id)
    else
      self.status = 404
    end
  end
end
```

In the above example, if the parameters are invalid, or the user cannot be authenticated, the action can simply render the login form, but the HTTP status will be respected and returned to the client browser or application.

When the developer wants to indicate an exception condition, this can be achieved through the use of `halt`:

```
module Web::Controllers::Users
  class Show
    include Hanami::Action

  params do
    param :id, presence: true
  end

  def call(params)
    if params.invalid?
      halt 400
    elsif @user = FindUser.call(params) 
      # do nothing; the associated template will be rendered
    else
      halt 404
    end
  end
end
```

Thus using `halt` in this way is similar to `raise`, i.e. an exception has occurred. In this case, the appropriate custom error page will be rendered.

jodosha:
### I want to show a page for a book.

Be sure to have an action, a view and a template. This is the default use case.

### I want to show a generic, application wide 404 page for a not found book.

If the book is not found, from the action `halt(404)`.

### I want to show a custom, application wide 404 page for a not found book.

Create a `apps/web/templates/404.html.erb` template and from the action: `halt(404)`.

### I want to show a custom 404 page **only** for not found books.

From the action `self.status = 404`, and then manage the `nil` book from the view/template.

```erb
# apps/web/templates/books/show.html.erb
<% if book.nil? %>
  <h1>Hey, I can't find the book!</h1>
<% else %>
  <!-- ... -->
<% end %>
```

### I want exceptions to be managed by a generic, application wide 500 page.

This is already handled by Hanami.

### I want exceptions to be managed by a custom, application wide 500 page.

Create a `apps/web/templates/500.html.erb` template.

### I want exceptions to be managed by a specific page for a specific action

Make sure to handle exceptions in that action, with an exception handler. Then just assign the status.

```ruby
module Web::Controllers::Books
  class Create
    include Web::Action
    handle_exception FooError => :handle_foo_error

     def call(params)
    raise FooError.new
  end

  private

  def handle_foo_error(exception)
    self.status = 500
  end
end
```

Then handle the exception in the corresponding view/template.

```erb
# apps/web/templates/books/create.html.erb
<h1>Oh no!</h1>
```

### I want to return a 422 when validations fails

Ensure to set the status `self.status = 422`, and in the corresponding view indicate the template to render.

```ruby
module Web::Controllers::Books
  class Create
    include Web::Action

   def call(params)
      if params.valid?
      else
        self.status = 422
      end
    end
  end
end
```

```ruby
module Web::Views::Books
  class Create
    include Web::View
    template 'books/new'
  end
end
```

Please note that `params` will be used to fill form fields.

---

The bottom line is: use `#halt` to terminate the control flow and render immediately a generic page. For more control use `#status=`.

---

Closes #509 

@hanami/core-team @hanami/contributors Please review.

### DIFF
--- a/lib/hanami/rendering_policy.rb
+++ b/lib/hanami/rendering_policy.rb
@@ -15,7 +15,7 @@ module Hanami
     HANAMI_ACTION = 'hanami.action'.freeze
 
     SUCCESSFUL_STATUSES = (200..201).freeze
-    RENDERABLE_FORMATS  = [:all, :html].freeze
+    RENDERABLE_FORMATS = [:all, :html].freeze
 
     def initialize(configuration)
       @controller_pattern = %r{#{ configuration.controller_pattern.gsub(/\%\{(controller|action)\}/) { "(?<#{ $1 }>(.*))" } }}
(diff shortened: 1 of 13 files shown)


davydovanton:
Looks cool :+1:
What about updating documentation?

fidothe:
Looks pretty good at first glance, though it's not 100% obvious to me that my middleware usecase would work. I'll try this PR out with our code and see...

jodosha:
@fidothe Now Hanami doesn't try to override the body. I have to merge this to make progress on an project of mine.

PR #604, opened 2016-07-01
Propose an alternate default CSP policy

oreoshake:
First off, I want to congratulate y'all on applying a default CSP. The concept of CSP as a default has amazing returns. I've seen orgs gradually grasp content security policy while applying it over time simply because it was a default and opting out required thought (and approval).

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr"><a href="https://twitter.com/hanamirb">@hanamirb</a> respectfully I think an unrealistic policy will cause people to disable csp. I&#39;ll submit a pr for discussion.</p>&mdash; neil matatall (@ndm) <a href="https://twitter.com/ndm/status/748798325931913217">July 1, 2016</a></blockquote>

This is a follow up to https://twitter.com/ndm/status/748798325931913217 and is related to https://github.com/rails/rails/pull/24961 where I proposed a default CSP, but the idea was rejected due to lack of a CSP management API. The proposed policies are not the same and there is some room for debate. A lot of the reasoning comes from [GitHub's CSP Journey](http://githubengineering.com/githubs-csp-journey/).

With the exception of `object-src` becoming `'none'`, this policy should be a non-breaking change for applications. Considering it's in a generator, there's less risk of breaking things.

Reasonings:
- Changing `object-src` to `'none'` - `object-src` is very dangerous. Using a value of `'self'` opens up a lot of cross-origin bypasses mostly via flash or pdf bugs.
- `img-src` gets `https data:` added - while sourcing 3rd party images can lead to defacement if the underlying tags reference undesirable content, the goal here is to eliminated mixed content. `data:` is safe as a default.
- `style-src` gets `'unsafe-inline' https:'` - while "scriptless" attacks exist, restricting inline style is not a realistic goal. So much so that a 3rd `'unsafe'` value had been discussed but it was too complex and nobody cared enough to make it happen. Inline styles are safe enough. Similarly, almost any style is safe from a security point of view so let's just prevent mixed content here too.
- Removing `referrer` and `reflected-xss` - these are non-standard at this point. I don't think these are well supported either. Glad to add these back.

More proposed changes:
- (breaking change) default `script-src` to be `https:` - using `'self'` opens you up to jsonp attacks. In a lot of cases, people bundle assets on a CDN. Using `'self'` is mostly for convenience when developing locally. While closing the jsonp loophole is nice, maybe allowing any `https` worse. Using Subresource integrity with CSP alleviates the risk of using the broad `https:` value.
- Adding `https: data;` to font-src - this is _mostly_ safe.

### DIFF
--- a/lib/hanami/generators/app/application.rb.tt
+++ b/lib/hanami/generators/app/application.rb.tt
@@ -214,7 +214,7 @@ module <%= config[:classified_app_name] %>
         img-src 'self' https: data:;
         style-src 'self' 'unsafe-inline' https:;
         font-src 'self';
-        object-src 'none'; 
+        object-src 'none';
         plugin-types application/pdf;
         child-src 'self';
         frame-src 'self';
--- a/lib/hanami/generators/app/application.rb.tt
+++ b/lib/hanami/generators/app/application.rb.tt
@@ -206,17 +206,15 @@ module <%= config[:classified_app_name] %>
       #
       security.content_security_policy %{
         form-action 'self';
-        referrer origin-when-cross-origin;
-        reflected-xss block;
         frame-ancestors 'self';
         base-uri 'self';
         default-src 'none';
         script-src 'self';
         connect-src 'self';
-        img-src 'self';
-        style-src 'self';
+        img-src 'self' https: data:;
+        style-src 'self' 'unsafe-inline' https:;
         font-src 'self';
-        object-src 'self';
+        object-src 'none'; 
         plugin-types application/pdf;
         child-src 'self';
         frame-src 'self';
--- a/test/fixtures/commands/generate/app/application.rb
+++ b/test/fixtures/commands/generate/app/application.rb
@@ -206,17 +206,15 @@ module Admin
       #
       security.content_security_policy %{
         form-action 'self';
-        referrer origin-when-cross-origin;
-        reflected-xss block;
         frame-ancestors 'self';
         base-uri 'self';
         default-src 'none';
         script-src 'self';
         connect-src 'self';
-        img-src 'self';
-        style-src 'self';
+        img-src 'self https: data:';
+        style-src 'self 'unsafe-inline' https:;
         font-src 'self';
-        object-src 'self';
+        object-src 'none';
         plugin-types application/pdf;
         child-src 'self';
         frame-src 'self';
--- a/test/fixtures/commands/generate/app/application.rb
+++ b/test/fixtures/commands/generate/app/application.rb
@@ -211,8 +211,8 @@ module Admin
         default-src 'none';
         script-src 'self';
         connect-src 'self';
-        img-src 'self https: data:';
-        style-src 'self 'unsafe-inline' https:;
+        img-src 'self' https: data:;
+        style-src 'self' 'unsafe-inline' https:;
         font-src 'self';
         object-src 'none';
         plugin-types application/pdf;


jodosha:
@oreoshake Thanks for taking the time to study our CSP default policy and to write this detailed PR. We truly appreciate it.

I read all that GH blog post, which is interesting. I've got the sense of how complex is security and how many devs just ignore these topics. Even with an imperfect CSP, we're putting obstacles on the way of malicious users.

I agree with your proposed changes, but I'm not sure about `data:` for `img-src`. That could be source of attacks. My suggestion is to merge all the changes but this one. See https://developer.mozilla.org/en/docs/Web/Security/CSP/CSP_policy_directives (Data section), http://www.paladion.net/bypass-xss-filters-using-data-uris, http://cubalo.github.io/blog/2014/01/04/bypassing-xss-filters-using-data-uris/

@joneslee85 Thoughts?

oreoshake:
> Note: data: URIs are unsafe and can open your web site up to cross-site scripting vulnerabilities if allowed for script sources.

This is absolutely true, for **script sources** and **object sources** but not for image sources. `img-src data:` is safe.

joneslee85:
@jodosha @oreoshake the data directive seems to be only an issue of used for script; thus to me it is safe for img tag. I am good with the proposal on doing for don't tag too.

Regarding the proposal to set to https for script, I like the idea but it should be the setting only for production mode imo.

This PR looks good to me. However due to its security nature, I would like to invite another round of review from @jacobbednarz

jacobbednarz:
Agree with @oreoshake that the default (very strict) CSP is more likely to have people disable it then debug it. CSP can be a pain for those who don't understand it and just want to Get Things Done™.

> Using Subresource integrity with CSP alleviates the risk of using the broad https: value.

@jodosha @joneslee85 Does hanami have a SRI helper for assets? This might be something worth chucking in your backlog if it isn't since it is critical to locking down your known "safe" assets without it you are open to attack vectors like compromised CDN's.

> Removing referrer and reflected-xss - these are non-standard at this point. I don't think these are well supported either. Glad to add these back.

Yep - at this stage I don't think this has fully been implemented and last time I tried this I was still getting a lot of inconsistencies between browsers. 👍  to this decision. 

> `img-src data:` is safe.

I am still on the fence about this one because I'm not sure if you also need to have mime type sniffing disabled to achieve it safely. Any thoughts there @oreoshake?

> Regarding the proposal to set to https for script, I like the idea but it should be the setting only for production mode imo.

Only enabling HTTPS in production can cause unexpected side effects due to the way the applications can behave differently in the two scenarios. If you're using HTTPS in production, I'd be recommending it in all your environments to catch regressions early. 

One thing that I learnt the hard way is that a CSP should really only be used for HTTPS enabled applications. Trying to rollout a CSP on a HTTP only application you end up hitting weirdness with things like Safari not liking cross protocol requests and throwing up a violation. At a minimum, it is probably worth documenting this for the users to avoid the headaches encountered.

Huge thanks @oreoshake for putting this proposal together!

jodosha:
@jacobbednarz Thanks for joining this conversation. I really appreciate it. 😄 

> Agree with @oreoshake that the default (very strict) CSP is more likely to have people disable it then debug it. CSP can be a pain for those who don't understand it and just want to Get Things Done™.

Isn't this true for all security mechanisms? There is always a dichotomy between security vs get things done.

The general principle behind our CSP settings is to start strict and let developers to whitelist their exceptions.

Again, I'm in favor of the changes that @oreoshake proposed here. I just want to make clear that shipping a relaxed CSP and then expect developers to make it stricter is more unrealistic. We all know how hard is to make CSP stricter for apps that are already running in production. That's why Hanami ships with this feature in pre 1.0.

---

> @jodosha @joneslee85 Does hanami have a SRI helper for assets?

Yes, we'll ship SRI with the next version (v0.8), together with this proposed CSP changes.

---

> I am still on the fence about this one because I'm not sure if you also need to have mime type sniffing disabled to achieve it safely.

Yup, I think we should include `X-Content-Type-Options: nosniff` as default response header.

---

> At a minimum, it is probably worth documenting this for the users to avoid the headaches encountered.

Very good suggestion. 👍 

---

> Huge thanks @oreoshake for putting this proposal together!

Yes, thank you again.

oreoshake:
> Yes, we'll ship SRI with the next version (v0.8), together with this proposed CSP changes.

Y'all rock!

> Yup, I think we should include X-Content-Type-Options: nosniff as default response header.

Definitely. It's the safest header that closes a huge whole.

>> img-src data: is safe.

> I am still on the fence about this one because I'm not sure if you also need to have mime type sniffing disabled to achieve it safely. Any thoughts there @oreoshake?

I'm not sure if mime sniffing comes into play here (and I wouldn't be surprised if it did) but there's no reason to allow mime sniffing anyways. 

Not that this is a good argument, but nearly all CSP deployments have `img-src data:` and as far as I know, there has never been an attack of this form. I would think any attack would have come in as a bounty report.

jacobbednarz:
> Isn't this true for all security mechanisms? There is always a dichotomy between security vs get things done.

It is more of a pain with a CSP since browsers differ in support and unless you're keeping an eye on it, can quickly create issues for your end users (I'm looking at you safari). Even if you have a 100% bulletproof CSP, safari often has it's own mind when it comes to interpreting these policies. 

> Yes, we'll ship SRI with the next version (v0.8), together with this proposed CSP changes.

So. So. Cool. Major kudos to you folks!

> Yup, I think we should include X-Content-Type-Options: nosniff as default response header.

Such a quick turn around - very impressive.

> Not that this is a good argument, but nearly all CSP deployments have img-src data: and as far as I know, there has never been an attack of this form. I would think any attack would have come in as a bounty report.

Happy to 👍  this until proven otherwise. I'm in the process of rolling out a CSP to quite a large application so I'll keep an eye on reports for this too.

jodosha:
With `X-Content-Type-Options` in place, we all agree that the proposed changes worth to be merged. Thank you @oreoshake @joneslee85 @jacobbednarz for this constructive discussion. 👍

PR #586, opened 2016-05-28
Set application to nil by default and throw error if arch is a container

maheshm:
Fixes #584

PR #606, opened 2016-07-02
Destroy spec

### LINKED ISSUE #605: Destroy command - more specs required
There are no specs associated with destroy command.

maheshm:
This has to be merged post merging #586 
Fixes #605


jodosha:
@maheshm thanks for these extra tests, they are always welcome!

PR #608, opened 2016-07-11
X-Content-Type-Options

jodosha:
Include `X-Content-Type-Options: nosniff` in default response headers.

Ref #604


joneslee85:
:+1:

PR #609, opened 2016-07-11
Enable Rack middleware mounted in apps to render

### LINKED ISSUE #602: Hanami errors when a custom 404 page has been rendered by a middleware
Following on from the discussion on #509 and #599, I created a minimal app to see what happens with the new non-200 status rendering code with my 404-page-rendered-by-middleware use case.

Unfortunately Hanami now 500's (before it would escape the rendered HTML and render the standard default page with the custom HTML rendered as source, effectively). The minimal app is at https://github.com/fidothe/middleware_404.
## Expected behaviour

Custom 404 page rendered by middleware is displayed when 404 encountered
## Actual behaviour

Hanami returns 500 status code
## Reproduction steps

Run the specs in the demonstrator app, or run `hanami server` and visit any page except `/`
(See https://github.com/fidothe/middleware_404/blob/master/README.md for instructions on reproduction and workaround)

jodosha:
Ref #509 
Ref #599 
Fix #602

### DIFF
--- a/lib/hanami/routing/default.rb
+++ b/lib/hanami/routing/default.rb
@@ -14,6 +14,10 @@ module Hanami
 
         def call(env)
         end
+
+        def renderable?
+          false
+        end
       end
 
       def call(env)
--- a/test/fixtures/middleware_stack/application.rb
+++ b/test/fixtures/middleware_stack/application.rb
@@ -1,6 +1,9 @@
 module MiddlewareStack
   class Application < Hanami::Application
     configure do
+      # Test rendering from middleware
+      middleware.use 'Middlewares::LegacyNotFound'
+
       # Test lazy loading with relative class name
       middleware.use 'Middlewares::Runtime'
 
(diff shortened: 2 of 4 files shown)


joneslee85:
LGTM 👍

cllns:
What does 'Legacy Not Found' mean? I see 'legacy' referred to in `test/` but not in `lib/`.

jodosha:
Ironically, this "fix" breaks other tests and so other behaviors. I'm struggling to find an implementation to make all the cases to work.

jodosha:
@cllns "Legacy Not Found" is just a dumb message test. It's inspired by https://github.com/fidothe/middleware_404, which is a demo app by @fidothe to reproduce #602. In production he has a Rack middleware to return custom 404 page for legacy routes.

In our tests there is nothing special, we're asserting for that message to prove that the rendered response comes from the [test Rack middleware](https://github.com/hanami/hanami/pull/609/files#diff-ace3ac0b55a40ba87cedec211e137b5eR26).

fidothe:
@jodosha yeah, i was having similar problems on my fork, and then I got swamped with work. Hoping to get another few hours to have a crack at it later this week. 

Here's where I got to: https://github.com/fidothe/hanami/pull/1

I was trying to check for non-empty response body before checking for `hanami.action`.

I'm beginning to wonder if there should be a two-step process - explicitly rendering in the action sets a `hanami.view` key in the env, implicit rendering could do that in the wrapping of `Hanami::Action.call`, but only in the case of e.g. a 2XX status code. Then the `Hanami::RenderingPolicy` check could check for non-empty body as an override, followed by a check for `hanami.view`

Needs more thought and poking, sorry to have been tardy on this :-(

jodosha:
After two days working on this fix, I can tell this is isn't a viable solution.

PR #616, opened 2016-07-18
X-XSS-Protection

jodosha:
Include `X-XSS-Protection: 1; mode=block` in default response headers.

### DIFF
--- a/lib/hanami/config/security.rb
+++ b/lib/hanami/config/security.rb
@@ -16,6 +16,12 @@ module Hanami
       # @see Hanami::Loader#_configure_controller_framework!
       X_CONTENT_TYPE_OPTIONS_HEADER = 'X-Content-Type-Options'.freeze
 
+      # @since x.x.x
+      # @api private
+      #
+      # @see Hanami::Loader#_configure_controller_framework!
+      X_XSS_PROTECTION_HEADER = 'X-XSS-Protection'.freeze
+
       # @since 0.3.0
       # @api private
       #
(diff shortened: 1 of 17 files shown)


simi:
`X-` prefix was [deprecated](https://www.rfc-editor.org/info/rfc6648) for HTTP headers. I wonder if it can be possible to use at lease both variants (prefixed and non-prefixed.

jodosha:
@simi That is true, `X` prefixed headers has been deprecated.

But I haven't found any alternative header for this. For instance `X-Content-Security-Policy` is now `Content-Security-Policy`, because that deprecation has been implemented by browsers vendors. But it looks like they didn't the same for `X-XSS-Protection`.

Again, I can't find references to `XSS-Protection`. [OWASP](https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#X-XSS-Protection), [securityheaders.io](https://securityheaders.io/?q=hanamirb.org&followRedirects=on), [`secure_headers`](https://github.com/twitter/secureheaders) gem, are just a few examples that are still using the `X` prefixed version.

Can you please help with this? Thanks.

simi:
After reading whole RFC and some browser docs, there's probably nothing you can switch `X-XSS-Protection` for right now. According to that RFC it looks like there's no pressure to migrate currently used headers and it is only recommended to not introduce new prefixed HTTP headers.

Sorry for false alarm.

jodosha:
@simi No worries, I just wanted to be sure. Thanks for checking the RFC and starting the discussion! 👍

PR #622, opened 2016-07-28
Allow generating an app into current folder, not under it

### LINKED ISSUE #621: generate a new project
with v0.8.0 `hanami new .` command generates project files inside a new folder nested to the current folder. this new folder has the same name of the current folder.
with v0.7.3 `hanami new .` command generates project files regularly inside current folder.

bruz:
Fixes #621.

There was a test for when `.` is supplied as the project name that checked that the app was put into a subfolder, so perhaps there is an argument for the current behavior? But it makes sense to me that it'd generate directly into the current folder.

### DIFF
--- a/lib/hanami/commands/new/abstract.rb
+++ b/lib/hanami/commands/new/abstract.rb
@@ -35,8 +35,8 @@ module Hanami
         end
 
         def start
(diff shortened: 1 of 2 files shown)


jodosha:
@bruz Thank you!

Issue #623, opened 2016-07-31
--help options does not work correctly for some subcommands


igneus:
`hanami new --help`, instead of describing available options, prints

```
ERROR: "hanami new" was called with no arguments
Usage: "hanami new APPLICATION_NAME"
```

This discovery made me try the `--help` option for all other subcommands (at least those listed by `hanami help`). This way I found a bug
- `hanami db --help`
- `hanami destroy --help`
- `hanami generate --help`

All these commands print

```
Usage:
  hanami help [COMMAND]

Describe subcommands or one specific subcommand
```

which is help for `hanami help`, not for the respective subcommands.

**(ADDED LATER)**

Help for sub-subcommands prints misleading usage synopsis: e.g. `hanami db create --help` prints

```
Usage:
  hanami create
```

instead of expected

```
Usage:
  hanami db create
```

jodosha:
Fixed by #624

igneus:
@jodosha It isn't really fixed as a whole. #624 only fixed `hanami new --help`, not the other similar issues mentioned above.

(I looked into these as well, but I wasn't able to solve them in a reasonable amount of time, so I submitted at least the fix for `hanami new --help`, which was easy enough to produce.)

jodosha:
@igneus Sorry, you're right. Do you think you will have the time to work on this? Thank you very much!

igneus:
I hope so, but I don't know when exactly. Anyone knowing how to fix it or willing to dig into it is most welcome to take it over.

akhramov:
Regarding
> Help for sub-subcommands prints misleading usage synopsis

Other Thor-based projects experience the same issue as well.
calabash/run_loop#226

As a workaround we could override [Thor.banner](https://github.com/erikhuda/thor/blob/v0.19.1/lib/thor.rb#L367-L369) method:

```ruby
module Hanami
  module CliBase
    def banner(command, *)
      if namespace == 'hanami:cli'
        super
      else
        super(command, namespace, true)
      end
    end
  # ...
  end
end
```

It should fix the Usage banner:
```fish
% bundle exec bin/hanami db create --help

Usage:
  hanami db create
  ...

% bundle exec  bin/hanami new  --help
Usage:
  hanami new APPLICATION_NAME
  ...
```

beauby:
Possibly related: `$ hanami help generate model` gives the same message as `$ hanami help generate`.

jodosha:
[...]

PR #624, opened 2016-07-31
`hanami new --help` does not fail

igneus:
Here is a partial fix for #623

### DIFF
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -105,10 +105,10 @@ module Hanami
     method_option :hanami_head, desc: 'Use hanami HEAD (true/false)', type: :boolean, default: false
     method_option :help, desc: 'Displays the usage method'
     def new(application_name=nil)
-      if application_name.nil? && !options[:help]
-        raise InvocationError.new("\"hanami new\" was called with no arguments\nUsage: \"hanami new APPLICATION_NAME\"")
-      elsif options[:help]
+      if options[:help]
         invoke :help, ['new']
+      elsif application_name.nil?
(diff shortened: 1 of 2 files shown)


lib/hanami/cli.rb
```diff
@@ -104,8 +104,10 @@ def console
     method_option :test, desc: "Project test framework (#{Hanami::Generators::TestFramework::VALID_FRAMEWORKS.join('/')})", default: Hanami::Hanamirc::DEFAULT_TEST_SUITE
     method_option :hanami_head, desc: 'Use hanami HEAD (true/false)', type: :boolean, default: false
     method_option :help, desc: 'Displays the usage method'
-    def new(application_name)
-      if options[:help]
+    def new(application_name=nil)
+      if application_name.nil? && !options[:help]
+        raise InvocationError.new("\"hanami new\" was called with no arguments\nUsage: \"hanami new APPLICATION_NAME\"")
+      elsif options[:help]
```
tejasbubane: It would be better to move the second condition branch `if options[:help]` as first, that way you don't have to check for `!options[:help]` in the first branch.

tejasbubane:
You should also add some test for this.

igneus:
> @tejasbubane: You should also add some test for this.
I would love to. I actually looked into the tests even before I wrote the fix, but I haven't found anything like CLI integration tests and don't feel like creating the infrastructure for them in a project I am not yet familiar enough with. Advice highly appreciated.

tejasbubane:
@igneus You can add the tests here: https://github.com/hanami/hanami/blob/master/test/cli_test.rb

jodosha:
@igneus I manually tested it: it works. Thanks!

Issue #625, opened 2016-07-31
'hanami secret' command


Phitherek:
I would like to see a 'hanami secret' command - it would help with setting up session secrets in production environment. My proposal for this command is simple: just invoke SecureRandom.hex(32) as in generate app command and return the result to console.

jodosha:
This is a convenient alternative to:

```shell
➜ ruby -rsecurerandom -e "puts SecureRandom.hex(32)"
b240729645bf9e54a85136f3c9524cac13c2862e8b84ae3cbb9942db52ff343b
```

uaihebert:
Hey, I have created a PR with this implementation. I believe that is an useful and easy to create feature. ( :

toch:
what about my PR previously submitted #627 ?

uaihebert:
Ops, Did not notice your PR. Sorry.

toch:
@uaihebert no worries. Maybe there is good thing we can merge together?

uaihebert:
Mine approach is different from yours. I will close my pr. ( :

jodosha:
Implemented by #627

PR #627, opened 2016-08-04
Generate command for secret token in production

toch:
As I wanted something simple to contribute for a first time, I decided to implement a generate subcommand for the issue #625. It appends a secret token to the `.env` file for a given app.

As #625 was not already acknowledged as something the hanami team would like, I understand that contribution could be rejected. At least, it helped me to dig more into the hanami code :).

Considering it's something you'd like, what do you think of the changes? Have I missed anything?

I have a side question: is it right that the documentation/guide is to submit to https://github.com/hanami/hanami.github.io/?

### DIFF
--- a/lib/hanami/cli_sub_commands/generate.rb
+++ b/lib/hanami/cli_sub_commands/generate.rb
@@ -123,6 +123,22 @@ module Hanami
           Hanami::Commands::Generate::App.new(options, application_name).start
         end
       end
+
+      desc 'secret APPLICATION_NAME', 'Print a fresh secret token for production'
+      long_desc <<-EOS
+        `hanami generate secret` prints a new secret for a given app
+
+        > $ hanami generate secret admin
+
+      EOS
+      def secret(application_name = nil)
+        if options[:help]
+          invoke :help, ['secret']
+        else
+          require 'hanami/commands/generate/secret_token'
+          Hanami::Commands::Generate::SecretToken.new(application_name).start
+        end
+      end
     end
   end
 end
--- a/lib/hanami/commands/generate/secret_token.rb
+++ b/lib/hanami/commands/generate/secret_token.rb
@@ -1,0 +1,31 @@
+require 'hanami/commands/generate/abstract'
+require 'hanami/application_name'
+require 'securerandom'
+
+module Hanami
+  module Commands
+    class Generate
+      class SecretToken
+
+        def initialize(application_name)
+          @application_name = application_name
+        end
+
+        def start
+          if Hanami::Utils::Blank.blank?(@application_name)
+            puts SecureRandom.hex(32)
+          else
+            puts "Set the following environment variable to provide the secret token:"
+            puts %(#{ upcase_app_name }_SESSIONS_SECRET="#{ SecureRandom.hex(32) }")
+          end
+        end
+
+        private
+        def upcase_app_name
+          @application_name.upcase
+        end
+
+      end
+    end
+  end
+end
--- a/test/cli_test.rb
+++ b/test/cli_test.rb
@@ -4,6 +4,7 @@ require 'hanami/commands/generate/migration'
 require 'hanami/commands/generate/model'
 require 'hanami/commands/generate/action'
 require 'hanami/commands/generate/app'
+require 'hanami/commands/generate/secret_token'
 require 'hanami/commands/new/container'
 require 'hanami/commands/new/app'
 
@@ -250,6 +251,18 @@ describe Hanami::Cli do
         options = {'application_base_url' => '/backend'}
 
         assert_cli_calls_command(Hanami::Commands::Generate::App, options, 'admin')
+      end
+    end
+
+    describe 'secret' do
+      it 'calls the generator with app name' do
+        ARGV.replace(%w(generate secret admin))
+        assert_cli_calls_command(Hanami::Commands::Generate::SecretToken, 'admin')
+      end
+
+      it 'calls the generator with nothing' do
+        ARGV.replace(%w(generate secret))
+        assert_cli_calls_command(Hanami::Commands::Generate::SecretToken, nil)
       end
     end
   end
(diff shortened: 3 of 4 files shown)


bruz:
This is some great work!

I'm wondering though if `rake secret` should write the secret to stdout rather than putting it into a file? `dotenv` is only loaded in the development and test environments by default, which I've assumed is because Hanami is following the [12 factor app](http://12factor.net/config) methodology which advocates not having production configuration written to files.

The way I was imagining `rake secret` being used is to, either manually or through automation, generate a secret and set it as an environment variable in the production environment.

@Phitherek since you filed the issue originally, what was your expectation for this?

Phitherek:
@bruz I've described it in the original issue - I wanted it to be basically 'hanami secret' which would be a basic command that writes the secret to stdout, as for example I don't use .env at all in production but rather define the variables in the server app configuration there. I looked at the code quickly and I think it could be usable to someone, but in my opinion it is better to leave it to the user where to declare the secret in production. And even copy-pasting the secret from stdout to .env or doing it with the quick command is not a huge hassle for server admin in my opinion.

toch:
if I understand you well, it's better to make it a Rake task and to output the secret on stdout. right?

If so, I'll change it in that way. I have also to fix that failing test.

Phitherek:
@toch Yes, but to be more specific it is not a Rake task but a Hanami CLI command ;)

tejasbubane:
Here is how rails does it: https://github.com/rails/rails/blob/master/railties/lib/rails/tasks/misc.rake#L2

jodosha:
@toch thanks for taking care of this, but this solution has security / devops implications and so it can't be merged.

First of all, sensitive data should be stored in ENV vars. In the latest versions of Hanami we don't generate `.env` for new projects, to encourage to use ENV vars in production.

For deployments in readonly file systems (Heroku), this would command would fail, unless it's run before the production environment has started. So you're unable to reset the session token while the application is running.

---

As @Phitherek mentioned, it's useful to have a CLI command to print a random string to stdout. I always generated it via `ruby -rsecurerandom -e "puts SecureRandom.hex(32)"`, but I have nothing against a `hanami secret` command.

toch:
Thanks @jodosha, @Phitherek  for the feedback

Here a new version that just prints the secret as following:

```
$ hanami generate secret admin
Set the following environment variable to provide the secret token:
ADMIN_SESSIONS_SECRET="ba6967c1864c4894cb04b7ec672553dc6df12a6c5cb4775742566bf7b90f7dce"
```

As it's just about printing the secret token, I also removed the dependency on Abstract.

is it better?

Phitherek:
@toch The only thing I would change is to make application name optional - without it the command would print out just the secret string without the information about the environment variable. I like the version with the app name though as it can be easier to set up environment variables in some cases and I also like that it does not enforce any kind of configuration model.

toch:
I've just push new changes where the application name is optional. It behaves as @Phitherek suggested.

Phitherek:
@toch This is great! I approve of this pull request and consider it to resolve #625. I hope to see it merged soon!

joneslee85:
@toch thanks for reply, I find the CLI interface is a bit confusing as it is geared toward generation of session secret. I think the class name should changed to reflect that intent.

toch:
@joneslee85 thanks for the feedback. I'm not sure to follow. Which class should be renamed?

toch:
ping

lib/hanami/commands/generate/secret.rb
```diff
@@ -0,0 +1,31 @@
+require 'hanami/commands/generate/abstract'
+require 'hanami/application_name'
+require 'securerandom'
+
+module Hanami
+  module Commands
+    class Generate
+      class Secret
```
b264: `class SessionSecret` maybe?

duykhoa: Or `SecretToken`? 🐔

lib/hanami/commands/generate/secret.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        def initialize(application_name)
+          @application_name = application_name
+        end
+
+        def start
```
duykhoa: Just my personal opinion, the `start` method isn't quite relevant for this case, I think `execute!` is a better one.

toch:
Thanks @b264 and @duykhoa for the suggestions. I've just pushed a renaming.

@joneslee85 is it better? was it what you meant?

toch:
ping?

mereghost:
Seems ok to me. 👍 

@toch I think most of the effort is going into smoothing out the new version of hanami-model (featuring ROM) into the framework. After the said integration they should get back to all the PRs. =)

cllns:
👋 @toch You bought me and other people tacos last year at Rocky Mountain Ruby :) Thanks again for that! 🌮 🍹 

This looks good to me. 

What are the use cases for this command? Is it pretty much only for session secrets, or is it for other things too?

If it's only session secrets, I think we should be more explicit about *how* to set it, like: using `.env` in development and test and other means in production. Rather than including that info in the error message, I wonder if the right solution would be adding a link to a new guide for Environment Variables? We could mention this and also `SERVE_STATIC_ASSETS`.

If we want to do that, we could link to it *both* when users provide an app_name and when they don't?

Thoughts?

toch:
@mereghost and @cllns thanks for your feedback :)

> What are the use cases for this command? Is it pretty much only for session secrets, or is it for other things too?

The use case is described in #625 

> If it's only session secrets, I think we should be more explicit about how to set it, like: using .env in development and test and other means in production. Rather than including that info in the error message, I wonder if the right solution would be adding a link to a new guide for Environment Variables? We could mention this and also SERVE_STATIC_ASSETS.

Agreed. It's my intention to add an entry in the guide when it's done.

> If we want to do that, we could link to it both when users provide an app_name and when they don't?
>
>Thoughts?

I Like the idea to link the guide directly. I'll start to write the entry.

jodosha:
@toch Thank you very much for this feature. I love the explanatory line when app name is provided.
I'm gonna merge it straight away. 👍

toch:
@jodosha Cool :grin: I'll submit an entry in the guide tomorrow.

PR #628, opened 2016-08-04
Skip preloading when code reloading is enabled

bruz:
This issue was originally reported by @opan in chat, but I was able to confirm it using an [example app](https://github.com/bruz/bookshelf-delivery-example), with code changes in either the controller or view not reflecting on refresh when running in development mode.

The problem appears to have started as of https://github.com/hanami/hanami/commit/027cd29f58845933e65ac4dfab0d18ed0a384cb9, due to running `Hanami::Application#preload!`. I'm not sure if this proposed fix is the best one possible or if code reloading should be fixed at the per-framework level. Also, is there any way to test that code reloading is happening?

### DIFF
--- a/lib/hanami/commands/server.rb
+++ b/lib/hanami/commands/server.rb
@@ -78,7 +78,7 @@ module Hanami
       # @see Hanami::Environment::CODE_RELOADING
       def detect_strategy!
         @strategy = :rackup
-        if Hanami::Environment.new(@options).code_reloading?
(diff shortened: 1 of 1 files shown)


jodosha:
@bruz Thank you for this fix. I can confirm both that that the problem is there and that the PR solves it.

> Also, is there any way to test that code reloading is happening?

Not at this time.

---

Merging it! 👍

PR #630, opened 2016-08-06
Update commentary for example with file system adapter

max-si-m:
Use app name in example, instead of default `bookshelf_development`


jodosha:
@max-si-m I like it, thanks!

Issue #631, opened 2016-08-09
const_defined? should be scoped in loader


pascalbetz:
Otherwise classes like ```Web::Routes``` will not get defined if a const called ```Routes``` already exists.

Same goes for ```Logger```, ```View```, ```Controller```, ```Assets```, ```Model```.

Reproduce the problem:

* hanami new sample
* create a (empty) class ```Routes``` and make sure it is required before ```Loader```
* bundle exec hanami console
* Web::Routes #> will lead to an exception
```
NameError: uninitialized constant Web::Routes
Did you mean?  Routes 
```

This is because ```const_defined?``` automatically checks the parent namespace for existence of the constant if constant was not found incurrent namespace. 

So lines like ```unless application_module.const_defined?('Routes')``` should actually be 
```unless application_module.const_defined?('Routes', false)````
(Example from https://github.com/hanami/hanami/blob/master/lib/hanami/loader.rb#L219)

See http://ruby-doc.org/core-2.2.0/Module.html#method-i-const_defined-3F for details on ```const_defined?```

jodosha:
@pascalbetz Great catch! Would you mind to create a PR to fix this issue? Thanks!

pascalbetz:
@jodosha here you go:-)

PR #633, opened 2016-08-09
Scope constant check to application module

pascalbetz:
Fixes https://github.com/hanami/hanami/issues/631#issuecomment-238513397

All specs pass. Additionally tested manually.

### DIFF
--- a/lib/hanami/loader.rb
+++ b/lib/hanami/loader.rb
@@ -51,7 +51,7 @@ module Hanami
 
     def _configure_controller_framework!
       config = configuration
-      unless namespace.const_defined?('Controller')
+      unless namespace.const_defined?('Controller', false)
         controller = Hanami::Controller.duplicate(namespace) do
           handle_exceptions config.handle_exceptions
           default_request_format config.default_request_format
@@ -88,7 +88,7 @@ module Hanami
 
     def _configure_view_framework!
       config = configuration
-      unless namespace.const_defined?('View')
+      unless namespace.const_defined?('View', false)
         view = Hanami::View.duplicate(namespace) do
           root   config.templates
           layout config.layout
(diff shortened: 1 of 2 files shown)


jodosha:
@pascalbetz Thanks for this PR! Can you please add a constant in one of the fixtures applications to prove that this proposed change, actually fixes the name clashing problem? Thanks!

pascalbetz:
Good catch with the tests. The mere presence of a class ```Controller``` breaks hanami in some weird way. Investigating.

pascalbetz:
I think one of the problems is https://github.com/hanami/utils/issues/151

pascalbetz:
@jodosha 
What is the expected behavior when a ```Namespace::View``` is defined _but_ it is not a ```Hanami::View``` ?

For namespaced apps I think it's OK to fail. But if the user modifies ```view_pattern``` to be rooted in the top level namespace (```view_pattern       '%{controller}::%{action}'```) then its quite possible that a ```Object::View``` already exists.

WDYT?

jodosha:
@pascalbetz I think it's safe to ignore "namespaced" apps. That is an legacy mechanism that I put in place in the early days of the project. The goal at that time was to make possible arbitrary project structure via that configuration.

Fast forward to today, there is no value to keep that feature, because nobody (myself included) cares about it. The idea is to cleanup it after 1.0.

pascalbetz:
@jodosha 

Glad to hear! I immediately disliked the feature because I don't think a user should be able to change such an important concept.
Don't you think it should be removed before 1.0 because it is a breaking change?

About this PR: I'll make sure that the classes for the "conflict" test are not loaded in said scenario and update the PR.

jodosha:
> Don't you think it should be removed before 1.0 because it is a breaking change?

I don't think anyone is using it.

pascalbetz:
@jodosha I've added a test. Please review.

jodosha:
@pascalbetz This looks good to me. Thanks! Once the CI will be green, I'll merge it. 😉

pascalbetz:
CI better be green because "it works on my machine"©️  
😄

PR #641, opened 2016-08-21
Don't duplicate middlewares in the stack

### LINKED ISSUE #638: Duplicate middlewares in the stack
Hello!

I have setup Hanami with [omniauth-google-oauth2](https://github.com/zquestz/omniauth-google-oauth2) and experienced weird behavior of callback_request, it was called twice according to console.

After some investigation, I found, that Hanami pushes each middleware [to the `@stack` variable](https://github.com/hanami/hanami/blob/master/lib/hanami/middleware.rb#L66) twice. First time when the server starts, and for second time at request handling. Thus, we have duplicate middlewares in the stack.

Here is naive fix for this issue:

``` ruby
def use(middleware, *args, &blk)
  @stack.push [middleware, args, blk] unless @stack.find { |m| m.first == middleware  }
end
```

This fix has helped me with omniauth.

Is this behavior used for some functionality in Hanami or is this bug?

deepj:
Closes #638.

Middleware stack was changed to `Set` from `Array`. This guarantees uniqueness for free. That avoids to have duplicated middlewares and their other values in the stack.

### DIFF
--- a/lib/hanami/middleware.rb
+++ b/lib/hanami/middleware.rb
@@ -15,7 +15,7 @@ module Hanami
     #
     # @see Hanami::Configuration
     def initialize(configuration)
-      @stack = []
+      @stack = Set.new
       @configuration = configuration
     end
 
@@ -64,7 +64,7 @@ module Hanami
     #
     # @see Hanami::Middleware#prepend
     def use(middleware, *args, &blk)
-      @stack.push [middleware, args, blk]
+      @stack.add [middleware, args, blk]
     end
 
     # Prepend a middleware to the stack.
@@ -79,7 +79,7 @@ module Hanami
     #
     # @see Hanami::Middleware#use
     def prepend(middleware, *args, &blk)
-      @stack.unshift [middleware, args, blk]
+      @stack = Set[[middleware, args, blk]] | @stack
     end
 
     # @api private
--- a/test/middleware_test.rb
+++ b/test/middleware_test.rb
@@ -32,7 +32,7 @@ describe Hanami::Middleware do
 
   it 'appends new added middleware' do
     middleware.use MockMiddleware
-    middleware.stack.last.must_equal [MockMiddleware, [], nil]
+    middleware.stack.to_a.last.must_equal [MockMiddleware, [], nil]
   end
 
   describe "with container architecture" do
@@ -110,8 +110,9 @@ describe Hanami::Middleware do
       end
 
       it 'prepends sessions' do
-        sessions_position   = middleware.stack.index(["Rack::Session::Cookie", [{:domain=>nil, :secure=>false}], nil]).to_i
-        middleware_position = middleware.stack.index([MockMiddlewareClass, [], nil])
+        middleware_stack    = middleware.stack.to_a
+        sessions_position   = middleware_stack.index(["Rack::Session::Cookie", [{:domain=>nil, :secure=>false}], nil]).to_i
+        middleware_position = middleware_stack.index([MockMiddlewareClass, [], nil])
 
         assert sessions_position < middleware_position,
           "Expected sessions middleware to be prepended"


joneslee85:
Thanks for the PR. Can you please provide the test for the duplication in middleware stack edge case?

fidothe:
@joneslee85 @deepj I think there's a deeper underlying problem that this doesn't address – the config block gets executed twice, which is why middleware is being duplicated (see #643).

Sets are also, by definition, unordered, which breaks the idea of an ordered middleware stack. 99% of the time it'd work like you expect, but sometimes it wouldn't (prepending would be particularly vulnerable, I imagine).

deepj:
@fidothe Yes, good point. I started to think about after @joneslee85 comment. I just wanted to make a test (de facto if a stack is `Set`), then I realized things which you mentioned. My basic assumption was wrong here.

deepj:
@fidothe What do you think about using `SortedSet`?

fidothe:
I think `SortedSet` is a problem because the order is controlled by `<=>` of the entries and not the order you inserted them in.

I wonder if maybe the best thing would be to stick with `Array` as the backing structure and `warn` if an identical `[Middleware, args, block]` tries to be added.

The thing I worry about there is that an equality test will always fail on `block` comparison:

```ruby
$ irb
irb(main):001:0> b1 = ->() { true }
=> #<Proc:0x007fb8391a7338@(irb):1 (lambda)>
irb(main):002:0> b2 = ->() { true }
=> #<Proc:0x007fb83919d040@(irb):2 (lambda)>
irb(main):003:0> b1 == b2
=> false
```

To be honest, I'm not convinced mandating uniqueness is worth it. The real problem here is that the config block is being executed twice when we only expect it to be executed once...

deepj:
@fidothe You can't compare procs since Ruby 2.0 https://bugs.ruby-lang.org/issues/4559 or https://github.com/ruby/ruby/blob/f031aec4233d7a6d4622c048abed3e86eb5dd6c2/NEWS#L127 :(

beauby:
Provided [this information](https://github.com/hanami/hanami/issues/643#issue-172488047) is accurate, then this PR should not be merged as it merely fixes the symptoms of an actual bug (that may introduce other unexpected behaviors down the road).

deepj:
@beauby Agree, I wanted to refactor this PR based on the information but due to lack of time I can't make it in near month...

jodosha:
I think `Array` is the right data structure in this case because it's ordered and it can keep unique entries (via `#uniq!`).

---

It isn't accurate to say that we can't compare two procs. We can, not for _value equality_, but _identity equality_.

```ruby
-> { true } == -> { true } # => false, value equality

block1 = -> { true }
block1 == block1 # => true, identity equality
```

In a Hanami application's `configure` block, the procs are always the same objects (See https://github.com/hanami/hanami/commit/4de8fd5ba80f14f38ddccee3ad0a696cef5b3bb3).

---

Because `Proc#==` will return true for identical configuration blocks, we can rely on `Array#uniq!`, which will iterate through all the elements an discard duplicated entries.

```ruby
block1 = ->() {}
block2 = ->() {}

stack = []
stack.push([Object, { foo: :bar }])
stack.push([Object, { foo: :bar }]) # duplicate
stack.push([Object, { baz: :bat }])
stack.push([Object, block1])
stack.push([Object, block1]) # duplicate
stack.push([Object, block2])
stack.uniq!

puts stack.count
puts stack.inspect

__END__
4
[[Object, {:foo=>:bar}], [Object, {:baz=>:bat}], [Object, #<Proc:0x007fd2bc0a7080@middleware.rb:1 (lambda)>], [Object, #<Proc:0x007fd2bc0a7008@middleware.rb:2 (lambda)>]]
```

This is the concept behind the commit that fixes the problem: https://github.com/hanami/hanami/commit/115b3302a20bc7717538d1d1580ea62a66ed3a1d

---

@deepj Thanks for your help, but I'm closing this in favor of the commit above.

deepj:
@jodosha Thanks. And a good point about procs. I just discovered something new 😸

PR #645, opened 2016-09-02
Fix typos in static documentation

davydovanton:


davydovanton:
@joneslee85 thanks :star

PR #647, opened 2016-09-26
Adding secret feature with tests

uaihebert:
I have created this PR based in this conversation: https://github.com/hanami/hanami/issues/625

It is an simple and useful feature

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ test/fixtures/rake/rake_tasks/public/assets*
 test/fixtures/rake/rake_tasks_app/public/assets*
 test/fixtures/rake/rake_tasks/db/*.sqlite
 test/fixtures/rake/rake_tasks_app/db/*.sqlite
+*.idea
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -140,7 +140,7 @@ module Hanami
       # secure random generates a string as length * 2.
       # We want to return the required length by the developer
       secret = SecureRandom.hex(length.to_i)
-      puts secret[0..(length.to_i - 1)]
+      puts secret[0...length.to_i]
     end
 
     require 'hanami/cli_sub_commands/db'
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -131,6 +131,16 @@ module Hanami
       end
     end
 
+    desc 'secret', 'Prints a secure hash'
+    long_desc <<-EOS
+    `hanami secret` will generate a secure key to be used by the developer as he wants, e.g. cookie sessions.
+     The secret will have the desired length * 2.
+    EOS
+    def secret(length = '64')
+      require 'securerandom'
+      puts "#{ SecureRandom.hex(length.to_i) }"
+    end
+
     require 'hanami/cli_sub_commands/db'
     register Hanami::CliSubCommands::DB, 'db', 'db [SUBCOMMAND]', 'Manage set of DB operations'
 
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -133,12 +133,13 @@ module Hanami
 
     desc 'secret', 'Prints a secure hash'
     long_desc <<-EOS
-    `hanami secret` will generate a secure key to be used by the developer as he wants, e.g. cookie sessions.
-     The secret will have the desired length * 2.
+    `hanami secret` will generate a secure key to be used by the developer as they want, e.g. cookie sessions.
     EOS
     def secret(length = '64')
       require 'securerandom'
-      puts "#{ SecureRandom.hex(length.to_i) }"
+      # must divide by 2 because SecureRandom.hex multiplies it by 2
+      divided_length = length.to_i / 2
+      puts SecureRandom.hex(divided_length)
     end
 
     require 'hanami/cli_sub_commands/db'
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -137,9 +137,10 @@ module Hanami
     EOS
     def secret(length = '64')
       require 'securerandom'
-      # must divide by 2 because SecureRandom.hex multiplies it by 2
-      divided_length = length.to_i / 2
-      puts SecureRandom.hex(divided_length)
+      # secure random generates a string as length * 2.
+      # We want to return the required length by the developer
+      secret = SecureRandom.hex(length.to_i)
+      puts secret[0..(length.to_i - 1)]
     end
 
     require 'hanami/cli_sub_commands/db'
--- a/test/cli_test.rb
+++ b/test/cli_test.rb
@@ -292,26 +292,24 @@ describe Hanami::Cli do
   end
 
   describe 'secret' do
-    describe 'when `secret` command' do
-      it 'prints Hanami secret using default 64 length' do
-        assert_output(/^[a-zA-Z0-9]{64}$/) do
-          ARGV.replace(%w{secret})
-          Hanami::Cli.start
-        end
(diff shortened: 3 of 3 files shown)


lib/hanami/cli.rb
```diff
@@ -131,6 +131,16 @@ def routes
...
     end
 
+    desc 'secret', 'Prints a secure hash'
+    long_desc <<-EOS
+    `hanami secret` will generate a secure key to be used by the developer as he wants, e.g. cookie sessions.
+     The secret will have the desired length * 2.
+    EOS
+    def secret(length = '64')
+      require 'securerandom'
+      puts "#{ SecureRandom.hex(length.to_i) }"
```
cllns: Don't need string interpolation here :) `puts SecureRandom.hex(length.to_i)`

lib/hanami/cli.rb
```diff
@@ -131,6 +131,16 @@ def routes
       end
     end
 
+    desc 'secret', 'Prints a secure hash'
+    long_desc <<-EOS
+    `hanami secret` will generate a secure key to be used by the developer as he wants, e.g. cookie sessions.
+     The secret will have the desired length * 2.
```
cllns: I'd personally prefer `hanami secret 20` to output a 20 character string, rather than a 40 character one.

If we want to keep it the same as how `SecureRandom` works, then can we at least explain that they're specifying the number of bytes, rather than number of characters?

lib/hanami/cli.rb
```diff
@@ -131,6 +131,16 @@ def routes
       end
     end
 
+    desc 'secret', 'Prints a secure hash'
+    long_desc <<-EOS
+    `hanami secret` will generate a secure key to be used by the developer as he wants, e.g. cookie sessions.
```
cllns: We should be gender neutral here

uaihebert: Good point. I will replace it with `it`.

lib/hanami/cli.rb
uaihebert: Ops, my bad. -_-'

lib/hanami/cli.rb
uaihebert: I believe that the ``hanami secret`` result  could be decoupled from the ``SecureRandom`` result. I will do ``length / 2`` instead.

lib/hanami/cli.rb
cllns: 'they' is better than 'it' as a gender neutral pronoun. for example: `as they want` sounds better

uaihebert: great. Thanks for the tip.

lib/hanami/cli.rb
```diff
@@ -131,6 +131,17 @@ def routes
...
     end
 
+    desc 'secret', 'Prints a secure hash'
+    long_desc <<-EOS
+    `hanami secret` will generate a secure key to be used by the developer as they want, e.g. cookie sessions.
+    EOS
+    def secret(length = '64')
+      require 'securerandom'
+      # must divide by 2 because SecureRandom.hex multiplies it by 2
+      divided_length = length.to_i / 2
```
cllns: would rather name this `number_of_bytes` rather than `divided_length`, but I also realized his solution might not be as good as I thought....

Not sure if this is likely, but if someone runs `hanami secret 65`, they'll get a string that's 64 characters long. Maybe we should keep it as it was? What do you think? 😇

uaihebert: Well, I could update the code to something like: ``SecureRandom.hex(length)[0..(number_of_bytes - 1)].length``. (I will not write all in a single command like here)

uaihebert: What do you think? about the suggestion above? ^

uaihebert: @cllns the full method would be like:
```
    def secret(length = '64')
      require 'securerandom'
      # secure random generates a string as length * 2.
      # We want to return the required length by the developer
      secret = SecureRandom.hex(length.to_i)
      puts secret[0..(length.to_i - 1)]
    end
```

.gitignore
```diff
@@ -29,3 +29,4 @@ test/fixtures/rake/rake_tasks/public/assets*
 test/fixtures/rake/rake_tasks_app/public/assets*
 test/fixtures/rake/rake_tasks/db/*.sqlite
 test/fixtures/rake/rake_tasks_app/db/*.sqlite
+*.idea
```
cllns: 👻

lib/hanami/cli.rb
```diff
@@ -131,6 +131,18 @@ def routes
...
+    desc 'secret', 'Prints a secure hash'
+    long_desc <<-EOS
+    `hanami secret` will generate a secure key to be used by the developer as they want, e.g. cookie sessions.
+    EOS
+    def secret(length = '64')
+      require 'securerandom'
+      # secure random generates a string as length * 2.
+      # We want to return the required length by the developer
+      secret = SecureRandom.hex(length.to_i)
+      puts secret[0..(length.to_i - 1)]
```
cllns: Simpler to use triple dot (an exclusive range) instead of double dot (an inclusive range) and subtracting one: `puts secret[0...length.to_i]` :)

lib/hanami/cli.rb
```diff
@@ -131,6 +131,18 @@ def routes
...
 
+    desc 'secret', 'Prints a secure hash'
+    long_desc <<-EOS
+    `hanami secret` will generate a secure key to be used by the developer as they want, e.g. cookie sessions.
+    EOS
+    def secret(length = '64')
+      require 'securerandom'
+      # secure random generates a string as length * 2.
+      # We want to return the required length by the developer
+      secret = SecureRandom.hex(length.to_i)
```
cllns: Can we do `length.to_i / 2` here? That way, in the next line, the most we throw away is one digit instead of over half of them

test/cli_test.rb
```diff
@@ -291,6 +291,24 @@
...
+        end
+      end
+
+      it 'prints Hanami secret using 32 length' do
+        assert_output(/^[a-zA-Z0-9]{32}$/) do
+          ARGV.replace(%w{secret 32})
+          Hanami::Cli.start
+        end
+      end
+    end
```
cllns: Can you add a test for an odd length too, please?

.gitignore
uaihebert: I was with vim, but started using rubymine to run some plugins for static code analyzes. :3

lib/hanami/cli.rb
uaihebert: The problem with i / 2 would be that: 

```
number_of_bytes = 31 / 2 # (it results in 15)
secret = SecureRandom.hex(number_of_bytes)
# secret.length would be 30 and not 31
```
After the * operation the odd would return 30 instead of 31.

Did I misunderstand you?

lib/hanami/cli.rb
uaihebert: Nice!

test/cli_test.rb
uaihebert: Sure! Done.

test/cli_test.rb
```diff
@@ -291,6 +291,31 @@
     end
   end
 
+  describe 'secret' do
+    describe 'when `secret` command' do
```
Ninigi: The second describe line is a little redundant, since there is only the command and no other case, right?

uaihebert: Yup! Done!

jodosha:
@uaihebert Thanks for this PR 👍  I've read all the discussion that lead you to close this. We hope to see you to contribute in the future 😉

PR #652, opened 2016-10-13
Add rubocop linter to travis.

beauby:
+ Runs the linter first on travis.
+ Existing code is forgiven for now.
+ The exact cops/parameters should probably be discussed.

Note: Had to remove some local cop disabling as corresponding cop was globally disabled by the auto-generated config.


.rubocop.yml
```diff
@@ -0,0 +1 @@
+inherit_from: .rubocop_todo.yml
```
TiteiKo: Having the inheritance here instead of in the .rubocop_todo.yml will prevent linters in editors to function correctly, and it will be almost pointless to run rubocop before committing.

For example, writing any new method will allow an ABC complexity of 52 (which is a lot). 

It'll also make people not see existing errors, as they won't show up during the development process, and so they likely won't be fixed.

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
+# This configuration was generated by
```
TiteiKo: The rubocop_todo should inherit .rubocop.yml, see comment on .rubocop.yml for further explanations.

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+    - 'test/commands/generate/model_test.rb'
+    - 'test/commands/new/app_test.rb'
+    - 'test/commands/new/container_test.rb'
+    - 'test/fixtures/config.ru'
+
+# Offense count: 40
+# Cop supports --auto-correct.
+# Configuration parameters: AllowUnusedKeywordArguments, IgnoreEmptyMethods.
+Lint/UnusedMethodArgument:
+  Enabled: false
```
TiteiKo: Why disable this cop?

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+
+# Offense count: 2
+Lint/UselessAssignment:
+  Exclude:
+    - 'test/cli_test.rb'
+    - 'test/fixtures/commands/generate/mailer/forgot_password_spec.minitest.rb'
+
+# Offense count: 18
+Metrics/AbcSize:
+  Max: 52
```
TiteiKo: A list of wrongdoing files would help the CI catch new problems

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+
+# Offense count: 5
+# Configuration parameters: CountComments.
+Metrics/ClassLength:
+  Max: 204
```
TiteiKo: Same comment as Metrics/AbcSize

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+
+# Offense count: 1
+Metrics/CyclomaticComplexity:
+  Max: 7
```
TiteiKo: Same comment as Metrics/AbcSize

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+
+# Offense count: 846
+# Configuration parameters: AllowHeredoc, AllowURI, URISchemes.
+# URISchemes: http, https
+Metrics/LineLength:
+  Max: 369
```
TiteiKo: Same comment as Metrics/AbcSize

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+
+# Offense count: 23
+# Configuration parameters: CountComments.
+Metrics/MethodLength:
+  Max: 45
```
TiteiKo: See Metrics/AbcSize

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+
+# Offense count: 2
+Style/ConstantName:
+  Exclude:
+    - 'test/commands/console_test.rb'
+    - 'test/hanami_setup_test.rb'
+
+# Offense count: 36
+Style/Documentation:
+  Enabled: false
```
TiteiKo: Why disable it when most of the code is correctly documented? Having methods and classes documented is something the hanami team seem to care a lot about.

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+    - 'test/integration/body_parsers_test.rb'
+    - 'test/integration/full_stack_test.rb'
+    - 'test/loader_test.rb'
+
+# Offense count: 18
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, SupportedStyles.
+# SupportedStyles: empty_lines, no_empty_lines
+Style/EmptyLinesAroundClassBody:
+  Enabled: false
```
TiteiKo: Why disable it?

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+    - 'test/integration/cli/database_test.rb'
+    - 'test/integration/configuration_test.rb'
+    - 'test/integration/full_stack_test.rb'
+    - 'test/middleware_test.rb'
+    - 'test/support/helper.rb'
+
+# Offense count: 36
+# Configuration parameters: MinBodyLength.
+Style/GuardClause:
+  Enabled: false
```
TiteiKo: Why disable it?

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+    - 'lib/hanami/commands/generate/action.rb'
+    - 'lib/hanami/commands/generate/app.rb'
+    - 'lib/hanami/commands/new/abstract.rb'
+    - 'lib/hanami/generators/test_framework.rb'
+    - 'lib/hanami/middleware.rb'
+
+# Offense count: 2
+# Cop supports --auto-correct.
+Style/NumericLiterals:
+  MinDigits: 8
```
TiteiKo: Same comment as Metrics/AbcSize

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+# SupportedStyles: space, no_space
+Style/SpaceAroundEqualsInParameterDefault:
+  Exclude:
+    - 'lib/hanami/cli.rb'
+
+# Offense count: 44
+# Cop supports --auto-correct.
+# Configuration parameters: AllowForAlignment.
+Style/SpaceAroundOperators:
+  Enabled: false
```
TiteiKo: Why disable it?

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+    - 'lib/hanami/configuration.rb'
+    - 'lib/hanami/routing/default.rb'
+    - 'test/controller/csrf_protection_test.rb'
+
+# Offense count: 99
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, EnforcedStyleForEmptyBraces, SupportedStyles.
+# SupportedStyles: space, no_space, compact
+Style/SpaceInsideHashLiteralBraces:
+  Enabled: false
```
TiteiKo: Why disable it?

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+
+# Offense count: 240
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, SupportedStyles.
+# SupportedStyles: space, no_space
+Style/SpaceInsideStringInterpolation:
+  Enabled: false
```
TiteiKo: Why disable it?

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+    - 'lib/hanami/rake_helper.rb'
+    - 'test/commands/console_test.rb'
+    - 'test/support/helper.rb'
+
+# Offense count: 334
+# Cop supports --auto-correct.
+# Configuration parameters: EnforcedStyle, SupportedStyles, ConsistentQuotesInMultiline.
+# SupportedStyles: single_quotes, double_quotes
+Style/StringLiterals:
+  Enabled: false
```
TiteiKo: Why disable it?

.rubocop_todo.yml
```diff
@@ -0,0 +1,841 @@
...
+Style/TrailingWhitespace:
+  Exclude:
+    - 'lib/hanami/root.rb'
+
+# Offense count: 17
+# Cop supports --auto-correct.
+# Configuration parameters: ExactNameMatch, AllowPredicates, AllowDSLWriters, IgnoreClassMethods, Whitelist.
+# Whitelist: to_ary, to_a, to_c, to_enum, to_h, to_hash, to_i, to_int, to_io, to_open, to_path, to_proc, to_r, to_regexp, to_str, to_s, to_sym
+Style/TrivialAccessors:
+  Enabled: false
```
TiteiKo: Why disable it?

script/ci
```diff
@@ -2,6 +2,10 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+run_linter() {
+  bundle exec rubocop
```
TiteiKo: Should be 
```
bundle exec rubocop --config .rubocop_todo.yml
```

.rubocop.yml
TiteiKo: All disableling of cops should be in this file. If a cop is disabled, it means it's not part of the project's style guidelines.

script/ci
TiteiKo: I don't know much about travis (I use gitlab), but I think the linter should be a test on its own, as to not be run for every ruby/jruby/other version that the gem is tested against. The result won't be different on different interpreters.

test/support/coverage.rb
```diff
@@ -24,7 +24,6 @@ def self.enabled?
     !ENV['COVERAGE'].nil?
   end
 
-  # rubocop:disable Metrics/MethodLength
```
TiteiKo: The rubocop comments should stay here, as the method can't be shortened. It's not a problem in the code, it's a method dealing with configuration.

.rubocop_todo.yml
beauby: This is the configuration generated by rubocop. Haven't looked into it, but I was assuming it would generate a more fine-grained list of forgiven offences, rather than disable cops altogether.

script/ci
beauby: Agreed.

cllns:
See this Issue: https://github.com/hanami/hanami/issues/514

I'd prefer to use something like https://houndci.com/ than integrate `rubocop` into Travis. Prefer nice comments rather than conflating a style guide with the test suite.
Either way, not sure we'll go for it because of #514.

TiteiKo:
@cllns  +1 for HoundCI, but this PR is not completely out of scope:

https://houndci.com/ uses rubocop (see https://houndci.com/configuration#ruby ), so having clean `.rubocop.yml` and `.rubocop_todo.yml` files has its importance.

Having a precise `.rubocop_todo.yml` will satisfy the need to not cleanup the entire codebase right now (in #514), but it will ensure that new contributions are following the guidelines.

beauby:
@cllns +1 for HoundCI as well. I mainly opened this PR in order to discuss style in hanami, as my personal rubocop config cries whenever I open Hanami code in emacs :)

jodosha:
@beauby Hi, I agree on using Rubocop for our codebase. Indeed, most of our frameworks are linted on master with it.

However, after experimenting with it for a couple of month, I've realized that adding it to the build isn't the best solution because people who don't integrate Rubocop with their editors don't notice the style violations. Then when the build fails on CI, they don't look at it. So even if the code is working (including unit tests), it isn't mergeable.

---

@cllns I'm sorry: you were right and I was wrong 😉 . 👍 for HoundCI.

PR #655, opened 2016-10-18
Add .sort to glob, to fix OS X / Linux discrepancy

### LINKED ISSUE #646: Issue with load_paths
Hello everyone :v:

We've encountered some odd behavior when trying to configure our application. At the moment, I am not sure how `load_paths` works. Does it load files immediately or not?

So in the `class Application < Hanami::Application` I had to add `require_relateve` per each `include` under `controller.prepare`, whereas my friend doesn't `require` these modules and it works just fine. We have the same versions of gems and ruby, both of us uses `rbenv`. The main difference is that he uses macbook and I am a laptop with derivative of Ubuntu 14.04 as OS. 

The folders that get loaded:
```ruby
load_paths << [
  'helpers',
  'controllers',
  'views'
]
```

The problem code looks like:
```ruby
controller.prepare do
  # require_relative 'controllers/authentication'
  # require_relative 'controllers/authorize'
  # require_relative 'controllers/csrf_protection'

  include Web::Controllers::Authorize
  include Web::Controllers::Authentication
  include Web::Controllers::CsrfProtection
  expose :flash
end
```

Those are all modules, which mix some common behaviour. They do not `require` or `load` anything. If I'll remove those `require_relative` I'll get `NameError: uninitialized constant Web::Controllers::Authorize` when running `bundle exec hanami s`. So that my server doesn't even start.

We checked this on the other machines with same results. On the macbook everything is ok, on ubuntu distros the same problem. 

I've created a gist of our [Gemfile.lock](https://gist.github.com/dikond/7ae4b86fcb0073b014df1fcbdc10dd7d) and a [stack trace](https://gist.github.com/dikond/3b4745dcbcd3508ed3d8358a72788a66).

cllns:
Closes #646. @kkolotyuk confirmed it fixes the problem for them

We should look through all examples of `Dir.glob` (and its alias `Dir[]`) and see if adding `.sort` would be helpful. Looks like we do that 7 other times in `hanami/hanami`. Happy to do that for this PR if it's deemed necessary. Or we could make it a separate issue that would be great for a first time contribution :)

### DIFF
--- a/lib/hanami/config/load_paths.rb
+++ b/lib/hanami/config/load_paths.rb
@@ -13,7 +13,7 @@ module Hanami
         @root = root
 
         each do |path|
-          Dir.glob(path.join(PATTERN)).each { |file| require file }
+          Dir.glob(path.join(PATTERN)).sort.each { |file| require file }
         end
       end
 


jodosha:
@cllns Thank you! 👍

Issue #643, opened 2016-08-22
Application configuration double-executed


fidothe:
Application configurations are double-executed when running `hanami server`.  The first time is when https://github.com/hanami/hanami/blob/master/lib/hanami/commands/server.rb#L25 runs, and the second by starting the app (https://github.com/hanami/hanami/blob/master/lib/hanami/commands/server.rb#L33 or https://github.com/hanami/hanami/blob/master/lib/hanami/commands/server.rb#L35).

This produces the double-middleware behaviour reported in #638.

I'll add a simple reproducing app shortly.

If this is a bug, then the root cause seems to be that there's an implicit assumption that `Hanami::Loader#load_configuration!` is memoized and subsequent calls will not re-execute the already-loaded configuration.

jodosha:
@fidothe Hello. This is fixed by https://github.com/hanami/hanami/commit/4de8fd5ba80f14f38ddccee3ad0a696cef5b3bb3, which will be merged soon.

fidothe:
@jodosha fabulous, thanks!

jodosha:
As #657 was merged, we can close this too.

PR #620, opened 2016-07-22
Revert "Require gems in bin/hanami file"

jodosha:
This reverts commit 9a8db0445fec53046c2b65e96987569a2bf1c005.

We use `hanami` bin both inside and outside Hanami projects. Inside, with the presence of a `Gemfile`, requiring `hanami/setup` just works fine. But outside, without a `Gemfile`, `hanami` crashes.

```bash
 ➜ hanami version
/Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:170:in `rescue in root': Could not locate Gemfile or .bundle/ directory (Bundler::GemfileNotFound)
	from /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:166:in `root'
	from /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:74:in `bundle_path'
	from /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:414:in `configure_gem_home_and_path'
	from /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:60:in `configure'
	from /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:121:in `definition'
	from /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:91:in `setup'
	from /Users/luca/.gem/ruby/2.3.1/gems/bundler-1.12.5/lib/bundler.rb:102:in `require'
	from /Users/luca/.gem/ruby/2.3.1/gems/hanami-0.8.0/lib/hanami/setup.rb:3:in `<top (required)>'
	from /Users/luca/.rubies/ruby-2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from /Users/luca/.rubies/ruby-2.3.1/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from /Users/luca/.gem/ruby/2.3.1/gems/hanami-0.8.0/bin/hanami:4:in `<top (required)>'
	from /Users/luca/.gem/ruby/2.3.1/bin/hanami:22:in `load'
	from /Users/luca/.gem/ruby/2.3.1/bin/hanami:22:in `<main>'
```

[...]

PR #657, opened 2016-10-24
Cleanup Hanami boot process

jodosha:
This is a proposal of a huge cleanup of the Hanami boot process.

## The problems

Until today we had chaotic, out-of-control and poor tested way to start Hanami projects. The now defunct `Hanami::Loader` was a dump of technical debt, and that was mostly my fault.

There wasn’t an easy way to inject new functionalities (eg. Project wide `config/initializers`) without the danger of break the existing code.

Soft dependencies like `dotenv` weren’t confined behind abstractions to guarantee a safe fallback in case they were missing.

Integration tests and unit tests of `hanami/hanami` used to mess with global state (`ENV`, loaded gems), which led to brittle CI builds.

The test suite used to target the code under `lib/`, which is fine for unit tests, but for integration tests we need to simulate installed gems to work on developers' computers. There was a [pull request](https://github.com/hanami/hanami/pull/566) that let the tests to pass, but once Hanami was installed as a gem, it was [broken](https://github.com/hanami/hanami/pull/620). We needed a test suite to catch these problems earlier.

There wasn’t a solid set of integration tests to install `hanami` and use its CLI commands for real.

Minitest is great for small libraries, but it had become not enough for our use cases. It hasn’t builtin features to running a focused spec, to include modules only for specific test files, and to have _shared examples_.

We used to maintain a lot of Hanami projects as [fixtures](https://github.com/hanami/hanami/tree/fd4043b8afae36ff158c0fc5ee681ac08cb3f744/test/fixtures), which constituted a maintenance problem each time something changed: Hanami version, `hanami-model` version or a feature. We had to manually update them.

To support multiple architectures was a mistake. Hanami is used in real world project with: 1. a subset of its gems or 2. with the _container architecture_ (the default). The _application architecture_ isn't used and it only contributed to increase the maintenance burden with endless conditionals in the code. In the future we can think of having this architecture in a separated gem.

### Components

`Hanami::Loader` is removed by this PR. It’s replaced by a set of focused components that have the purpose of load single pieces of a Hanami project. They are orchestrated by CLI commands implementations that explicitly declare on which components they depend on.

For instance, to start `hanami server`, the command activates the component named `all`, which loads the framework, the configurations of the project and the code.

For commands like `hanami routes`, it requires to _”resolve”_ the `routes.inspector` component, which only loads the routes of the mounted applications.

Same thing goes for `hanami assets precompile`, which just loads the assets configurations, or `hanami db console` which loads the `hanami-model` configuration and extract the connection string, **without** paying the price of loading the entire project (controllers, views, repositories, etc..) without using them.

We can easily hook into the process to add new functionalities like the project wide `config/initializers`.

**TL;DR:** Fine grained control over boot process via a new concept called _component_. Unit tests (in isolation) for these components to guarantee their behavior.

### Abstraction For Soft Dependencies

This is mainly for `dotenv` and `hanami-model` may or may not be bundled in a project.

The first is now abstracted by the new `Hanami::Env`, which is a wrapper of `ENV`. Its `#load!` method is responsible to dump into `ENV` the `.env.*` files, only if `dotenv` gem is bundled. Otherwise it’s a safe _no-op_. This is the case of deployment environments, where env vars aren’t handled via Hanami.

`hanami-model` management is now delegated to components, that safely ignore this framework if it isn’t bundled.

**TL;DR:** We have abstractions to guarantee Hanami to work fine without bundled soft dependencies.

### Unit, Isolation, Integration tests

We now have three types of tests.

#### Unit

Classic unit tests (see `spec/unit`).

#### Isolation

Similar to unit tests, but each file is run by a separated Ruby process (see `spec/isolation`). When we need to test something that changes the global state, we want to make sure to use this kind of tests.

For instance, we want to test if trying to load `model` component, **without** the `hanami-model` gem is around, there isn’t other safer way to run this scenario in its own Ruby process.

Other cases are the loading of `.hanamirc`, `.env.*`, `Hanami::Environment`, which change the state of `ENV`.

#### Integration

They simulate features from the user perspective. They are run in isolation too (see `spec/integration`).

We use now `aruba` gem to assert stdout/stderr and exit code of a command is the one we expect.

With the help of Bundler, during these tests we vendor all the Hanami gems (and its dependencies) in `vendor/cache` and install them. So we’re as much as close to a developer’s computer that installed Hanami via `gem install hanami`.

If a change [breaks](https://github.com/hanami/hanami/pull/566) the CLI, we can find this problem at the build time.

**TL;DR:** Tests that change the global state are run in isolation in their own Ruby process. During _integration tests_, we use Bundler to vendor and install all the Hanami gems and their dependencies to guarantee the right behavior of CLI commands and high level features.

### RSpec

I ported the test suite from Minitest (too minimal) to RSpec, which offers a rich set of features that we need.

There are third-party gems to add these functionalities to Minitest, but these gems can potentially go out-to-date, while RSpec always guarantees to have them baked in.

**TL;DR:** We now use RSpec.

### Au Revoir, Fixtures

Fixtures are gone! We have now a method: `with_project` that generates a **complete** and **updated** Hanami project. Then it bundles the gems and changes the current working directory to the root of the project. With this technique we simulate a developer working on their project.

**Each test (`it` block) uses its own fresh Hanami project, generated at the runtime.**

With this method, we have fine control on the default template (`ERB`/`HAML`/`Slim`), on the bundled gems in `Gemfile`, on the chosen database, etc..

A few examples:

Simulate `hanami version` in a project.

```ruby
RSpec.describe 'hanami version', type: :cli do
  it 'prints current version' do
    with_project do
      run_command 'hanami version', "v#{Hanami::VERSION}"
    end
  end
end
```

Simulate db commands with a project that uses `sqlite`.

```ruby
RSpec.describe "hanami db", type: :cli do
  describe "migrate" do
    it "migrates database" do
      with_project("bookshelf_db_migrate", database: :sqlite) do
        # …
      end
    end
  end
end
```

**TL;DR:** We use a set of test helpers to generate Hanami projects, bundle the gems and run commands from its root. Each test gets a fresh project to work with.

### Application Architecture Is No Longer Supported

The new set of tests only cover _container architecture_. The CLI implementation is mostly left untouched by this refactoring, so it’s still technically possible to generate an _application architecture_ project, but we just don’t want to support it anymore. In a separated PR, we should work to eliminate the leftover code.

**TL;DR:** We want to focus only on one architecture: _container_. _application_ is gone.

—

## Fixes

Alongside with this cleanup, I had the chance to fix a few issues:

  * Ensure consistent loading order of the code across different operating systems (See #646, partially fixed by @cllns with #655)
  * Ensure to evaluate applications’ configurations only once (See #643)
  * Don’t duplicate middleware in the stack (See #638)

#### `Hanami.app`

Use `Hanami.app` instead of `Hanami::Container.new`. This affects `config.ru` and `spec/features_helper.rb`.

```ruby
# config.ru
require './config/environment'

run Hanami.app
```

```ruby
# spec/features_helper.rb
# …
Capybara.app = Hanami.app
# …
```

It reads better and clarifies that there isn’t only one instance of it per Ruby process.

#### `Hanami.boot`

Use `Hanami.boot` istead of `Hanami::Application.preload!`, as the first is cleaner and less detailed of the latter.

```ruby
# spec/spec_helper.rb
# …
Hanami.boot
# …
```

#### `Hanami.configure`

Global configurations are now handled by this new method. It includes: mounted apps, model and mailer configurations.

```ruby
require 'bundler/setup'
require 'hanami/setup'
require 'hanami/model'

Hanami.configure do

  model do
    ##
    # Database adapter
    #
    # Available options:
    #
    #  * File System adapter
    #    adapter type: :file_system, uri: 'file:///db/bookshelf_development'
    #
    #  * Memory adapter
    #    adapter type: :memory, uri: 'memory://localhost/bookshelf_development'
    #
    #  * SQL adapter
    #    adapter type: :sql, uri: 'sqlite://db/bookshelf_development.sqlite3'
    #    adapter type: :sql, uri: 'postgres://localhost/bookshelf_development'
    #    adapter type: :sql, uri: 'mysql://localhost/bookshelf_development'
    #
    adapter type: :sql, uri: ENV['DATABASE_URL']

    ##
    # Migrations
    #
    migrations 'db/migrations'
    schema     'db/schema.sql'

    ##
    # Database mapping
    #
    # Intended for specifying application wide mappings.
    #
    mapping do
      # collection :users do
      #   entity     User
      #   repository UserRepository
      #
      #   attribute :id,   Integer
      #   attribute :name, String
      # end
    end
  end

  mailer do
    root 'lib/bookshelf/mailers'

    delivery do
      development :test
      test        :test
      # production :smtp, address: ENV['SMTP_PORT'], port: 1025
    end
  end
end
```

As consequence, `lib/bookshelf.rb` is cleaned up from Hanami configurations:

```ruby
# lib/bookshelf.rb
Hanami::Utils.require!("#{__dir__}/bookshelf")
```

NOTE: this new `Hanami::Utils.require!` is a _syntactic sugar_ to recursively require Ruby files under that directory. It guarantees the same order of loading across operating systems.

#### `Web.routes` and `Web.logger`

We used to generate constants at the runtime to access single application’s routes and logger. For a given application `Web::Application` we used to define `Web::Routes` and `Web::Logger`. This was confusing for newcomers and this is a technique that we should abandon for other frameworks too (eg. `Web::Action`, `Web::View`, etc..)

This is the time of introducing `Web.routes` and `Web.logger` as replacement of that two constants.

—

I hope this was an exhaustive explanation of the reasons behind this work and how it is implemented. I’d love to hear @hanami/core-team and @hanami/contributors  voice here. Thanks! :green_heart:


TiteiKo:
👍

lib/hanami/generators/test_framework.rb
```diff
@@ -25,12 +25,13 @@ def minitest?
...
         if !supported_framework?
-          raise ArgumentError.new("Unknown test framework '#{ framework}'. Please use one of #{ valid_test_frameworks.join(', ')}")
+          warn "`#{framework}' is not a valid test framework. Please use one of: #{valid_test_frameworks.join(', ')}"
+          exit(1)
         end
       end
 
       def valid_test_frameworks
-        VALID_FRAMEWORKS.map { |name| "'#{ name }'"}
+        VALID_FRAMEWORKS.map { |name| "`#{name}'"}
```
AlfonsoUceda: `'?

lib/hanami/app.rb
```diff
@@ -0,0 +1,72 @@
...
+    def mount(configuration)
+      configuration.mounted.each do |klass, app|
+        routes.mount(klass, at: app.path_prefix)
+      end
+    end
+
+    # @since x.x.x
+    # @api private
+    def middleware(environment)
+      if middleware = environment.static_assets_middleware # rubocop:disable Lint/AssignmentInCondition
```
jodosha: @AlfonsoUceda I wanted to disable this **only** in this place.

PR #665, opened 2016-11-11
Only run git-related commands if git available.

beauby:
Following [this PR](https://github.com/hanami/hanami/pull/163), running `$ hanami new foo` will run `git init` inside the created directory. If git is not available, the generator crashes. Although git is widely used, hanami should probably not assume it is available on every developper's machine.

This PR makes the generators add git-related files and run git-related commands only if git is available.
Moreover, all `.gitkeep` files are renamed to `.keep`.


lib/hanami/commands/new/abstract.rb
```diff
@@ -93,6 +93,10 @@ def git_dir_present?
           File.directory?(target.join('.git'))
         end
 
+        def git_available?
+          system('which', 'git', out: File::NULL)
```
AlfonsoUceda: @beauby thanks for the PR, it looks good, the only thing I'm concerned that this line doens't work in windows but I can't tell you if hanami works in windows.

cc @jodosha

beauby: @AlfonsoUceda You are right. Maybe a first step could be to have `git_available?` always return `true` on windows for the moment, and only do the check on *nix.

jodosha:
@beauby Can you please add tests for this? CLI tests are powered by Aruba, please check its documentation to simulate the lack of `git`.

beauby:
@jodosha Sure, will do it over the weekend :+1:

lib/hanami/commands/new/abstract.rb
katafrakt: Question: in order to have it work both on Unix and Windows, why can't we just use `system('git --version')`? It would return `true` if git is available and `nil` if it's not. If I recall correctly, we used it widely in the code that was supposed to work on both systems.

beauby: @katafrakt That's a great suggestion – will update PR accordingly.

beauby:
After browsing through the docs of Aruba, it does not appear clearly that there's a way to simulate a command is not available. Any Aruba connaiseur around?

jodosha:
@beauby I can't recall where I found it a few months ago.

But we can build it in house. Try to add these methods to `spec/support/cli.rb`:

```ruby
require_relative 'env'

def without_command(cmd)
  cmd_path = which(cmd)
  path = RSpec::Support::Env['PATH']
  RSpec::Support::Env['PATH'] = path.split(":").reject { |p| p == cmd_path }.join(":")

  yield
ensure
  RSpec::Support::Env['PATH'] = path
end

# Cross-platform way of finding an executable in the $PATH.
#
#   which('ruby') #=> /usr/bin/ruby
#
# Adapted from http://stackoverflow.com/a/5471032/498386
def which(cmd)
  exts = RSpec::Support::Env['PATHEXT'] ? RSpec::Support::Env['PATHEXT'].split(';') : ['']
  RSpec::Support::Env['PATH'].split(File::PATH_SEPARATOR).each do |path|
    exts.each { |ext|
      exe = File.join(path, "#{cmd}#{ext}")
      return exe if File.executable?(exe) && !File.directory?(exe)
    }
  end
  
  return nil
end
```

Then add a new integration test: `spec/integration/cli/new/without_git_spec.rb` that wraps the generation of the project with: `without_command('git')`.

Please let me know if this does the trick.

beauby:
@jodosha Thanks – will take care of it later tonight.

beauby:
@jodosha Done (c.f. https://github.com/hanami/hanami/pull/665/files#diff-4dfb6fbfe5cdb09c8725a5bea1909f78 https://github.com/hanami/hanami/pull/665/files#diff-90e737db900dee751c4076fd6b4a7a72).

beauby:
@jodosha Actually, this solution is not really satisfying: if `bundle` happens to be in the same path as `git`, the spec will fail. We could add a fake `git` executable that returns non-zero in an artificial path in front (though it wouldn't be exactly the same, and it would take a bit of work to make it work cross-platform (even though Hanami tests are currently not ran on anything else than ubuntu)).

mereghost:
@beauby have you found a way to simulate a missing command?

beauby:
@mereghost As you can see I tried several things, none of which proved satisfactory. I'll try to come up with something, unless someone has a solution?

igneus:
I'm not sure if this is the right place for fundamental objections. If not, feel free to delete/move it:

I consider automatic creation of a git repository by `hanami new` slightly annoying. Not because it has made me any real harm, but (1) because it attempts to decide for me if and by which means I will version-control the new app, (2) because it brings no real value (I don't find being saved a single invocation of `git init` really valuable) and (3) because it violates "single responsibility principle esthetics" of doing a single task well. As I see it, initializing version control is out of scope of application scaffolding.

So, while being really excited about the problem of simulating git non-availability in aruba, I would prefer removing the feature which triggered this testing adventure altogether.

AlfonsoUceda:
I'm with @igneus IMO, the developer should use the control version that they want (subversion, git, mercurial...) so I think we can get rid everything related to git init and ignore files, sometimes I create hanami apps for testing and I don't need the version control.

Just my opinion folks ;)

igneus:
> I think we can get rid everything related to git init and ignore files

I would only get rid of `git init`, not of generating `.gitignore`. I actually quite appreciate ignore lists generated with respect to the framework's conventions, especially when I am new to the particular framework. AFAIK, Rails and Symfony generate `.gitignore`, too. (But neither of them initializes a git repository.)

beauby:
@AlfonsoUceda @igneus I agree that initializing a git repo is unnecessary and brings very little value so I would support removal of that feature. However, keeping the `.keep` files makes sense, as it is a generic pattern. Keeping the `.gitignore` as well (possibly only when git is available).

katafrakt:
Just so you know: Rails will init git repo on `rails new` from now on: https://github.com/rails/rails/pull/27632

igneus:
@katafrakt Good to know. It doesn't change anything about the fact that I don't like the idea, though.

("`rails new` initializes a git repo? One more reason for Hanami not to do so!" :stuck_out_tongue: )

The reasons I expressed above are obviously rather esthetic than anything else. It seems that @AlfonsoUceda sort of shares my concept of "software behaviour beauty". But if other developers find initializing a git repo useful/convenient, I have no really convincing arguments against it.

AlfonsoUceda:
I would go keeping the `.keep` files and not generating a git repo, `git init` is a very straightforward command.

@beauby could you keep those files but remove everything else related to `git init` command? of course could you squash all your commits?

cc @jodosha

beauby:
@AlfonsoUceda This PR renames the `.gitkeep` files to `.keep`. Should I just remove the git commands then?

AlfonsoUceda:
@beauby no sorry, I think we should keep the `.keep` files, but remove the `git_availability?` method for problems with Windows, then in the future, think if we should keep initialize git or not.

katafrakt:
What are the problems with `git_available?` method on Windows?

@igneus I don't claim that this changes anything. I also don't think that initializing empty repo is a good thing. However, it was mentioned as an argument in this discussion that "Rails does not do it" so I thought it might be worth to know that it no longer stands as true :wink:

beauby:
@katafrakt The issue with `git_available?` is that we can't seem to find a satisfying way to test it (even on linux).

AlfonsoUceda:
Folks, I'm going to close this PR, the reason is we'll rewrite CLI with `optparse` and then we can add an option to avoid git, but thanks for this PR ;)

PR #678, opened 2016-11-29
Ensure `:environment` Rake task to load the entire project

### LINKED ISSUE #677: Environment rake task depends on preload, which does not exist anymore
Well... as the title says. The `:environment` rake task depends on the `:preload` task that was removed just before the 0.9 release.

As far as I understand, the `:preload` task shouldn't even be needed anymore so the simple fix would be remove `:environment`'s dependency on it.

jodosha:
It also eliminates the dependency with the `:preload` Rake task which was removed from v0.9.0.

---

Closes #677

/cc @mereghost @hanami/core-team @hanami/contributors for review. Thanks

### DIFF
--- a/lib/hanami/rake_helper.rb
+++ b/lib/hanami/rake_helper.rb
@@ -21,8 +21,10 @@ module Hanami
     # rubocop:disable Metrics/MethodLength
     def install
       desc "Load the full project"
-      task environment: :preload do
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
LGTM :+1:

TiteiKo:
👍

PR #679, opened 2016-11-30
Lazily load the project when `hanami server` isn't used

### LINKED ISSUE #675: LoadError on action execution when not started by hanami server
Hello everyone,

On a fresh hanami application when running the server directly (i.e `bundle exec puma` , using `bundle exec` as soon as a Repository is called, the server throws a LoadError.

It seems that the Hanami boot process (`Components.resolve('all')`) isn't called by `Hanami.app`, that leads to `hanami-model` not to be correctly initialized.

A simple fix is to call `Hanami.boot` on the `config.ru` just the before the `run Hanami.app` or change `Hanami::Application` (or the `Hanami.app` method) to resolve all components. I do think the later might be more elegant. =)

Steps to reproduce the problem [here](https://github.com/tadejm/bookshelf/commit/b20cfd16ef20accc7ed5fa236096eddfb9f9d042)

kudos to @tadejm for the repo that helped me isolate the problem.

jodosha:
Our CLI commands guarantee to load only certain parts of a project.

For instance `hanami routes` loads only the routes of the project, **without** loading other useless parts like repositories or actions.

In the case of `hanami server`, it loads the entire project (routes, actions, models, etc..). But when the project is loaded without this command (eg. `rackup` or `puma`), the boot process doesn't happen (as reported by #675).

---

Personally, I use this configuration for Puma:

```ruby
# config/puma.rb
workers Integer(ENV['WEB_CONCURRENCY'] || 2)
threads_count = Integer(ENV['MAX_THREADS'] || 5)
threads threads_count, threads_count

preload_app!

rackup      DefaultRackup
port        ENV['PORT']     || 2300
environment ENV['RACK_ENV'] || ENV['HANAMI_ENV'] || 'development'

on_worker_boot do
  Hanami.boot
end
```

Which is used like this:

```shell
bundle exec puma -C config/puma.rb
```

That `on_worker_boot` hook is used to (re)boot the project.

---

While the configuration above solves the problem, the lack of lazy boot can be: confusing (see #675) or inelegant (`bundle exec rackup --eval "require './config/environment'; Hanami.boot"` would solve the problem)

**This PR aims to solves these frictions by introducting lazy boot support.**

---

Closes #675 

/cc @mereghost @tadejm @hanami/core-team @hanami/contributors for review

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -77,7 +77,15 @@ module Hanami
     end
   end
 
-  # Boot Hanami project
+  # Boot your Hanami project
+  #
+  # NOTE: In case this is invoked many times, it guarantees that the boot
+  #   process happens only once.
+  #
+  # NOTE: There is no reason to cache the result with `@_booted`, because it
+  #   already caches it internally.
+  #
+  # NOTE: This MUST NOT be wrapped by a Mutex, because it would cause a deadlock.
   #
   # @since 0.9.0
   # @api private
@@ -92,11 +100,20 @@ module Hanami
   #   * `config.ru` (`run Hanami.app`)
   #   * Feature tests (`Capybara.app = Hanami.app`)
   #
+  #
+  #
+  # It lazily loads your Hanami project, in case it wasn't booted on before.
+  # This is the case when `hanami server` isn't invoked, but we use different
+  # ways to run the project (eg. `rackup`).
+  #
   # @return [Hanami::App] the app
   #
   # @since 0.9.0
   # @api private
+  #
+  # @see Hanami.boot
   def self.app
+    boot
     App.new(configuration, environment)
   end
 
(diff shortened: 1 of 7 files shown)


AlfonsoUceda:
@jodosha LGTM, I've seen you deleted `setup_model` in server_spec.rb, I think you can remove the following methods:
https://github.com/hanami/hanami/blob/45a244dabed7c817c9478c0ddf632bb150350804/spec/integration/cli/console_spec.rb#L28
https://github.com/hanami/hanami/blob/27516f9028d15b4251284fccc73670a53df82b1b/spec/integration/rake/with_minitest_spec.rb#L36
https://github.com/hanami/hanami/blob/27516f9028d15b4251284fccc73670a53df82b1b/spec/integration/rake/with_rspec_spec.rb#L34

PR #680, opened 2016-11-30
Ensure force SSL to redirect to the right port

### LINKED ISSUE #666: Force SSL includes default port in redirect
Experienced in Hanami v0.8.

### Scenario

When the `force_ssl` option is set but `port` and `HANAMI_PORT` are not set the [port passed to `Hanami::Router`](https://github.com/hanami/hanami/blob/0.8.x/lib/hanami/loader.rb#L240) is the default (`2300`) this causes the result http => https redirect to have the port `2300` which causes the response to hang.

See `curl` output below:

```
$ curl -sv http://a-force-ssl-site.com
* Rebuilt URL to: http://a-force-ssl-site.com/
*   Trying x.x.x.x...
* Connected to a-force-ssl-site.com (x.x.x.x) port 80 (#0)
> GET / HTTP/1.1
> Host: a-force-ssl-site.com
> User-Agent: curl/7.49.1
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Date: Tue, 15 Nov 2016 01:33:29 GMT
< Location: https://a-force-ssl-site.com:2300/
< Content-Length: 0
< Connection: keep-alive
<
* Connection #0 to host a-force-ssl-site.com left intact
```

Setting the port to `443` for the production environment resolves issue although this conflicts with [generator comment]( https://github.com/hanami/hanami/blob/0.8.x/lib/hanami/generators/app/application.rb.tt#L49-L53) which also appears to suggest that the port may be figured out by the scheme however, [this does not appear to be the case](https://github.com/hanami/hanami/blob/0.8.x/lib/hanami/environment.rb#L326-L328).

>URI port used by the routing system to generate absolute URLs
>Argument: An object coercible to integer, default to 80 if the scheme is http and 443 if it's https
>This SHOULD be configured only in case the application listens to that non standard ports

jodosha:
When force SSL option is active and the port is not set, the application configuration returns a wrong port value.

```ruby
module Web
    # ...
    configure :production do
      force_ssl true
      scheme 'https'
      host 'example.org'
      # port 443
    end
  end
end
```

In this case the returned port is `2300`, but it should be the default SSL port `443`.

---

This PR fixes this problem, by implementing the following rules:

  * When nothing is set, return `2300`
  * When port is set via `HANAMI_PORT` env var, use it
  * When port is set via `port` setting, use it
  * When nothing is set, but `force_ssl` is, use `443`
  * When `force_ssl` is set but one of `HANAMI_PORT` or `port` are used, give them the priority

---

Closes #666 

/cc @createdbypete @hanami/core-team @hanami/contributors for review. Thanks. 😄

### DIFF
--- a/spec/unit/application_configuration_spec.rb
+++ b/spec/unit/application_configuration_spec.rb
@@ -78,6 +78,14 @@ RSpec.describe Hanami::ApplicationConfiguration do
             expect(subject.port).to eq(8200)
           end
         end
+
+        context "and forced by env var" do
+          let(:hanami_port) { 4321 }
+
+          xit "returns the one from the env var" do
+            expect(subject.port).to eq(4321)
+          end
+        end
       end
 
       context "only in the current environment configuration" do
--- a/spec/unit/application_configuration_spec.rb
+++ b/spec/unit/application_configuration_spec.rb
@@ -1,0 +1,109 @@
+RSpec.describe Hanami::ApplicationConfiguration do
+  before do
+    env['HANAMI_ENV']  = hanami_env
+    env['HANAMI_PORT'] = hanami_port unless hanami_port.nil?
+  end
+
+  subject { described_class.new(namespace, configurations, path_prefix, env: environment) }
+
+  let(:namespace)      { ApplicationConfigurationTesting }
+  let(:configurations) { Hanami::EnvironmentApplicationConfigurations.new }
+  let(:path_prefix)    { '/' }
+
+  let(:environment)    { Hanami::Environment.new(env: env) }
+  let(:env)            { Hash[] }
+
+  let(:hanami_env)  { 'development' }
+  let(:hanami_port) { nil }
+
+  describe "#port" do
+    context "when not configured" do
+      it "returns the default value" do
+        expect(subject.port).to eq(2300)
+      end
+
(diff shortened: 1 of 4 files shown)


davydovanton:
looks good 👍

spec/unit/application_configuration_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+          it "returns the current environment value" do
+            expect(subject.port).to eq(8200)
+          end
+        end
+
+        context "and forced by env var" do
+          let(:hanami_port) { 4321 }
+
+          xit "returns the one from the env var" do
```
AlfonsoUceda: @jodosha why this pending test?

jodosha: @AlfonsoUceda Because it doesn't pass. Unfortunately local `port` setting takes the priority over the `HANAMI_PORT` env var. Luckily it's an edge case: `port` isn't always used.

---

This led me to think that we may have a design flaw here. We have a project wide mode to set the port via `HANAMI_PORT` and an application setting `port` for that. Shouldn't everything be at the project/process level?

jodosha: This could be food for thought after 1.0.

PR #682, opened 2016-12-02
Add create_table migration file to model generator

### LINKED ISSUE #670: Consider generating a migration when generating a model
Generating a model (via `$ hanami generate model foo`) currently generates an entity along with a repository (but no migration). However, repository schema inference expects a corresponding table to exist, so generating a model effectively leaves the app in broken state (unable to boot).

yjukaku:
This adds the ability to generate a basic table creation migration when generating a model through the `hanami generate model ***` command.

It also adds a --skip-migration option to allow skipping the migration.

Closes #670 

Notes: I had to require the `Migration::Generate` class into the `Model::Generate` class, so that I could use the constants defined about migration file naming - not sure if this is bad practice because it couples those together.


cllns:
Can you add a test for `skip_migration` too, please? :)

yjukaku:
I'm not sure how, to be honest. 😄   I would probably have to remove all of the current migration file checking from the shared "new model" spec?

yjukaku:
There's also the issue that once you create the model with a migration, using hanami destroy model_name doesn't delete the migration.

jodosha:
@yjukaku Hi and thanks for this PR.

You should reuse the current generator class for the migration. The same that powers `hanami generate migration`. So we can avoid duplication.

With this PR, if you generate a model, the corresponding migration is something like this:

```ruby
Hanami::Model.migration do
  change do
    create_table :dns_records do
      primary_key :id

      column :created_at, DateTime, null: false
      column :updated_at, DateTime, null: false
    end
  end
end
```

Which looks good. But if you run only the migration generator (`bundle exec generate migration create_domains`), we have something like this:

```ruby
Hanami::Model.migration do
  change do
  end
end
```

---

The same thing is true for destroying the migration contextually to the model:

```shell
# Generate
➜ bundle exec hanami generate model dns_record
      create  lib/bookshelf/entities/dns_record.rb
      create  lib/bookshelf/repositories/dns_record_repository.rb
      create  db/migrations/20161206162136_create_dns_records.rb # current timestamp, right
      create  spec/bookshelf/entities/dns_record_spec.rb
      create  spec/bookshelf/repositories/dns_record_repository_spec.rb
```

It generates the migration with the current timestamp (right). But when it comes to destroy the model, it uses the current timestamp of the command execution.

```shell
# Destroy
➜ bundle exec hanami destroy model dns_record
      remove  lib/bookshelf/entities/dns_record.rb
      remove  lib/bookshelf/repositories/dns_record_repository.rb
      remove  db/migrations/20161206162202_create_dns_records.rb # current timestamp, wrong
      remove  spec/bookshelf/entities/dns_record_spec.rb
      remove  spec/bookshelf/repositories/dns_record_repository_spec.rb
```

The command `hanami destroy migration create_dns_records` already works as expected, please reuse it.

yjukaku:
Good point on using the `Migration` class. The logic in that class that I need to use is `destination_path`, and `destroy`.   


## Generate
The only method I need from the Migrate generator is the `destination_path` that gives the path + timestamped file name for the migration. Unfortunately `destination_path` is a private method of `Migrate`.  Should I make that public?

## Destroy

If I just override Generate::Model's destroy and do:

```ruby
def destroy
  migration = Migration.new({}, "create_#{@table_name}").destroy.start
  super
end
```

There is a problem, since we've already called `add_mapping` in the Model class.  That automatically knows to delete the file when destroy is called, so if I override destroy as above, we get two delete attempts on `hanami destroy model ***`. I don't know how to stop that from happening. 😕

yjukaku:
A completely different approach could be moving logic into the `CliSubCommands::Generate` class which seems like it wouldn't be a great way to do it:

```ruby
      def model(name)
        if options[:help]
          invoke :help, ['model']
        else
          require 'hanami/commands/generate/model'
          model = Hanami::Commands::Generate::Model.new(options, name)
          model.start
         # Add all of this logic... :(
          unless options.fetch(:skip_migration, false)
            require 'hanami/commands/generate/migration'
            # migration_name being 'create_users' or whatever
            migrate = Hanami::Commands::Generate::Migration.new({}, model.migration_name)
            migrate.start
          end
          # end new logic
        end
      end
```

and the corresponding code in the `CliSubCommands::Destroy` class:

```ruby
      def model(name)
        if options[:help]
          invoke :help, ['model']
        else
          require 'hanami/commands/generate/model'
          model = Hanami::Commands::Generate::Model.new(options, name)
          model.destroy.start
          # new code
          migrate = Hanami::Commands::Generate::Migration.new({}, model.migration_name)
          migrate.destroy.start
          # end new code
        end
      end
```

This of course loses the ability to add table creation specific migration code... 🙃  unless we add some options to the migrator to use a specific template.

yjukaku:
Any suggestions or guidance on which way would be more acceptable?

mereghost:
My gut reaction would be go for the second approach, but @jodosha might have a different view here.

jodosha:
I just merged this in order to work on the destroy part myself. Thank you @yjukaku !

PR #685, opened 2016-12-07
Replaced hash merge in generator.rb

Newman101:
Hi.

Looks like there was a hash merge in lib/hanami/generators/generator.rb. I replaced this merge with faster version. I am aware that destructive method (such as ```merge!```) does not duplicate the hash but this should make the operation even faster.

Thank you.


AlfonsoUceda:
thanks @Newman101, I'm merging it because the errors are related with other thing in master.

PR #687, opened 2016-12-08
Use poltergeist for all platforms

cllns:
Linux and JRuby: `poltergeist` depends on `phantomjs`, which is 16+ MB.

Everything else: `capybara-webkit` depends on `QT`, which is 700+ MB.

Using two different libraries based on platform could cause problems down the road if the drivers behave differently.

Can we just use poltergeist for all platforms?

@jodosha What do we need these for? Do we have feature tests that need JS? I'd prefer using `Rack::Test` (no dependencies) if possible!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,11 +41,7 @@ if RUBY_DESCRIPTION =~ /linux/
   gem 'therubyrhino', require: false, platforms: :jruby
 end
 
-if RUBY_DESCRIPTION =~ /linux|jruby/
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
to me it's ok! it works and we can get rid of capybara-webkit (QT!!!) cc @jodosha

jodosha:
@cllns Thank you very much for this PR. This facilitates newcomers' life. 💯

PR #684, opened 2016-12-07
Fix links in comments

mgrachev:

PR #688, opened 2016-12-08
Fix build for #684

cllns:
My fault 😇 I merge #684 when the build was broken.

I thought it was just a documentation change (because it is) but I didn't realize we have assertions based on code comments. Now I know :)

### DIFF
--- a/spec/integration/cli/new_spec.rb
+++ b/spec/integration/cli/new_spec.rb
@@ -647,7 +647,7 @@ module Web
 
         # Subresource Integrity
         #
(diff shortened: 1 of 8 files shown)


AlfonsoUceda:
@cllns you need to change this [line](https://github.com/hanami/hanami/blob/master/spec/integration/cli/new_spec.rb#L650), see the [template](https://github.com/hanami/hanami/blob/master/lib/hanami/generators/app/application.rb.tt#L343)

cllns:
Merging so master build isn't broken :)

PR #694, opened 2016-12-14
Delimited %w literal in cli.rb

Newman101:
Hi. 

Looks like there was ```%w``` literal which was not delimited in lib/hanami/cli.rb. 

This should remove a minor bug risk.

Thank you.

### DIFF
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -17,7 +17,7 @@ module Hanami
       require 'hanami/version'
       puts "v#{ Hanami::VERSION }"
     end
-    map (%w{--version -v}) => :version
+    map %w({--version -v}) => :version
 
     desc 'server', 'Starts a hanami server'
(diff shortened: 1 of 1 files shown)


TiteiKo:
Wasn't this `%w` already delimited by `{}`?

> If you are using “(”, “[”, “{”, “<” you must close it with “)”, “]”, “}”, “>” respectively. You may use most other non-alphanumeric characters for percent string delimiters such as “%”, “|”, “^”, etc.
> https://ruby-doc.org/core-2.0.0/doc/syntax/literals_rdoc.html#label-Percent+Strings

Newman101:
Hi. Looks like you are right. I reviewed the document and noticed that it is indeed already delimited. Thanks for notification and sorry for meaningless PR. =)

TiteiKo:
No problem, thanks to you for your time!

PR #700, opened 2016-12-21
Add spec helper for each generated spec file

davydovanton:
I found that `generation` doesn't add `spec_helper` that's why I have some troubles with rspec-hanami and that's why I need to put `require 'spec_helper'` to each file.

/cc @hanami/core


mereghost:
On the generated `hanami` project, there's already a `.rspec` file that requires `spec_helper`, so requiring it on each file should be unnecessary. 

Can you provide a sample of the troubles you had?

davydovanton:
hmm. @mereghost you're right. My application didn't have this line in `.rspec` that's why I had an error. Thanks for the help. I'll close it 👍

Issue #696, opened 2016-12-14
file not found error / router bypassed


kaikuchn:
Greetings, let me first state what a joy it has been to work with Hanami so far! Thank you all so much for putting in the time and effort to build something so grand!

Today, however, I've encountered a weird issue that is probably related to #480 #488 and their fixes #493 and #506. I also believe it or rather its solution to be conceptually related to #640.

I'm building an app called `notifications` and I want all of it's dynamic and static content to be scoped under `/notifications`. I therefore tried the following in my `config/routes.rb`:
```ruby
namespace :notifications do
  ...
end

### alternatively
mount Web::Application, at: '/notifications'
```

Either of which works fine until I also prefixed the assets in my `apps/web/application.rb` like this
```ruby
# frozen_string_literal: true
require 'hanami/helpers'
require 'hanami/assets'

module Web
  class Application < Hanami::Application
    configure do
      # ...
      assets do
        # ...
        prefix '/notifications/assets'
      end
      # ...
    end
    # ...
  end
end
```
After the first successful loading of a page the assets were written to the public directory and henceforth the router is bypassed and a file not found error is returned as http response.

I'd expect the existence of a folder in the public directory to not preclude the router's matching. Much less so when the full path cannot be resolved.

To reproduce this:
- run `hanami new filenotfound`
- cd into the new project and run bundle install
- run `echo "get '/hello/goodbye', to: ->(env) { [200, {}, ['Hello from Hanami!']] }" > apps/web/config/routes.rb`
- visit `http://localhost:2300/hello/goodbye` to check it works
- run `mkdir public/hello`
- visit `http://localhost:2300/hello/goodbye` again to now see the error

I'm using hanami 0.9.1. Here's my full [Gemfile.lock](https://gist.github.com/kaikuchn/814301d497218f3de0cf41366e9d7103).

kaikuchn:
Digging into the source code I see that it has changed somewhat since the other issues were resolved.
[Here](https://github.com/hanami/hanami/blob/master/lib/hanami/static.rb#L46) you are explicitly passing in all folders contained in `public/` which -- as stated by the [rack/static docs](https://github.com/rack/rack/blob/master/lib/rack/static.rb#L13) -- results in _all_ requests where the path starts with those folders to be intercepted.

[...]

PR #704, opened 2016-12-25
Serve only existing assets with Rack::Static

kaikuchn:
Closes https://github.com/hanami/hanami/issues/696

### DIFF
--- a/lib/hanami/static.rb
+++ b/lib/hanami/static.rb
@@ -34,10 +34,6 @@ module Hanami
 
     # @since 0.8.0
     # @api private
-    EXCLUDED_ENTRIES = %w(. ..).freeze
-
-    # @since 0.8.0
-    # @api private
     URL_PREFIX = '/'.freeze
 
     # @since 0.6.0
@@ -53,10 +49,10 @@ module Hanami
     def _urls(root)
       return [] unless root.exist?
 
-      Dir.entries(root).sort.map do |entry|
-        next if EXCLUDED_ENTRIES.include?(entry)
-        "#{URL_PREFIX}#{entry}"
-      end.compact
+      asset_files = Dir.chdir(root) do
+        Dir['**/*'].select { |path| File.file? path }
+      end
+      Hash[asset_files.map { |entry| ["#{URL_PREFIX}#{entry}", entry] }]
     end
   end
 end
--- a/spec/integration/route_namespaces_spec.rb
+++ b/spec/integration/route_namespaces_spec.rb
@@ -1,0 +1,51 @@
+RSpec.describe 'route namespaces', type: :cli do
+  it "can handle asset prefixes" do
+    with_project do
+      generate "action web home#index --url=/"
+
+      replace(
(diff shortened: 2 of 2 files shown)


kaikuchn:
Instead of passing an array of folders that exist in the `root` directory into `Rack::Static` we now pass a hash which maps URIs to files as described in the [Rack docs](https://github.com/rack/rack/blob/master/lib/rack/static.rb#L23).
We thereby only trigger the rack fileserver for URIs resolvable to static files, else we delegate to later middlewares to handle the request.

AlfonsoUceda:
@kaikuchn we need a test or tests proving the failing scenario, could you add one? If you don't know how to start, maybe @jodosha could help you to point out what you need for that.

kaikuchn:
Sure, I actually wrote a failing test first. But then I realized that it isn't executed when I run `bundle exec rake test`. I then looked into the README on how to run the test suite and did as explained. However several tests would fail on my local machine.

Any pointers to where I can read up on what I need to do to get the tests on `HEAD` to pass?

mereghost:
@kaikuchn You need a `phantomjs` binary on you `PATH` to make all the tests pass.

Which ones are failing?

kaikuchn:
```
Failed examples:

rspec ./spec/isolation/components/all/component_spec.rb:2 # Components: all loads all
```
Here's a [gist](https://gist.github.com/kaikuchn/c2274d20448748a0e2d22e1cf6d9cbcc) containing all the output from running `./script/ci`

phantomjs is on my path:
```bash
> which phantomjs
/usr/local/bin/phantomjs
> phantomjs --version
2.1.1
```

kaikuchn:
@mereghost or @AlfonsoUceda any ideas what might be wrong?

mereghost:
@kaikuchn =/ I was unable to reproduce, even against your fork.

kaikuchn:
Tracking IO I can see that the generate command never generates anything (at least it does not write to disk!) on my system. I'm on maxOS Sierra 10.12.2.. maybe it doesn't work for OS X?
Anyway I'll just do this on my linux system this weekend and get back to you.
Anyway, thank you for your help but I'll just give up fixing this on my OS X device.

jodosha:
@kaikuchn Can you just please paste here as a diff your test? I'll apply it to this PR.

kaikuchn:
Hey, sorry for not getting back to you guys for weeks!
This [gist](https://gist.github.com/kaikuchn/a42b8efb56bce721b871770be1c79c12) contains a test that might fail on the current master branch's head and pass with my changes. It's hypothetical though.. I've also added this spec to this PR after rebasing onto the current head of hanami/hanami.

jodosha:
@kaikuchn I verified the PR, it works great. Thank you very much! FYI have a look at https://github.com/hanami/hanami/commit/415c8fdbe6800402fabe17f75f00617c0bd9957d to see how I made it to work.

kaikuchn:
Awesome! Thank you for helping out!

PR #705, opened 2016-12-26
Setup Hanami::Mailer as a component

### LINKED ISSUE #703: Hanami does not configure or load the hanami-mailer as one of its components
It seems since we released the 0.9 series, using a component model for loading dependencies that `hanami-mailer` stayed out of the load scheme.

It can be seem by creating a new app, generating a mailer and trying to deliver it. By default it should use `Mail::TestMailer` but it tries to use the `smtp` delivery method.

mereghost:
`hanami-mailer` was only called by the `Glue` module and it would not load the configurations (no calls to `.load!`) passed to the `Hanami::Application`. A side effect created by this was that the mailer root was setup, so all calls to `.deliver` would fail.

This PR fixes #703 by setting `hanami-mailer` as a component and loading it. There's still a weird behaviour that the global config is not reflected on the mailers, but that's a separate issue with `hanami-mailer` (working on a fix).

### DIFF
--- a/spec/isolation/components/mailer/with_hanami_mailer_spec.rb
+++ b/spec/isolation/components/mailer/with_hanami_mailer_spec.rb
@@ -1,5 +1,5 @@
 RSpec.describe "Components: mailer", type: :cli do
-  context "with hanami-model" do
+  context "with hanami-mailer" do
     it "resolves mailer" do
       with_project do
         require Pathname.new(Dir.pwd).join("config", "environment")
--- a/spec/isolation/components/mailer/with_hanami_mailer_spec.rb
+++ b/spec/isolation/components/mailer/with_hanami_mailer_spec.rb
@@ -1,0 +1,13 @@
+RSpec.describe "Components: mailer", type: :cli do
+  context "with hanami-model" do
+    it "resolves mailer" do
+      with_project do
+        require Pathname.new(Dir.pwd).join("config", "environment")
+        Hanami::Components.resolve('mailer')
+
+        expect(Hanami::Components['mailer']).to               be(true)
+        expect(Hanami::Components['mailer.configuration']).to be_truthy
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


mereghost:
@AlfonsoUceda I updated the issue, it's not very trivial to write a test for that (at least I could not fathom how to do that) specific case. I updated the issue with a more detailed description of the bug.

That exposes a kinda weird behaviour on `Hanami::Mailer`. It sets `Hamami::Mailer.configuration` at include time, so if you required your mailer first and them configured the lib (exactly how it works on `hanami`), while the mailer will work because it re-reads the global state, the `YourMailer.configuration` is always the default one which IMO defies the POLS.

spec/isolation/components/mailer/with_hanami_mailer_spec.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.describe "Components: mailer", type: :cli do
+  context "with hanami-model" do
```
AlfonsoUceda: hanami-model? :P

AlfonsoUceda:
Forget what I said, as you pointed me out, hanami mailer isn't a soft dependency in hanami so it's fine :)

mereghost:
The failing spec is about migrations. =/

I'll merge this, since the error seems unrelated.

jodosha:
@mereghost First of all, thanks for this PR. But..

> I'll merge this, since the error seems unrelated.

This creates a dangerous trend here: broken windows. If this happens in the future, please hold on with merging the PR. Fix master or ask somebody else to do so, and only then merge.

You can't never tell if it's unrelated or not, until you haven't proven so with a fix in master.

Does it makes sense?

mereghost:
@jodosha perfect sense. Well, I can revert it.

The issue on Travis is a migration text difference (model is issuing an `autoincrement` keyword that the test does not expect) that I'm failing to reproduce.

jodosha:
@mereghost No, don't revert it. It doesn't make sense, master is broken. The goal is to fix that. :wink:

mereghost:
So everyone is informed: #708 fixes the broken master.

PR #707, opened 2016-12-28
Remove bundler dependency

sakuro:
In a Gemfile generate by `hanami new` contains a line

```ruby
gem 'bundler'
```

I think it is redundant.

Since Gemfiles are intended to be interpreted by bundler itself, when it can tell you the project depends on bundler, you already have it.

Consider removing this line from generated apps.

### DIFF
--- a/lib/hanami/generators/application/app/Gemfile.tt
+++ b/lib/hanami/generators/application/app/Gemfile.tt
@@ -1,6 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'bundler'
 gem 'rake'
 <%- if config[:hanami_head] -%>
 
--- a/lib/hanami/generators/application/container/Gemfile.tt
+++ b/lib/hanami/generators/application/container/Gemfile.tt
@@ -1,6 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'bundler'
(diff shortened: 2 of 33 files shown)


cllns:
Thanks for opening this @sakuro. 

`hanami` depends on `bundler`, so I think this is a good change. Can anyone think of any issues this could cause?

jodosha:
@sakuro Thanks for this PR. :+1: 

Ref Twitter thread: https://twitter.com/sakuro/status/813953428254760960

jodosha:
`hanami` depends on `bundler`. We declare this dependency in `hanami.gemspec`. When a developer (Alice) installs Hanami via `gem install hanami`, Rubygems looks up to the deps declared in the gemspec and installs Bundler if missing.

---

When Alice generates a new project, she needs to run `bundle install`, which implies that at this point she **already has bundler installed**. There is no need to declare this dependency in the `Gemfile` as well.

---

If Bob (a coworker of Alice) needs to work on the same project, he has to install **both Ruby and Bundler** in order to run `bundle install` on that said project.

---

We have no way to install indirectly install Bundler, **unless** a developer is installing Hanami via `gem install hanami` and **then** they run `bundle install`.

PR #713, opened 2017-01-25
Fix code reloading with components

### LINKED ISSUE #702: Code reloading not work when repository file change
When I make changes to repository file, code reloading doesn't work.
I have to manually restart the server to load latest code.

hanami version: 0.9.1
env: development mode

jodosha:
Since Hanami 0.9, we have introduced the internal concept of component. A component is responsible to load a Hanami part. It could be the Model configuration, or the routes inspector. Components build a graph of dependencies which are triggered by `Hanami.boot`. This system is thread-safe and guarantees that once a component is loaded, it ignores subsequent attempt to load it.

This is great for production, where we want to ensure that after the boot process code and configurations can't be altered anymore.

---

Unfortunately, this introduced problems for development environment. This proposal relax a bit the component behavior when code reloading is activated.

Hanami still relies on `shotgun` to reload the code in `apps/`, but for the code in `lib/` (entities, repositories, mailers) we had to introduce an internal mechanism: `Hanami::Utils.reload!`.

It uses `Kernel.load` to force the Ruby VM to reload the code for a given file.

Each time a developer refresh the browser, Hanami recursively scans `lib/` to reload the code.

---

Closes #702 

/cc @hanami/issues @hanami/ecosystem for review.

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -90,6 +90,7 @@ module Hanami
   # @since 0.9.0
   # @api private
   def self.boot
+    Components.release if code_reloading?
     Components.resolve('all')
   end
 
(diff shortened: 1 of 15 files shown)


AlfonsoUceda:
@jodosha I tested in a new app, I made changes in a repository and I saw in the view without restarting the server ;)

One thing I'm wondering, I saw in this [line](https://github.com/hanami/hanami/pull/713/files#diff-84b2703c42b75893bd8873537cbcba12R41) and we only allow the lib folder, why not add a config for this? maybe an user could add more folders but I don't know an use case now hehe.

jodosha:
@AlfonsoUceda Let's stick for this convention for now.

PR #715, opened 2017-01-26
Do not overwrite shell declared ENV vars

### LINKED ISSUE #712: Hanami rewrite env variables from .env file
## The problem
When you have  EVN variable in `.env.*` file and try to set this variable from the shell you will have the variable value from the file.

## Reproduce
1. Create new hanami app
2. Add any variable to `.env.development` file (`echo "TEST_VALUE=" >> .env.development`)
3. Run console with new variable value  (`TEST_VALUE="TEST HERE" bundle exec hanami c`)
4. Call `ENV['TEST_VALUE']` in console (now it returns empty sting)

/cc @hanami/core

gizotti:
As mentioned on #712, if an `ENV` var, with the same name as one declared on `.env.*` is declared on the shell when calling a hanami command, we will end up with the value declared on `.env.*, which is not the desired outcome.

This PR changes the initialization of `Hanami::Environment` to ensure that the `ENV` vars declared by the user on the shell are not overwritten when `.env.*` is loaded.

While my solutions works, there are some I would like to discuss if it is the best solution or not. 

My other solution was to change `Hanami::Env#load!` to only load the variables that yet not loaded on the instance. Now that I am writing this PR and thinking better, this is actually a better solution then the actual, but I wait for some feedback before I act on it.

Closes #712

### DIFF
--- a/lib/hanami/env.rb
+++ b/lib/hanami/env.rb
@@ -59,6 +59,8 @@ module Hanami
       parsed   = Dotenv::Parser.call(contents)
 
       parsed.each do |k, v|
+        next if @env.has_key?(k)
+
         @env[k] = v
       end
       nil
--- a/lib/hanami/environment.rb
+++ b/lib/hanami/environment.rb
@@ -193,6 +193,10 @@ module Hanami
     #   # is "development". The settings defined in this last file override
     #   # the one defined in the parent (eg `FOO` is overwritten). All the
     #   # other settings (eg `XYZ`) will be left untouched.
+    #   # Variables declared on `.env` and `.env.development` will not override
+    #   # any variable declared on the shell when calling a `hanami` command.
+    #   # Eg. In `FOO="not ok" bundle exec hanami c` `FOO` will not be overwritten
+    #   # to `"ok"`.
     def initialize(options = {})
       opts     = options.to_h.dup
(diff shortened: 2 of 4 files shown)


jodosha:
@gizotti Hey Gabriel thanks for this PR.

Probably delegating this responsibility to `Hanami::Env` is better because this accidental override happens only when `.env.*` files are loaded.

Can you please verify it it's possible? Thanks! 👍

gizotti:
@jodosha no worries at all.

I was just having a look at it and it's as simple as skipping the assigning of the variable on `Hanami::Env#load!` if that variable is already loaded in the instance.
I can add an `allow_overwrite` flag to the `#load!` method, but I am not sure if it's that important right now.

jodosha:
@gizotti 
> I can add an allow_overwrite flag to the #load! method, but I am not sure if it's that important right now.

I avoid boolean flags in methods because if you look closer they are a symptom of a SRI violation. A method supposed to do X, unless the flag is true, then it does Y. That means too many responsibilities for a method. 😉 

The PR is good as it is, merged. Thank you very much! 👍

gizotti:
@jodosha 

> I avoid boolean flags in methods because if you look closer they are a symptom of a SRI violation. A method supposed to do X, unless the flag is true, then it does Y. That means too many responsibilities for a method. 😉

Never really thought about it this way. Thanks for the feedback!

PR #716, opened 2017-01-27
Hanami destroy integration tests

jodosha:
@hanami/issues @hanami/ecosystem for review


AlfonsoUceda:
tested and it works but as you pointed out in tests, if you run destroy command for unknown migration/action/app then the command doesn't "fail", I mean always return the output saying we removed the files.

jodosha:
@AlfonsoUceda Yeah, that was marked as pending.

PR #717, opened 2017-01-27
Adds the right url for postgres on JRuby.

aemadrid:
Currently when generating a new app with `hanami new bookshelf --database=postgres` the default database url is `DATABASE_URL="jdbc:postgres://localhost/bookshelf_development"` when it should be `DATABASE_URL="jdbc:postgresql://localhost/bookshelf_development"`. 

The reason is that Sequel requires the `jdbc:postgresql` database adapter instead of `jdbc:postgres`.

### DIFF
--- a/lib/hanami/application_configuration.rb
+++ b/lib/hanami/application_configuration.rb
@@ -1398,7 +1398,7 @@ module Hanami
     #   module Bookshelf
     #     class Application < Hanami::Application
     #       configure do
-    #         adapter       :sql, 'postgres://localhost/database'
+    #         adapter       :sql, 'postgresql://localhost/database'
     #         model.adapter :sql, 'sqlite://db/bookshelf_development'
     #       end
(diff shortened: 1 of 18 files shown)


jodosha:
@aemadrid Hi and thanks for this PR. Does `"postgresql"` scheme works for MRI too? If so, we can avoid that `if` conditional and use `"postgresql"` for all the Ruby implementations.

AlfonsoUceda:
@jodosha I tested in MRI and it works with `postgresql`, @aemadrid can you remove the `if` and append `ql`? thanks!

aemadrid:
It's done.

AlfonsoUceda:
@aemadrid please fix the build and I'll merge the PR

aemadrid:
It's finally done. Found all the places where it was using `postgres:` and changed them to `postgresql:`.

PR #718, opened 2017-01-28
Remove old tests (Minitest)

jodosha:
As of 0.9.0 we switched from Minitest to RSpec for our tests.

I left over that old tests as reference. Now they are safe to be removed.

---

/cc @hanami/issues @hanami/ecosystem for review.


AlfonsoUceda:
👍 16k lines :D

cllns:
` +0 −16,095 ` now that's what I like to see!

jodosha:
For the record, these 16k LOCs aren't all tests. The number is big, cause it includes tons of fixtures. Each fixture was a full Hanami project.

As of 0.9 we use a different strategy: instead of maintaining all these fixtures, we let the single spec to generate a Hanami project at the runtime. Read more at: http://lucaguidi.com/2017/01/20/how-to-test-ruby-cli-the-setup/

PR #720, opened 2017-01-30
Generate new projects with config/boot.rb

### LINKED ISSUE #695: config/environment.rb not requiring initializers nor files in load paths
While setting up hanami + sidekiq, I came across two hurdles:
+ requiring `config/environment.rb` does not load initializers (which may or may not be the right thing to do, but since sidekiq does not boot the app, but simply requires that file, it forces one to require the sidekiq config directly from `config/environment`)
+ requiring `config/environment.rb` does not load files in the load paths declared within apps.

@jodosha Any thoughts on this?

jodosha:
Closes #695 

---

/cc @beauby @hanami/issues @hanami/ecosystem for review. 👍

### DIFF
--- a/lib/hanami/commands/new/container.rb
+++ b/lib/hanami/commands/new/container.rb
@@ -43,6 +43,7 @@ module Hanami
           add_mapping('.env.test.tt', '.env.test')
           add_mapping('Gemfile.tt', 'Gemfile')
           add_mapping('config.ru.tt', 'config.ru')
+          add_mapping('config/boot.rb.tt', 'config/boot.rb')
           add_mapping('config/environment.rb.tt', 'config/environment.rb')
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
You were quick haha, LGTM but I think someone using sidekiq could try this.

jodosha:
@AlfonsoUceda This was already on my list. 😉 

We already had feedback from devs using Sidekiq when we were suggesting this patch.

@davydovanton Can you please confirm too? Thanks.

jodosha:
@davydovanton Something like?

davydovanton:
@jodosha maybe something like this?
```
# you can use this file for boot hanami apps in other services like sidekiq, etc
```

jodosha:
@davydovanton I'd go for documenting it. By the time that you discover that file, you already figured out. 😉

PR #721, opened 2017-01-30
Use SMTP_HOST env in address instead SMTP_PORT for mailer config

artofhuman:
* Use SMTP_PORT env for port configuration

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -52,7 +52,7 @@ module Hanami
   #       delivery do
   #         development :test
   #         test        :test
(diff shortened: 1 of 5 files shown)


cllns:
Should `SMTP_PORT` and `SMTP_HOST` be in the generated `.env.development` and `.env.test` files?

artofhuman:
@cllns Now we use env only for production. In tests and development value is `:test`.

jodosha:
@artofhuman Thanks for this PR. It makes sense. Merged 👍

PR #724, opened 2017-02-01
Add Hanami::Configuration#environment for environment specific settings

jodosha:
Add Hanami::Configuration#environment for environment specific settings. Add different logger settings according to the current env.

With this proposal, the generated `config/environment.rb` will look like this:

```ruby
require 'bundler/setup'
require 'hanami/setup'
require 'hanami/model'

Hanami.configure do

  model do
    ##
    # Database adapter
    #
    # Available options:
    #
    #  * SQL adapter
    #    adapter :sql, 'sqlite://db/bookshelf_development.sqlite3'
    #    adapter :sql, 'postgresql://localhost/bookshelf_development'
    #    adapter :sql, 'mysql://localhost/bookshelf_development'
    #
    adapter :sql, ENV['DATABASE_URL']

    ##
    # Migrations
    #
    migrations 'db/migrations'
    schema     'db/schema.sql'
  end

  mailer do
    root 'lib/bookshelf/mailers'

    delivery :test
  end

  environment :development do
    # See: http://hanamirb.org/guides/projects/logging
    logger level: :debug
  end

  environment :production do
    logger level: :info, formatter: :json

    mailer do
      delivery :smtp, address: ENV['SMTP_HOST'], port: ENV['SMTP_PORT']
    end
  end
end
```

@hanami/issues @hanami/ecosystem I have one doubt about the `test` environment. Do we want to have a logger redirected to `log/test.log` or do we want to skip logging at all? Please let me know what you think. Thanks!

### DIFF
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -17,7 +17,6 @@ module Hanami
 
     def initialize(&blk)
       @settings = Concurrent::Map.new
-      # @settings[:environments] = Concurrent::Hash.new { |h, k| h[k] = [] }
       instance_eval(&blk)
     end
 
@@ -49,14 +48,13 @@ module Hanami
 
     def logger(options = nil)
       if options.nil?
-        settings.fetch(:logger)
+        settings.fetch(:logger, nil)
       else
         settings[:logger] = options
       end
     end
 
     def environment(name)
-      # settings[:environments][name.to_sym] << blk
       yield if ENV['HANAMI_ENV'] == name.to_s
     end
 
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -17,6 +17,7 @@ module Hanami
 
     def initialize(&blk)
       @settings = Concurrent::Map.new
+      # @settings[:environments] = Concurrent::Hash.new { |h, k| h[k] = [] }
       instance_eval(&blk)
     end
 
@@ -50,8 +51,13 @@ module Hanami
       if options.nil?
         settings.fetch(:logger)
       else
-        settings.put_if_absent(:logger, options)
+        settings[:logger] = options
       end
+    end
+
+    def environment(name)
+      # settings[:environments][name.to_sym] << blk
+      yield if ENV['HANAMI_ENV'] == name.to_s
     end
 
     private
--- a/lib/hanami/commands/new/container.rb
+++ b/lib/hanami/commands/new/container.rb
@@ -64,6 +64,7 @@ module Hanami
 
         def add_empty_directories
           add_mapping('.gitkeep', 'public/.gitkeep')
+          add_mapping('.gitkeep', 'log/.gitkeep')
           add_mapping('.gitkeep', 'config/initializers/.gitkeep')
           add_mapping('.gitkeep', "lib/#{ project_name }/entities/.gitkeep")
           add_mapping('.gitkeep', "lib/#{ project_name }/repositories/.gitkeep")
--- a/lib/hanami/commands/new/container.rb
+++ b/lib/hanami/commands/new/container.rb
@@ -64,7 +64,6 @@ module Hanami
 
         def add_empty_directories
           add_mapping('.gitkeep', 'public/.gitkeep')
-          add_mapping('.gitkeep', 'log/.gitkeep')
           add_mapping('.gitkeep', 'config/initializers/.gitkeep')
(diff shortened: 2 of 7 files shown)


davydovanton:
@jodosha I prefer to skip logger for test env. You can use `puts` for debugging if you want. Also, all CIs have web interface with all tests output

lib/hanami/configuration.rb
```diff
@@ -17,6 +17,7 @@ def initialize(app, path_prefix)
 
     def initialize(&blk)
       @settings = Concurrent::Map.new
+      # @settings[:environments] = Concurrent::Hash.new { |h, k| h[k] = [] }
```
davydovanton: maybe is better to drop this line?

lib/hanami/configuration.rb
```diff
@@ -50,10 +51,15 @@ def logger(options = nil)
       if options.nil?
         settings.fetch(:logger)
       else
-        settings.put_if_absent(:logger, options)
+        settings[:logger] = options
       end
     end
 
+    def environment(name)
+      # settings[:environments][name.to_sym] << blk
```
davydovanton: and drop this line too?

lib/hanami/configuration.rb
```diff
@@ -50,10 +51,15 @@ def logger(options = nil)
...
+      yield if ENV['HANAMI_ENV'] == name.to_s
```
davydovanton: Just a question: can we use `Hanami.env?(name.to_sym)` here?

mereghost: I don't think so. The `Hanami::Environment` is created after all the components are resolved and that happens prior to that.

mereghost:
I'd skip logging during tests as @davydovanton said and just because it's another unecessary IO slowing my tests. ;p

mereghost:
(Probably dumb) question: if it's just for the `Logger`, why not add a `logger` block? 

Do we intend to be able to change other settings in them (like the connection string of the `sql` adapter) or that would be better on the blocks belonging to each library (kinda like `mailer` already do with delivery methods)?

AlfonsoUceda:
LGTM 👍 

@mereghost I think that is the reason.

lib/hanami/configuration.rb
jodosha: Yes @mereghost is right.

lib/hanami/configuration.rb
jodosha: Sorry this is a leftover. Yes it should be removed.

jodosha:
@mereghost Good point. I'm still not sure which API is the best in the long term. Probably to have the blocks per environment, like proposed in this PR.

---

A bit of background.

Model uses env vars because the value that changes for each environment is the connection string. This eliminates the need of having env blocks.

Mailer uses env settings because it arrived later to the party and I still didn't wanted to add env blocks.

---

Now it's probably the time to have env blocks as first citizens and generate something like this:


```ruby
Hanami.configure do

  # See: http://hanamirb.org/guides/projects/logging
  logger level: :info

  model do
    adapter :sql, ENV['DATABASE_URL']

    migrations 'db/migrations'
    schema     'db/schema.sql'
  end

  mailer do
    root Hanami.root.join("lib", "bookshelf", "mailers")
    delivery :test
  end

  environment :production do
    logger level: :info, formatter: :json

    mailer do
      delivery :smtp, address: ENV['SMTP_HOST'], port: ENV['SMTP_PORT']
    end
  end
end
```

# or

```ruby
Hanami.configure do

  # See: http://hanamirb.org/guides/projects/logging
  logger level: :info

  model.adapter :sql, ENV['DATABASE_URL']
  model.migrations 'db/migrations'
  model.schema     'db/schema.sql'

  mailer.root Hanami.root.join("lib", "bookshelf", "mailers")
  mailer.delivery :test

  environment :production do
    logger level: :info, formatter: :json

    mailer.delivery :smtp, address: ENV['SMTP_HOST'], port: ENV['SMTP_PORT']
  end
end
```

@hanami/issues @hanami/ecosystem WDYT?

mereghost:
While the later seems cool, I do like the block style of the first example (and I think it would require a little less "magic" to make it work :P )

AlfonsoUceda:
I prefer the first one!

jodosha:
Folks, I updated the code and the description of this PR, as requested by you. Please review again 😄 .

AMHOL:
Think it's worth noting somewhere the default configurations will need to be defined before environment specific configurations, otherwise the defaults will overwrite the environment specific configs.

Issue #729, opened 2017-02-14
It will be nice to comment that 'unsafe-eval' needed for vue.js in application.rb


mytram:
Hi, 

I was trying to use vue.js with hanami. I basically copied over built js files for hanami to serve them. However, at the beginning, it appeared as if the vuejs scripts stopped half way mysteriously whilst the same approach would work with rails 5. After some digging, I found that rails didn't have the content-security-policy header by default compared with hanami. 

I found out that 'unsafe-eval' was needed for vuejs to do its work. I will be nice to add some comments in the apps/web/application.rb below: 

```  
 # Inline JavaScript is NOT allowed. To enable it, please use:
 # "script-src 'unsafe-inline'".

```
For instance, 

```
# For some javascript frameworks, such as vuejs, 'unsafe-eval' needs to be enabled for eval(), function() to work. 
# Here are the details: 
# https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
```

Just a suggestion that may save other people's time. 

Many thanks, 

Mytram.

davydovanton:
@mytram hey, could you update issue? :)

mytram:
@davydovanton Hi, I was just adding. accidentally hit some key before I finished :) Thanks

davydovanton:
@mytram good proposal. I [added](https://github.com/ossboard-org/ossboard/blob/master/apps/web/application.rb#L234) this policy too (and spend some time for this) 👍

Could you create PR with updated comments? 😊

mytram:
@davydovanton , no problems. Will do. Thanks

jodosha:
@mytram I discovered that this is already documented. Please see https://github.com/hanami/hanami/pull/731#issuecomment-283114410

cllns:
@jodosha This and #731 should not have been closed. There's [a difference between `unsafe-inline` and `unsafe-eval`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). `unsafe-inline` allows inline scripts, but not **not** `eval` (and similar) functions. `unsafe-eval` allows inline function to use `eval` and similar.

[Vue.js uses `new Function()`](https://vuejs.org/v2/guide/installation.html#CSP-environments), which is a `eval`-type function.

jodosha:
[...]

PR #731, opened 2017-02-16
Add comments that 'unsafe-eval' needed for vue.js to work in application.rb

mytram:
Documentation improvement (#729) Thanks, -Mytram

### DIFF
--- a/lib/hanami/generators/app/application.rb.tt
+++ b/lib/hanami/generators/app/application.rb.tt
@@ -205,6 +205,12 @@ module <%= config[:classified_app_name] %>
       # Inline JavaScript is NOT allowed. To enable it, please use:
       # "script-src 'unsafe-inline'".
       #
+      # Javascript eval() and similar methods are NOT allowed. It needs to be enabled
+      # for Javascript frameworks that rely on
+      # such features, such as vue.js. To enable it, please use:
+      # "script-src unsafe-eval'".
+      #
+      #
       # Content Security Policy introduction:
       #
       #  * http://www.html5rocks.com/en/tutorials/security/content-security-policy/


lib/hanami/generators/app/application.rb.tt
```diff
@@ -205,6 +205,12 @@ module <%= config[:classified_app_name] %>
       # Inline JavaScript is NOT allowed. To enable it, please use:
       # "script-src 'unsafe-inline'".
       #
+      # Javascript eval() and similar methods are NOT allowed. It needs to be enabled
+      # for Javascript frameworks that rely on
+      # such features, such as vue.js. To enable it, please use:
+      # "script-src unsafe-eval'".
```
cllns: What do you think about wording this as:
```
      # Inline JavaScript is NOT allowed. To enable it, please use:
      # "script-src 'unsafe-inline'".
      #
      # If you need to use JavaScript's eval() and similar functions, 
      # which some frameworks require, please use:
      # "script-src unsafe-eval'".
```
?

I'd prefer to stay away from mentioned Vue.js specifically. The place to handle that is in a guide for Vue.js + Hanami (which is planned).

davydovanton: @cllns what do you think about this:

```
# Some js frameworks (Vue, react, etc) use inline JavaScript
# And hanami don't allow it bu default. To enable it, please use:
# "script-src 'unsafe-inline'".
```

cllns: I don't know much about how JS MVC frameworks work in the browser. But there's `unsafe-inline` and `unsave-eval`. 

Vue needs the latter (according to @mytram), but I'm not sure about the others :)

davydovanton:
@mytram thanks for PR! 💛 
Could you fix @cllns issue and update tests?

lib/hanami/generators/app/application.rb.tt
jodosha: I'd avoid a direct mention to Vue.js as well.

jodosha:
I've just noticed that we already have documented this issue here: https://github.com/hanami/hanami/blob/master/lib/hanami/generators/app/application.rb.tt#L205-L206

```ruby
      # Inline JavaScript is NOT allowed. To enable it, please use:
      # "script-src 'unsafe-inline'".
```

The changes of this PR will be redundant. I'm closing this. Thank you @mytram

PR #737, opened 2017-02-27
Remove architecture flag from CLI and .hanamirc

cllns:
We used to support a `container` architecture (default) and an `application` architecture. 
Now we only support what was called `container`, which we refer to as a a Hanami "project".

I'm not sure if this is ready to merge. 
Basically I'm not sure if I went too far, or not far enough in removing references to `container`. 
It's still a useful concept in the back-end, but perhaps it should be renamed to 'project'. I'm also not sure if we want to leave some of the configuration options around to make it easy for people to add their own architectures later (mostly if someone wants to add `application` architecture back).

Thoughts?

### DIFF
--- a/lib/hanami/cli.rb
+++ b/lib/hanami/cli.rb
@@ -88,16 +88,15 @@ module Hanami
     desc 'new PROJECT_NAME', 'Generate a new hanami project'
     long_desc <<-EOS
 `hanami new` creates a new hanami project.
-You can specify various options such as the database to be used as well as the path and architecture.
+You can specify various options, like which database type, testing framework or templating engine the project will use.
 
 $ > hanami new fancy_app --application_name=admin
 
-$ > hanami new fancy_app --arch=app
+$ > hanami new fancy_app --database=postgresql
 
 $ > hanami new fancy_app --hanami-head=true
     EOS
     method_option :database, aliases: ['-d', '--db'], desc: "Application database (#{Hanami::Generators::DatabaseConfig::SUPPORTED_ENGINES.keys.join('/')})", default: Hanami::Generators::DatabaseConfig::DEFAULT_ENGINE
-    method_option :architecture, aliases: ['-a', '--arch'], desc: 'Project architecture (container/app)', default: Hanami::Commands::New::Abstract::DEFAULT_ARCHITECTURE
     method_option :application_name, desc: 'Application name, only for container', default: Hanami::Commands::New::Container::DEFAULT_APPLICATION_NAME
     method_option :application_base_url, desc: 'Application base url', default: Hanami::Commands::New::Abstract::DEFAULT_APPLICATION_BASE_URL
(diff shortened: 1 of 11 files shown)


lib/hanami/environment.rb
```diff
@@ -420,17 +408,8 @@ def code_reloading?
...
-        puts "Cannot recognize Hanami architecture, please check `.hanamirc'"
-        exit 1
-      end
-    end
-
-    # @since 0.4.0
-    # @api private
     def container?
-      architecture == CONTAINER
+      true
```
cllns: This is the smelliest part of the code. We only support CONTAINER, so I just made it return true.

I could go further and remove all references to `.container?` but there are quite a few and it involves several different sections of the codebase.

davydovanton:
Sould we drop all app architecture generators too?

mereghost:
@davydovanton Already droped.

lib/hanami/environment.rb
mereghost: I'd say go for it, specially as it simplifies the whole lib.

lib/hanami/cli.rb
```diff
@@ -88,16 +88,15 @@ def console
...
 $ > hanami new fancy_app --application_name=admin
 
-$ > hanami new fancy_app --arch=app
+$ > hanami new fancy_app --database=postgresql
 
 $ > hanami new fancy_app --hanami-head=true
     EOS
     method_option :database, aliases: ['-d', '--db'], desc: "Application database (#{Hanami::Generators::DatabaseConfig::SUPPORTED_ENGINES.keys.join('/')})", default: Hanami::Generators::DatabaseConfig::DEFAULT_ENGINE
-    method_option :architecture, aliases: ['-a', '--arch'], desc: 'Project architecture (container/app)', default: Hanami::Commands::New::Abstract::DEFAULT_ARCHITECTURE
     method_option :application_name, desc: 'Application name, only for container', default: Hanami::Commands::New::Container::DEFAULT_APPLICATION_NAME
```
ixti: As there are no more non-container arch, it worth to leave documentation of this options simply:

``` ruby
method_option :application_name, desc: 'Application name'. default: ...
```

AlfonsoUceda: I agree, @cllns could you take care of it?

mereghost:
@cllns Any progress on this?

jodosha:
@cllns As soon you'll resolve the conflicts, please merge it.

jodosha:
@cllns I don't know if you have already noticed it but there are still references to container:

```shell
(cllns-remove-architecture-flags) hanami ➜ grep -inr container lib | wc -l
      29
```

cllns:
Yea, there's a bit more to do on this. The current state is a pretty superficial change (see `container?` returning true above). Per our conversation last week, I'll go through and fully remove references to the different architectures

mereghost:
@cllns any news on this?

AlfonsoUceda:
Folks, I'm going to close it because with the new CLI, we won't support architecture flag, thanks anyway ;)

cllns:
thanks @AlfonsoUceda! I'm going to go through and clean up the references to architectures (`container`/`app`) in the back-end, once #808 is merged

AlfonsoUceda:
thanks @cllns ;)

PR #741, opened 2017-03-05
Make application.rb conform to 80 char line length

### LINKED ISSUE #734: Generated `application.rb` file has long lines
Generating a new app and running ` rubocop --only Metrics/LineLength` (which checks for lines greater than 80 characters) on a new Hanami app turns up 13 offenses,  all of which are in the generated `application.rb` file. In general, we should make an attempt to keep lines to 80 characters or less, regardless of rubocop.

`rubocop --autocorrect` doesn't fix this offense, so it's something we should handle when generating the project, in my opinion.

```
apps/web/application.rb:16:81: C: Line is too long. [82/80]
      # Relative load paths where this application will recursively load the code.
                                                                                ^^
apps/web/application.rb:50:81: C: Line is too long. [107/80]
      # Argument: An object coercible to integer, default to 80 if the scheme is http and 443 if it's https
                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
apps/web/application.rb:51:81: C: Line is too long. [97/80]
      # This SHOULD be configured only in case the application listens to that non standard ports
                                                                                ^^^^^^^^^^^^^^^^^
apps/web/application.rb:60:81: C: Line is too long. [89/80]
      #          :path     - Restrict cookies to a relative URI (String - nil by default)
                                                                                ^^^^^^^^^
apps/web/application.rb:61:81: C: Line is too long. [95/80]
      #          :max_age  - Cookies expiration expressed in seconds (Integer - nil by default)
                                                                                ^^^^^^^^^^^^^^^
apps/web/application.rb:63:81: C: Line is too long. [105/80]
      #                      (Boolean - Automatically set on true if currently using a secure connection)
                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^
apps/web/application.rb:65:81: C: Line is too long. [82/80]
      #          :httponly - Prevent JavaScript access (Boolean - true by default)
                                                                                ^^
apps/web/application.rb:88:81: C: Line is too long. [86/80]
      # Default format for responses that doesn't take into account the request format
                                                                                ^^^^^^
apps/web/application.rb:95:81: C: Line is too long. [97/80]
      # Argument: Symbol, which represent the format of the mime type (only `:json` is supported)
                                                                                ^^^^^^^^^^^^^^^^^
apps/web/application.rb:101:81: C: Line is too long. [81/80]
      # it redirects to the secure equivalent resource (https). Default disabled.
                                                                                ^
apps/web/application.rb:183:81: C: Line is too long. [82/80]
      # X-XSS-Protection is a HTTP header to determine the behavior of the browser
                                                                                ^^
apps/web/application.rb:193:81: C: Line is too long. [84/80]
      # Content-Security-Policy (CSP) is a HTTP header supported by modern browsers.
                                                                                ^^^^
apps/web/application.rb:201:81: C: Line is too long. [83/80]
      # The default value allows images, scripts, AJAX, fonts and CSS from the same
```

Could be a neat to make an integration test for making sure new applications don't have any line over 80 characters.

alcesleo:
Honestly, I find some of these changes to make the formatting look slightly worse, but it does pass the default RuboCop `Metrics/LineLength` check mentioned in #734. Please have a look if this is okay, or if there's a nicer way of formatting it that still passes this check - if not I'm also okay with closing the PR. :) 

Closes #734

### DIFF
--- a/lib/hanami/generators/app/application.rb.tt
+++ b/lib/hanami/generators/app/application.rb.tt
@@ -13,7 +13,9 @@ module <%= config[:classified_app_name] %>
       #
       root __dir__
 
-      # Relative load paths where this application will recursively load the code.
+      # Relative load paths where this application will recursively load the
+      # code.
+      #
       # When you add new directories, remember to add them here.
       #
       load_paths << [
@@ -47,8 +49,10 @@ module <%= config[:classified_app_name] %>
       # host 'example.org'
 
       # URI port used by the routing system to generate absolute URLs
-      # Argument: An object coercible to integer, default to 80 if the scheme is http and 443 if it's https
-      # This SHOULD be configured only in case the application listens to that non standard ports
+      # Argument: An object coercible to integer, defaults to 80 if the scheme
+      # is http and 443 if it's https
+      #
+      # This should only be configured if app listens to non-standard ports
       #
       # port 443
 
@@ -56,13 +60,16 @@ module <%= config[:classified_app_name] %>
       # Argument: boolean to toggle the feature
       #           A Hash with options
       #
-      # Options: :domain   - The domain (String - nil by default, not required)
-      #          :path     - Restrict cookies to a relative URI (String - nil by default)
-      #          :max_age  - Cookies expiration expressed in seconds (Integer - nil by default)
-      #          :secure   - Restrict cookies to secure connections
-      #                      (Boolean - Automatically set on true if currently using a secure connection)
-      #                      See #scheme and #ssl?
-      #          :httponly - Prevent JavaScript access (Boolean - true by default)
+      # Options:
+      #   :domain   - The domain (String - nil by default, not required)
+      #   :path     - Restrict cookies to a relative URI
+      #               (String - nil by default)
+      #   :max_age  - Cookies expiration expressed in seconds
+      #               (Integer - nil by default)
+      #   :secure   - Restrict cookies to secure connections
+      #               (Boolean - Automatically true when using HTTPS)
+      #               See #scheme and #ssl?
+      #   :httponly - Prevent JavaScript access (Boolean - true by default)
       #
       # cookies true
       # or
@@ -81,24 +88,25 @@ module <%= config[:classified_app_name] %>
       # middleware.use Rack::Protection
 
       # Default format for the requests that don't specify an HTTP_ACCEPT header
-      # Argument: A symbol representation of a mime type, default to :html
+      # Argument: A symbol representation of a mime type, defaults to :html
       #
       # default_request_format :html
 
-      # Default format for responses that doesn't take into account the request format
-      # Argument: A symbol representation of a mime type, default to :html
+      # Default format for responses that don't consider the request format
+      # Argument: A symbol representation of a mime type, defaults to :html
       #
       # default_response_format :html
 
       # HTTP Body parsers
       # Parse non GET responses body for a specific mime type
-      # Argument: Symbol, which represent the format of the mime type (only `:json` is supported)
+      # Argument: Symbol, which represent the format of the mime type
+      #             (only `:json` is supported)
       #           Object, the parser
       #
       # body_parsers :json
 
       # When it's true and the router receives a non-encrypted request (http),
-      # it redirects to the secure equivalent resource (https). Default disabled.
+      # it redirects to the secure equivalent (https). Disabled by default.
       #
       # force_ssl true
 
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
@alcesleo thanks for the PR but you should amend the test ;)

alcesleo:
Will do, I was lazy and waited for CI 😀

alcesleo:
Before I go around and fix the specs, do you think this is actually worth merging? I feel like it compromises on the formatting a little bit to follow this arbitrary rule. Thoughts @AlfonsoUceda?

AlfonsoUceda:
@alcesleo to me it's ok, what do think @hanami/core

lib/hanami/generators/app/application.rb.tt
```diff
@@ -180,7 +193,9 @@ module <%= config[:classified_app_name] %>
       #
       security.x_content_type_options 'nosniff'
 
-      # X-XSS-Protection is a HTTP header to determine the behavior of the browser
+      # X-XSS-Protection is a HTTP header to determine the behavior of the
+      # browser
+      #
```
cllns: lines 197 thru 199 can be merged into a single line :)

lib/hanami/generators/app/application.rb.tt
```diff
@@ -47,22 +49,29 @@ module <%= config[:classified_app_name] %>
       # host 'example.org'
 
       # URI port used by the routing system to generate absolute URLs
-      # Argument: An object coercible to integer, default to 80 if the scheme is http and 443 if it's https
-      # This SHOULD be configured only in case the application listens to that non standard ports
+      # Argument: An object coercible to integer, default to 80 if the scheme
+      # is http and 443 if it's https
+      #
+      # This SHOULD be configured only in case the application listens to that
+      # non standard ports
```
cllns: 55 and 56 can combined this into one line by rewording to:
`     # This should only be configured if app listens to non-standard ports`

lib/hanami/generators/app/application.rb.tt
```diff
@@ -47,22 +49,29 @@ module <%= config[:classified_app_name] %>
...
-      #          :httponly - Prevent JavaScript access (Boolean - true by default)
+      # Options:
+      #   :domain   - The domain (String - nil by default, not required)
+      #   :path     - Restrict cookies to a relative URI
+      #               (String - nil by default)
+      #   :max_age  - Cookies expiration expressed in seconds
+      #               (Integer - nil by default)
+      #   :secure   - Restrict cookies to secure connections
+      #               (Boolean - Automatically set on true if currently using a
+      #               secure connection)
```
cllns: 71 and 72 can be combined into one line by rewording to:
`     #               (Boolean - Automatically true when using HTTPS)`

lib/hanami/generators/app/application.rb.tt
```diff
@@ -85,20 +94,24 @@ module <%= config[:classified_app_name] %>
       #
       # default_request_format :html
 
-      # Default format for responses that doesn't take into account the request format
+      # Default format for responses that doesn't take into account the request
+      # format
```
cllns: 97 and 98 can be combined into one line by rewording to:  
`    # Default format for responses that don't consider the request format`

lib/hanami/generators/app/application.rb.tt
```diff
@@ -85,20 +94,24 @@ module <%= config[:classified_app_name] %>
...
       # Argument: A symbol representation of a mime type, default to :html
       #
       # default_response_format :html
 
       # HTTP Body parsers
       # Parse non GET responses body for a specific mime type
-      # Argument: Symbol, which represent the format of the mime type (only `:json` is supported)
+      # Argument: Symbol, which represent the format of the mime type
+      #             (only `:json` is supported)
       #           Object, the parser
```
cllns: Can we clarify what this means? I interpret it means that you can pass an Object which parses the body, but it's not super clear

lib/hanami/generators/app/application.rb.tt
```diff
@@ -85,20 +94,24 @@ module <%= config[:classified_app_name] %>
...
       #
       # body_parsers :json
 
       # When it's true and the router receives a non-encrypted request (http),
-      # it redirects to the secure equivalent resource (https). Default disabled.
+      # it redirects to the secure equivalent resource (https).
+      # Default disabled.
```
cllns: 113 and 114 can be combined into one line by rewording to:
` # it redirects to the secure equivalent (https). Disabled by default.`
(dropping 'resource' because it redirects for all routes, not just RESTful/resourceful routes)

alcesleo:
Great suggestions @cllns, thanks! I've added new commit with pretty much exactly what you wrote, apart from the `Object` bit, not sure exactly how to clarify it, but it might also be out of scope for fixing the line lengths

cllns:
awesome, good work! I agree that `Object` bit is out-of-scope.

this is good to me once the tests pass :)

alcesleo:
I fixed all the specs and squashed it into a single commit :)

cllns:
thank you! 🌸

alcesleo:
My pleasure!

Issue #740, opened 2017-03-05
Removing mailer config causes showstopper exception


andrew:
Just started playing around with Hanami, making a small app that doesn't need to do any email, so I removed the `mailer` section from `config/environment.rb`, which looks like this (also removed the model section but hanami seems to work fine without that bit):

```ruby
require 'bundler/setup'
require 'hanami/setup'
require_relative '../lib/code_analyser'
require_relative '../apps/web/application'

Hanami.configure do
  mount Web::Application, at: '/'

  environment :development do
    # See: http://hanamirb.org/guides/projects/logging
    logger level: :debug
  end

  environment :production do
    logger level: :info, formatter: :json
  end
end
```

which resulted in following exception when loading the homepage:

```
Boot Error

Something went wrong while loading /Users/andrew/code/code_analyser/config.ru

ArgumentError: wrong number of arguments (given 0, expected 1..3)

/usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/hanami-mailer-1.0.0.beta1/lib/hanami/mailer.rb:61:in `instance_eval'
/usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/hanami-mailer-1.0.0.beta1/lib/hanami/mailer.rb:61:in `configure'
/usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/hanami-1.0.0.beta2/lib/hanami/components/components.rb:181:in `block (2 levels) in <module:Components>'
/usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/hanami-1.0.0.beta2/lib/hanami/components/component.rb:42:in `call'
/usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/hanami-1.0.0.beta2/lib/hanami/components.rb:85:in `block (2 levels) in resolve'
/usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/concurrent-ruby-1.0.5/lib/concurrent/map.rb:133:in `block in fetch_or_store'
/usr/local/var/rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/concurrent-ruby-1.0.5/lib/concurrent/map.rb:122:in `fetch'
[...]

PR #742, opened 2017-03-05
Ensure project to work without mailer configuration

davydovanton:
I tried to solve https://github.com/hanami/hanami/issues/740, and I think my patch will work. But I don't understand how I can test this case. @jodosha (@hanami/core) could you help me with tests?

### DIFF
--- a/lib/hanami/components/components.rb
+++ b/lib/hanami/components/components.rb
@@ -177,9 +177,11 @@ module Hanami
       end
 
       resolve do |configuration|
-        Hanami::Mailer.configuration = Hanami::Mailer::Configuration.new if Hanami.code_reloading?
-        Hanami::Mailer.configure(&configuration.mailer)
-        Hanami::Mailer.configuration
+        if configuration.mailer
+          Hanami::Mailer.configuration = Hanami::Mailer::Configuration.new if Hanami.code_reloading?
+          Hanami::Mailer.configure(&configuration.mailer)
+          Hanami::Mailer.configuration
+        end
       end
     end
 
--- a/lib/hanami/components/components.rb
+++ b/lib/hanami/components/components.rb
@@ -177,7 +177,7 @@ module Hanami
       end
 
       resolve do |configuration|
-        if configuration.mailer
+        unless configuration.mailer.nil?
           Hanami::Mailer.configuration = Hanami::Mailer::Configuration.new if Hanami.code_reloading?
           Hanami::Mailer.configure(&configuration.mailer)
           Hanami::Mailer.configuration
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -365,10 +365,10 @@ EOF
     end
   end
 
-  context "without email" do
+  context "without mailer" do
     it "returns page" do
       with_project do
-        replace "config/environment.rb", "delivery :test", ""
+        remove_block "config/environment.rb", "mailer do"
 
         server do
           visit "/"
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -364,4 +364,17 @@ EOF
       end
     end
   end
+
+  context "without email" do
+    it "returns page" do
+      with_project do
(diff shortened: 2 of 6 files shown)


lib/hanami/components/components.rb
```diff
@@ -177,9 +177,11 @@ module Components # rubocop:disable Metrics/ModuleLength
       end
 
       resolve do |configuration|
-        Hanami::Mailer.configuration = Hanami::Mailer::Configuration.new if Hanami.code_reloading?
-        Hanami::Mailer.configure(&configuration.mailer)
-        Hanami::Mailer.configuration
+        if configuration.mailer
```
AlfonsoUceda: @davydovanton I think should be like this line https://github.com/hanami/hanami/blob/9863c0949930f93155ed9362e2ea27bf732498c6/lib/hanami/components/components.rb#L105

AlfonsoUceda:
@davydovanton thanks for this PR, to test this you should use as foundation this spec https://github.com/hanami/hanami/blob/master/spec/isolation/components/model_configuration/without_hanami_model_spec.rb and add to this folder https://github.com/hanami/hanami/tree/master/spec/isolation/components/mailer_configuration

can you try this? ;)

lib/hanami/components/components.rb
mereghost: The trouble is that `mailer` is a hard dependency of `hanami`, unlike `model`.

If we ignore the configuration silently we may happen into issues complaining about it not working at all.

Either we make it a soft-dependency or put in a warn that strange stuff may happen.

davydovanton: I prefer to warn. @mereghost @AlfonsoUceda @jodosha WDYT?

jodosha:
@davydovanton Please try with @AlfonsoUceda's suggestion. Create a new test file like this: https://github.com/hanami/hanami/blob/master/spec/isolation/components/mailer_configuration/with_hanami_mailer_spec.rb named `without_hanami_mailer_spec.rb`.

davydovanton:
@AlfonsoUceda @jodosha thanks for help. I'll do it today 👍

spec/integration/cli/server_spec.rb
```diff
@@ -364,4 +364,17 @@ def listing
       end
     end
   end
+
+  context "without email" do
+    it "returns page" do
+      with_project do
+        replace "config/environment.rb", "delivery :test", ""
```
jodosha: @davydovanton I'm afraid this isn't enough. If you look at the original [bug report](https://github.com/hanami/hanami/issues/740#issue-211972961), there was more things removed. Specifically, all the `mailer` blocks were stripped from the original configuration. We should reproduce that scenario.

There are other examples here that show how to rewrite an entire file. That could be a solution.

spec/integration/cli/server_spec.rb
```diff
@@ -364,4 +364,17 @@ def listing
       end
     end
   end
+
+  context "without email" do
```
jodosha: s/without email/without malier/

spec/isolation/components/mailer_configuration/without_hanami_mailer_spec.rb
```diff
@@ -0,0 +1,15 @@
+RSpec.describe "Components: mailer.configuration", type: :cli do
+  context "without hanami-mailer" do
```
jodosha: s/without hanami-mailer/without mailer configuration

spec/isolation/components/mailer_configuration/without_hanami_mailer_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    it "resolves mailer configuration" do
+      with_project do
+        environment_file = Pathname.new(Dir.pwd).join("config", "environment")
+        replace "#{environment_file}.rb", "delivery :test", ""
```
jodosha: Same of https://github.com/hanami/hanami/pull/742/commits/5523b86ca13a8f3145f3a41257b1b2c61f2b1f3b#r105839060

davydovanton:
@jodosha thanks for help 💚

PR #748, opened 2017-03-10
Disconnect database during the boot process.

jodosha:
This is useful to restart projects in production, so the system can prune stale database connections.

### DIFF
--- a/spec/isolation/components/model/with_hanami_model_spec.rb
+++ b/spec/isolation/components/model/with_hanami_model_spec.rb
@@ -10,5 +10,38 @@ RSpec.describe "Components: model", type: :cli do
         expect(Hanami::Model.configuration.logger).to eq(Hanami.logger)
       end
     end
+
+    it "disconnects database connections on reboot" do
+      with_project do
+        require Pathname.new(Dir.pwd).join("config", "environment")
+
+        # Simulate previous connection
+        Hanami::Components.resolve('model.configuration')
+        Hanami::Components.resolve('model.sql')
+        Hanami::Model.load!
+
+        #
+        Hanami::Model.configuration.connection.test_connection
+        # Current connections count is 1
+
+        #
+        Hanami::Components.resolve('model')
+        # Current connections count is 0
+
+        expect(Hanami::Components['model']).to     be(true)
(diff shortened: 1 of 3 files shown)


spec/isolation/components/model/with_hanami_model_spec.rb
```diff
@@ -10,5 +10,38 @@
...
+        expect(Hanami::Model.configuration.logger).to eq(Hanami.logger)
+
+        # This is tight coupled to Sequel
+        #
+        # When `.disconnect` is invoked, it returns a collection of disconnected
+        # connections. Here we want to assert that `.disconnect` was invoked as
+        # part of the boot process.
+        #
+        # Invoking disconnect again **here in the test** should return an empty
+        # collection, because we haven't tryied to connect to the database again.
```
cllns: typo: `tryied` => `tried`

spec/isolation/components/model/with_hanami_model_spec.rb
```diff
@@ -10,5 +10,38 @@
...
+        # Current connections count is 0
+
+        expect(Hanami::Components['model']).to     be(true)
+        expect(Hanami::Components['model.sql']).to be(true)
+        expect(Hanami::Model.configuration.logger).to eq(Hanami.logger)
+
+        # This is tight coupled to Sequel
+        #
+        # When `.disconnect` is invoked, it returns a collection of disconnected
+        # connections. Here we want to assert that `.disconnect` was invoked as
```
cllns: Could we do `expect(Hanami::Model).to receive(:disconnect)` instead? I'm OK with this if there's a good reason we can't do that, just want to check :)

mereghost: I think that he's testing if `disconnect` is doing it's thing and not it it's being called (as he's calling it explicitly)

jodosha: @cllns I preferred to not mock `disconnect` and test if the connections get wiped for real.

cllns: makes sense :)

Issue #744, opened 2017-03-06
Don't mount Hanami::CommonLogger if Hanami.logger is nil


jodosha:

sandelius:
@jodosha I've fix this but have no idea where to put my tests :)

jodosha:
@sandelius Please go ahead with the fix, I'll take care of the tests. Are you able to provide a PR in the next couple of days? We want to ship beta3 soon. Thank you.

sandelius:
@jodosha I can do it later today but I would really love to add a test as well. Would a unit or integration be the better option?

I'm trying to "learn" how you guys write, and place, your tests so I can contribute more.

jodosha:
@sandelius I asked to provide only the fix implementation, but not the tests because I recognised we didn't had an integration test for `hanami server` logging.

With https://github.com/hanami/hanami/commit/3b435ae1c65afa2aaa3fcbec9c0bfee98b7b6cf4 I added an integration testing for logging. Please use it as a starting point for testing your fix.

PR #749, opened 2017-03-13
Do not mount CommonLogger if logging is disabled

sandelius:
Fixes https://github.com/hanami/hanami/issues/744

### DIFF
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -63,7 +63,7 @@ module Hanami
     # @since 0.9.0
     # @api private
     def middleware(environment)
-      builder.use Hanami::CommonLogger, Hanami.logger
+      builder.use Hanami::CommonLogger, Hanami.logger if Hanami.logger
       builder.use Rack::ContentLength
 
       if middleware = environment.static_assets_middleware # rubocop:disable Lint/AssignmentInCondition
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -63,7 +63,7 @@ module Hanami
     # @since 0.9.0
     # @api private
     def middleware(environment)
-      builder.use Hanami::CommonLogger, Hanami.logger if Hanami.logger
+      builder.use Hanami::CommonLogger, Hanami.logger unless Hanami.logger.nil?
       builder.use Rack::ContentLength
 
       if middleware = environment.static_assets_middleware # rubocop:disable Lint/AssignmentInCondition
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -85,10 +85,10 @@ EOF
   end
 
   context "logging" do
-    context "when enabled" do
-      let(:log)     { "log/development.log" }
-      let(:project) { "bookshelf" }
+    let(:log)     { "log/development.log" }
+    let(:project) { "bookshelf" }
 
+    context "when enabled" do
       it "logs request" do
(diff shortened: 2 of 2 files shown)


sandelius:
Perhaps we should check that `$stdout` does not receive any output as well?

davydovanton:
hey, I found that you have failed specs. Could you check this?
https://travis-ci.org/hanami/hanami/jobs/210641947#L1125

jodosha:
@davydovanton We ignore jruby failures..

spec/integration/cli/server_spec.rb
```diff
@@ -105,6 +105,23 @@ def call(params)
         end
       end
     end
+
+    context "when not enabled" do
+      it "does not log request" do
+        with_project(project) do
+          touch log
```
jodosha: This shouldn't be `touch`'d

spec/integration/cli/server_spec.rb
```diff
@@ -105,6 +105,23 @@ def call(params)
...
+          replace "config/environment.rb", "logger level: :debug", "logger stream: false"
```
jodosha: This should be removed completely:

```ruby
replace "config/environment.rb", "logger level: :debug", ""
```

And so all the instances of logging.

spec/integration/cli/server_spec.rb
```diff
@@ -105,6 +105,23 @@ def call(params)
...
+
+          server do
+            visit "/"
+            expect(page).to have_title("Hanami | The web, with simplicity")
+          end
+
+          content = contents(log)
```
jodosha: This can be removed

spec/integration/cli/server_spec.rb
```diff
@@ -105,6 +105,23 @@ def call(params)
...
+          expect(content).to be_empty
```
jodosha: This should be:

```ruby
expect(log).to_not be_existing_file
```

lib/hanami/app.rb
```diff
@@ -63,7 +63,7 @@ def mount(configuration)
     # @since 0.9.0
     # @api private
     def middleware(environment)
-      builder.use Hanami::CommonLogger, Hanami.logger
+      builder.use Hanami::CommonLogger, Hanami.logger if Hanami.logger
```
jodosha: `unless Hanami.logger.nil?`

If we introduce a _null object_ this test will still pass because of Ruby's truthiness:

```ruby
class NullLogger
  def nil?
    true
  end
end

Hanami.logger = NullLogger.new

if Hanami.logger
  # it enters this block, but it's unwanted
end

unless Hanami.logger.nil?
  # it doesn't enter this block as expected
end
```

sandelius: @jodosha good point!

spec/integration/cli/server_spec.rb
sandelius: @jodosha I'm getting `expected "log/development.log" to respond to `existing_file?` with this.

davydovanton:
@jodosha it was 2.3.3 ruby error 🤔

jodosha:
@davydovanton I will never comment again without a decent coffee. That comment supposed to go in #750 😉

spec/integration/cli/server_spec.rb
jodosha: That's because is `be_an_existing_file`, not `be_existing_file`. Sorry about the confusion.

Issue #656, opened 2016-10-19
Allow host to be set when mounting apps in the container


sandelius:
It would be really nice to set a host (String, Regexp) in the container like this:

``` ruby
Hanami::Container.configure do
  mount Api::Application, at: '/', host: /\Aapi\.*/
  mount Auth::Application, at: '/auth'
end
```

jodosha:
@hanami/core-team @hanami/contributors WDYT?

AlfonsoUceda:
@sandelius could you provide an example? I mean, what do you want to do with that option? I can imagine urls will be generated to that endpoint, but if you can explain a little more.
Thanks!

sandelius:
Let's say I have two apps, one called web and one called api. I want bort to be served from `/` but the API will only be served from the `api.` subdomain.

```ruby
Hanami::Container.configure do
  mount Api::Application, at: '/', host: /\Aapi\.*/
  mount Web::Application, at: '/'
end
```

ping @jodosha @AlfonsoUceda

mereghost:
@sandelius A rewrite rule (on the http server) wouldn't suffice for that?

sandelius:
@mereghost yeah we can do it there but not everyone have access to it easily. I can do it like that but I think it's a feature that should be in Hanami.

Also it would make it a bit easier during development.

TiteiKo:
As far as I've tried, rewrite rules don't necessarily do well with non-GET http methods, so that might be a problem for an api

IMHO, it's a feature that should be in Hanami, as it is in line with the container architecture and it's common to have/need subdomains like `admin.example.com, dashboard.example.com, api.example.com, ...` or even an entirely other domain.

mereghost:
@sandelius I see. The thing is that a server will need to be setup hear to that subdomain and whatnot, so a bit of Ops work goes there so it would be simpler to leave routing subdomains (and internal rewrites) to the proven tech that already does it. =) Besides it's less code to maintain.

 That being said, if implementing that would be a breeze, I'm all for it. 👍 

@TiteiKo I'm not sure that I follow you on the whole *in line with the container arch* tho 😞 Could you elaborate a bit (can be over gitter so we do not pollute here with a tangent discussion).

pascalbetz:
@mereghost 

Hanami Supports the "Container Architecture". You can think of it as multiple, separated apps inside one Application.

mereghost:
[...]

PR #753, opened 2017-03-16
Allow optional host parameter when mounting apps

robbl-as:
This pull-request introduces the possibility to mount an application using an optional host parameter. IMHO this closes a gap to the implementation of hanami/router, that allows the host parameter.

```ruby
Hanami.configure do
  mount(Api::Application, host: 'api.myapp.org', at: '/v1')
  mount(Web::Application, host: 'myapp.org', at: '/v1')
end
```


robbl-as:
This pull-request also addresses the issue #656.

lib/hanami/configuration.rb
```diff
@@ -42,7 +43,7 @@ def initialize(&blk)
     #     # ...
     #   end
     def mount(app, options)
-      mounted[app] = App.new(app, options.fetch(:at))
+      mounted[app] = App.new(app, options.fetch(:at), options.fetch(:host) { nil })
```
artofhuman: Can we use `options[:host]` ?
Should we add this option to documentation and add example?

robbl-as: @artofhuman Good point, I simplified the read access for the host option. Actually we got quite confused with the existing documentation http://hanamirb.org/guides/routing/overview/#mounting-on-a-subdomain. It implies, that there is already the possibility to mount an application with a given host. Prior to my pull-request the host options was ignored and not delegated to hanami/router.

cllns:
Thanks for this!

What are `account` and `history` here? subdomains or full host names?  If they're just subdomains, I wonder if we should name the option `subdomain` instead? 

I could imagine a use-case of a single app on completely different host names. I'm not sure how we balance allowing people to mount apps on entirely different domains while letting people mount on a subdomain without having to hard-code the rest of the URL in the router. thoughts?

robbl-as:
@cllns In my example account and history are complete hostnames. I am coming from a docker environment perspective, where hostnames like these are not unusual. Anyhow I updated my pull-request description. When referring to the complete hostname, mounting to subdomains is possibile, too, of cause.

robbl-as:
Also I am not sure, why the travis build to this pull-request for the last commit failed. Is it possible to rerun the build?

cllns:
great, thanks! Does providing a regex, instead of a string work too?

So you could mount to a subdomain without providing the rest of the domain?
```
Hanami.configure do
  mount(Api::Application, host: %r{^api.\w+}, at: '/v1')
  mount(Web::Application, at: '/v1')
end
```

sandelius:
@cllns don't think so, what I can see it's just passed along to HttpRouter and there only a string is supported.

robbl-as:
@cllns @sandelius There is no support for regex. I'd really like to see this pull request merged without an implementation that supports regex. In fact this pull request should be handled like a bugfix since the current documentation with in the hanami guides is not correct. The section http://hanamirb.org/guides/routing/overview/#mounting-on-a-subdomain describes mounting an application to a subdomain using the host-option. The host options is ignored. This pull-request fixes the issue.

jodosha:
@robbl-as Hi and thanks for this PR.

---

Here's the steps I used to manually test it. I'm on Mac OS.

```
➜ pwd
/Users/luca/Code/hanami/hanami
➜ git checkout -b robbl-as-feature/allow-host-option-when-mounting-apps master
➜ git pull https://github.com/robbl-as/hanami.git feature/allow-host-option-when-mounting-apps
```

Then I generated a project that uses Hanami HEAD: https://github.com/hanami/hanami#how-to-use-hanami-head

I edited `/etc/hosts` and flushed the DNS cache:

```
127.0.0.1 bookshelf.local api.bookshelf.local # ...
```

```
➜ dscacheutil -flushcache
```

From the root of the test project:

```shell
➜ pwd
/Users/luca/Code/hanami/hanami/bookshelf
➜ bundle exec hanami generate app api
# ...
```

I edited `config/environment.rb` and started the server

```ruby
Hanami.configure do
  mount Api::Application, at: '/api', host: 'api.bookshelf.local'
  # ...
end
```

```shell
➜ bundle exec hanami server
# ...
```

At this point I was able to hit http://api.bookshelf.local:2300/

### Problem 1

At that URL I see the welcome page that tells me how to generate the first action for `Api` application:

```shell
➜ bundle exec hanami generate action web home#index --url=/
# ...
```

If I visit again that URL, **I still see the welcome page**. This is wrong.

If I inspect the routes, there is something wrong:

```shell
➜ bundle exec hanami routes
                Name Method     Path                           Action

                     GET, HEAD  /api                           Api::Controllers::Home::Index

                Name Method     Path                           Action

```

It looks like the URL that Hanami sees is http://api.bookshelf.local:2300/api, but we expect http://api.bookshelf.local:2300/. Indeed, by visiting http://api.bookshelf.local:2300/api, I can see the new generated page.

### Problem 2

If I generate a new action:

```shell
➜ bundle exec hanami generate action api books#index --url=/books
# ...
```

And I assign a name to that route:

```ruby
# apps/api/config/routes.rb
get '/books', to: 'books#index', as: :books
```

When I use the routes helper from a template it returns a wrong URL:

```erb
# apps/api/templates/home/index.html.erb
<%= routes.books_url %>
```

It returns http://localhost:2300/api/books, but we expect http://api.bookshelf.local:2300/books.

### Problem 3

If I add a stylesheet at `apps/api/assets/stylesheets/application.css` and refer it from `apps/api/templates/application.html.erb`:

```erb
<%= stylesheet "application" %>
```

It generates the wrong URL for it:

```html
<link href="/assets/api/application.css" type="text/css" rel="stylesheet">
```

---

Before to merge this we want to fix these problems above. Thank you.

robbl-as:
@jodosha Thanks for diving in! Acctually I cannot recreate some of your problems, because it just works for me, even by following the same steps. Maybe I have to setup a repository with a test application we can work with? I had no issues with the routes or the assets.

The only issue I was able to reproduce: The routes url helper does not include the assigned host but still the localhost. I could look into this, though.

TiteiKo:
I haven't tested it but I have a question...

ATM, assets path are generated based on the routing of the assets app. Meaning that if you have a `Web` app mounted at `/` and an `Admin` app mounted at `/admin`, `Web` assets paths will look like `/assets/application.css` and `Admin` assets paths will look like `/assets/admin/application.css`

With your change, if we have a `Web` app at `/` and an `Admin` app at `/` on a subdomain, both having an `application.css` asset, what will happen? What will be the assets paths?

robbl-as:
@TiteiKo Looks like we have another issue. :) I tried what you suggested and both asset paths point to the same relative url `/assets/application.css`. Since we use the suggested changes from this pull-request for two API-only-applications mounted to `/api` but with different host-options we didn't run into any assets related issues. :smile: I failed to fix this issue by setting an explicit asset host for both applications. The assets paths are now legit `http://api.myhost.de/assets/application.css` but still not resolved to the correct file. Maybe this could be fixed somewhere [over here](https://github.com/hanami/hanami/blob/master/lib/hanami/assets/asset.rb#L35). Looks like assets are only resolved based on the path but not the host.

```
module Api
    configure do
      host 'api.myhost.de'

      assets do
        host 'api.myhost.de'
      end
    end
  end
end
```

jodosha:
Closing this as stale contribution. I'm marking it "for future consideration", because I'd like to see it in 2.0.

PR #754, opened 2017-03-17
Removed deprecated `ApplicationConfiguration#default_format`

jodosha:


jodosha:
@davydovanton 
> Should we change something in specs?

I haven't ported all the unit specs when migrated from minitest to rspec. It shouldn't be a problem.

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ The web, with simplicity.
 - [Anton Davydov] Don't configure mailers, if it's mailing is disabled for the project.
 - [Marcello Rocha] Ensure code reloading don't misconfigure mailer settings
 
+### Changed
```
TiteiKo: Shouldn't it be `Removed` instead of `Changed` ?

jodosha: We only have three categories:

  - "Added" for new features
  - "Fixed" for fixed bugs
  - "Changed" for breaking changes

PR #755, opened 2017-03-21
Remove `--architecture` option from `new` command

ixti:
From what I saw (on forum), seems like `app` architecture support
was dropped. Also passing `--arch=app` fails as there's no such
template anymore.


davydovanton:
hey, thanks for PR but we have this :)
https://github.com/hanami/hanami/pull/737

/cc @cllns

ixti:
Oh! Awesome! #737 definitely covers more parts than this one. Closing this in favour of #737.

Issue #733, opened 2017-02-23
Logger: add form submission parameters


cllns:
Our logger records query parameters, but not parameters submitted via form submissions. 

Current output:
```
[test_app] [INFO] [2017-02-23 16:34:24 -0700] HTTP/1.1 POST 405 ::1 /users/new  - 0.00902
```

Proposed (example) output:
```
[test_app] [INFO] [2017-02-23 16:34:24 -0700] HTTP/1.1 POST 405 ::1 /users/new - 0.00902
[test_app] [INFO]     Parameters: { "email" => "hello@example.com", "password" => "[FILTERED] } 
```

Note that we ought to filter out `password`. We should also filter out `password_confirmation` (maybe just filter out all keys that match `/^password/`?). 
We should also allow people to specify other field names (or regexes?) to filter out. For example, so people could filter out `credit_card_number`.

When printing to STDOUT, like in development mode, I personally prefer displaying the parameters hash with `pp`, so it's easy to read. Printing a hash on a single line is incredibly error-prone, in my experience. 
Should we also `deep_symbolize` the params before outputting them?

Note: this is a proposed enhancement for `v1.1`, **not** `v1.0`.

What do you all think? 💁

---

Currently assigned to @gizotti

b264:
I agree with all of that, especially the `v1.1` part.

I do have a tip-- as a Hanami beginner, if I saw fields being filtered out, it would be great if the search term on how to change the setting was the thing displayed in the log, example:
```
[test_app] [INFO]     Parameters: { "email" => "hello@example.com", "password" => "[KEYS_HIDDEN] }
```
so then if I search for, example, `https://ddg.gg/hanami%20%22KEYS_HIDDEN%22` then I get a search result saying where to adjust the regex `Hanami::Logger::KEYS_HIDDEN` or whatnot.  Just a suggestion that how to change it is searchable from the filtered text output itself with a word a little less generic than "FILTERED" and especially not the same word Rails uses.

davydovanton:
Great idea 👍
I think we need to use `pp` style for displaying params (because sometimes params are too long).

gizotti:
@cllns @davydovanton I'll have a go at this one. Can you guys point out a good start point, I am looking for it now but a little help is always welcome :).

As for `pp`, should we also have an option if people want to disable `pp` and display the params in a single line?

gizotti:
Hey guys. Just a heads up that I started on this one.
I'll try to put something out sometime this week :)

[...]

PR #756, opened 2017-03-23
Log form submission parameters

gizotti:
Relates to #733

This PR introduces logging of form parameters in a request, if any is passed. It does so by doing the following:

- Introduces `filtered_parameters` configuration so custom keys, or regexes, can be configured to match against form params.
- Introduces `Hanami::FormParams` class to handle the processing, and filtering, of param hashes
- update `CommonLogger` to log the params processed by `Hanami::FormParams` if any.

This is only the first pass for this enhancement and I would like some feedback on it. 

Here's a list of considerations on why I have done things this way (so far):

Timestamp is still logged with the Parameters. I have done this way so we don't need to update `hanami-utils`, at least not for now

I am using `JSON.pretty_generate` for the `pretty_print` effect, but that turns the hash into a `json` so I am not sure if there's much gain on `deep_symboize!` the hash before logging.

@b264 suggested on #733 to instead of filtering the params with `[FILTERED]` to have the location of where the filters can be customized. While I think that we should display where it can be customized, I am not convinced that the replacing `[FILTERED]`  with the location is the best solution. I would prefer that message to be another logged line.

Still on the filtering, I did so it always depend on the `filtered_parameters` configuration. But on second thought, should we always filter anything that matches `password` + anything that matches the custom filters? Is that what you mean in the issue @cllns?

Right now this is how the logging is looking:

```
[hanami_blog] [INFO] [2017-03-23 22:49:25 +1000] HTTP/1.1 POST 302 ::1 /posts  5 0.017046
[hanami_blog] [INFO] [2017-03-23 22:49:25 +1000] Parameters: {
  "post": {
    "title": "You won't believe what he has done....",
    "body": "All about a random guys achievements",
    "password": "[FILTERED]",
    "password_confirmation": "[FILTERED]"
  }
}
```


gizotti:
@davydovanton @cllns do you guys have any feedback on this one?

davydovanton:
@gizotti example looks awesome 💚 
Give me a day, I'll check a code. @hanami/core check this PR please 👍

gizotti:
@davydovanton thanks!! Didn't meant to rush anybody though :)

AlfonsoUceda:
@gizotti thank you so much for this PR, one think I don't like it, the use of `pretty_print`, IMO we shouldn't modify the behaviour of the log, this should be responsibility of the log's formatter. Why? this behaviour can clash with production environments, because it adds unnecessary spaces.

gizotti:
@AlfonsoUceda thanks for the feedback. I was looking at the PR again back this weekend and I had thoughts similar yours. I just decided to wait for a first round of feedback here first.

How about if, as a first step, remove the pretty_print and just pass the unformatted string to the formatted, maybe with a key that the formatted will understand as formatable in the near future, and afterwards I open a PR on utils to address the formatting of the log message?

gizotti:
@AlfonsoUceda I removed all the formatting part from the PR and just sent a separate log message to the logger with the raw filtered params to it.

I'll raise another PR in the `hanami-utils` to handle the formatting of it.

gizotti:
@jodosha everything looks exactly like the steps I took to test my changes, weird that it's not working. No reason for me to check it though if we are moving the filtering to `Hanami::Logger`.

Wouldn't making the filtering `Hanami::Logger`'s responsibility add the complexity of us having to ensure all `Formatters` are filtering the passed `hash` (which is not much of  a problem really), and that people implementing custom loggers would have to implement the filtering themselves on their loggers?

jodosha:
@gizotti 
> and that people implementing custom loggers would have to implement the filtering themselves on their loggers?

This can be transparent to formatters. They will already receive the filtered Hash. This line can be a good extension point: https://github.com/hanami/utils/blob/master/lib/hanami/logger.rb#L155

gizotti:
@jodosha awesome. I'll start from there.

I have had a few busy days and couldn't get to it. I'll have a free day Monday so I'll be able to get the whole lot done :)

gizotti:
@jodosha I've moved the filtering responsibility to `utils`. The PR is also ready to review hanami/utils#207.

Once that PR is merged, and `hanami` is using the version of `utils` with those changes, I can finish up this PR to ensure that new apps are created with filters for production.

jodosha:
Closing in favor of #818

PR #761, opened 2017-03-29
Allow mailer settings to be evaluated multiple times

### LINKED ISSUE #757: First mailer config rejects any further mods
Let's start with [initial](https://github.com/hanami/hanami/blob/master/lib/hanami/generators/application/container/config/environment.rb.tt) (leaving relevant parts only) configuration hanami generates for us:

``` ruby
Hanami.configure do
  mailer do
    root 'lib/example/mailers'
    delivery :test
  end

  environment :production do
    mailer do
      delivery :smtp, address: ENV['SMTP_HOST'], port: ENV['SMTP_PORT']
    end
  end
end
```

In this configuration, `mailer` block inside of `environment :production` will be silently ignored. And as result `delivery :test` will be applied to all environments. The reason is that once `mailer` block is called, all further calls will be silently ignored due to `put_if_absent` [under the hood](https://github.com/hanami/hanami/blob/master/lib/hanami/configuration.rb#L94-L96).

The only way to make mailer work right now is to ensure that you call it in environment specific block first:

``` ruby
Hanami.configure do
  environment :production do
    mailer do
      root 'lib/example/mailers'
      delivery :smtp, address: ENV['SMTP_HOST'], port: ENV['SMTP_PORT']
    end
  end

  mailer do
    root 'lib/example/mailers'
    delivery :test
  end
end
```

Notice that `root` needs to be duplicated, as only one of those blocks will be called in fact, so each of them needs to contains full configuration.

jodosha:
Closes #757 

/cc @hanami/core and @ixti for review.

### DIFF
--- a/lib/hanami/components/components.rb
+++ b/lib/hanami/components/components.rb
@@ -178,12 +178,15 @@ module Hanami
       end
 
       resolve do |configuration|
-        unless configuration.mailer.nil?
+        unless configuration.mailer_settings.empty?
           if Hanami.code_reloading? && !Hanami::Mailer.configuration.nil?
             Hanami::Mailer.configuration = Hanami::Mailer.configuration.dup
           end
 
-          Hanami::Mailer.configure(&configuration.mailer)
+          configuration.mailer_settings.each do |settings|
+            Hanami::Mailer.configure(&settings)
+          end
+
           Hanami::Mailer.configuration
         end
       end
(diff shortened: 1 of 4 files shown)


ixti:
Works perfectly for me!

ixti:
Thanks!

jodosha:
@ixti 🎩

PR #763, opened 2017-03-31
Improve files templates

ixti:


ixti:
This is mostly a discussion starter. :D

AlfonsoUceda:
Hi @ixti thanks for this PR, I'm against to add this string preprocessor in all files because if a developer wants to add it globally, it should use the following tip: http://stackoverflow.com/a/35122812 IMO this gets dirty the files, but it just my opnion.

@hanami/core what do you think?

ixti:
@AlfonsoUceda proposed on SO solution affects ALL required gems. My proposal is to "prepare" hanami for being that option-friendly and as a result being Ruby 3.0+ firendly, where this behaviour will be default.

ixti:
Oh. Seems like I just only thought that my proposal is about to add those magic comments to hanami itself as well... And forgot about that :D

In any case. Regarding templates. Isn't adding what considered best practice (nowadays) is pretty inline with "convention over configuration"?

ixti:
Also, IMO encouraging using immutable strings we are going along with idea of mutable entities. :D

jodosha:
@ixti Thanks for this PR. I agree with this change, but unfortunately this isn't the time to introduce it yet.

Hanami mutates strings as well (mainly in `hanami-utils`). We're moving away from this first of all for Hanami source code (see hanami/mailer#69) and then for projects using Hanami.

---

I would target Hanami 2.0 for that change, rather than having it now. It could be dangerous to enable it now.

For this reason, we can't keep this PR open for long time, but act later on this.

Does it makes sense?

ixti:
@jodosha 100% agree!

jodosha:
@ixti Thanks for understanding.

PR #771, opened 2017-04-19
CLI: `hanami version` rewrite

jodosha:
Implement `hanami version` with the new `hanami-cli` gem.

### DIFF
--- a/bin/hanami
+++ b/bin/hanami
@@ -1,5 +1,4 @@
 #!/usr/bin/env ruby
-
 require 'bundler'
 require 'hanami/command_line'
 
--- a/bin/hanami
+++ b/bin/hanami
@@ -1,5 +1,9 @@
 #!/usr/bin/env ruby
 
 require 'bundler'
-require 'hanami/cli'
-Hanami::Cli.start
+require 'hanami/command_line'
+
+result = Hanami::CommandLine.run
+
+return if result
+Hanami::OldCommandLine.start
--- a/bin/hanami
+++ b/bin/hanami
@@ -4,5 +4,5 @@ require 'hanami/command_line'
 
 result = Hanami::CommandLine.run
 
-return if result
+exit(0) if result
 Hanami::OldCommandLine.start
--- a/lib/hanami/command_line.rb
+++ b/lib/hanami/command_line.rb
@@ -1,3 +1,4 @@
+require 'hanami/cli'
 require 'thor'
 require 'hanami/cli_base'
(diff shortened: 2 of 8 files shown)


bin/hanami
```diff
@@ -1,5 +1,8 @@
...
-
 require 'bundler'
-require 'hanami/cli'
-Hanami::Cli.start
+require 'hanami/command_line'
+
+result = Hanami::CommandLine.run
+
+exit(0) if result
+Hanami::OldCommandLine.start
```
davydovanton: awesome 👏

lib/hanami/command_line.rb
```diff
@@ -1,26 +1,28 @@
...
+  module CommandLine
+    include Hanami::Cli
+
+    class Version
+      def call
+        puts "v#{Hanami::VERSION}"
+      end
+    end
+
+    register "version", Version
```
davydovanton: What will happen if I call `hanami v` or `hanami --version`?

davydovanton: and I understand right that command description will we create in future?

lib/hanami/command_line.rb
```diff
@@ -1,26 +1,28 @@
...
 require 'hanami/commands/new/app'
 require 'hanami/commands/new/container'
 
 module Hanami
+  module CommandLine
+    include Hanami::Cli
+
+    class Version
+      def call
+        puts "v#{Hanami::VERSION}"
```
AlfonsoUceda: I think we need `require 'hanami/version'` isn't it?

lib/hanami/command_line.rb
AlfonsoUceda: @davydovanton for now hanami-cli is very naive, this is a PR for testing the new Cli, I think we shouldn't merge yet.

lib/hanami/command_line.rb
jodosha: It's automatically required by `hanami.gemspec`. I think it's fine.

jodosha:
@AlfonsoUceda If you're talking about `--version` and `-v` aliases, I'm fine to wait for them. If so, just push new commits here.

AlfonsoUceda:
@jodosha I've implemented aliases for -v and --version

jodosha:
@AlfonsoUceda @oana-sipos As long you will merge hanami/cli#2, please update the branch in this `Gemfile` and merge it.

jodosha:
:shipit:

Issue #651, opened 2016-10-07
.env not working in production


rafaels88:
I just updated my Hanami application from `0.7.0` to `0.8.0` and `.env` file is not been loading in production. After reviewing `environment.rb` file in `0.8.0` version, I saw this line:

```
 DEFAULT_DOTENV_ENV = '.env.%s'.freeze
```

So, I tried to change my `.env` file to `.env.production` and it worked.

It seems that the solution for this is pretty simple. =)

cllns:
Thanks for opening this issue! 🌸 

We use the `dotenv` gem for this functionality. It's _only_ in the `:development` and `:test` groups, so it shouldn't be loaded in production. See here: https://github.com/hanami/hanami/blob/master/lib/hanami/generators/application/app/Gemfile.tt#L40-L42

We want to discourage people from having production environment variables in source control. I think this experience can be improved though, since a lot of people run into this problem. At the very least, this behavior should be explicity documented, but ideally we'd be able to show an error to the user to help them. Any ideas?

rafaels88:
Humm I see. That's fair enough. But, as you said, it would be nice to have a documentation mentioning a good way to set ENV variables in production, without using `.env` file. I mean, in a AWS-like environment probably there will be a dashboard where you just configure that, but in a Digital Ocean-like it is not trivial (at least for me) how to solve that without a `.env` file in production.

Also, there are some places where documentation mention that `.env` works on production, it would be nice remove that.

What do you think?

cllns:
Just added an issue to create an Environment Variables guide :) https://github.com/hanami/hanami.github.io/issues/224

Feel free to open a PR against https://github.com/hanami/hanami.github.io/ for any error you see in the Guides, and in each repo if it's in Rdoc/comments :)

Also, I'd probably recommend using `.env.production` instead of just `.env`, just to signify that it's really just for production (and create an entry in `.gitignore` for that too).

b264:
What if folks have `.env.production` ignored in source control?  ( `.gitignore` for example )

jodosha:
@rafaels88 Hi 😄 . I think it's just a matter of documentation, we can write a clear statement about the `.env` expectations in production, why we it's a bad idea to have it in production.

We can cover most common vendors (Heroku, AWS) too. But we expect people to understand their deployment environment.

---

[...]

PR #772, opened 2017-04-22
Evaluate configuration from master .env file

linchus:
Documentation says that both `.env` and `.env.*` files are evaluated, but actually `.env` does not.

This pull request fixes it.

### DIFF
--- a/lib/hanami/environment.rb
+++ b/lib/hanami/environment.rb
@@ -48,11 +48,17 @@ module Hanami
     # @api private
     RACK_ENV_DEPLOYMENT = 'deployment'.freeze
 
-    # Default `.env` per environment file name
+    # Default `.env` file name
     #
     # @since 0.2.0
     # @api private
-    DEFAULT_DOTENV_ENV = '.env.%s'.freeze
+    DEFAULT_DOTENV_ENV = '.env'.freeze
+
+    # Specific `.env` per environment file name
+    #
+    # @since 1.0.0
+    # @api private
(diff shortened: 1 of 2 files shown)


cllns:
Hm, I think the documentation here might be wrong, not the code. See this issue: https://github.com/hanami/hanami/issues/651 

We seek to encourage good practices and discourage bad ones. Putting ENV variables used in production into source code is generally a bad practice.

That being said, I think most people who have use `dotenv` would be surprised to find out `.env` is not being included

linchus:
Now I see, why it was checked in specs )

The inheritance from .env is described here at least https://github.com/hanami/hanami/blob/master/lib/hanami/environment.rb#L162

In my opinion keeping some common settings in `.env` file allows you to dry your dotenv files.
I agree with you that production settings must be out of source code but it is about .env.production - not .env =)

jodosha:
@linchus It used to be like that: `.env` first and then override via `.env.*`.

We removed it because people were using default `.env` the wrong way: they put sensitive data in it and then they used to add that file to git. We hope that now with a clear name (eg `.env.test`) people are discouraged to apply this bad practice.

DRY turned to be non-secure practice for our `.env` files. I'm closing this because security > duplication.

Thanks for understanding Sergey!

cllns:
@linchus would you be able to update the documentation comments in ` lib/hanami/environment.rb` so that they're accurate?

linchus:
@cllns sure

I'll update comments and send new PR with changes.

PR #773, opened 2017-04-25
Actualize Hanami::Environment comments about .env file

linchus:
Update documentation in Hanami::Environment to actualize dotenv file inheritance information.

Based on discussion  in #651 and #772

### DIFF
--- a/lib/hanami/environment.rb
+++ b/lib/hanami/environment.rb
@@ -148,8 +148,9 @@ module Hanami
     # configurations and discourage putting sensitive information into source
     # control.
     #
-    # The format of those `.env.<environment>` files is compatible with `dotenv`
-    # and `foreman` gems.
+    # The format of those `.env.<environment>` files follows UNIX and UNIX-like
+    # operating system environment variable declaration format and compatible
+    # with `dotenv` and `foreman` gems.
     #
     # @param options [Hash] override default options for various environment
     #   attributes
--- a/lib/hanami/environment.rb
+++ b/lib/hanami/environment.rb
@@ -140,12 +140,14 @@ module Hanami
     # When initialized, it sets standard `ENV` variables for Rack and Hanami,
     # such as `RACK_ENV` and `HANAMI_ENV`.
     #
-    # It also evaluates configuration from `.env` and `.env.<environment>`
+    # It evaluates configuration ONLY from `.env.<environment>` file
     # located under the config directory. All the settings in those files will
     # be exported as `ENV` variables.
     #
-    # The format of those `.env` files is compatible with `dotenv` and `foreman`
-    # gems.
+    # Master `.env` file is ignored for security reason.
+    #
+    # The format of those `.env.<environment>` files is compatible with `dotenv`
+    # and `foreman` gems.
     #
     # @param options [Hash] override default options for various environment
     #   attributes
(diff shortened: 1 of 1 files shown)


lib/hanami/environment.rb
```diff
@@ -140,12 +140,14 @@ class Environment
...
     # such as `RACK_ENV` and `HANAMI_ENV`.
     #
-    # It also evaluates configuration from `.env` and `.env.<environment>`
+    # It evaluates configuration ONLY from `.env.<environment>` file
     # located under the config directory. All the settings in those files will
     # be exported as `ENV` variables.
     #
-    # The format of those `.env` files is compatible with `dotenv` and `foreman`
-    # gems.
+    # Master `.env` file is ignored for security reason.
```
kaikuchn: Will people lacking the context of the discussions that lead to this PR understand this?

linchus: I'm not sure that it is the right place to put full explanation why it is bad.
I'd say it's enough to understand code behaviour.

What text do you suggest to add here to improve understanding?

kaikuchn: What about something along the lines of this
> Master `.env` file is ignored to suggest clear separation of environment configurations and discourage putting sensitive information into source control.

jodosha: I'd mention also that `.env` files are only for development and not for production.

mereghost:
@TiteiKo can you check this, and if okay, merge it?

lib/hanami/environment.rb
```diff
@@ -140,12 +140,16 @@ class Environment
...
     # be exported as `ENV` variables.
     #
-    # The format of those `.env` files is compatible with `dotenv` and `foreman`
-    # gems.
+    # Master .env file is ignored to suggest clear separation of environment
+    # configurations and discourage putting sensitive information into source
+    # control.
+    #
+    # The format of those `.env.<environment>` files is compatible with `dotenv`
+    # and `foreman` gems.
```
TiteiKo: dotenv and foreman, yes, but it's also compatible with systemd EnvironmentFile format and probably a lot of others, I'd add that it's a universal unix environment variable declaration format

davydovanton:
@TiteiKo hey, any progress here? :)

TiteiKo:
@davydovanton well I'm waiting on anybody to agree or disagree on my comment. I'm not ok merging it while it's not addressed :)

@linchus WDYT?

linchus:
@TiteiKo What do you think about my update? Is it ok?

b264:
Maybe could add some more info about jodosha's comment  `I'd mention also that .env files are only for development and not for production.`  but specifically also Why.  Could we add a link to why this is discouraged for further reading if they want. (even if on another site)

TiteiKo:
LGTM :+1: 

@b264 To me, this comment should be in a guide and not in the source code: If you're not using Heroku or other similar service, chances are that your deployment strategy will generate an .env file, to be read by your service (for example using the `EnvironmentFile` directive in systemd). 

Because in truth, `.env` files were not invented by `Dotenv`, but are a common thing in UNIX-based systems, so in a way, yes, `.env` files are for production. **With a huge but**: production environment variables should **never** be committed in the same repository as the source code. 

But it **should** be in a guide :)

b264:
Yes, just a lot of folks get confused when they see lots of "Don't do thing X" but never any "Do thing Y instead" or at a minimum a link to somewhere else that explains it in more details

b264:
@TiteiKo If it's to be in a guide, they may not be in the guide when they see this.  But if this were to _link_ to a guide that explained it, that's kind-of what I mean :-)

PR #777, opened 2017-05-01
Generator template fix

sovetnik:
By default, generated minitest _spec files raises error  
```
cannot load such file -- spec_helper (LoadError)
```
 if run it individually, like
```ruby spec/foo/bar_spec.rb```.

### CODE REVIEW

**Comment by AlfonsoUceda** [comment]
@sovetnik thanks for the PR.

I think we should keep `require 'spec_helper'` instead using the relative, I know you can't run with ruby command but you can run it like this:
```
be rake test TEST=spec/web/views/root/hello_spec.rb
```

What do you think @hanami/core

**Comment by sovetnik** [comment]
@AlfonsoUceda but running from ruby, without rake, is a one of key benefits of minitest. Or exists one or more reasons to keep this as it is now?

**Comment by AlfonsoUceda** [comment]
@sovetnik yeah, I think it's useful use ruby version, let's wait what they say.

**Comment by akhramov** [comment]
Hi, @sovetnik. Just wondering, why don't you use -I switch?

> -Idirectory     specify $LOAD_PATH directory (may be used more than once)

So you could launch these tests like so:
```bash
ruby -Ispec spec/foo/bar_spec.rb
```

In fact, many editor plugins do it this way. For instance, Emacs minitest plugin ([link](https://github.com/arthurnn/minitest-emacs/blob/2997ba81456f2a0f3e25b6555b491586865ebb61/minitest.el#L42))

sovetnik:
@akhramov because it shorter.
If this can work by design, why not?
 `-I` is switch, runtime option. Why we should have incorrect require statement and every time rewrite it via switch?

AlfonsoUceda:
@hanami/core what do you think? I like the idea.

mereghost:
I'm indifferent to it. Anyone else want to chime in @hanami/core ?

davydovanton:
@mereghost +1

Usually I use rspec that's why I can agree or not agree with this changes

TiteiKo:
Using minitest, I think this is a spendid idea! Thanks @sovetnik

davydovanton:
@sovetnik hey! I want to merge your PR. But we need to merge it to `develop` branch. Could you resolve conflicts and after that I'll merge? Thanks! 👏

sovetnik:
@davydovanton yes, of course

sovetnik:
@davydovanton what wrong with travis?

TiteiKo:
Strange, the icon doesn't match the result, which is that the build passed https://travis-ci.org/hanami/hanami/builds/259432522?utm_source=github_status&utm_medium=notification

Issue #778, opened 2017-05-04
Allow to start only specific hanami apps


davydovanton:
I think one of the killer features of hanami is isolated applications. That's why I believe that it will be awesome if we allow options for start hanami server only with specific apps. Something like this:
```
$ hanami s --only=api
$ HANAMI_APPS=account hanami s
```

## Why is so important? 

With this feature, we will scale our monolith hanami application by domain-specific servers. For example:

1. Set only API cluster independent of main web app
2. Display admin (or analytic) apps only in private network only 

## Progress

I [created a simple commit](https://github.com/hanami/hanami/commit/4bf099f3bc8add02b2718ae98580b7aeb1eee849) with `HANAMI_APPS` logic, and if it is reasonable, I can continue work on it.

WDYT @hanami/core?

/cc @beauby

beauby:
As discussed, I :+1: the idea. Not sure about the implementation though. I'd suggest something along the lines of:
```ruby
# config/environment.rb
require 'bundler/setup'
require 'hanami/setup'
require 'hanami/model'
require_relative '../lib/bookshelf'
require_relative '../apps/web/application' if Hanami::Utils.application_enabled?(:web)
require_relative '../apps/api/application' if Hanami::Utils.application_enabled?(:api)

Hanami.configure do
  mount Web::Application, at: '/' if Hanami::Utils.application_enabled?(:web)
  mount Api::Application, at: '/api' if Hanami::Utils.application_enabled?(:api)
  # ...
end
```

where `application_enabled?` is something along the lines of:
```ruby
def application_enabled?(app_name)
  return true unless ENV.key?('HANAMI_APPS')
  ENV['HANAMI_APPS'].split(',').include?(app_name.to_s)
end
```

davydovanton:
Closed

PR #781, opened 2017-05-08
Introduce Hanami.app?(:app_name) method

davydovanton:
It's a simple proof of concept for https://github.com/hanami/hanami/issues/778.

With this changes, you can create something like:

```ruby
require_relative '../apps/api/application'

Hanami.configure do
  mount Api::Application, at: '/api'

  if Hanami.app?(:web)
  end
end
```

And now, if you call `HANAMI_APPS=api hanami s` in memory you will have only the api app.

@jodosha WDYT? If it is good for you, I'll add tests, documentation and refactor `ENV` constant call

/cc @hanami/core @beauby

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -118,6 +118,33 @@ module Hanami
     App.new(configuration, environment)
   end
 
+  # Check if an application is allowed to load.
+  #
+  # The list of applications to be loaded can be set via the `HANAMI_APPS`
+  # env variable. If the HANAMI_APPS env variable is not set, it defaults
+  # to loading all applications.
+  #
+  # @return [TrueClass,FalseClass] the result of the check
+  #
+  # @since x.x.x
+  #
+  # @example
+  #
+  #   # Mount hanami app for specific app
+  #   Hanami.configure do
+  #     if Hanami.app?(:web)
+  #       require_relative '../apps/web/application'
+  #       mount Web::Application, at: '/'
+  #     end
+  #   end
+  #
+  def self.app?(app)
+    return true unless ENV.key?('HANAMI_APPS')
+
+    allowed_apps = ENV['HANAMI_APPS'].to_s.split(',')
+    allowed_apps.include?(app.to_s.downcase)
+  end
+
   # Return root of the project (top level directory).
   #
   # @return [Pathname] root path
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -506,7 +506,7 @@ OUT
 
         RSpec::Support::Env['HANAMI_APPS'] = 'web'
 
-        server do
+        server(port: 9999) do
           visit "/"
           expect(page).to have_title("Hanami | The web, with simplicity")
 
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -492,4 +492,66 @@ OUT
       run_command 'hanami server --help', output
     end
   end
+
+  context "with HANAMI_APPS ENV variable" do
+    it "loads only specific application" do
+      with_project do
+        generate "app admin"
+
+        remove_line "config/environment.rb", "require_relative '../apps/admin/application'"
+        remove_line "config/environment.rb", "mount Admin::Application"
+
+        remove_line "config/environment.rb", "require_relative '../apps/web/application'"
+        remove_line "config/environment.rb", "mount Web::Application"
+
+        inject_line_after "config/environment.rb", "Hanami.configure", <<-EOL
+  if Hanami.app?(:admin)
+    require_relative '../apps/admin/application'
+    mount Admin::Application, at: '/admin'
+  end
+
+  if Hanami.app?(:web)
+    require_relative '../apps/web/application'
+    mount Web::Application, at: '/'
+  end
+EOL
+        generate "action web home#index --url=/"
+
+        rewrite "apps/web/controllers/home/index.rb", <<-EOF
+module Web::Controllers::Home
+  class Index
+    include Web::Action
+
+    def call(params)
+      self.body = "app: web"
+    end
+  end
+end
+EOF
+        generate "action admin home#index --url=/"
+
+        rewrite "apps/admin/controllers/home/index.rb", <<-EOF
+module Web::Controllers::Home
+  class Index
+    include Admin::Action
+
+    def call(params)
+      self.body = "app: admin"
+    end
+  end
+end
+EOF
+
+        RSpec::Support::Env['HANAMI_APPS'] = 'web'
+
+        server do
+          visit "/"
+          expect(page).to have_content("app: web")
+
+          visit "/admin"
+          expect(page).to have_content("Not Found")
+        end
+      end
+    end
+  end
 end
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -498,7 +498,6 @@ OUT
       with_project do
         generate "app admin"
 
-
         replace "config/environment.rb", "require_relative '../apps/admin/application'", ""
         replace "config/environment.rb", "mount Admin::Application, at: '/admin'", "mount :admin, at: '/admin' do\nrequire_relative '../apps/admin/application'\nAdmin::Application\nend"
 
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -492,28 +492,4 @@ OUT
       run_command 'hanami server --help', output
     end
   end
-
-  context "with HANAMI_APPS ENV variable" do
-    it "loads only specific application" do
-      with_project do
-        generate "app admin"
-
-        replace "config/environment.rb", "require_relative '../apps/admin/application'", ""
-        replace "config/environment.rb", "mount Admin::Application, at: '/admin'", "mount :admin, at: '/admin' do\nrequire_relative '../apps/admin/application'\nAdmin::Application\nend"
-
-        replace "config/environment.rb", "require_relative '../apps/web/application'", ""
-        replace "config/environment.rb", "mount Web::Application, at: '/'", "mount :web, at: '/' do\nrequire_relative '../apps/web/application'\nWeb::Application\nend"
-
-        RSpec::Support::Env['HANAMI_APPS'] = 'web'
-
-        server(port: 9999) do
-          visit "/"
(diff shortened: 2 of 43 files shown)


AlfonsoUceda:
@davydovanton I like it, maybe it everything goes well, I think we should change container generator template for using it like your example.

AlfonsoUceda:
@jodosha what do you think? as I said, I like the idea.

TiteiKo:
I like the idea, but:
* one could want to run the project without any app (think console or background workers)
* I think it should be way more explicit

davydovanton:
@TiteiKo 
1. I think we can run server without any apps
2. do you have any ideas how to improve this?

davydovanton:
@TiteiKo and about the first point:
I checked it now and you can use `HANAMI_APPS=nothing hanami s` for start hanami server without any apps.

Also, about background processing: usually they use own servers (like `bundle exec sidekiq` or `QUEUE=file_serve rake resque:work`). That's why I don't understand why we need this but it works well 🤔

TiteiKo:
@davydovanton 

About background processing: as the apps loading process is done at the same time as the lib and env, background workers startup will boot Hanami and require the configuration file. And as they don't serve web pages, they don't need the apps, so thats why we need to be able to load 0 apps.

I hadn't thought about putting one app name that didn't match an actual app. Dirty but efficient... 

To make it cleaner I propose something that would look like this:

```ruby
Hanami.configure do
  if Hanami.app?(:web)
  end
end
```

Adding 2 methods that could be used inside the project:

```ruby
def self.app?(name)
  apps.include?(name.to_s)
end

def self.apps
  @apps ||= ENV['HANAMI_APPS'].to_s.split(',')
end
```

Anyway, it's not `.mount` job to check for this :)

davydovanton:
hm, @TiteiKo good idea with `.app?` 👍

davydovanton:
@jodosha @TiteiKo okay. I created `Hanami.app?` method. Added specs for this method and also updated CLI spec. WDYT now is it good for merge?

lib/hanami.rb
```diff
@@ -118,6 +118,30 @@ def self.app
     App.new(configuration, environment)
   end
 
+  # Check to see if application allowed to load.
+  #
+  # You can set allowed application in `HANAMI_APPS` env variable.
+  # If `HANAMI_APPS` variable is emply - any apps allowed.
```
AlfonsoUceda: s/emply/empty

any allowed apps.

lib/hanami.rb
```diff
@@ -118,6 +118,30 @@ def self.app
     App.new(configuration, environment)
   end
 
+  # Check to see if application allowed to load.
+  #
+  # You can set allowed application in `HANAMI_APPS` env variable.
```
AlfonsoUceda: s/application/applications

lib/hanami.rb
```diff
@@ -118,6 +118,30 @@ def self.app
     App.new(configuration, environment)
   end
 
+  # Check to see if application allowed to load.
```
AlfonsoUceda: Check if an application is allowed to load.

spec/integration/cli/server_spec.rb
```diff
@@ -492,4 +492,29 @@ def call(params)
       run_command 'hanami server --help', output
     end
   end
+
+  context "with HANAMI_APPS ENV variable" do
+    it "loads only specific application" do
+      with_project do
+        generate "app admin"
+
```
AlfonsoUceda: extra line

lib/hanami.rb
beauby: Nitpicking: "should be loaded" vs "is allowed to load". Thoughts?

lib/hanami.rb
beauby: Suggestion: "The list of applications to be loaded can be set via the `HANAMI_APPS` env variable."

lib/hanami.rb
beauby: Suggestion: "If the `HANAMI_APPS` env variable is not set, it defaults to loading all applications."

Which brings the question of differentiating between explicitly setting `HANAMI_APPS=""` vs not setting it. It might be worth allowing to load no app at all.

lib/hanami.rb
```diff
@@ -118,6 +118,30 @@ def self.app
...
+  #
+  # You can set allowed application in `HANAMI_APPS` env variable.
+  # If `HANAMI_APPS` variable is emply - any apps allowed.
+  #
+  # @return [TrueClass,FalseClass] the result of the check
```
beauby: Not sure "the result of the check" adds much value.

lib/hanami.rb
```diff
@@ -118,6 +118,30 @@ def self.app
...
+  # @since x.x.x
+  #
+  # @example
+  #
+  #   # Mount hanami app for specific app
+  #   Hanami.configure do
+  #     if Hanami.app?(:web)
+  #       require_relative '../apps/web/application'
+  #       mount Web::Application, at: '/'
+  #     end
```
beauby: I suspect this chunk of code will be repeated, and might benefit from being extracted into some macro `load_app(name, path)` (naming could be improved though).

lib/hanami.rb
```diff
@@ -118,6 +118,30 @@ def self.app
...
+  #   end
+  #
+  def self.app?(app)
+    allowed_apps = ENV['HANAMI_APPS'].to_s.split(',')
+    allowed_apps.empty? || allowed_apps.any? { |a| app.to_s.downcase[a] }
+  end
```
beauby: As mentioned above, 
> Which brings the question of differentiating between explicitly setting HANAMI_APPS="" vs not setting it. It might be worth allowing to load no app at all.

This method would become:
```ruby
def self.app?(app)
  return false unless ENV.key?('HANAMI_APPS')
  # ...
end
```

lib/hanami.rb
```diff
@@ -118,6 +118,30 @@ def self.app
...
+  #   # Mount hanami app for specific app
+  #   Hanami.configure do
+  #     if Hanami.app?(:web)
+  #       require_relative '../apps/web/application'
+  #       mount Web::Application, at: '/'
+  #     end
+  #   end
+  #
+  def self.app?(app)
+    allowed_apps = ENV['HANAMI_APPS'].to_s.split(',')
```
beauby: Might be worth extracting this to a cached method:
```ruby
def self.apps
  @allowed_apps ||= ENV['HANAMI_APPS'].to_s.split(',')
end
```

lib/hanami.rb
davydovanton: I like idea with allowed 👍

lib/hanami.rb
davydovanton: In future. Now I want to merge this feature and if we need we can add all helpers

lib/hanami.rb
davydovanton: I don't like this idea because it's instance variable in Model scope 😞

lib/hanami.rb
davydovanton: I think we should say that: "if HANAMI_ENV missing it loads all apps, if HANAMI_ENV='' it loads nothing". WDYT?

lib/hanami.rb
davydovanton: and with this cache we will create global state

davydovanton:
@AlfonsoUceda @beauby thanks for your review ❤️ 
I'll update docs, logic and specs for new logic 👍

lib/hanami.rb
beauby: This state is inherently global as we rely on an environment variable. Am I missing something?

lib/hanami.rb
```diff
@@ -118,6 +118,33 @@ def self.app
...
+  #       require_relative '../apps/web/application'
+  #       mount Web::Application, at: '/'
+  #     end
+  #   end
+  #
+  def self.app?(app)
+    return true unless ENV.key?('HANAMI_APPS')
+
+    allowed_apps = ENV['HANAMI_APPS'].to_s.split(',')
+    allowed_apps.any? { |a| app.to_s.downcase[a] }
```
beauby: My comment somehow disappeared but what's the rationale behind `allowed_apps.any? { |a| app.to_s.downcase[a] }` vs `allowed_apps.include?(app.to_s.downcase)`?

davydovanton: great suggestion 👍

lib/hanami.rb
beauby: I believe the current wording implies this sufficiently clearly but if you feel it's not then some other might as well, so up you.

lib/hanami.rb
davydovanton: hm, let's wait for @jodosha comment :)

lib/hanami.rb
beauby: Also, try to avoid force-pushing commits because it messes up the reviews :)

davydovanton: okay, sorry 😅

lib/hanami.rb
beauby: Nitpicking: you can remove the "to see".

lib/hanami.rb
```diff
@@ -118,6 +118,33 @@ def self.app
...
+  #       require_relative '../apps/web/application'
+  #       mount Web::Application, at: '/'
+  #     end
+  #   end
+  #
+  def self.app?(app)
+    return true unless ENV.key?('HANAMI_APPS')
+
+    allowed_apps = ENV['HANAMI_APPS'].to_s.split(',')
+    allowed_apps.include?(app.to_s.downcase)
```
jodosha: @davydovanton You should use `Utils::String#underscore` here: `CamelCaseApp` is referenced as `camel_case_app` by convention.

davydovanton: Yep, but why we need to underscore? We compare `HANAMI_APSS` values with the argument of `app?` method. Also, I think we can drop `#downcase` call too 🤔

jodosha: @davydovanton You're right, I got confused by `downcase`.

One thing we should still make clear. if the code won't return `true` for this:

```ruby
Hanami.app?("CamelCaseApp") # => false
Hanami.app?("camel_case_app") # => true
```

Where is gonna to be defined this rule? Do we want to document this? 😉

davydovanton: I think we should not support this cases. Only downcased app names. And yep, we should document this :)

davydovanton:
great work everyone 🎉

PR #783, opened 2017-05-16
Correctly removing migration file when model destroying

jarosluv:
When I generate and then destroy model through generator, migration file doesn't removed.

Please, see [issue](https://github.com/hanami/hanami/issues/784).

With this fix we have identical names of migration file for generating and destroying.

### DIFF
--- a/lib/hanami/commands/generate/model.rb
+++ b/lib/hanami/commands/generate/model.rb
@@ -103,9 +103,24 @@ module Hanami
         # @since 0.9.1
         # @api private
         def migration_path
+          existing_migration_path || new_migration_path
+        end
+
+        # @since 1.1.0
+        # @api private
+        def new_migration_path
           timestamp = Time.now.utc.strftime(Migration::TIMESTAMP_FORMAT)
-          filename = Migration::FILENAME_PATTERN % { timestamp: timestamp, name: "create_#{table_name}"}
+          filename = Migration::FILENAME_PATTERN % { timestamp: timestamp, name: migration_name_underscored }
+
           Hanami::Model.configuration.migrations.join(filename)
+        end
+
+        # @since 1.1.0
+        # @api private
+        def existing_migration_path
+          Utils::FileList["#{Hanami::Model.configuration.migrations}/*.rb"].find do |filename|
+            %r{/\d{14}_#{migration_name_underscored}.rb\z} =~ filename
+          end
         end
 
         # @since 0.5.0
@@ -125,6 +140,12 @@ module Hanami
         # @api private
         def model_name_underscored
           input
+        end
+
+        # @since 1.1.0
+        # @api private
+        def migration_name_underscored
+          "create_#{table_name}"
         end
       end
     end
--- a/README.md
+++ b/README.md
@@ -52,7 +52,9 @@ You can give back to Open Source, by supporting Hanami development via a [donati
 
 ### Supporters
 
+  * [Trung Lê](https://github.com/joneslee85)
   * [James Carlson](https://github.com/jxxcarlson)
+  * [Creditas](https://www.creditas.com.br/)
 
 ## Contact
 
--- a/spec/integration/cli/destroy/model_spec.rb
+++ b/spec/integration/cli/destroy/model_spec.rb
@@ -5,13 +5,12 @@ RSpec.describe "hanami destroy", type: :cli do
     it "destroys model" do
       with_project do
         generate "model user"
-        _migration = Dir.glob(Pathname.new("db").join("migrations", "*_create_users.rb")).first.to_s
+        migration = Dir.glob(Pathname.new("db").join("migrations", "*_create_users.rb")).first.to_s
 
         output = [
           "remove  lib/bookshelf/entities/user.rb",
           "remove  lib/bookshelf/repositories/user_repository.rb",
-          # FIXME: with Hanami 1.1
-          # "remove  #{migration}",
+          "remove  #{migration}",
           "remove  spec/bookshelf/entities/user_spec.rb",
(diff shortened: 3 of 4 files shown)


kaikuchn:
Isn't it actually wrong to destroy the migrations?
I guess most of the time one uses the destroy action because of a typo or one wants to rename things. In that case it may be okay to just remove all the files.
However, if it is about removing everything that pertains to a model, irrespective of when one wants to do that, we should actually generate a migration that drops the table, instead of removing the `create_table`.

WDYT?

jarosluv:
@kaikuchn 

I think this is actually bug, since generator still tries to remove the migration file, but with wrong timestamp. Plus `destroy` action must be full opposite for `generate`.

The most use case, as you say, is destroying model because of a typo or wishing to rename. And removing of model after a long time with a generator is a totally bad practice.

jarosluv:
@AlfonsoUceda 

Thanks for your answer!

I'm happy to finish it soon.

jarosluv:
@AlfonsoUceda 

Sorry, but I looked at the code and didn't understand what is the "test with unknown migration file"?

AlfonsoUceda:
@jarosluv this is the file: https://github.com/hanami/hanami/blob/master/spec/integration/cli/destroy/model_spec.rb#L13

jarosluv:
@AlfonsoUceda 

I understood this moment. But next you write about additional `test with unknown migration file`.
I already fixed line 13, but I have no idea about your second request.

AlfonsoUceda:
@jarosluv I just wanted to have a test to know what happens if the migration doesn't exist, Imagine you create the model and the migration, run the migrations, and in the future for some reason you deleted the migration and then run the destroy command, it shouldn't fail, isn't it?

jarosluv:
@AlfonsoUceda 

Thanks for your explaining!

Done! Please, review new changes.

lib/hanami/commands/generate/model.rb
```diff
@@ -103,11 +103,24 @@ def repository_path
...
-          filename = Migration::FILENAME_PATTERN % { timestamp: timestamp, name: "create_#{table_name}"}
+          filename = Migration::FILENAME_PATTERN % { timestamp: timestamp, name: migration_name_underscored }
+
           Hanami::Model.configuration.migrations.join(filename)
         end
 
+        # @since 1.1.0
+        # @api private
+        def existing_migration_path
+          Utils::FileList["#{Hanami::Model.configuration.migrations}/[0-9]*_#{migration_name_underscored}.rb"].first
```
kaikuchn: FileList is a wrapper around `Dir.glob` which takes a pattern that is not a regex, but more like a shell glob. (See the [docs](https://ruby-doc.org/core-2.4.0/Dir.html#method-c-glob))

The way you wrote this pattern one is tempted to think that it would only match if a sequence of 0 or more numbers is followed by `_` and the underscored migration name `.rb`. However, it would in fact match anything that starts with one number and ends with the migration name. E.g., `20170420080151_revert_create_products.rb` would match too if the underscored migration name were `create_products`.

kaikuchn: Since you're using `first` on the result you'd get the oldest migration which probably is always the one that was created by the generate model statement. Therefore this probably is not a real-world concern.

jarosluv: @kaikuchn 

Thanks for your thoughts. This code was here before my changes, but you're right - this is little mismatch. 

We can do something like 
```
Hanami::Utils::FileList["db/migrations/*.rb"].find { |filename| %r{/\d{14,}_create_campaigns.rb\z} =~ filename }
```

This is looks more complex, but works more accurately.

kaikuchn: I'm all for it. @AlfonsoUceda wdyt?

mereghost: Actually the second version has a higher overhead the whole file tree is read and then searched for the right file. The original (which albeit weird works perfectly) let the OS handle the matching meaning less objects and probably less time spent on IO.

kaikuchn: No, the globbing is a simple `*.rb`, that's non-recursive. Granted, it returns more strings (all migrations) than the previous version, but I'd argue that it is a negligible allocation overhead.

jarosluv: It's not such big overhead, because this is one time operation, not a part of high performance server side core.

AlfonsoUceda: @jarosluv I prefer the second version ;)

jarosluv:
@AlfonsoUceda 

All edits are done. :)

jarosluv:
@AlfonsoUceda 

Sorry, but what can I do for merging my pull request? CI for some reasons doesn't pass. And I don't know why.

AlfonsoUceda:
@jarosluv there is a problem with one test, and it is because of this change. The code works but I need to research why it's failing.

kaikuchn:
The only side effects of this change should be that the regex match sets global variables (`$&`, `$~` and `$n` where n is the match number). How that could result in the other test failing.. I have no clue. Maybe that isn't even the cause.

jarosluv:
But no tests are failed. It's so weird. What can I do for solving this problem?

kaikuchn:
The problem ist that `/home/travis/build/hanami/hanami/spec/isolation/rake/environment_spec.rb` hangs indefinitely.

AlfonsoUceda:
exactly @kaikuchn

jarosluv:
@AlfonsoUceda @kaikuchn Thanks for your thoughts!

I've just changed only the name of model in failing test and it all passed.

PR #806, opened 2017-07-17
Add some examples for Hanami::Components

davydovanton:


jodosha:
@davydovanton `Hanami::Components` is private API. It's gonna be renamed internally soon in order to make it available as public API (see the API docs). We also need to standardize the name of the internal components before to make it public.

I'm fine to merge those examples, but keep in mind that this is subject to change. 👍

davydovanton:
yes, I know that it's private. I just thinking that examples will be helpful for any methods and we can change it in future :)

PR #808, opened 2017-07-22
CLI cleanup

jodosha:
### Notable changes

  * Introduces the latest changes from `hanami-cli`
  * Removes `thor` as dependency
  * Migrates all the commands as "components": they can declare dependencies. Eg. `hanami db migrate` depends on the `model.configuration` component. See `Hanami::Cli::Commands::Command` for details
  * Adds more control on errors and exit codes (eg. `Kernel.exit(1)`)
  * Let third party developers can add or override commands. (See **Third party CLI** section below)
  * Adds extended `--help` banner with examples. (See **Command banner** section below)
  * Introduces new file manipulation helpers from `hanami-utils`.
  * Introduces a centralized module to determine where the files should be generate: Eg. where an action for the app `web` should go? `Hanami::Cli::Commands::Project` gives you the answer via: `.action` function

#### Third party CLI

Let's imagine someone wants to implement a `hanami-webpack` gem. Because of this PR, they can now do:

```ruby
module Hanami
  module Webpack
    module CLI
      class Generate < Hanami::Cli::Commands::Command
        desc "Generate Webpack configuration"
        def call(*)
          # ...
        end
      end
    end
  end
end

Hanami::Cli.register "generate webpack", Hanami::Webpack::CLI::Generate
```

From a `bookshelf` project that bundles `hanami-webpack`:

```shell
% bundle exec hanami generate
Commands:
  hanami generate action APP ACTION                    # Generate an action for app
  hanami generate app APP                              # Generate an app
  hanami generate mailer MAILER                        # Generate a mailer
  hanami generate migration MIGRATION                  # Generate a migration
  hanami generate model MODEL                          # Generate a model
  hanami generate secret [APP]                         # Generate session secret
  hanami generate webpack                              # Generate Webpack configuration
```

#### Command banner

Here's some examples of the new `--help` banners

```shell
% bundle exec hanami new --help
Command:
  hanami new

Usage:
  hanami new PROJECT

Description:
  Generate a new Hanami project

Arguments:
  PROJECT             	# REQUIRED The project name

Options:
  --database=VALUE, -d VALUE      	# Database (mysql/mysql2/postgresql/postgres/sqlite/sqlite3), default: "sqlite"
  --application-name=VALUE        	# App name, default: "web"
  --application-base-url=VALUE    	# App base URL, default: "/"
  --template=VALUE                	# Template engine (erb/haml/slim), default: "erb"
  --test=VALUE                    	# Project testing framework (minitest/rspec), default: "minitest"
  --[no-]hanami-head              	# Use Hanami HEAD (true/false), default: false
  --help, -h                      	# Print this help

Examples:
  hanami new bookshelf                     # Basic usage
  hanami new bookshelf --test=rspec        # Setup RSpec testing framework
  hanami new bookshelf --database=postgres # Setup Postgres database
  hanami new bookshelf --template=slim     # Setup Slim template engine
  hanami new bookshelf --hanami-head       # Use Hanami HEAD
```

```shell
% bundle exec hanami server --help
Command:
  hanami server

Usage:
  hanami server

Description:
  Start Hanami server (only for development)

Options:
  --server=VALUE                  	# Force a server engine (eg, webrick, puma, thin, etc..)
  --host=VALUE                    	# The host address to bind to
  --port=VALUE, -p VALUE          	# The port to run the server on
  --debug=VALUE                   	# Turn on debug output
  --warn=VALUE                    	# Turn on warnings
  --daemonize=VALUE               	# Daemonize the server
  --pid=VALUE                     	# Path to write a pid file after daemonize
  --[no-]code-reloading           	# Code reloading, default: true
  --help, -h                      	# Print this help

Examples:
  hanami server                     # Basic usage (it uses the bundled server engine)
  hanami server --server=webrick    # Force `webrick` server engine
  hanami server --host=0.0.0.0      # Bind to a host
  hanami server --port=2306         # Bind to a port
  hanami server --no-code-reloading # Disable code reloading
```

```shell
% bundle exec hanami db prepare --help
Command:
  hanami db prepare

Usage:
  hanami db prepare

Description:
  Drop, create, and migrate the database (only for development/test)

Options:
  --help, -h                      	# Print this help
```

---

### Technical notes

Most of the new code don't have API docs yet and need smaller, subsequent refactorings. But I decided to stop at a "good enough" point and revisit it during the 1.1.beta period.

This PR points to non-stable branches of `hanami-utils` and `hanami-cli`. It's my duty to finish the work on these branches, but I wanted to give the team the opportunity to review the `hanami` code in the meantime.


lib/hanami/cli/commands/new.rb
```diff
@@ -0,0 +1,457 @@
...
+        end
+
+        def generate_app(context)
+          Hanami::Cli::Commands::Generate::App.new.call(app: context.application_name, application_base_url: context.application_base_url, **context.options)
+        end
+
+        def init_git
+          return if git_dir_present?
+
+          say(:run, "git init . from \".\"")
```
AlfonsoUceda: I think we need to add `target` here, isn't it?

spec/integration/cli/console_spec.rb
```diff
@@ -48,6 +59,45 @@ class AccessToken < Dry::Struct
     end
   end # irb
 
+  xit "returns error when known engine isn't bundled" do
```
AlfonsoUceda: doesn't it work for now?

spec/integration/cli/generate/secret_spec.rb
```diff
@@ -19,5 +19,36 @@
         end
       end
     end
+
+    xit 'prints help message' do
```
AlfonsoUceda: remove the pending?

spec/integration/cli/generate_spec.rb
```diff
@@ -0,0 +1,17 @@
+RSpec.describe "hanami generate", type: :cli do
```
AlfonsoUceda: I think you need to add a test using the alias?

spec/integration/cli/destroy_spec.rb
```diff
@@ -0,0 +1,16 @@
+RSpec.describe "hanami destroy", type: :cli do
```
AlfonsoUceda: I think you need to add a test using the alias, isn't it?

lib/hanami/cli/commands/destroy/app.rb
```diff
@@ -0,0 +1,119 @@
...
+            files.delete_directory(destination)
+            say(:remove, destination)
+          end
+
+          def base_url(context)
+            content     = "mount #{context.app.classify}::Application"
+            destination = project.environment(context)
+
+            line  = files.read_matching_line(destination, content)
+            *, at = line.split(/at\:[[:space:]]*/)
```
kaikuchn: On the left of the assignment a destructuring is happening. If you used `_` then it would match for only one element of the array, `at` would then match the second element. `*` captures as many elements as it can, depending on what comes before and after it. In this case `*, at = ...` means that `at` will be the last element of the array and `*` captures all the rest. Since it doesn't have a name we discard it.
I think the style guide says to use `*_` in that case. But I'd argue that here we should just say `at = line.split(/at\:[[:space:]]*/).last` since that appears to be what we're interested in and it is a more revealing way of stating that.

AlfonsoUceda: @kaikuchn thank you!! IMO if we can replace that with your suggestion, it would improve the read of the code.

lib/hanami/cli/commands/db/migrate.rb
```diff
@@ -0,0 +1,33 @@
...
+      module Db
+        class Migrate < Command
+          requires "model.sql"
+
+          desc "Migrate the database"
+
+          argument :version, desc: "The target version of the migration (see `hanami db version`)"
+
+          example [
+            "               # Migrate to the last version",
```
jodosha: yup

jodosha:
@AlfonsoUceda I know there are some small things to cleanup, but as stated in the description, I stopped at a "good enough" point, as we need to get this merged to focus on other features for 1.1.beta1.

AlfonsoUceda:
@jodosha it’s good to me, I know there are sometimes we need to stop and split issues. Go ahead ;)

Issue #785, opened 2017-05-19
Error when db/migrations folder is not present


arni1981:
Steps to reproduce:
1. Manually remove `db/migrations` folder.
2. Run `bundle exec hanami g model user`

Fails with this error message:

> Errno::ENOENT: No such file or directory @ realpath_rec - /home/arni1981/www/hanami/sandbox/db/migrations

arni1981:
Perhaps I should have filed this in Hanami::Model, since it seems to boil down to this line 

[](https://github.com/hanami/model/blob/master/lib/hanami/model/configuration.rb#L75)

artofhuman:
@cllns What behavior for this issue. If we don't have migrations directory which setup in the config, should we create this directory or use project root?

cllns:
@arni1981 I think this belongs here, actually, since it's a generator problem.

@artofhuman I think we should create the `db/migrations/` folder, if it doesn't exist.

jodosha:
Fixed by #809

PR #809, opened 2017-07-25
Create migrations folder if not present

ddymko:
Hey, this PR relates to #785 which was saying that if your migrations folder is missing within the db folder it would throw an error when you would try to create a model. It indeed would throw an error since the directory does not exist.

I wasn't sure what the proper way of getting the path of the project is, so if there is a more defined way of getting it please let me know. (or if you think there is a better solution available)


cllns:
Thanks for this PR 🎉 !

Do you mind writing a spec for this? One that fails without this change, and passed with this change :) That way we can be sure this behavior won't break in the future.

ddymko:
@cllns yeah no problem

PR #811, opened 2017-07-28
Fix prepare mailer

### LINKED ISSUE #797: Mixin in mailer prepare is not included in mailers
Hi the following code does not work for me:

environment.rb:
```
  mailer do
    root 'lib/some-app/mailers'

    delivery :test

    # This should work
     prepare do
       include SomeAPP::Mailers::I18n
     end
  end
```

I can confirm that doing a direct include in a mailer works. For example:
```
module Mailers::Users
  class Welcome
    include Hanami::Mailer
    include SomeApp::Mailers::I18n
  end
end
```

AlfonsoUceda:
This PR fixes #797

### DIFF
--- a/spec/integration/mailer_spec.rb
+++ b/spec/integration/mailer_spec.rb
@@ -1,0 +1,30 @@
+RSpec.describe "Mailer", type: :cli do
+  it "use a mailer" do
+    with_project do
+      generate "mailer welcome"
+      write "lib/bookshelf/mailers/default_user.rb", <<-EOF
+module Mailers
+  module DefaultUser
+    def user_name
+      "Alfonso"
+    end
+  end
+end
+EOF
+
+      replace "config/environment.rb", "delivery :test", <<-EOF
+    delivery :test
+
+    prepare do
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
cc @jodosha

spec/integration/mailer_spec.rb
```diff
@@ -0,0 +1,30 @@
+RSpec.describe "Mailer", type: :cli do
+  it "use a mailer" do
```
AlfonsoUceda: I think I need to amend the test's name.

AlfonsoUceda:
cc @hanami/core

spec/integration/mailer_spec.rb
cllns: tiny grammar change: should be `it "uses..."`

jodosha:
@AlfonsoUceda Thank you. I've changed the target branch from `develop` to `master` and the milestone from `1.1.0` to `1.0.1` as we want to this fix to be shipped soon.

PR #812, opened 2017-07-31
Change CLI capitalization

cllns:
To reflect https://github.com/hanami/cli/pull/8, once it's merged.

I also changed the Gemfile here to use that branch (on `hanami/cli`) so the test suite can run and pass🤞 . That line should be changed upon merge, or immediately afterwards in a cleanup commit.


jodosha:
Technical note: I can't change the `hanami-cli` branch in `Gemfile`, because this PR uses a fork of this repository.

I'm gonna merge it, and then fix it. As result, we should see a CI failure, because the referenced branch doesn't exist anymore.

PR #813, opened 2017-07-31
Start hanami server in tests on random ports

davydovanton:
Regarding our gitter discussion, I set up random port value for each `hanami server` call in specs

### DIFF
--- a/spec/support/hanami_commands.rb
+++ b/spec/support/hanami_commands.rb
@@ -1,6 +1,7 @@
 require_relative 'bundler'
 require_relative 'files'
 require_relative 'retry'
+require_relative 'random_port'
 require 'hanami/utils/string'
 
 module RSpec
@@ -9,13 +10,16 @@ module RSpec
       private
 
       def rackup(args = {}, &blk)
-        bundle_exec "rackup -p 2300" do |_, _, wait_thr|
+        args[:port] ||= RandomPort.call
+
+        bundle_exec "rackup -p #{args[:port]}" do |_, _, wait_thr|
           exec_server_tests(args, wait_thr, &blk)
         end
       end
 
       def server(args = {}, &blk)
-        args[:port] ||= rand(2300..9999)
+        args[:port] ||= RandomPort.call
+
         hanami "server#{_hanami_server_args(args)}" do |_, _, wait_thr|
           exec_server_tests(args, wait_thr, &blk)
         end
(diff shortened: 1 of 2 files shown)


spec/support/hanami_commands.rb
```diff
@@ -15,6 +15,7 @@ def rackup(args = {}, &blk)
       end
 
       def server(args = {}, &blk)
+        args[:port] ||= rand(1000..9000)
```
cllns: 1000 thru 1024 won't work without sudo :) We should start at 1025

davydovanton: great point 👍

b264: It'd actually be ideal to use only prime numbers between 1025 and 9999 if there's a good way to code that.  Many folks may have other development servers running on ports like 3000, 5000, 8000, 8100, et cetera.  It's very unlikely the prime numbers will be used already, because it's not in human nature.

cllns: Good point, though I think primes might be overkill. Maybe we can skip anywhere the last digit is 0, or is a one or two digit number repeated? e.g. `4242`?

spec/support/hanami_commands.rb
```diff
@@ -15,6 +15,7 @@ def rackup(args = {}, &blk)
       end
 
       def server(args = {}, &blk)
+        args[:port] ||= rand(2300..9999)
```
b264: `require 'prime'`

```ruby
args[:port] ||= Prime.each(9999).to_a.map{|p| p if p > 2300}.compact.sample
```
could work and have a smaller chance of collision with another port being used

cllns: Ooh, neat. I didn't know about `Prime` :)

I like this. Could be shortened to:  `Prime.each(9999).select{ |p| p > 2300}.sample`

b264: Oh yeah.  I always forget about the ...`ect` methods

TiteiKo: I don't see the advantage here... Prime is costly and the select is costly...

As Prime will only return the prime numbers, there will be fewer ports to choose from, so a greater risk of it being used, isn't it?

Couldn't we instead just check if the port is being used and take another random port if it is?

davydovanton: I agree with @TiteiKo point here 👍

jodosha:
Folks, please review this alternative solution. 👍

PR #818, opened 2017-08-09
Send params to logger

TiteiKo:


cllns:
@TiteiKo Do you think we should filter some defaults: like `password` and `password_confirmation`? Some others we could include too: https://rollbar.com/docs/notifier/rollbar-gem/#data-sanitization-scrubbing

BenBach:
HI @TiteiKo .

Thanks for adding this functionality.

One question: Is it intended they JSON params are not logged? (JSON requests with type application/json and a json body)

jodosha:
@BenBach They aren't logged yet. We're planning to add this feature: https://github.com/hanami/hanami/issues/898

PR #820, opened 2017-09-01
Add some description for Hanamirc#file_options

dmitriy-strukov:


dmitriy-strukov:
@jodosha  please review

dmitriy-strukov:
@cllns can you please review ))?

lib/hanami/hanamirc.rb
```diff
@@ -140,7 +140,12 @@ def symbolize(hash)
       Utils::Hash.new(hash).symbolize!
     end
 
+    # Checks file existence and return parsed values or empty hash
```
davydovanton: `return` -> `returns`

davydovanton: WDYT about this: `Returns options from hanamirc file`?

dmitriy-strukov: @davydovanton thank you for review! Yea I think it better `Returns options from hanamirc file`

PR #821, opened 2017-09-04
Fix hanami destroy app description

profx5:


profx5:
http://hanamirb.org/guides/head/command-line/destroy/

lib/hanami/cli_sub_commands/destroy.rb
```diff
@@ -76,11 +76,11 @@ def model(name)
         end
       end
 
-      desc 'application NAME', 'Destroy an application'
+      desc 'app NAME', 'Destroy an application'
```
davydovanton: maybe is better change to `application_name`?

cllns:
I wonder if we should update this documentation instead of this? 

Should add this as an alias, rather than changing it? Changing the command would be a breaking change

profx5:
Ok, add alias is a good idea. 

It would be good that destroy command would have the same options as a generate command.

AlfonsoUceda:
@profx5 thanks for the fix but it isn't needed. CLI was rewrote with gem [hanami-cli](https://github.com/hanami/cli). You can check the code in develop branch: https://github.com/hanami/hanami/tree/develop/lib/hanami/cli/commands

jodosha:
@profx5 I want to inform you that we have fixed this in Hanami v1.1.0:

```shell
➜ bundle exec hanami version
v1.1.0.beta3

➜ bundle exec hanami destroy app --help
Command:
  hanami destroy app

Usage:
  hanami destroy app APP

Description:
  Destroy an app

Arguments:
  APP                 	# REQUIRED The application name (eg. `web`)

Options:
  --help, -h                      	# Print this help

Examples:
  hanami destroy app admin # Destroy `admin` app
```

profx5:
@jodosha Awesome!

Issue #603, opened 2016-06-27
Be able to rollback migrations


greggilbert:
So we have `hanami db migrate`. It'd be great if we could do `hanami db rollback` to rollback the most recent migration. I think right now you have to invoke Sequel directly to do any sort of rollbacks (or write a Rake script to do that), so having this would be a huge convenience.

jodosha:
@greggilbert This feature is already implemented: if you provide a version number, it "migrates down" to that version. Please have a look here: http://hanamirb.org/guides/command-line/database/#migrate

cllns:
Ah, thanks Luca! I searched and found nothing in `hanami/hanami` and `hanami/model` but not the guides. I'd still be in favor of `hanami db rollback` command, but I don't think it comes up often enough to be very important.

katafrakt:
Still no chance for `rollback` command? I keep forgetting it's not there and missing it during casual development.

cllns:
@katafrakt I'm in favor of adding `rollback`. 

What do others think @hanami/core?

b264:
To be clear, this is just syntactic sugar on for "db migrate (max minus one)", right?  Why not go all-in and use git-inspired `bundle exec rake db migrate HEAD~1`

AlfonsoUceda:
I think we could support it 👍

b264:
There is a distinct concept of `HEAD`: the latest migration.  It may not be called that, but the concept is very present.  In a chain of migrations, there is a newest one

 Also, what is the desired behaviour if there are new migrations that have never been run?  Say I wrote 3 new migrations that weren't ran, then executed `bundle exec hanami db rollback 2`  What is the outcome?  Does it migrate forward one (MAX-2 aka CURRENT+1) or does it migrate backward two (MAX-5 aka CURRENT-2)?  The latter seems more appropriate, I think

cllns:
True, the concept exists but the name doesn't.

I agree `bundle exec hanami db rollback 2` should do CURRENT-2 (so MAX-5)

jodosha:
We have a concept of `HEAD`, it's `hanami db version`.

AlfonsoUceda:
@jodosha if you don't mind, I'm working on this feature.

jodosha:
Implemented by https://github.com/hanami/hanami/pull/822

PR #822, opened 2017-09-11
Implement rollback command

AlfonsoUceda:
This PR implements #603 

Model PR: https://github.com/hanami/model/pull/439

As discussed in that issue, we support migrate a version using `hanami db migrate VERSION` but for better user experience, we are implementing `hanami db rollback` by default it will rollback only one version but you can pass a number `hanami db rollback 3`, it will rollback 3 versions.

IMO this will improve the user experience because you don't have to see the version, sometime you'll know to search all the version and then copy&paste, I think it's better make a simple rollback using a command for that purpose and using a number of steps.

cc @hanami/core

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,11 +12,11 @@ gem 'hanami-validations', '1.1.0.beta1', require: false, git: 'https://github.co
 gem 'hanami-router',      '1.1.0.beta1', require: false, git: 'https://github.com/hanami/router.git',      branch: 'develop'
 gem 'hanami-controller',  '1.1.0.beta1', require: false, git: 'https://github.com/hanami/controller.git',  branch: 'develop'
 gem 'hanami-view',        '1.1.0.beta1', require: false, git: 'https://github.com/hanami/view.git',        branch: 'develop'
-gem 'hanami-model',       '1.1.0.beta1', require: false, git: 'https://github.com/hanami/model.git',       branch: 'develop'
+gem 'hanami-model',       require: false, git: 'https://github.com/hanami/model.git',       branch: 'develop'
 gem 'hanami-helpers',     '1.1.0.beta1', require: false, git: 'https://github.com/hanami/helpers.git',     branch: 'develop'
 gem 'hanami-mailer',      '1.1.0.beta1', require: false, git: 'https://github.com/hanami/mailer.git',      branch: 'develop'
 gem 'hanami-assets',      '1.1.0.beta1', require: false, git: 'https://github.com/hanami/assets.git',      branch: 'develop'
-gem 'hanami-cli',         '0.1.0.beta1', require: false, git: 'https://github.com/hanami/cli.git',         branch: 'develop'
+gem 'hanami-cli',         require: false, git: 'https://github.com/hanami/cli.git',         branch: 'develop'
 
 platforms :ruby do
   gem 'sqlite3'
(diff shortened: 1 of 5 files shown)


lib/hanami/cli/commands/db/rollback.rb
```diff
@@ -0,0 +1,39 @@
...
+    module Commands
+      module Db
+        # @since x.x.x
+        # @api private
+        class Rollback < Command
+          requires "model.sql"
+
+          desc "Rollback the database"
+
+          option :steps, desc: "Steps to rollback the database", default: 1
```
kaikuchn: In an effort to achieve a ubiquitous language I'd suggest to decide on a term for the concept of version/step and stick to that one term. That has the added benefit of being less confusing for the uninitiated 😄 .

kaikuchn: Also you could just use the generic `:num` as a name for this option then it reads better if you were to decide on `version` as the name for the concept.

AlfonsoUceda: Thanks for the feedback ;) maybe I try to support 'hanami db rollback 2', then we can avoid use steps, version or whatever, but first I should implement default in arguments, what do you think.?

Gemfile
```diff
@@ -12,11 +12,11 @@ gem 'hanami-validations', '1.1.0.beta1', require: false, git: 'https://github.co
 gem 'hanami-router',      '1.1.0.beta1', require: false, git: 'https://github.com/hanami/router.git',      branch: 'develop'
 gem 'hanami-controller',  '1.1.0.beta1', require: false, git: 'https://github.com/hanami/controller.git',  branch: 'develop'
 gem 'hanami-view',        '1.1.0.beta1', require: false, git: 'https://github.com/hanami/view.git',        branch: 'develop'
-gem 'hanami-model',       '1.1.0.beta1', require: false, git: 'https://github.com/hanami/model.git',       branch: 'develop'
+gem 'hanami-model',       require: false, git: 'https://github.com/hanami/model.git',       branch: 'develop'
```
AlfonsoUceda: @jodosha can we merge this PR with this branches and update it after in develop branch?

lib/hanami/cli/commands/db/rollback.rb
```diff
@@ -0,0 +1,39 @@
...
+        # @api private
+        class Rollback < Command
+          requires "model.sql"
+
+          desc "Rollback the database"
+
+          argument :versions, desc: "Number of versions to rollback the database", default: 1
+
+          example [
+            "  # Rollback only one version (default)",
```
davydovanton: WDYT about: `Rollback only latest migration version` or something like this?

lib/hanami/cli/commands/db/rollback.rb
```diff
@@ -0,0 +1,39 @@
...
+            "2 # Rollbacks two versions"
+          ]
+
+          # @since x.x.x
+          # @api private
+          def call(versions:, **options)
```
davydovanton: we really need `options` here? Oo

AlfonsoUceda: nope, I'll remove it.

lib/hanami/cli/commands/db/rollback.rb
AlfonsoUceda: `Rollback lastest version`?

jodosha:
@AlfonsoUceda I merged it, thank you very much 👍

PR #825, opened 2017-09-15
Update with_shotgun_spec.rb

fewtaeyeon:


AlfonsoUceda:
Hi @fewtaeyeon why this change? Thanks

cllns:
@AlfonsoUceda I guess just to make the spec cleaner, to keep consistent indentation.

This means that file has a bunch of leading spaces, which is not how it would ever be for users, so I don't think this is a good change.

Since we only use Ruby 2.3+, we could use squiggly heredoc `<<~` here, which strips the leading space from the output string.

We have 40 (at least) files with HEREDOCs https://github.com/hanami/hanami/search?utf8=✓&q=EOF&type= though, so if we do that, we should do it across the board.

AlfonsoUceda:
@cllns I'd do that, I prefer change them all of them in one PR, that one file per PR, what do you think?

fewtaeyeon:
Sorry, I just want it to look cleaner so I does that and I agree with you @cllns  @AlfonsoUceda .

Thank you for comment and correcting me.

cllns:
@fewtaeyeon Do you want to submit a PR for changing all the HEREDOCS in this repo to be indented properly, with the squiggly heredoc? It's OK if not, just let us know :)

AlfonsoUceda:
I'm going to close this PR, @fewtaeyeon please can you follow @cllns's suggestion?

jodosha:
@fewtaeyeon If you're keen to, please update all the occurrences with @cllns suggestion. Thanks 👍

PR #828, opened 2017-09-18
Hanami logger creates automatically all folders for the stream

AlfonsoUceda:
When using `stream` option in the logger, we can use a path for our file but if that path doesn't exist, application will crash. 

This test ensure that logger will create the path. 

Logger PR: https://github.com/hanami/utils/pull/231


AlfonsoUceda:
Merging because this is just one line to remove, no review required.

PR #830, opened 2017-09-26
Cleanup generated readme

cllns:
1) Titleizes header from `# bookshelf` to `# Bookshelf`

2) Simplifies `db create` and `db migrate` as `db prepare`

@davydovanton I was going to fix the hardcoded `1.0.0` to the docs page at the bottom, but it looks like that's the only docs version we are generating. What are the plans for that? What versions of docs will we generate? (Maybe not alphas/betas/rcs, but just proper releases?)


davydovanton:
good job @cllns 👏

PR #832, opened 2017-09-27
Gemfile plugins

jodosha:
Introduce a new convention for Hanami plugin gems: the `:plugins` `Gemfile` group.

---

## The problem

With 1.1 we'll introduce a way to register custom CLI commands for third party gems.

Let's say we have a fictional `hanami-foo` gem, which wants to register a `hanami foo setup` command. This gem was added by a developer in their `bookshelf` project `Gemfile`.

When we start `hanami` command, we **don't** eager load all the gems in `Gemfile`, because it isn't efficient to do so.

But if we don't load the bundled gems, that means our `hanami-foo` isn't loaded as well, so the commands cannot be registered.

This is a chicken-egg problem.

## The solution

Instead of eager loading **all** the gems in `Gemfile`, we can load **only a subset** of them that is strictly necessary: the `:plugins` group.

Here's how should look the `Gemfile` of a `bookshelf` project:

```ruby
source "https://rubygems.org"

# ...

group :plugins do
  gem "hanami-foo"
end
```

When we'll run `bundle exec hanami` the command `hanami foo setup` will be listed in the output.

### DIFF
--- a/bin/hanami
+++ b/bin/hanami
@@ -2,4 +2,5 @@
 require 'bundler'
 require 'hanami/cli/commands'
 
+Bundler.require(:plugins)
 Hanami::CLI.new(Hanami::CLI::Commands).call
--- a/lib/hanami/environment.rb
+++ b/lib/hanami/environment.rb
@@ -391,6 +391,7 @@ module Hanami
     # @since 0.4.0
     # @api private
     def require_application_environment
+      Bundler.setup(*bundler_groups)
       require project_environment_configuration.to_s # if project_environment_configuration.exist?
     end
 
--- a/lib/hanami/rake_helper.rb
+++ b/lib/hanami/rake_helper.rb
@@ -17,7 +17,6 @@ module Hanami
     # @since 0.6.0
     # @api private
     #
-    # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
     def install
       desc "Load the full project"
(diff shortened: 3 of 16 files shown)


davydovanton:
@jodosha quick question: what I need to do if I want to add cli gem only for dev environment?

jodosha:
@davydovanton It's not possible at this time. 😢

AlfonsoUceda:
Removing `needs-documentation` label. You can check dock in: https://github.com/hanami/hanami.github.io/pull/399

jodosha:
@davydovanton I found a workaround for this problem using `install_if`: https://bundler.io/v1.10/man/gemfile.5.html#INSTALL_IF-install_if-

I have added an env variable in production: `HANAMI_STAGE="deployment"`. Please note that this name is  **arbitrary**, it can be anything.

Then In my `Gemfile`:

```
group :plugins do
  install_if -> { ENV["HANAMI_STAGE"] != "deployment" } do
    gem "hanami-reloader"
  end
end
```

It does its job, but I think it's too convoluted.

mereghost:
It looks hackish as hell, but that's beyond our power to fix. Stupid question: did passing a lambda that returns a boolean to require not work?

jodosha:
@mereghost 
> Stupid question: did passing a lambda that returns a boolean to require not work?

It should work, but the problem here isn't the require of a gem, but the installation. In the case of `hanami-reloader` it installs in production development gems like `guard`, which is unwanted.

Issue #823, opened 2017-09-13
How to use custom inflections with hanami generate commands


kaikuchn:
Hi,

@imago38 encountered [trouble using custom inflections](https://gitter.im/hanami/chat?at=59b815fd210ac269202212c3). He wanted to generate and use a model with a custom inflection rule ('stimulus' -> 'stimuli') and defined that inflection in a initialiser.

```ruby
# config/initializers/inflections.rb
Hanami::Utils::Inflector.inflections do
  exception 'stimulus', 'stimuli'
end
```

However, it seems that initialisers are not run when executing hanami commands. Therefore the custom inflection rule is not picked up by the inflector and the generated migration is wrong. I.e.:

```bash
❯ bundle exec hanami g model stimulus
      ...
      create  db/migrations/20170913072014_create_stimuluses.rb
      ...

❯ cat db/migrations/20170913072014_create_stimuluses.rb
Hanami::Model.migration do
  change do
    create_table :stimuluses do
    ...
```

This may be related to https://github.com/hanami/hanami/issues/787

If this is intentional (that initialisers are not run when executing hanami commands), then where should people put their custom inflection rules?

cllns:
Thanks for opening this. It's intentional that all of Hanami isn't loaded for generators. Inflectors are perfect a case where we'd want the initializers run though.

I just reproduced this both with `1.0` and `1.1` (which is a beta release at the moment, and includes a re-write of the CLI).

AlfonsoUceda:
thanks @kaikuchn, we aren't loading initializers but we can change that because it seems some users are experiencing the same problem.

kaikuchn:
I think it is good that we do not load all of Hanami for running generators. It is what makes them start up quickly and feel snappy. I love that and I think we should endeavour to keep it that way.

However, I think most people would assume that initializers are run whenever they do something with Hanami. So we either need to make very clear that they don't and point people to an alternative for running code that needs to execute even for hanami commands (documentation). Or we do run initializers, which probably requires us to load all of Hanami, which sucks.

What are your thoughts on this?

cllns:
I wonder if a workaround for this could be allowing people to specify the table name in the model generator?

```
# proposed, not real at the moment
bundle exec hanami g model stimulus --table-name=stimuli
```

Not ideal, since people would expect Inflections to work, but makes working around it simple enough. 
[...]

PR #834, opened 2017-09-28
Add --table-name options to model generator (Issue #823)

likeath:
Hello!
This commit provides  `--table-name` option to model generator as described in #823.


AlfonsoUceda:
@likeath thanks for the PR but it can't be merged, CLI was rewritten in develop branch, you should take that branch as base and then create a pull request against develop branch.

Can you do that? I'm going to close this PR.

PR #837, opened 2017-10-03
Use compact module definition in application_layout.rb generated by cli

unhappychoice:
Any other templates are using compact module definition like `Application::Views`.
So, I unified it.

FYI, Rubocop has a rule for module definition. And I was noticed by that.
https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/class_and_module_children.rb

### DIFF
--- a/lib/hanami/generators/app/views/application_layout.rb.tt
+++ b/lib/hanami/generators/app/views/application_layout.rb.tt
@@ -1,7 +1,5 @@
-module <%= config[:classified_app_name] %>
-  module Views
-    class ApplicationLayout
-      include <%= config[:classified_app_name] %>::Layout
-    end
+module <%= config[:classified_app_name] %>::Views
+  class ApplicationLayout
+    include <%= config[:classified_app_name] %>::Layout
   end
 end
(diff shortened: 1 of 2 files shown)


jodosha:
@unhappychoice Hi and thanks for this PR. We'll move on nested style in the future, so there isn't the need of this change.

The choice isn't arbitrary, but it's requested by Ruby constant lookup. With compact style the following code will crash:

```ruby
# apps/web/controllers/authentication.rb
module Web
  module Controllers
    module Authentication
      module Skip
        # ...
      end
    end
  end
end
```

```ruby
module Web::Controllers::Home
  class Index
    include Web::Action

     # Ruby will fail to find this module
    include Authentication::Skip

     # To make it to work, you should reference the full name constant
    include Web::Controllers::Authentication::Skip

    # ...
  end
end
```

This scenario above doesn't look good. If we use nested style, we can have a more natural syntax:

```ruby
module Web
  module Controllers
    module Home
      class Index
        include Web::Action

         # This will work now
        include Authentication::Skip

        # ...
      end
    end
  end
end
```

---

Also, to support compact style, we need to generate the `Web::Controllers` module dynamically during the boot process. Which is an unnecessary magic that we want to get rid of.

---

Thanks for understanding. 👍

unhappychoice:
Thank you for your detailed description :smiley:
I completely agree to that

Issue #838, opened 2017-10-03
hanami beta needs gemfile installed


dsisnero:
bin\hanami in beta requires bundler and uses Bundle.require.  The gem doesn't include a Gemfile so the command fails on windows

c:\Dominic\programming\ruby\hanami\hanami-1.1.0.beta2\bin>hanami new bookshelf
Bundler::GemfileNotFound: Could not locate Gemfile or .bundle/ directory
dir

c:\Dominic\programming\ruby\hanami\hanami-1.1.0.beta2\bin>gem list hanami

*** LOCAL GEMS ***

hanami (1.1.0.beta2)
hanami-assets (1.1.0.beta2)
hanami-cli (0.1.0.beta2)
hanami-controller (1.1.0.beta2)
hanami-helpers (1.1.0.beta2)
hanami-mailer (1.1.0.beta2)
hanami-router (1.1.0.beta2)
hanami-utils (1.1.0.beta2)
hanami-validations (1.1.0.beta2)
hanami-view (1.1.0.beta2)

From an unpacked hanami -- gem unpack hanami

 Directory of c:\Dominic\programming\ruby\hanami\hanami-1.1.0.beta2

10/03/2017  04:59 PM    <DIR>          .
10/03/2017  04:59 PM    <DIR>          ..
10/03/2017  04:59 PM    <DIR>          bin
10/03/2017  04:59 PM            24,064 CHANGELOG.md
10/03/2017  04:59 PM             3,230 CODE_OF_CONDUCT.md
10/03/2017  04:59 PM             9,163 FEATURES.md
10/03/2017  04:59 PM             1,792 hanami.gemspec
10/03/2017  04:59 PM    <DIR>          lib
10/03/2017  04:59 PM             1,071 LICENSE.md
10/03/2017  04:59 PM             4,694 README.md
               6 File(s)         44,014 bytes
               4 Dir(s)   8,635,719,680 bytes free

jodosha:
@dsisnero thanks for reporting it. I'm working on a fix.

jodosha:
Another case reported by @malin-as in [chat](https://gitter.im/hanami/chat?at=59d479bcb59d55b8236e911f):

```shell
$ echo '2.4.2' > .ruby-version
$ echo 'hanami-beta2' > .ruby-gemset
$ cd .
ruby-2.4.2 - #gemset created /home/malina/.rvm/gems/ruby-2.4.2@hanami-beta2
[...]

PR #839, opened 2017-10-04
Don't require `:plugins` `Gemfile` group when running `hanami new`

jodosha:
Fixes: https://github.com/hanami/hanami/issues/838

### DIFF
--- a/bin/hanami
+++ b/bin/hanami
@@ -2,5 +2,5 @@
 require 'bundler'
 require 'hanami/cli/commands'
 
-Bundler.require(:plugins)
+Bundler.require(:plugins) if File.exist?(ENV["BUNDLE_GEMFILE"] || "Gemfile")
 Hanami::CLI.new(Hanami::CLI::Commands).call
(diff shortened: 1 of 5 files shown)


bin/hanami
```diff
@@ -2,5 +2,5 @@
 require 'bundler'
 require 'hanami/cli/commands'
 
-Bundler.require(:plugins)
+Bundler.require(:plugins) if File.exist?(ENV["BUNDLE_GEMFILE"] || "Gemfile")
```
davydovanton: WDYT maybe is better to wrap this line by method `load_plugins`?

jodosha: @davydovanton I considered that too, but I'm not sure yet about that API. `load_plugins` looks like a complete plugins system, which is not there yet.

davydovanton: @jodosha agree. We can fix it in future (of course, if we need it)

cllns:
Not necessary right now, but we shouldn't rely on `Gemfile`, hard-coded, since bundler already supports using `gems.rb` instead and apparently that will be the default in the future. (Will have to other changes in Hanami too, most likely)

jodosha:
@cllns My idea is to make Hanami 2.0 to support `gems.rb`.

PR #840, opened 2017-10-05
Add routing helpers doc

dmitriy-strukov:
@davydovanton Hi! Can you please review?


jodosha:
@dmitriy-strukov Hi and thanks for this PR. Because that helper is generated with meta-programming (via `.included`) this isn't an effective API doc.

It also exposes low level implementation details, whereas the API doc should only indicate how to use a method, not how's implemented. See the API docs for that module.

I'm closing this. Thanks for understanding. 👍

dmitriy-strukov:
@jodosha Ok, thank you for review!

PR #843, opened 2017-10-07
Remove unused block argument

masatooba:
Remove unused block argument.


jodosha:
@masatooba Thanks for this PR, it looks good. Merging. 👍 

PS: please don't focus on the CLI implementation that we have right now in `master` branch, as we rewrote it (see `develop`) branch.

PR #844, opened 2017-10-10
Add type tags to rspec templates

unhappychoice:
Like [rspec-rails](https://github.com/rspec/rspec-rails), I added `type` tags to rspec templates which is generated by `hanami generate` command.
 
I think users can include shared_contexts, shared_examples or helpers, by using this tag.

### DIFF
--- a/lib/hanami/cli/commands/generate/action/action_spec.rspec.erb
+++ b/lib/hanami/cli/commands/generate/action/action_spec.rspec.erb
@@ -1,4 +1,4 @@
-RSpec.describe <%= app.classify %>::Controllers::<%= classified_controller_name %>::<%= action.classify %> do
+RSpec.describe <%= app.classify %>::Controllers::<%= classified_controller_name %>::<%= action.classify %>, type: :action do
   let(:action) { described_class.new }
   let(:params) { Hash[] }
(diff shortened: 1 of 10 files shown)


davydovanton:
I like this idea 👍 
In future we can update rspec-hanami for load only type specific matchers.
@hanami/core WDYT?

AlfonsoUceda:
@unhappychoice thanks for this PR, I agree too, I think it's useful to provide a foundation of rspec tags

unhappychoice:
Good to know that. 👍 
I wiil make the branch based  on `develop` branch 😄

unhappychoice:
BTW, which do you like to use `type: :action` or `type: :controller` for an action class ?
I'm currently using `type: :action`, but `:controller` also sounds good :eyes:

What do you think?

davydovanton:
@unhappychoice I think `:controller` will more understandable 🤔

AlfonsoUceda:
IMO we shouldn't use controllers, we don't have controller word in hanami we use actions, so I think `action` is ok. cc @jodosha

jodosha:
@unhappychoice This looks good now. Merging 👍

PR #853, opened 2017-10-25
Set release date of version 1.1.0 in FEATURES file

cveneziani:
Looks like it's been released :)
http://hanamirb.org/blog/2017/10/25/annoucing-hanami-110.html

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -3,7 +3,7 @@
 
 ## Features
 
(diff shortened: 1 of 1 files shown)


jodosha:
@cveneziani Yes, it still have a vivid memory of that day. It felt like yesterday.. 😜 
Thank you Cecile for all your help. ✨

Issue #851, opened 2017-10-25
Mailer generator - options template output is broken


cllns:
In a new app, with `1.1.0.rc1`, running: 

`bundle exec hanami generate mailer blah --from=fromuser@example.com --to=touser@example.com`

Generates:
```
class Mailers::Boo
  include Hanami::Mailer

  from    fromuser@example.com
  to      touser@example.com
  subject 'Hello'
end
```

Which isn't valid Ruby. Instead, we should put the email addresses inside string quotes.

Also, adding one level of quotes doesn't fix it either. You have to do:
`bundle exec hanami generate mailer blah --from="'fromuser@example.com'" --to="'touser@example.com'"` in order to get it to work properly (or switch the order of the nested quotes).

AlfonsoUceda:
Fixed by #854

PR #854, opened 2017-10-26
Fix mailer generator when use or not quotes in options

AlfonsoUceda:
This PR implements #851


lib/hanami/cli/commands/generate/mailer.rb
```diff
@@ -87,6 +87,12 @@ def generate_html_template(context)
             files.touch(destination)
             say(:create, destination)
           end
+
+          # @since x.x.x
+          # @api privete
```
cllns: Should be `private`

Issue #857, opened 2017-10-29
Run bundle using option after new project is created.


AlfonsoUceda:
@citizen428 [proposed](https://github.com/hanami/hanami/issues/849#issuecomment-338906112) to run bundle after a new Hanami project is created.

cllns:
I think we should run `bundle install` in the new folder by default. And have a `--skip-bundle-install`. What do you think?

AlfonsoUceda:
I wouldn't do that because some people use gemsets, I think it's better to be conservative but I would like to hear more opinions.

citizen428:
I'm with @AlfonsoUceda on this one, there shouldn't be too many assumptions regarding people's environments and preferences built in.

jodosha:
Me and @cllns had a conversation about this. Probably it's not a good idea. People may want to fiddle with their `Gemfile` first. Or if they are newcomers, they may understand that Hanami is slow to generate a new project, but it's the wait for Rubygems instead.

As last thing, I'd defer this, because we're going probably to do more work on CLI hooks.

AlfonsoUceda:
Closing this issue because I think bundler should be run manually.

PR #858, opened 2017-10-30
WIP: Add  option for project generator #857

jarosluv:
Proof of concept for #857.

For now bundler runs `bundle install` command not in new project directory, as I think. Spec fails.


AlfonsoUceda:
@jarosluv thanks for this PR but for now we are decided to bundle should be run manually and don't add more options to CLI, maybe in the future we can reconsider this.

PR #861, opened 2017-11-12
Avoid using Hanami::Utils.reload! when code_reloading is disabled

kj:
Code reloading causes 'already initialized constants' warnings for me, due to the use of `load`, so I want to disable it. It appears the way to do this is to remove the shotgun gem, but the check that decides whether to use `Hanami::Utils.reload!` or `Hanami::Utils.require!` seems to be wrong. I think it should use either `Hanami::Components['code_reloading']` or `Hanami.code_reloading?`, and the latter is used elsewhere in this file for all other checks while this is the only exception (so I'm assuming it was an oversight). With this change, when shotgun is not available, it will use `Hanami::Utils.require!`.

Please correct me if I've misunderstood the way this is supposed to work!

### DIFF
--- a/lib/hanami/components/components.rb
+++ b/lib/hanami/components/components.rb
@@ -61,7 +61,7 @@ module Hanami
       run do
         directory = Hanami.root.join('lib')
 
-        if Components['environment'].code_reloading?
+        if Hanami.code_reloading?
           Utils.reload!(directory)
         else
           Utils.require!(directory)


jodosha:
Thanks for this PR, I'll have to test it manually ASAP. In the meantime, if you're interested into code reloading feature, please consider to try hanami-reloader gem.

It's an experiment that I'm running, as an alternative to shotgun. If it goes well, we'll integrate it in Hanami itself.

http://lucaguidi.com

kj:
Yeah I saw that gem while I was looking into this, and I like the idea behind it. I will definitely give it a try!

AlfonsoUceda:
hey @kj did you try `hanami-reloader` gem?

kj:
Hi @AlfonsoUceda, I've tried it briefly, and it seems to work well. I'm still getting the constant warnings I described above though, presumably due to the same issue causing it to use `load` instead of `require` even when shotgun in uninstalled. Personally, I've been using `entr` for my reloading which works well enough for me (I think this was the method used earlier in Hanami's life?). I guess most Rubyists will expect a Ruby-only solution though, so hanami-reloader probably makes the most sense for an official solution! It might be nice if it could remain a separate gem, even if it is included by default in new Hanami apps (like hanami-model etc.).

kj:
Hi @xvonabur, I don't use JRuby, just standard MRI, so I'm not sure if my approach will work or not. This is what I have in my shell startup file:

```
alias hanami-dev-server='while true; do find -name \*.rb | entr -dr bundle exec hanami server; done'
```

Note of course that this reloads the whole server whenever a file has changed (or been created/removed), so might not always be suitable depending on the impact of that.

kj:
@jodosha @AlfonsoUceda just checking if there was a reason this wasn't merged? It looks like it won't make it into 1.2 I guess.

jodosha:
@kj Sorry for the long wait. I'm manually testing it right now.

In which circumstances `Components['environment'].code_reloading?` has a different value from `Hanami.code_reloading?`.

I'm trying both with and w/o code reloading and both the methods are returning results that are coherent with the context.

---

Can you please tell me how to reproduce your issue (the warnings)? And also why did you decide to change that line? I can't see nothing different when using this PR.

kj:
No worries @jodosha, I'll see if I can get back to you with something reproducible.

kj:
@jodosha 

Okay, I have something you can test now. I didn't bother about the constant warnings, as I can't remember the exact situation I had originally. The following should show that the wrong branch is being executed in the code reloading check.

---

First, create a new Hanami app. Edit the Gemfile to remove the shotgun gem and set the hanami gem to some local path. Run `bundle install`.

Now, in the local hanami repo, edit lib/hanami/components/components.rb. You will notice that the conditional at line 64 implies that `Util.reload!` should only be called when code reloading is enabled:

    if Components['environment'].code_reloading?
      Utils.reload!(directory)
    else
      Utils.require!(directory)
    end

Edit it to raise an exception:

    if Components['environment'].code_reloading?
      raise 'code reloading should be disabled'

      Utils.reload!(directory)
    else
      Utils.require!(directory)
    end

Now run the app you created earlier. The exception is raised immediately, before the server starts, even though we removed shotgun. Now edit the file again, with the change from this pull request:

    if Hanami.code_reloading?
      raise 'code reloading should be disabled'

      Utils.reload!(directory)
    else
      Utils.require!(directory)
    end

Now the app starts, as expected, meaning it correctly used `Util.require!` instead of `Util.reload!`. You can reintroduce the shotgun gem now, and run the app again. This time the server starts, but when you try to load a page, the exception is raised as expected, meaning it is correctly using `Util.require!` when code reloading _is_ enabled.

jodosha:
@kj Thanks for the quick reply. I can reproduce it locally. Merging this and adding some code comments. This is hard to test and will be gone in a few months. With `hanami-reloader`, we're gonna remove code reloading knowledge from `hanami`.

PR #863, opened 2017-11-20
Overwrite when generating files

cllns:
Adding this PR to show how https://github.com/hanami/utils/pull/239 would be implemented in this repo. (I created another branch, `add-files-overwrite-flag` used here, with the same content, because the old one had an inaccurate name `fix-write-to-append`).

This PR fails when using the `master` branch, and it also failed when `overwrite: true` is not added. The value is `false` default. An exception is raised in that case, which causes `run_command` to have a non-zero exit status, which causes the `run_command` expectation to fail.

We _could_ add more tests for this same behavior, but this behavior (nuances of generating files) should moved to `Hanami::CLI`. It's not specific to `action` at all, but rather than base `Command`. Want to get this fixed first though, before extracting the generating code into that gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ end
 
 gem 'i18n'
 
-gem 'hanami-utils',       '~> 1.1', require: false, git: 'https://github.com/hanami/utils.git',       branch: '1.1.x'
+gem 'hanami-utils',       '~> 1.1', require: false, git: 'https://github.com/hanami/utils.git',       branch: 'add-files-overwrite-flag'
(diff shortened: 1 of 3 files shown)


cllns:
@AlfonsoUceda In the future, we'll have `overwrite: false` at first, then rescue the `FileAlreadyExistsException`. Then we can ask `Do you want to overwrite? [y/N,etc.]` and send `overwrite: true` if they say yes.

AlfonsoUceda:
Closing because right now it doesn't apply, PR https://github.com/hanami/utils/pull/239 has been merged with another solution.

PR #864, opened 2017-11-20
Use '<<~' and indent appropriately

cllns:
I added the `Style/IndentHeredoc` cop, but I think we should move this rule to https://github.com/hanami/devtools Can take care of that, just want to make sure this is something we want. 

I think indenting HEREDOC's is much more readable. We only support Ruby 2.3+, and this feature was added in 2.3.

I used `bundle exec rubocop --autocorrect` to do this, by the way :)


kj:
I think you can also indent the final EOF line when you use `<<~` (rather than setting it flush left), if that is desirable.

```
    my_string = <<~EOF
      example
    EOF

p my_string
# -> "example\n"
```

AlfonsoUceda:
@cllns thank you for this, it will make the code more readable, can you apply @kj suggestion? you can indent `EOF`.

I'm with you, that cop can be applied to all our gems, please can you take care of that? ;)

cllns:
Yes! I think I'll try to fix `rubocop`'s auto-correct for this, too, so we can be good Ruby ecosystem citizens :)

AlfonsoUceda:
@cllns can you finish this? Thanks man ;)

AlfonsoUceda:
I've changed base from `master` to `develop`

cllns:
@AlfonsoUceda all set!

I opened [a PR](https://github.com/bbatsov/rubocop/pull/5190) with Rubocop to fix this behavior, but I used my local version to clean this up so we can push this live sooner.

One thing that's possibly an issue is that this adds a number of empty lines with spaces to indent them to match the other lines. What should we do about that? I was going to clean them up, but we have other violations of that rule, so I'm not sure if it's desired.

`develop` is fine, but I intentionally targeted `master`, since it has no new features. This PR can be included as part of a patch release, in my opinion :)

spec/integration/cli/generate/action_spec.rb
```diff
@@ -16,27 +16,21 @@
...
-END
+        expect('apps/web/controllers/authors/index.rb').to have_file_content <<~END
+          module Web::Controllers::Authors
+            class Index
+              include Web::Action
+          
+                  #
+                  # apps/web/views/authors/index.rb
+                  #
+                  expect('apps/web/views/authors/index.rb').to have_file_content <<~END
```
AlfonsoUceda: can you fix these lines?

AlfonsoUceda:
@cllns thanks! don't worry about the other rules, can you fix the problem and make the suite green?

AlfonsoUceda:
@cllns thanks man ;)

PR #866, opened 2017-12-04
Render 500 custom template if an exception happens in a view or template

### LINKED ISSUE #865: Custom error page not rendered when an exception is raised inside view or template sections.
The original bug was opened by @cveneziani in https://github.com/hanami/view/issues/138 but that bug belongs to hanami gem.

When an exception is raised inside view or template section, we don't display the custom error template so the application will render the exception with the backtrace.

AlfonsoUceda:
This PR fixes #865 

It will render the 500 custom template if an exception happens in a view or template, in addition it sets `handle_exceptions` to true by default, in controller level it was true for not in hanami application level, https://github.com/hanami/hanami/blob/master/lib/hanami/application_configuration.rb#L1272

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,8 @@ AllCops:
     - "vendor/**/*"
     - "spec/support/**/*"
     - "**/*.gemspec"
+Style/IndentHeredoc:
+  Enabled: true
 Style/MixinGrouping:
   Exclude:
     - "spec/support/**/*"
--- a/lib/hanami/application_configuration.rb
+++ b/lib/hanami/application_configuration.rb
@@ -40,10 +40,11 @@ module Hanami
     # @since 0.1.0
     # @api private
     def initialize(namespace, configurations, path_prefix, env: Environment.new)
-      @namespace      = namespace
-      @configurations = configurations
-      @path_prefix    = path_prefix
-      @env            = env
+      @namespace         = namespace
+      @configurations    = configurations
+      @path_prefix       = path_prefix
+      @env               = env
+      @handle_exceptions = true
 
       evaluate_configurations!
     end
--- a/lib/hanami/rendering_policy.rb
+++ b/lib/hanami/rendering_policy.rb
@@ -24,13 +24,13 @@ module Hanami
     SUCCESSFUL_STATUSES = (200..201).freeze
     # @api private
     RENDERABLE_FORMATS = [:all, :html].freeze
+    # @api private
+    ERROR_STATUS = 500
 
     # @api private
     def initialize(configuration)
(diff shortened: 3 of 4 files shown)


AlfonsoUceda:
@cveneziani this is a work in progress but it should render 500 page in production when an error happens, can you give it a try?

cveneziani:
I confirm, it's working. I do have the 500 page on errors happening in the view or the template.

The only thing I'm missing is the bugtracker being notified for these errors.

AlfonsoUceda:
Hi @cveneziani what bug error tracker are you using? I'd like to test it.

cveneziani:
I'm using Bugsnag. 
The tracker is notified for any action error but not for view or template errors.

AlfonsoUceda:
Did you set bugsnag using this doc? https://docs.bugsnag.com/platforms/ruby/rack/ I ask this because I'll test using that

cveneziani:
Yes exactly.
BTW if you want to have a look, it's here: https://github.com/nfilzi/housing-list/commit/bc1bca59716e5ee0f9fc21ad240ac3a45f9435a1

AlfonsoUceda:
For now I can only check the code since I don't have my laptop with me. We set 'rack.exception' and bugsnag middleware should notify but I'll debug an application when I come back, thanks @cveneziani

cveneziani:
Thanks again.

AlfonsoUceda:
@cveneziani where did you define `Bugsnag::Rack` middleware? in `application.rb` inside `apps/APP` folder? please try to move the middleware to config.ru:
```ruby
require './config/environment'

use Bugsnag::Rack
run Hanami.app
```

cveneziani:
I've defined the Bugsnag middleware in the `application.rb`, see here: https://github.com/nfilzi/housing-list/commit/bc1bca59716e5ee0f9fc21ad240ac3a45f9435a1#diff-945ab852b87f99f59478ad5c2092a40bR93

I tried to move it within the `config.ru` and it's working. 
Bugsnag is now being notified wherever the error is triggered: in the action, in the view or in the template.

Is it normal that it's not working (except for an error within an action) when the bug tracker is defined within the application? What if I want to enable Bugsnag middleware only from within a specific app?

AlfonsoUceda:
@cveneziani nop, it isn't normal and I'll try to fix that problem because as you said, you could add only in one app.

cveneziani:
Thanks for the clarification :)

AlfonsoUceda:
@cveneziani sorry for the late answer. In hanami 1.2.0 we support middlewares per application you will be able to add only bugsnag to one app or apps you want. I'll retest again this PR and make test passes. I'll come back to you shortly

cveneziani:
Nice to hear. Thanks again :)

PR #875, opened 2017-12-20
Handle exceptions at the project level

jodosha:
**TL;DR:** This PR moves the exception handling from the app level to the project level. 

---

Specifically, it removes `Hanami::Application::Configuration#handle_exceptions` in favor of `Hanami::Application::Configuration#handle_exceptions`.

This work is based on https://github.com/hanami/controller/pull/246. On that PR we removed `Hanami::Controller::Configuration#handle_exceptions`, which was the setting that `Hanami::Application::Configuration#handle_exceptions` used. With that setting gone in `hanami-controller`, we wanted to remove the same setting in `hanami`.

---

### Handled exceptions

They will always be handled by the actions. We can turn an `ArgumentError` into a `400 Bad Request` HTTP status.

Example:

```ruby
module Web
  module Controllers
    module Authors
      class Create < Hanami::Action
        handle_exception ArgumentError => 400

        def call(req, res)
          raise ArgumentError
        end
      end
    end
  end
end
```

If the request is made via a browser, we ensure to render an error page. This wasn't working fine in 1.x series: sometimes Hanami returned only the message as a text response.

### Unhandled exceptions

The raised exception will be referenced in `rack.exception` in Rack env. In this way vendors like Bugsnag can report it.

In development/test a person will see an error page with the stack trace. In production (with the `handle_exceptions` on `true`), the user will see an error page similar to the one rendered for handled exceptions.


jodosha:
> One question @jodosha, what happens if an user wants to enable just error tracker for one app?

That won't be possible directly, but with some workarounds. I don't see a huge demand for that.
Another option would be to filter exceptions on with the bug tracker tool.

PR #878, opened 2017-12-21
Revert #873 (adding better_errors gem)

cllns:
Thanks to @igneus for pointing out a security vulnerability in the `better_errors` gem: https://github.com/hanami/hanami/commit/46de8b1596e70daca4a960d6098d423fe7217be6#commitcomment-26418678

It looks like this issue has _not_ been fixed in `better_errors`: https://github.com/charliesome/better_errors/issues/350

### DIFF
--- a/lib/hanami/cli/commands/new/Gemfile.erb
+++ b/lib/hanami/cli/commands/new/Gemfile.erb
@@ -33,9 +33,6 @@ gem 'haml'
 
 <%- if code_reloading -%>
 group :development do
-  gem 'better_errors'
-  gem 'binding_of_caller'
-
   # Code reloading
   # See: http://hanamirb.org/guides/projects/code-reloading
   gem 'shotgun'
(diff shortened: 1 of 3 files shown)


davydovanton:
Hey, sorry for my delay. I know about this problem (check [this lightning talk] (https://www.youtube.com/watch?v=WnlgKWCt8wQ)).

I think we can introduce simple DNSRebinding middleware for hanami apps for fixing it on our side for all hanami devs. WDYT?

I mean something like this:
screenshot 2018-01-24 17 51 01

davydovanton:
@cllns after discussion with @jodosha we want to build hanami-webconsole gem with full better errors integration and DNC protections. I think we need to drop all better error integration from `hanami/hanami` now, that's why let's merge this PR 👏

AlfonsoUceda:
:+1: so let's merge this PR

PR #880, opened 2017-12-22
Introduce configurable project Rack middleware stack

jodosha:
In a nutshell:

```ruby
# config/environment.rb
Hanami.configure do
  middleware.use MyRackMiddleware
end
```

---

Please note that:

  1. This is kinda equivalent to add a middleware in `config.ru`.
  1. This PR does **NOT** changes the `config/environment.rb` that we generate for new projects. But a developer that needs to add a project wide middleware can use it.

---

Why this is important? For plugins.

A plugin may want to add a Rack middleware. On 1.1.x the only ways to add it are: 1. asking the developer to manual add it into `config.ru` 2. create a code generator that writes into `config.ru`. 

Both these solutions require manual intervention for newer versions of a plugin. Which isn't ideal both for developers and plugin authors. With this proposal, a plugin author can mount the middleware from the gem.

### DIFF
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -26,7 +26,7 @@ module Hanami
       @routes  = Hanami::Router.new
 
       mount(configuration)
-      middleware(environment)
+      middleware(configuration, environment)
       builder.run(routes)
     end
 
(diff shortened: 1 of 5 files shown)


oana-sipos:
Looks good to me, clearer anyway! 👍

PR #881, opened 2017-12-22
Introduce Hanami.plugin to let plugins to configure Hanami project

jodosha:
In a nutshell: a plugin can configure a Hanami project.

---

Imagine a plugin that wants to mount a Rack middleware (#880 ). From the plugin code in the gem, the author can do:

```ruby
Hanami.plugin do
  middleware.use MyPluginMiddleware
end
```

or

```ruby
Hanami.plugin do
  mount Hanami::Admin::App, at: "/admin"
end
```

---

The yielded context is the same of `Hanami.configure`. A plugin can override/add settings from `config/environment.rb`

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -21,6 +21,7 @@ module Hanami
   # @api private
   # @since 0.9.0
   @_mutex = Mutex.new
+  @_plugins = Concurrent::Array.new
 
   # Configure Hanami project
   #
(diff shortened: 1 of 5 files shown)


davydovanton:
@jodosha I understand right that we can use `Hanami.plugin`  instead `config.ru`? And just a idea for the future:

WDYT if we mix `Hanami.plugin` and `HANAMI_APPS` together? Something like:

```ruby
Hanami.plugin (apps: %[web admin]) do
  # ...
end
```

Issue #673, opened 2016-11-25
Consider adding `--api` option to generator.


beauby:
It would basically remove the view and assets-specific generated code, but also the hard-dependency on `hanami-view` and `hanami-assets` in the generated Gemfile.

Modifications include: 
- not generating views, templates, and specs for those
- in the generated application, no configuration for views and assets, no `views/` in the load paths

jodosha:
@beauby Generator for what?

beauby:
@jodosha I meant for `$ hanami new`.

mereghost:
Am I the only weird one that renders the JSON responses on the view layer? =/

jodosha:
@beauby We just removed one architecture named "application", which was similar to your idea. But the burden of maintaining it, wasn't worth the effort.

After 1.0, we can decide to have another gem only focused to it.

jodosha:
@mereghost 
> Am I the only weird one that renders the JSON responses on the view layer? =/

Nope, I'm delegating it to the view too. It's speculative to assume that all the devs won't use the view layer.

---

For now devs can generate a new app within a project, and threat it as an API.

PR #885, opened 2018-01-11
Adding new feature: --api option to app generator

Serhiy-Nazarov:
Quick generate API only application.
https://github.com/hanami/hanami/issues/673

TODO:
Auto add --skip-view option to action generator if app has --api option

### DIFF
--- a/lib/hanami/cli/commands/generate/app.rb
+++ b/lib/hanami/cli/commands/generate/app.rb
@@ -11,10 +11,12 @@ module Hanami
 
           argument :app, required: true, desc: "The application name (eg. `web`)"
           option :application_base_url, desc: "The app base URL (eg. `/api/v1`)"
+          option :api, type: :boolean, default: false, desc: "The application API mode"
 
           example [
             "admin                              # Generate `admin` app",
-            "api --application-base-url=/api/v1 # Generate `api` app and mount at `/api/v1`"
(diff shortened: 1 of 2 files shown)


davydovanton:
any specs for this feature?

davydovanton:
Agree with @cllns. Hanami-api gem will be awesome project. You can create json generator with presenters and with all useful logic and gems

Serhiy-Nazarov:
hanami-api - it's a good idea. But I think in future API pattern must be main for hanami as part of the core.
just thought - maybe there is a way to extract the current generator into a gem hanami-app and create gem hanami-api...

cllns:
Agree, as things separate into JSON API's, we maybe want to have Hanami itself be able to generate API's, without a plugin. But, first, I think `hanami-api` should be its own plugin, then we can integrate if if there's demand for it. :) 

How does that sound to you @Serhiy-Nazarov? Do you feel like you want to take that on?

Serhiy-Nazarov:
@cllns 
as I said, I am a beginner with hanami :)
I want and can join in to work with more experienced hamani-developers

cllns:
Closing this PR, since the appropriate place for this functionality is a Hanami plugin.

 @Serhiy-Nazarov If you want to work on this, I think you should see how far you can get creating a plugin, and then ask for help in the [Gitter chat](https://gitter.im/hanami/chat), and I'm sure someone will be able to help out :)

PR #891, opened 2018-01-18
Generate frozen strings

### LINKED ISSUE #879: Consider adding `# frozen_string_literal: true` to generated files
If hanami can work with it, it would be nice having `# frozen_string_literal: true` in generated files. What do you think?

graywolf:
Fixes #879


graywolf:
This pull request is wrong, I should PR to develop, not master.

PR #895, opened 2018-02-05
Integration tools

jodosha:
Extracted local RSpec helpers into `hanami-devtools` and use them from the gem.


jodosha:
@davydovanton https://github.com/hanami/webconsole/pull/1 and https://github.com/jodosha/hanami-reloader/pull/6 are using new `hanami-devtools` version.

jodosha:
@davydovanton Are you talking about a Hanami project like `bookshelf` (aka a final product built with Hanami)? If so, there is no need. These tools are designed to build Hanami and Hanami plugins, not projects.

davydovanton:
@jodosha no, I mean something like: "as a hanami plugin developer, I can take `hanami-devtools` repo and use some integration matchers for testing my hanami plugin/hanami project/etc" :)

jodosha:
@davydovanton I see. Well, in that case you have two examples linked above 😉 .

PR #900, opened 2018-02-21
Provide stack trace when exception is caught

Pistos:
Without this, the user has no idea what has caused the exception (what file, what line number).
This makes it much easier on the user.

### DIFF
--- a/lib/hanami/cli/commands/command.rb
+++ b/lib/hanami/cli/commands/command.rb
@@ -85,6 +85,7 @@ module Hanami
             super(options)
           rescue StandardError => e
             warn e.message
+            warn e.backtrace.join("\n\t")
             exit(1)
           end
         end


cllns:
:+1: Thanks for this improvement!

This actually helps a bug I've been stuck on :) I had something wrong in my migration and the migration just said:
```
...
[hanami] [INFO] (0.000412s) COMMIT
[hanami] [INFO] Finished applying migration 20180217224523_create_twitter_accounts.rb, direction: up, took 0.004545 seconds
[hanami] [INFO] Begin applying migration 20180217233918_add_foreign_key_from_twitter_accounts_to_users.rb, direction: up
[hanami] [INFO] (0.000192s) BEGIN
[hanami] [INFO] (0.000094s) ROLLBACK
no implicit conversion of Class into Hash
```
which is not very helpful.

And, with this change, it is followed with the backtrace:
```
...
no implicit conversion of Class into Hash
/Users/sean/.gem/ruby/2.5.0/gems/sequel-4.49.0/lib/sequel/database/schema_generator.rb:481:in `merge!'
	/Users/sean/.gem/ruby/2.5.0/gems/sequel-4.49.0/lib/sequel/database/schema_generator.rb:481:in `drop_column'
	/Users/sean/bookshelf/db/migrations/20180217233918_add_foreign_key_from_twitter_accounts_to_users.rb:4:in `block (3 levels) in <top (required)>'
...
```

@Pistos Do you think you could find a way to write a regression spec for this? So that we don't accidentally remove this functionality in the future.

Pistos:
> Do you think you could find a way to write a regression spec for this? So that we don't accidentally remove this functionality in the future.

Yes, I can do that, though not quite right away.  :+1:

jodosha:
@Pistos Thanks for this PR, this LGTM 👍

PR #905, opened 2018-02-26
Early Hints

jodosha:
## Historical Background

I experimented with Early Hints (EH) in the summer of 2015, when the project was still Lotus:

  * https://github.com/jodosha/instants
  * https://www.youtube.com/watch?v=XCgsXUKLsOc&feature=youtu.be&t=31m4s

At that time, I used the term "Push Promise" instead of "Early Hints".

We didn't ship this feature because there was the lack of support in Rack and any other Ruby web server. I had to write a toy HTTP/2 web server for Rack https://github.com/jodosha/panther

Fast forwarding to today, both Rack and Puma have support for Early Hints: see https://github.com/puma/puma/pull/1403

## Setup

In order to use it, you need to use Puma with Early Hints feature enabled.

```ruby
# config/puma.rb
# ...
early_hints true
```

Inside the project configuration you can simply enable like this:

```ruby
# config/environment.rb
Hanami.configure do
  # ...
  early_hints true
end
```

As last step, you need a web server that supports HTTP/2 and Early Hints: eg. [h2o](https://h2o.examp1e.net/).

**If you're looking for a full working example, please check this [demo app](https://github.com/jodosha/hall_of_fame)**.

## Usage

Google Chrome ignores the Early Hint responses after the server pushed 100 assets.
In order to avoid this limit I decided to not push all the assets by default.

### Automatic opted-in assets

Assets like JavaScripts and Stylesheets are **automatically** opted in to be pushed.

This will be pushed:

```erb
<%= stylesheet "application" %>
```

To opt-out:

```erb
<%= stylesheet "application", push: false %>
```

### Automatic opted-out assets

Other assets like images, audio, video, and generic assets are **not** pushed by default.

This won't be sent:

```erb
<%= audio "song.ogg" %>
```

To opt-in:

```erb
<%= audio "song.ogg", push: true %>
```

For a complete syntax reference, please have a look at https://github.com/hanami/assets/pull/85

---

Ref https://github.com/hanami/assets/pull/27
Ref https://github.com/hanami/assets/pull/85
Ref https://github.com/puma/puma/pull/1403
Ref https://github.com/hanami/hanami/pull/880
Ref https://trello.com/c/c4NGy33K/1-http-2-early-hints

### DIFF
--- a/lib/hanami/early_hints.rb
+++ b/lib/hanami/early_hints.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+
+module Hanami
+  # HTTP/2 Early Hints Rack middleware
+  #
+  # It sends extra responses **before** the main reponse is sent.
+  # These extra responses are HTTP/2 Early Hints (103).
+  # They specify the web assets (javascripts, stylesheets, etc..) to be "pushed",
+  # so modern browsers pre-fetch them in parallel with the main HTTP response.
+  #
+  # @see https://tools.ietf.org/html/draft-ietf-httpbis-early-hints-05
+  #
+  # @since x.x.x
+  # @api private
+  class EarlyHints
+    # @since x.x.x
+    # @api private
+    class NotSupportedByServerError < ::StandardError
+      # @since x.x.x
+      # @api private
+      def initialize
+        super("Current Ruby server doesn't support Early Hints.\nPlease make sure to use Puma with Early Hints enabled.")
+      end
+    end
+
+    # @since x.x.x
+    # @api private
+    def initialize(app)
+      @app = app
+    end
+
+    # @param env [Hash] Rack env
+    #
+    # @return [Array,Rack::Response] a Rack response
+    #
+    # @raise [Hanami::EarlyHints::NotSupportedByServerError] if the current Ruby
+    #   server doesn't support Early Hints
+    #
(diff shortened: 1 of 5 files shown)


lib/hanami/early_hints.rb
```diff
@@ -0,0 +1,129 @@
...
+  # @since x.x.x
+  # @api private
+  class EarlyHints
+    # @since x.x.x
+    # @api private
+    class NotSupportedByServerError < ::StandardError
+      # @since x.x.x
+      # @api private
+      def initialize
+        super("Current Ruby server doesn't support Early Hints.\nPlease make sure to use Puma with Early Hints enabled.")
```
AlfonsoUceda: Why not better say
>Please make sure to use a web server with Early Hints enabled
or
>Please make sure to use a web server with support for Early Hints.

jodosha: Yes, just Puma for now. That's why my suggestion is really specific. As long other vendors will add support, we can change this message.

jodosha: @AlfonsoUceda I rephrased this: https://github.com/hanami/hanami/pull/905/commits/44d25a35896b43c39326f87dbf979e66c7bdf61f

PR #906, opened 2018-02-26
[WIP] Colorize file generation

cllns:
Here's a work-in-progress for colorizing the CLI commands:

image

You can see what changes will be necessary to the CLI specs. It's _a lot_ of manual work (or could be automated) but it'll cause a lot of diff noise, and make it much harder to edit the output of the specs in the future. 

We (@jodosha and I) have talked about extracting the file generation commands into `hanami/cli`. If we did that, then we could test these exact CLI things, without repeating the shellcodes a bunch of times in this repo. We could even write a helper to strip the shellcodes, if we wanted.

Thoughts? @jodosha @hanami/core?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ end
 
 gem 'i18n'
 
-gem 'hanami-utils',       '~> 1.1', require: false, git: 'https://github.com/hanami/utils.git',       branch: '1.1.x'
+gem 'hanami-utils',       '~> 1.1', require: false, git: 'https://github.com/hanami/utils.git', branch: 'coloured_output'
 gem 'hanami-validations', '~> 1.1', require: false, git: 'https://github.com/hanami/validations.git', branch: '1.1.x'
 gem 'hanami-router',      '~> 1.1', require: false, git: 'https://github.com/hanami/router.git',      branch: '1.1.x'
 gem 'hanami-controller',  '~> 1.1', require: false, git: 'https://github.com/hanami/controller.git',  branch: '1.1.x'
--- a/lib/hanami/cli/commands/command.rb
+++ b/lib/hanami/cli/commands/command.rb
@@ -6,6 +6,7 @@ require 'hanami/cli/commands/project'
 require 'hanami/cli/commands/templates'
 require 'concurrent'
 require 'hanami/utils/files'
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -8,7 +8,7 @@ end
 
 gem 'i18n'
 
-gem 'hanami-utils',       '~> 1.1', require: false, git: 'https://github.com/hanami/utils.git',       branch: '1.1.x'
+gem 'hanami-utils',       '~> 1.1', require: false, git: 'https://github.com/hanami/utils.git', branch: 'coloured_output'
```
cllns: Temporary, until this is merged in.

lib/hanami/cli/commands/command.rb
```diff
@@ -160,7 +175,26 @@ def generate_file(source, destination, context)
...
+        ].join("  ")
+        end
+
+        # @since x.x.x
+        # @api private
+        def _colorize_operation(operation)
+          if OPERATION_COLORS.key?(operation)
+            _colorize_and_justify(operation)
+          else
+            operation.to_s.rjust(COLUMN_WIDTH)
```
cllns: We _could_ just colorize other stuff to black, to get rid of this conditional.

jodosha: @cllns This assumes a black terminal background.

jodosha:
@cllns This feature is just great to have, but it comes with the cost of changing all the specs under `spec/integration/cli`.

My suggestion is to:

  1. Close this PR
  1. Work on `hanami-cli` (after `v0.2.0`) to add the colorization feature over there. It has to be turned off by default
  1. Work on a new PR to add a **private** setting for `Hanami` (see **Snippet 1**)
  1. On the same PR [generate new projects](https://github.com/hanami/hanami/blob/master/spec/support/with_project.rb#L13) with colorization turned off. This avoid us to rewrite all the specs under `spec/integration/cli` **and** it keeps stack traces clean in case of CI failure.
  1. Still on the same PR, add colorization feature (from this PR), alongside with new integration tests with colorization **turned on** (only for these few specs).

#### Snippet 1

```ruby
# config/environment.rb
Hanami.configure do
  # ...
  cli colorize: false
end
```

cllns:
:+1:

PR #908, opened 2018-02-28
Require config/boot from spec helpers in new generated projects

sandelius:
Since we now have `config/boot` we can require that file instead of `config/environment`.

Not a big "thing" but we dont have to manually call `Hanami.boot`.

I've tested it and the suite run fine on both minitest and rspec.


sandelius:
I have no idea why Travis say it fails when the tests pass

jodosha:
@sandelius This is the reason why the build is failing: https://travis-ci.org/hanami/hanami/jobs/347422838#L2957

jodosha:
@sandelius Thanks for this PR. I'm gonna close it for now, as there wasn't any progress since 2 months. Feel free to resume the work here.

Issue #726, opened 2017-02-06
hanami on Windows x64 - creating new project throws error: thor-0.19.4 actions.rb line 254


harbirg:
Unable to run hanami on windows 10 using ruby 2.3.0 installed via https://rubyinstaller.org/ along with DevKit for native gems. 

When creating new project, thor gem gives an error. Has anyone else tried to use hanami on windows? I was able to get Sinatra working, but I really wanted to use Hanami. Can someone point me to what I should change to resolve the issue. 

D:\dev>hanami new bookshelf
      create  .hanamirc
      create  .env.development
      create  .env.test
      create  Gemfile
      create  config.ru
      create  config/environment.rb
      create  lib/bookshelf.rb
      create  public/.gitkeep
      create  config/initializers/.gitkeep
      create  lib/bookshelf/entities/.gitkeep
      create  lib/bookshelf/repositories/.gitkeep
      create  lib/bookshelf/mailers/.gitkeep
      create  lib/bookshelf/mailers/templates/.gitkeep
      create  spec/bookshelf/entities/.gitkeep
      create  spec/bookshelf/repositories/.gitkeep
      create  spec/bookshelf/mailers/.gitkeep
      create  spec/support/.gitkeep
      create  db/migrations/.gitkeep
      create  Rakefile
      create  spec/spec_helper.rb
      create  spec/features_helper.rb
      create  db/schema.sql
      create  .gitignore
         run  git init . from "."
C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/actions.rb:254:in ``': No such file or directory - git init . (Errno::ENOENT)
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/thor-0.19.4/lib/thor/actions.rb:254:in `run'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/commands/new/abstract.rb:90:in `init_git'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/commands/new/container.rb:34:in `post_process_templates'
        from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/hanami-0.9.2/lib/hanami/generators/generatable.rb:31:in `process_templates'
[...]

PR #914, opened 2018-03-20
Restrict shotgun gem in new projects' Gemfiles to working platforms

PeterHatch:
This adds `"platforms: :ruby"` to the Shotgun gem's options in the Gemfile of newly created Hanami projects, so that Hanami will work out of the box on Windows and JRuby, rather than needing to remove that line from the Gemfile manually, or start the server with `--no-code-reloading` every time.

This fixes the second issue described in https://github.com/hanami/hanami/issues/726, which I just ran into.

If this is accepted, it'd also make sense to update http://hanamirb.org/guides/projects/code-reloading/ at some point - maybe after a version with the change is released?

### DIFF
--- a/lib/hanami/cli/commands/new/Gemfile.erb
+++ b/lib/hanami/cli/commands/new/Gemfile.erb
@@ -35,7 +35,7 @@ gem 'haml'
 group :development do
   # Code reloading
   # See: http://hanamirb.org/guides/projects/code-reloading
-  gem 'shotgun'
+  gem 'shotgun', platforms: :ruby
 end
 
 <%- end -%>


PeterHatch:
It actually isn't being added when using JRuby - apparently there's a different Gemfile created when using JRuby, which I didn't notice until looking at the tests.

So right now it just helps with Windows, which is the issue I ran into and wanted to fix.  It could also be handled by making the Gemfile not include that line if running on Windows, but I think this is better, as it could help the same project work on multiple platforms.

jodosha:
@PeterHatch Thank you for the clarification. How this helps with Windows? Isn't `platforms: :ruby` true for MRI on Windows?

PeterHatch:
`platforms: :ruby` is true for "C Ruby (MRI) or Rubinius, but NOT Windows" - documented at http://bundler.io/v1.14/man/gemfile.5.html#PLATFORMS.

Issue #859, opened 2017-11-03
Provide more useful error message instead of "undefined local variable or method `flash'"


inem:
I am trying to create some very basic app with Hanami, and recently got this weird error message:
```
undefined local variable or method `flash' #<Web::Controllers::Sessions::Create:0x007fb84ef98068>
```

It took me a while to figure out what is wrong. I forgot to add `include Hanami::Action::Session` to my action, but nothing suggested that I should do so.

I guess a lot of newbies might fall into this trap as well.

My suggestion is to make error message more useful and self-explanatory

AlfonsoUceda:
@inem thanks for the suggestion, I think we should document the behaviour, what do you think @hanami/core?

cllns:
I think we could go a step further and look for `session` and `flash`, and raise errors when those happen with a helpful message saying something like "Did you forget to add `include Hanami::Action::Session` to your action?"

(This would be a change to [`hanami/controller`](https://github.com/hanami/controller))

davydovanton:
@cllns or "For using session add `include Hanami::Action::Session`. For more information check docs: doc_link" :)

AlfonsoUceda:
@wheresmyjetpack don't worry, just open the PR and we discuss there ;)

jodosha:
Implemented by #915

PR #915, opened 2018-03-22
Raise meaningful error when trying to access session/flash without enabled sessions

jodosha:
It prints one of the following error messages, if a dev tries to access `session` or `flash` from an action, without having enabled sessions for that app.

```
To use `session', please enable sessions for the current app.
```

```
To use `flash', please enable sessions for the current app.
```

---

Closes https://github.com/hanami/hanami/issues/859
Ref https://github.com/hanami/controller/pull/250

### DIFF
--- a/lib/hanami/frameworks.rb
+++ b/lib/hanami/frameworks.rb
@@ -14,3 +14,9 @@ Hanami::Controller.configure do
     include Hanami::Action::Glue
   end
 end
+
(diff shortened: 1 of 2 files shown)


jodosha:
This is failing because we're reaching the max build time allowed by Travis CI. They are so kind to increase the build time for us. By merging this, I expect the new build on `develop` to pass.

PR #917, opened 2018-03-27
Fix chdir warning

jtdowney:
Dir.chdir will change the directory for the whole process. If multiple threads are running, this means those threads have their current directory changed as well. During feature tests, this could cause an issue with rack servers running in a background thread.

Since this use of chdir was to calculate the relative path for static files, I refactored it to not change the current directory for the whole process as a side effect.


jtdowney:
Hi @jodosha. I wanted to check in on this and provide a bit more information. I only saw this with ruby 2.5 and it appears others saw it as well: https://discourse.hanamirb.org/t/confilicting-chdir-in-static-rb/422.

kaikuchn:
This problem was also discussed in the hanami chat:
glaszig had randomly failing feature tests and noticed warnings about chdir: [March 8, 2018 3:43 PM](https://gitter.im/hanami/chat?at=5aa14c06e4ff28713a0b7bcf)
I found that the above code is to blame and suggested as a quick fix the following: [March 9, 2018 3:32 PM](https://gitter.im/hanami/chat?at=5aa29b1ae4ff28713a11c974)

I'm not sure about relative paths @jtdowney I think right now it is absolute paths? But yeah, the `chdir` is completely unnecessary and I don't know what made me use it when I wrote that code..

jodosha:
Hi folks, I apologize for my delay here.

We're gonna to release `v1.2.0` on tomorrow, and didn't want to merge a PR that I haven't manually tested carefully.

Security is my main concern here: does this change allows evil users to access the local file system outside of the public directory? I'm not saying that this PR will cause this problem, but I need to verify before to merge.

@kaikuchn would you be so kind to check as well?

Thank you all.

kaikuchn:
What this piece of code does is register existing files relative to some root directory (by default the hanami public directory) and pass those along to Rack for serving static assets.

At no time is this code dependent on external(/evil user) input.

While the underlying rack-static middleware does make routing decisions based on the request, it cannot influence the root directory and any file outside of the root directory will not be registered, and therefore result in a 404 (or some other routing decision based on other middlewares).

jtdowney:
I mistakenly set the branch to develop but changed it back to master after I saw the note at the bottom of the 1.2.0 announcement blog post. Since this is a bug fix it seems like it should go against master. I also updated the test since it seems a few extra env params are now required in the setup.

jodosha:
@jtdowney I just merged it thank you again for this fix and the wait. 💯 
@kaikuchn Thanks for the review. 👍

Issue #924, opened 2018-04-21
Problems with dotenv 2.3.0


citizen428:
**Problem:**
Just initialize a new project with Hanami v1.2.0, installed the gems and got the following error when running `bundle exec hanami server`:

```
.../dotenv-2.3.0/lib/dotenv/parser.rb:33:in `call': wrong number of arguments (given 1, expected 2) (ArgumentError)
```

**Potential cause:**
I checked, dotenv 2.3.0 was released after Hanami 1.2.0 and potentially introduced some change that leads to this behavior. 

**Solution:**
I modified the Gemfile to depend on a 2.2. release of dotenv, everything works as expected now with dotenv 2.2.2.

hirokiraj:
Maybe it will help - after short investigation - problem occurs as new version of dotenv (2.3.0 - released 2 days ago), changed initializer for parser, difference can be seen here:

old: https://github.com/bkeepers/dotenv/blob/e388215d69fbf55e8031d8524edf9db22eb66c03/lib/dotenv/parser.rb#L35

new: 
https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/parser.rb#L33

Change was made with fix:
https://github.com/bkeepers/dotenv/commit/7f82f73420bb46a06bb3a3e2052c3b3a1f34a0f5

Current hanami release `1.2.0` still calls parser with just one argument:
https://github.com/hanami/hanami/blob/master/lib/hanami/env.rb#L59

spica831:
Hello,
I was suffering from the same thing
So I modified the dotenv itself
Since it has already been merged already, it got in a state to move once

https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/parser.rb#L33

citizen428:
Thanks @jodosha, so either stay with 2.2. or upgrade to 2.4.

tommyalvarez:
@citizen428 the error is still happening in 2.4.0 and Rails 5.1.5.

cllns:
@tommyalvarez This is the issue tracker for Hanami, which doens't have anything to do with Rails. If you're seeing problems still, you might want to open an issue in `dotenv-rails` (first I'd double-check with `bundle show dotenv` to ensure your project is actually using `2.4.0`)

tommyalvarez:
@cllns my bad I read the issue in google and came directly here without seeing the repo.

cllns:
[...]

PR #925, opened 2018-04-23
Add missing argument to Dotenv::Parser call

bonsoirq:
Dotenv 2.3 release has broken `hanami server` as mentioned in #924 I've added missing argument to parser call to get `hanami server` working again.

### DIFF
--- a/lib/hanami/env.rb
+++ b/lib/hanami/env.rb
@@ -56,7 +56,8 @@ module Hanami
       return unless defined?(Dotenv::Parser)
 
       contents = ::File.open(path, "rb:bom|utf-8", &:read)
-      parsed   = Dotenv::Parser.call(contents)
+      is_load  = false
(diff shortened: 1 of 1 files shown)


spica831:
Hello!
I also faced this problem.
Thank you for fixing in advance!
It worked as it was modified as described.
Thank you!

jodosha:
@bonsoirq Thank you very much for this fix and the fast reaction. 👍 

I'm closing this PR without merging it. In the meantime `dotenv` release `2.4.0` happened, with a rollback of the breaking changes. Merging this PR would mean to support only `dotenv` `2.3.0`, which is a broken and already outdated dependency.

---

I decided to bump the minimal version of `dotenv` to `~> 2.4` for new generated projects, so they won't run in this problem.

See https://github.com/hanami/hanami/pull/925

---

Thanks 🌸

PR #926, opened 2018-04-24
Generate new projects with dotenv ~> 2.4

jodosha:
`dotenv-2.3.0` introduced a change that breaks `hanami server` command, if `dotenv` is bundled (development and test envs).

In the meantime `dotenv-2.4.0` was released to fix that problem. To be clear, that version of `dotenv` broke a lot of apps, so their release is not specifically for Hanami.

---

This PR bumps the minimum version of `dotenv` to `~> 2.4`. In this way, new generated apps, won't run into the same problem. See https://github.com/hanami/hanami/issues/924

---

Fixes https://github.com/hanami/hanami/issues/924
Ref https://github.com/hanami/hanami/pull/925

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,7 @@ end
 gem 'sass',          require: false
 gem 'coffee-script', require: false
 
-gem 'dotenv',    '~> 2.0', require: false
+gem 'dotenv',    '~> 2.4', require: false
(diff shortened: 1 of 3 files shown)


joelvh:
FYI dotenv 2.5.0 just went out, which makes this backwards-compatible: https://github.com/bkeepers/dotenv/pull/345

PR #931, opened 2018-05-11
[WIP] Extract CLI generator

cllns:
We'd like to extract to file generation functionality from this repo to [`Hanami::CLI`](https://github.com/hanami/cli).

A few reasons:

1. So other users of `Hanami::CLI` can benefit from our work. They won't need to write their own file manipulation and output code.
2. We want to colorize the CLI file generation, and handle that at the `Hanami::CLI` level, so all our specs for this repo can be plaintext without being muddied by a bunch of shell-code for colorizing (since we can disable colorizing for specs, and delegate the testing of that behavior to `Hanami::CLI`'s test suite).
3. We can add a feature for prompting users when they'd potentially overwrite a file that already exists. We could've done this here, but it'll be simpler as its own feature in `Hanami::CLI`.

cc @AlfonsoUceda @hanami/core 

Some clean-up too:
- I changed some variables names to `path` from `destination` when there's only one page. It makes sense to have `destination` and `source` when they are two paths, but, for example, `destination` doesn't make sense for something we want to delete.

- The CLI said a route was `insert`ed, when really it was `append`ed. Either one works but it's simpler for the cade to just let it say the route was appended to the routes file.


AlfonsoUceda:
very good job @cllns it makes sense to extract this to CLI, for me it is ok :)

cllns:
Talking with @jodosha a little while ago, we decided this should be extracted more. Specifically, reading the templates should be extracted into `hanami-cli`.

So instead of:
```ruby
source = templates.find("README.md.erb")
destination = project.readme(context)
generator.create(source, destination, context)
```

The API would be:
```ruby
destination = project.readme(context)
generator.create("README.md.erb", destination, context)
```

Does that make sense? It's quite a bit more work on top of what's already in this PR.

AlfonsoUceda:
Removing milestone 1.3.0 because it isn't completed and is not a priority

jodosha:
Closing as stale initiative. @cllns Feel free to reopen, in case.

PR #935, opened 2018-05-18
Remove broken Gemnasium badge

thrgamon:
The Gemnasium badge is broken since they got bought by GitLab. This is explained in slightly more detail [here](https://docs.gitlab.com/ee/user/project/import/gemnasium.html#what-happened-to-my-badge)

This PR removes the broken badge until such a time that it can be updated or replace.


greysteil:
Also, if you'd like to automate the dependency update work that Gemnasium used to alert you about then I've build [Dependabot](https://dependabot.com) for that. 🤖 😉

jodosha:
@greysteil Thanks for the offer, we will consider that. 😄

Issue #936, opened 2018-05-20
`hanami new .` generate broken app


mariusz-kowalski:
> hanami version 1.2.0.

If you give the name in typical way e.g. `hanami new app_x` it will create folders like this:
```
lib/
    app_x/
          entities/
          mailers/
          repositories/
    app_x.rb
```
When the app name is `.` it puts folders `entities`, `mailers` and `repositories` directly in `lib/` and creates a file `..rb` (in `lib/`), which is then incorrectly required in `config/environment.rb`, like so:

```ruby
# ...
require_relative '../lib/.'
```
Name `.` should be translated into real name of the current folder.

cllns:
Related: https://github.com/hanami/hanami/issues/736

cllns:
I (still) think we should have this functionality, since it's a common pattern for command line utilities.

mariusz-kowalski:
I'm using hanami CLI that way because I like to put everything in rvm gemset. So I start new project like this:

```
mkdir project_x
cd project_x
rvm use ruby-2.5@project_x --create --ruby-version
gem install hanami
hanami new .
```

davydovanton:
@cllns WDYT if we just add some simple format validators to CLI library and will use it here?

cllns:
Just had another request for this (#939).

@davydovanton I'm not sure how format validators would solve this. We could use that to tell people they can't create a project at `.`, but, instead, I think we should allow it.

crdx:
@cllns Might be a silly question (I probably missed something) but what was wrong with @tzmfreedom's PR #939 which (as far as I can see) has an implementation of a solution?

jodosha:
Dear all, we support this behavior since long time, but due to an evident regression it wasn't working anymore. I've opened #947 To fix this. Please review.

Issue #736, opened 2017-02-27
Generate project in folder


cllns:
Right now, a hanami project is created in a new folder named based on the project. This is good for the vast majority of cases, but I think the command should be more flexible, to optionally allow other locations.

For example:
```
hanami new test_app test_app2
```
would create a `test_app` project in a new `test_app2` folder, that folder doesn't already exist. (This is the real usage I tried and it didn't work)

&

```
hanami new test_app .
```
would create a `test_app` project in the current folder, assuming the folder is empty.

&

```
mkdir test_app2 
touch test_app2/hello
hanami new test_app test_app2
# PROMPT: The folder 'test_app2' already has file(s) in it, are you sure you want to create a new hanami project in test_app2? Y/N?
```

I think the prompting is important, since we don't want to add a bunch of files to a folder based on a user's typo. A simpler solution could be to skip the prompt and only allow hanami projects to be created in folders that are empty.

Here's what the USAGE would look like, when running `hanami new --help`
```
Usage:
  hanami new PROJECT_NAME [PATH]
```

Note: this is proposed for `v1.1`, not `v1.0`.

Do people think this would be a useful feature, to improve developer experience?

davydovanton:
I think it will be useful for me and other developers 👍 
Because sometimes I need to create project with one name and other dir name.

AlfonsoUceda:
thanks @cllns I think it makes sense but after 1.0 we will rewrite CLI interface removing Thor and making it more flexible.

jodosha:
Lemme play the _grumpy cat_ ™️ here: do we need this? Can this be solved by manual intervention from a developer?

Creating a project in the current directory is it so common? It was introduced to add a deliverability mechanism for a current Ruby project. Aka add some web wrapper around your Ruby library. Which I don't know how much is it common.

</grumpy-cat>

---

We can add this change, if it won't complicate too much the implementation of the rewritten CLI.

mereghost:
I'm with @jodosha here. How often do you need this functionality? Why? Is this the common use case? Is there an easy way to circumvent the need for more code? 

What would we do if $PWD already has a directory of the same name? And git repo?

While it might help some cases, I don't think that it's *that common* to warrant the extra maintenance burden and having to deal with these edge cases.

jodosha:
@cllns WDYT?

cllns:
I don't think it's a very common workflow. But, more importantly, I think this functionality is what developers will expect from a command line utility. 

[...]

PR #939, opened 2018-06-04
Allow '.' as project name on new command

tzmfreedom:
On current hanami implementation, `bundle exec  hanami new .` generate application files incorrectly and results is as follow.
```
...
      create  lib/..rb
      create  public/.gitkeep
      create  config/initializers/.gitkeep
      create  lib/./entities/.gitkeep
      create  lib/./repositories/.gitkeep
      create  lib/./mailers/.gitkeep
      create  lib/./mailers/templates/.gitkeep
      create  spec/./entities/.gitkeep
      create  spec/./repositories/.gitkeep
      create  spec/./mailers/.gitkeep
...
```

`.` is interpreted as project name, but we intend '.' as current directory name.
So we want to current directory name as project name when executing `bundle exec hanami new .`.
On `Ruby on Rails`, this command(`bundle exec rails new .`) work correctly as we want.

This command is important, because we don't want to install gem in system global and we want to do it in local with bundler.

So we want to install and generate hanami application as follows.
1. `mkdir {project_directory}`
2. `cd {project_directory}`
3. `bundle init` and modify Gemfile to install `hanami`
4. `bundle install --path=vendor/bundle`
5. `bundle exec hanami new .`
6. `bundle install`

This pull request solve them.
If we put '.', current directory name is interpreted as project name.

### DIFF
--- a/lib/hanami/cli/commands/new.rb
+++ b/lib/hanami/cli/commands/new.rb
@@ -282,14 +282,15 @@ module Hanami
         # rubocop:disable Metrics/AbcSize
         # rubocop:disable Metrics/MethodLength
         def call(project:, **options)
-          project         = Utils::String.underscore(project)
-          database_config = DatabaseConfig.new(options[:database], project)
+          project_name    = Utils::String.underscore(project)
+          project_name    = File.basename(Dir.pwd) if project == '.'
+          database_config = DatabaseConfig.new(options[:database], project_name)
           test_framework  = TestFramework.new(hanamirc, options[:test])
           template_engine = TemplateEngine.new(hanamirc, options[:template])
-          options[:project] = project
+          options[:project] = project_name
 
           context = Context.new(
-            project: project,
+            project: project_name,
             database: database_config.type,
             database_config_hash: database_config.to_hash,
             database_config: database_config,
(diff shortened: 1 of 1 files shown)


cllns:
Thanks for opening this issue. This is something we've considered before and it's nice to know there's continuining interest in this feature. 😄 

I'm going to close this in favor of #936, since that's still open. (Duplicate of #736 too) 🌸

tzmfreedom:
Oh, I overlooked the issues. Sorry for your inconvenience 😓 
I watch the issues and I hope hanami will be more useful framework!

jodosha:
@cllns This is a PR, that means it was an attempt to fix the problem, not to report it (like in #936).

---

That being said, IMO this is **not** a duplicate of #736.

Because `hanami new .` is a feature that we already support since long time (even if it was accidentally broken), and so we need to fix it.

Whereas #736 is an enhancement proposal to generate a new project with a different directory than the project name. And in that case we didn't find an agreement if it was useful or not.

jodosha:
@tzmfreedom Sorry if we closed your PR too early. I've resumed your work in https://github.com/hanami/hanami/pull/947 and added an integration spec. I'll make sure to credit you in `CHANGELOG`. Thanks.

jodosha:
@cllns No worries, Hanami is complex.

PR #940, opened 2018-06-15
Use dry inflector

### LINKED ISSUE #922: Use `dry-inflector`
CLI generators use inflections to generate conventional names.
Internally these inflections are handled by `Utils::String.pluralize` and `.singularize`. These two methods will be removed in Hanami 2.0, alongside with `Utils::Inflector`.

For this reason, starting from 2.0, `hanami` should use `dry-inflector`. The main difference between `Utils::Inflector` and `dry-inflector` is the API: the first uses class level methods, whereas the second is an object, so it uses instance methods.

  - [x] Let `Hanami::Configuration` to hold and instance of `Dry::Inflector`
  - [x] Replace `Utils::String.pluralize` and `.singularize` to use that internal inflector instead
  - [x] Allow `Hanami::Configuration` to accept custom inflections. See the code snippet below.
  - [x] Pass that inflector to the router used by `Hanami::App`. See https://github.com/hanami/router/issues/159
  - [x] Pass that inflector to the model configuration. See https://github.com/hanami/model/issues/481

```ruby
# config/environment.rb
Hanami.configure do
  inflections do |rule|
    rule.plural      "virus",   "viruses"
    rule.singular    "thieves", "thief"
    rule.uncountable "dry-inflector"
  end
end
```

The block above should be forwarded to `Dry::Inflector#initialize`

---

Changes **MUST** go to `unstable` branch of this repository.

---

Ref https://trello.com/c/nJhd5awo
Ref https://github.com/hanami/utils/issues/267

artofhuman:
Closes #922


Gemfile
```diff
@@ -8,8 +8,8 @@ end
 
 gem 'i18n'
 
-gem 'hanami-utils',       '~> 2.0.alpha', require: false, git: 'https://github.com/hanami/utils.git',       branch: 'unstable'
-gem 'hanami-validations', '~> 2.0.alpha', require: false, git: 'https://github.com/hanami/validations.git', branch: 'unstable'
+gem 'hanami-utils',       '~> 2.0.alpha', require: false, git: 'https://github.com/hanami/utils.git', branch: 'feature/result'
+gem 'hanami-validations', '~> 2.0.alpha', require: false, git: 'https://github.com/artofhuman/validations.git', branch: 'unstable'
```
jodosha: @artofhuman Why is this pointing to your fork? Thanks.

PR #942, opened 2018-07-02
Allow to use logger not based on ::Logger class

davydovanton:
## What
Allow to use loggers which not based on `::Logger` instance. For example [semantic_logger](https://github.com/rocketjob/semantic_logger)

## Why
Now we check only `::Logger` instance, I think it will be better check logger interface too, something like `#info`, `#debug` and `#warn` methods. In this case we can register any logger types without any problems.

## How
Just create a list of base logger methods and check that instance allow to call each method

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,7 +49,6 @@ gem 'shotgun',   '~> 0.9', require: false
 
 gem 'hanami-devtools', require: false, git: 'https://github.com/hanami/devtools.git'
 gem 'hanami-webconsole', require: false, git: 'https://github.com/hanami/webconsole.git'
-gem 'semantic_logger'
 
 # https://github.com/hanami/hanami/issues/893
 gem 'builder'
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,7 @@ gem 'shotgun',   '~> 0.9', require: false
 
 gem 'hanami-devtools', require: false, git: 'https://github.com/hanami/devtools.git'
 gem 'hanami-webconsole', require: false, git: 'https://github.com/hanami/webconsole.git'
+gem 'semantic_logger'
 
 # https://github.com/hanami/hanami/issues/893
 gem 'builder'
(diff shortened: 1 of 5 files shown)


lib/hanami/components/components.rb
```diff
@@ -29,7 +29,7 @@ module Components # rubocop:disable Metrics/ModuleLength
 
       resolve do |configuration|
         if configuration.logger.is_a?(Array)
-          if configuration.logger.first.is_a?(::Logger)
+          if configuration.logger.first.is_a?(::Logger) || logger_interface?(configuration.logger.first)
```
timriley: Do we even need the `is_a?(::Logger)` check if the `logger_interface?` method will return true for `::Logger` objects anyway?

davydovanton: Yep, we can use this method for `::Logger` too. That's why we can drop `is_a?(::Logger)` 🎉

Gemfile
```diff
@@ -49,6 +49,7 @@ gem 'shotgun',   '~> 0.9', require: false
 
 gem 'hanami-devtools', require: false, git: 'https://github.com/hanami/devtools.git'
 gem 'hanami-webconsole', require: false, git: 'https://github.com/hanami/webconsole.git'
+gem 'semantic_logger'
```
jodosha: @davydovanton Can you please remove this dependency? A custom, non-working logger can be mocked like this:

```ruby
class SemanticLogger
  def initialize(*)
  end

  def info(*)
  end

  def warn(*)
  end

  def debug(*)
  end
end
```

spec/isolation/components/logger/with_custom_logger_spec.rb
```diff
@@ -9,4 +9,15 @@
       expect(Hanami.logger.class).to eq(::Logger)
     end
   end
+
+  it "allows loggers not base on ::Logger" do
```
jodosha: s/base/based 😉

Gemfile
davydovanton: Awesome, will fix it today

davydovanton:
@jodosha done. could you check it again after CI?

PR #945, opened 2018-08-01
Add support for local .env files

ianks:
One feature that I love from `dotenv-rails` is the ability to create a `.env.development.local` file which will take precedence over the `.env.development` file.

The main use case I have for this right now is being able to specify a custom `CAPYBARA_DRIVER` so I can run the specs on my machine without starting a browser. Since my window manager is works differently than my other co-workers, specs fail on my machine since the window resizing works differently. Thus, specifying a local env configuration is the perfect solution for me.

Let me know if you would like anything changed, I would love to see this merged!

- Ian

### DIFF
--- a/lib/hanami/cli/commands/new/gitignore.erb
+++ b/lib/hanami/cli/commands/new/gitignore.erb
@@ -1,2 +1,3 @@
 /public/assets*
 /tmp
+.env.*.local
--- a/lib/hanami/cli/commands/new/gitignore_with_sqlite.erb
+++ b/lib/hanami/cli/commands/new/gitignore_with_sqlite.erb
@@ -1,3 +1,4 @@
 /db/*.sqlite
 /public/assets*
 /tmp
+.env.*.local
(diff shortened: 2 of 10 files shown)


cllns:
Related: https://github.com/hanami/hanami/pull/287

I tend to agree with the reasoning that was rejected too: this seems like an edge case.

Could you solve this by adding an alias? Something like `alias rspec="CAPYBARA_DRIVER=mydriver rspec"`?

ianks:
I could solve it that way, but that would be somewhat annoying for a couple reasons. 
- It wouldn't play very well with editors or environments that don't rely on specific shell profiles.
- It becomes tedious to maintain when adding multiple variables
- It would only work with one command (I.e. not hanami console, rake tasks, etc

Although this may be an edge case, it's incredibly useful IMO, and a nice feature of dotenv-rails. I would argue that this feature is absolutely harmless to those that don't want it, but very useful to those that do. This, it's a net win. 

Please consider merging this patch 😀

lfmalves:
It has been 9 months, any news in this front? Otherwise, Admin, I think it would be reasonable to close this.

ianks:
We have a very large production Hanami app, and without this feature development would be incrediy tedious. So we apply this patch in an initializer now. This is find, but somewhat annoying since it bloats the code base with bespoke code.

jodosha:
Closing in favor of https://github.com/hanami/hanami/pull/1004

PR #949, opened 2018-08-08
Getting Started Guide 404'ing

baweaver:
Not sure if this is the _right_ change per-se, but the current link is 404'ing.


jodosha:
@baweaver This isn't the correct fix, but better a quick and dirty one, that a 404. 👍

Issue #898, opened 2018-02-19
Automatic logging of body parameters of a JSON request


smaximov:
Hanami [automatically logs](http://hanamirb.org/guides/1.1/projects/logging/#automatic-logging) parameters sent by regular form submissions but fails to do so for JSON requests. This issue can be reproduced on 1.1 and the current `master` (8b0fb2e), see [this repo](https://github.com/smaximov/hanami_automatic_json_parameter_logger) for details.

The reason is that [`Hanami::CommonLogger#extract_params`](https://github.com/hanami/hanami/blob/8b0fb2e5d89c8f9ff6d3cc71851c99e29e4f9dd4/lib/hanami/common_logger.rb#L79-L81) fetches `rack.request.form_hash` from the environment. This variable is populated by default only for "application/x-www-form-urlencoded", "multipart/form-data" (see [R::R::Helpers#form_data?](https://github.com/rack/rack/blob/b0818268fd9f4026a1b0fb205cc1060ee38a23cc/lib/rack/request.rb#L302-L306) and [R::R::Helpers::FORM_DATA_MEDIA_TYPES](https://github.com/rack/rack/blob/b0818268fd9f4026a1b0fb205cc1060ee38a23cc/lib/rack/request.rb#L103-L106)), for "application/json" one can use [R::PostBodyContentTypeParser](https://github.com/rack/rack-contrib/blob/74afe70f0e0ff600aa1dd5ccc56ee5662c2fa19e/lib/rack/contrib/post_body_content_type_parser.rb).

Hanami's parsers in `hanami-router` [populate](https://github.com/hanami/router/blob/028f301a5b50da6544d4e0f557aed320259dea24/lib/hanami/routing/parsers.rb#L76) the [`router.parsed_body`](https://github.com/hanami/router/blob/028f301a5b50da6544d4e0f557aed320259dea24/lib/hanami/routing/parsers.rb#L26) environment variable which is ignored by `Hanami::CommonLogger#extract_params`.

# Workarounds

1. Patch `Hanami::CommonLogger#extract_params` so it uses `router.parsed_body`:

    ``` ruby
    module CommonLoggerPatch
      ROUTER_PARSED_BODY = 'router.parsed_body'

      def extract_params(env)
        super.merge(env[ROUTER_PARSED_BODY] || {})
      end
    end

    Hanami::CommonLogger.prepend CommonLoggerPatch
    ```

[...]

PR #958, opened 2018-10-09
Automatically log body payload from body parsers

jodosha:
## Enhancement

Automatic log for body parsers body payload.

---

Ref https://github.com/hanami/hanami/pull/818
Closes https://github.com/hanami/hanami/issues/898

### DIFF
--- a/lib/hanami/common_logger.rb
+++ b/lib/hanami/common_logger.rb
@@ -49,10 +49,6 @@ module Hanami
     # @api private
     ROUTER_PARAMS        = 'router.params'.freeze
 
-    # @since 1.3.0
-    # @api private
-    BODY_PARSER_FALLBACK_KEY = '_'.freeze
-
     # @since 1.0.0
     # @api private
     #
@@ -91,7 +87,6 @@ module Hanami
       result = env.fetch(QUERY_HASH, {})
       result.merge!(env.fetch(FORM_HASH, {}))
       result.merge!(Utils::Hash.deep_stringify(env.fetch(ROUTER_PARAMS, {})))
-      result.delete(BODY_PARSER_FALLBACK_KEY)
       result
     end
   end
--- a/spec/integration/cli/server_spec.rb
+++ b/spec/integration/cli/server_spec.rb
@@ -149,18 +149,24 @@ EOF
           inject_line_after "config/environment.rb", "Hanami.configure", "require 'hanami/middleware/body_parser'\nmiddleware.use Hanami::Middleware::BodyParser, :json"
 
           run_command "hanami generate action web books#create --method=POST", []
-          inject_line_after "apps/web/controllers/books/create.rb", "Action", "accept :json"
           inject_line_after "apps/web/controllers/books/create.rb", "call", 'Hanami.logger.debug(request.env["CONTENT_TYPE"]);self.body = %({"status":"OK"})'
 
           server do
+            post "/books", book: { title: "Why we sleep" }
             post "/books", JSON.generate(book: { title: "Parsers" }), "CONTENT_TYPE" => "application/json", "HTTP_ACCEPT" => "application/json"
+            post "/books", JSON.generate(%w[this is cool]), "CONTENT_TYPE" => "application/json", "HTTP_ACCEPT" => "application/json"
           end
 
           content = contents(log)
           expect(content).to include("[#{project}] [INFO]")
           expect(content).to include("POST 200")
+
+          expect(content).to include("application/x-www-form-urlencoded")
+          expect(content).to include(%({"book"=>{"title"=>"Why we sleep"}}))
+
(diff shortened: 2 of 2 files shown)


mereghost:
Can't this become kind of a security issue, since there are no filters to the payload?

lib/hanami/common_logger.rb
```diff
@@ -73,11 +82,17 @@ def log(env, status, header, began_at)
...
+    # rubocop:enable Metrics/AbcSize
 
     # @since 1.1.0
     # @api private
     def extract_params(env)
-      (env[QUERY_HASH] || {}).merge(env[FORM_HASH] || {})
+      result = env.fetch(QUERY_HASH, {})
+      result.merge!(env.fetch(FORM_HASH, {}))
+      result.merge!(Utils::Hash.deep_stringify(env.fetch(ROUTER_PARAMS, {})))
+      result.delete(BODY_PARSER_FALLBACK_KEY)
```
davydovanton: Just for better understanding: could you explain why we need to drop it? and I understand right that we will drop all `'_'` keys?

jodosha:
@mereghost Really good call 👍 

I just manually tested the following scenarios:

```ruby
# config/environment.rb
Hanami.configure do
    logger level: :info, filter: ["book.title"] # 1
    logger level: :info, formatter: :json, filter: ["title"] # 2
end
```

```shell
$ curl http://localhost:2300/books -d '{"book":{"title": "GO"}}' -H "Content-Type: application/json"
```

They return

##### 1

```log
[bookshelf] [INFO] [2018-10-09 12:36:03 +0200] HTTP/1.1 POST 200 ::1 /books 175 {"book"=>{"title"=>"[FILTERED]"}} 0.00323
```

##### 2

```log
{"app":"bookshelf","severity":"INFO","time":"2018-10-09T10:32:52Z","http":"HTTP/1.1","verb":"POST","status":"200","ip":"::1","path":"/books","length":"175","params":{"book":{"title":"[FILTERED]"}},"elapsed":0.007452}
```

lib/hanami/common_logger.rb
jodosha: @davydovanton You made me spot a problem with this. Here's why I Initially dropped `_`. 

---

Rack/Hanami params are a key/value bucket. It may happen that a JSON payload is an array (e.g. `'["a", "b", "c"]'`), so we don't know which key to assign to the array. For this reason, we assign this fallback key for cases like this. (eg. `{ "_" => ["a", "b", "c"]`)

See https://github.com/hanami/router/issues/85 and https://github.com/hanami/router/pull/86

---

When a request has a `Content-Type` equal to `x-www-form-urlencoded` (a web form submission), the JSON body parser doesn't recognize the payload, and it adds it to the fallback key. See https://github.com/hanami/router/issues/179

---

My clumsy attempt here was to avoid the automatic logger to print a messed up log.

```log
 {"book"=>{"title"=>"GO"}, "_"=>"book%5Btitle%5D=GO"}
```

---

But I forgot two things:

  1. There is a PR by @timriley to fix the body parsers in the first place. See https://github.com/hanami/router/pull/180
  1. If we remove `_` from here, it won't log the JSON arrays

---

I'm gonna change this. Thanks. 👍

kaikuchn:
If I understand this change correctly, then the feature is always on, right?
Is it possible to opt-out of request body logging? If so, could this even be opt-in for production?

I'm asking because I am slightly worried about the security implications. Since I've not logged request bodies before, I haven't had to worry about the redaction catching all sensitive information.
I also believe that in general people expect web servers to not log request bodies and would be surprised if it does by default, even more so if they cannot opt-out of it.

On the other hand logging the full request would be super useful for debugging.

jodosha:
@kaikuchn With the current state of this PR it isn't possible to opt-out this feature.

Is your concern only about security? We have filters to filter out sensitive information. Please have a look at my comment above. https://github.com/hanami/hanami/pull/958#issuecomment-428233963

lib/hanami/common_logger.rb
jodosha: @davydovanton Thanks to https://github.com/hanami/router/pull/181, we don't need to remove the fallback key anymore. See https://github.com/hanami/hanami/pull/958/commits/10b3e455ef097a211cecfb249617d1ccdb4efdc7

kaikuchn:
Yes, my concerns are about security.
While filtering/redacting ameliorates the situation it could always happen that something slips through the filters. Not having to worry about it at all by not logging the body feels like the best option to me.

But I don't have a strong opinion on this issue. This is just my 2 cents :)

jodosha:
@kaikuchn IMO filtering is a an active task that goes beyond the scope of this PR. For instance, we use to log also body from POST requests coming from web forms.

This PR extends the feature to any incoming non-GET request coming from both web forms and HTTP APIs.

PR #959, opened 2018-10-16
Generate correct syntax for layout unit tests

jodosha:
## Problem

**TL;DR:** Until Hanami v1.2.0 we generate a wrong spec for layout unit tests.

This is a followup for https://github.com/hanami/view/pull/156, please read the complete problem description over there.

## Fix

This PR fixes the problem, by generating a new syntax for layout unit tests:

```ruby
require "spec_helper"

RSpec.describe Web::Views::ApplicationLayout, type: :view do
  let(:layout)   { Web::Views::ApplicationLayout.new({ format: :html }, "contents") }
  let(:rendered) { layout.render }

  it "contains application name" do
    expect(rendered).to include("Web")
  end
end
```

It's worth noting that the old, wrong, syntax will still work for backward compatibility. This backward compatible strategy has been implemented by https://github.com/hanami/view/pull/156.

### DIFF
--- a/lib/hanami/cli/commands/generate/app/layout_spec.minitest.erb
+++ b/lib/hanami/cli/commands/generate/app/layout_spec.minitest.erb
@@ -1,9 +1,8 @@
 require "spec_helper"
 
 describe <%= app.classify %>::Views::ApplicationLayout do
-  let(:layout)   { <%= app.classify %>::Views::ApplicationLayout.new(template, {}) }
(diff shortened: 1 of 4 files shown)


jodosha:
The Travis CI `push` failed because I manually terminated a `jruby-head` job that was taking too long. That job is in the list of the allowed failures, so it should've been not mark the entire build as failed.

PR #976, opened 2018-12-16
Fix :host mount option

### LINKED ISSUE #951: Mounting on a subdomain is broken
Specifying `host:` when mounting applications in Hanami seems to be broken.
I have the following setup:
```
Hanami.configure do
  mount TestOne::Application, at: '/', host: 'test-one'
  mount TestTwo::Application, at: '/', host: 'test-two'
# ...
end
```
In both applications, I define the following route:
```
get '/hello', to: ->(env) { [200, {}, ['Hello from Test One/Two!']] }
```
When I run my application with `hanami server --host=lvh.me`, as described in http://hanamirb.org/guides/1.2/routing/overview/#mounting-on-a-subdomain and try to call `test-two.lvh.me/hello`, I end up in TestOne.

Moreover, if I stick to the documentation and leave out the `at:` in mount, I end up getting
```
key not found: :at
/Users/user/.rvm/gems/ruby-2.5.1/gems/hanami-1.2.0/lib/hanami/configuration.rb:35:in `fetch'
```

Setting `at: ` to an empty String, it does not build any route (only getting 404s).

AlfonsoUceda:
This PR fix #951

### DIFF
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -62,7 +62,7 @@ module Hanami
     # @api private
     def mount(configuration)
       configuration.mounted.each do |klass, app|
-        routes.mount(klass, at: app.path_prefix)
+        routes.mount(klass, at: app.path_prefix, host: app.host)
       end
     end
 
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -62,7 +62,9 @@ module Hanami
     # @api private
     def mount(configuration)
       configuration.mounted.each do |klass, app|
-        routes.mount(klass, at: app.path_prefix, host: app.host)
+        # TODO: think if this line is in the correct place
+        host = [app.host, Components['environment'].host].compact.join('.') if Components['environment'].host != 'localhost'
+        routes.mount(klass, at: app.path_prefix, host: host)
       end
     end
 
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -62,9 +62,7 @@ module Hanami
     # @api private
     def mount(configuration)
       configuration.mounted.each do |klass, app|
-        # TODO: think if this line is in the correct place
-        host = [app.host, Components['environment'].host].compact.join('.') if Components['environment'].host != 'localhost'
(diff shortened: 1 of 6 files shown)


lib/hanami/app.rb
```diff
@@ -62,7 +62,9 @@ def call(env)
     # @api private
     def mount(configuration)
       configuration.mounted.each do |klass, app|
-        routes.mount(klass, at: app.path_prefix)
+        # TODO: think if this line is in the correct place
+        host = [app.host, Components['environment'].host].compact.join('.') if Components['environment'].host != 'localhost'
```
jodosha: @AlfonsoUceda Why we can't just forward the given `host`?

AlfonsoUceda: I've not added a spec yet but the problem is when you combine option `:host` in mount DSL  and option `--host` in server command.

checkbutton:
Why was this closed and never merged? @jodosha @AlfonsoUceda

cllns:
@checkbutton Sorry about that! I've restored the branch and I'm re-opening the PR. I'm not sure why it was closed. It might've been a mistake but if it's not @AlfonsoUceda can close it again and give a brief explanation.

Also, for your info, Hanami 2.0 will have an [entirely new router](https://github.com/hanami/router/tree/unstable). This should definitely work on 1.3 though, before that gets released.

AlfonsoUceda:
@checkbutton this was a proof of concept to understand the problem and if I could make it work. It works fine if I remember well but I never finished tests. I asked @jodosha if he can take a look ;)

lib/hanami/app.rb
glaszig: tested this patch _without_ `--host` param with lvh.me and it works. no need for this special joining of mount host and env host. if something like this is supposed to be a feature i'd introduce something like:

```rb
mount HelloWorld::Application, at: "/", subdomain: "hello-world"
```

which would serve requests like `https://hello-world.example.com` while the `host` depends on the request's host header if no `host: "..."` option was given for the mount.

PR #977, opened 2018-12-16
`body_parsers` is deprecated but still generated into new projects

### LINKED ISSUE #972: `body_parsers` is deprecated but still generated into new projects
According to https://guides.hanamirb.org/upgrade-notes/v130/ , `body_parsers` is deprecated but when I started new project using hanami 1.3, I still see that in `application.rb`.

```
$ cat apps/*/application.rb | grep body_parsers
      # body_parsers :json
```

acaporrini:
Closes #972

### DIFF
--- a/lib/hanami/cli/commands/generate/app/application.erb
+++ b/lib/hanami/cli/commands/generate/app/application.erb
@@ -97,14 +97,6 @@ module <%= app.classify %>
       #
       # default_response_format :html
 
-      # HTTP Body parsers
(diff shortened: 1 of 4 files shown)


acaporrini:
> @acaporrini Thanks for this PR. Can you please restore the integration tests?
> 
> Body parsers are deprecated, but still present as feature, and we need to keep testing them. Thanks.

@jodosha I updated the PR, please let me know if it's ok now. Thank you!

PR #986, opened 2019-01-28
Update copyright to 2019 [ci skip]

kaspermeyer:
Time for the annual copyright bump. Let me know if you want me to update the individual gems as well.


kaspermeyer:
I'm just realizing now that I should've added `[ci skip]` to the commit message, not the PR description. Sorry about that.

cllns:
For what it's worth, I don't think we need to keep doing this; we can drop the date entirely: https://stackoverflow.com/a/2391555

PR #996, opened 2019-04-18
Remove unnecessary spaces

matttrix1999:
I think one space after `~>` would be better


mereghost:
Thanks for catching that. :D

Issue #1003, opened 2019-06-12
Support default dotenv rails files


gmanley:
The dotenv integration with Rails supports [a number of default files](https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/rails.rb#L58) that the Hanami integration does not. Specifically the .env.local style files. I like to use the .env.local files as a place to store things I don't want to commit (I have it in the .gitignore). If I make a PR adding this functionality is that a contribution you all would be willing to accept? From what I can tell it will be a change to [this method](https://github.com/hanami/hanami/blob/master/lib/hanami/environment.rb#L489). I just wanted to check in before implementing it in case this was intentional or there is an opinion to keep the one default file pattern of .env.development.

Thanks!

davydovanton:
hey, thanks for report

Not sure that it can be helpful but I use `.env.development.sample` in my projects for commiting env file to the git repository and `.env.development` for working locally. You can [check my example](https://github.com/davydovanton/rubyjobs.dev) if you want.

@jodosha WDYT about this?

jodosha:
@davydovanton I think @gmanley is telling something slightly different.

Is the proposal to make create the following hierarchy from the lower to the higher priority?

  1. Parse `.env`
  1. Parse `.env.development` (potentially override `.env`)
  1. Parse `.env.development.local`  (potentially override `.env` and/or `.env.development`)

The idea is to keep generating the first two with `hanami new`, but potentially supporting also the third case, if that file is present.

If this is the idea, @gmanley please proceed. 👍 

---

Sample files (e.g. `.env.sample`) should be designed better as another feature, which I think goes beyond this specific need.

gmanley:
@jodosha That's correct. That is the intent. Thanks for laying it out so clearly. I'll start working on it this weekend. Thanks!

gmanley:
@davydovanton I do use sample files as well but as @jodosha explained the idea here is to have hierarchy that can allow overrides locally. I appreciate the example, however.

gmanley:
@jodosha @davydovanton https://github.com/hanami/hanami/pull/1004 here is my first pass to get some feedback.

jodosha:
Implemented by https://github.com/hanami/hanami/pull/1004

PR #1004, opened 2019-06-14
Standardize file loading for .env

gmanley:
This is based off of the conversation here: https://github.com/hanami/hanami/issues/1003

It adds support for same default dotenv files as the Rails integration: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use

The odd case is that `.env.local` isn't loaded in test mode. My guess on the rational is  that they want to prevent things like api keys for 3rd party services being inadvertently used while running specs. This, however, seems like it could be solved by just using a `.env.test.local` file which has higher priority.  I wanted to first implement exactly like it was in the rails integration and then have a further discussion here on if Hanami wants the same.

Also it seems like there are [some docs](https://github.com/hanami/hanami/blob/master/lib/hanami/environment.rb#L135) that are pretty clear that `.env` was specifically excluded so that people would use environment specific files right away. If that is still the case I can remove the `.env` parsing. 

If an environment variable is already set [the code](https://github.com/hanami/hanami/blob/master/lib/hanami/env.rb#L62) does not overwrite it. This means if a variable is in multiple files the variable in first file loaded is what is used. This is why the constant listing the files is ordered by descending priority.

I still need to modify the docs as well but I wanted to first make sure that this behavior is what is desired.


lib/hanami/environment.rb
```diff
@@ -487,10 +492,11 @@ def set_hanami_env_vars!
     # @since 0.2.0
     # @api private
     def set_application_env_vars!
-      dotenv = root.join(DEFAULT_DOTENV_ENV % environment)
-      return unless dotenv.exist?
-
-      env.load!(dotenv)
+      DEFAULT_DOTENV_ENV_FILES.each do |filename_format|
+        next if filename_format == '.env.local' && environment == 'test'
```
gmanley: I really don't like this. This method could also look like:

```ruby
    def set_application_env_vars!
      [
        root.join(".env.#{environment}.local"),
        (root.join(".env.local") unless environment == 'test'),
        root.join(".env.#{environment}"),
        root.join(".env")
      ].compact.each do |path|
        env.load!(path) if path.exist?
      end
    end
```

But that gets away from the use of constants which I was trying to adhere to.

lib/hanami/environment.rb
```diff
@@ -52,7 +52,12 @@ class Environment
     #
     # @since 0.2.0
     # @api private
-    DEFAULT_DOTENV_ENV = '.env.%s'.freeze
+    DEFAULT_DOTENV_ENV_FILES = [
```
davydovanton: WDYT about `DEFAULT_DOTENV_ENV_PATTERNS`?

gmanley: I like that better. I'll make the change.

gmanley:
@jodosha Is this looking ok from your perspective? Should I go ahead and update the comments/docs with this behavior?

lib/hanami/environment.rb
```diff
@@ -132,9 +138,8 @@ class Environment
     # located under the config directory. All the settings in those files will
     # be exported as `ENV` variables.
     #
-    # Master .env file is ignored to suggest clear separation of environment
-    # configurations and discourage putting sensitive information into source
-    # control.
+    # This table: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
```
gmanley: I'm not sure the nest place to put this info but the table on this page is very useful so I think it's good to include it somewhere.

lib/hanami/environment.rb
```diff
@@ -487,10 +492,11 @@ def set_hanami_env_vars!
     # @since 0.2.0
     # @api private
     def set_application_env_vars!
-      dotenv = root.join(DEFAULT_DOTENV_ENV % environment)
-      return unless dotenv.exist?
-
-      env.load!(dotenv)
+      DEFAULT_DOTENV_ENV_PATTERNS.each do |filename_format|
+        next if filename_format == '.env.local' && environment == 'test'
```
jodosha: @gmanley why `.env.local` isn't loaded if env is test?

lib/hanami/environment.rb
jodosha: The second line will produce a `nil` value if `environment == "test"` is true. That means `path.exist?` will crash.

lib/hanami/environment.rb
```diff
@@ -48,11 +48,17 @@ class Environment
...
+    # to lowest priority.
     #
     # @since 0.2.0
     # @api private
-    DEFAULT_DOTENV_ENV = '.env.%s'.freeze
+    DEFAULT_DOTENV_ENV_PATTERNS = [
+      '.env.%{environment}.local'.freeze,
+      '.env.local'.freeze,
+      '.env.%{environment}'.freeze,
+      '.env'.freeze
```
jodosha: @gmanley Is `.env` here really needed?

lib/hanami/environment.rb
gmanley: @jodosha I was trying to match behavior with dotenv rails. If you look at the chart: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use `.env.local` is ignored in test. I assume the rational is you don't want real api keys set. I can change this, though it may be confusing to have different behavior between dotenv-rails and hanami. That is kind of what I was trying to get away from.

lib/hanami/environment.rb
gmanley: Yeah, you are right. I would to need to do `path &&  path.exist?`. Do you prefer this way of doing it?

lib/hanami/environment.rb
gmanley: Are you suggesting not loading `.env` at all? Do you prefer to have less options and guide people to always use .env.local or environment specific ones?

gmanley:
@jodosha I've responded to your questions, could you take another look when you get the chance?

lib/hanami/environment.rb
jodosha: Yes, but why do we need to exclude `.env.local` for `test` environment?

jodosha: @gmanley Initially I missed the call to `Array#compact` which resolves that problem. However, `path & path.exist?` allocates less memory: https://gist.github.com/jodosha/4c6629f870a5d78aec8116e301140447

jodosha: @gmanley I'm fine to keep the code as it is with `DEFAULT_DOTENV_ENV_PATTERNS`, but we should extract a private method for that special check for test env.

Something like:

```ruby
next unless dotenv_applicable?(filename_format)
# ...

# @api private
# @see https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use
def dotenv_applicable?(filename_format)
  # ...
end
```

lib/hanami/environment.rb
jodosha: @gmanley Fine with me, thanks for the clarification. 👍

jodosha:
@gmanley I was offline for a few weeks. Sorry for the long wait.

lib/hanami/environment.rb
jodosha: Addressed here: https://github.com/hanami/hanami/pull/1004/commits/6dcc93417abc81f6d6bf2de729d0581dfb50392a

lib/hanami/environment.rb
jodosha: Addressed here: https://github.com/hanami/hanami/pull/1004/commits/6dcc93417abc81f6d6bf2de729d0581dfb50392a

PR #1007, opened 2019-07-11
Drone CI

jodosha:


script/test
```diff
@@ -0,0 +1,36 @@
...
+  done
+}
+
+run_test() {
+  local root=$1
+  local test=$2
+  local hash="$(shasum "$test" | cut -b 1-40)"
+
+  printf "\n\n\nRunning: $test\n"
+  SIMPLECOV_COMMAND_NAME=$hash bundle exec rspec $test
```
landongrindheim: I know this change is specific to CI, but since the filename here is not CI-specific, what do you think about issuing this command prefixed with `HANAMI_ENV=test`?

I ran the `script/ci` script locally to test something and ran into this:

```bash
  1) Hanami::MiddlewareStack#load! loads the middleware stack without duplicates
     Failure/Error: expect(middleware.__send__(:stack).count).to be(6)

       expected #<Integer:13> => 6
            got #<Integer:15> => 7

       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     # ./spec/isolation/middleware/ensure_unique_stack_spec.rb:52:in `block (4 levels) in <top (required)>'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/with_project.rb:26:in `block (2 levels) in with_project'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/within_project_directory.rb:28:in `block in within_project_directory'
     # ./vendor/bundle/gems/aruba-0.14.10/lib/aruba/platforms/local_environment.rb:22:in `call'
     # ./vendor/bundle/gems/aruba-0.14.10/lib/aruba/platforms/unix_platform.rb:78:in `with_environment'
     # ./vendor/bundle/gems/aruba-0.14.10/lib/aruba/api/core.rb:191:in `with_environment'
     # ./vendor/bundle/gems/aruba-0.14.10/lib/aruba/api/core.rb:71:in `cd'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/within_project_directory.rb:18:in `within_project_directory'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/with_project.rb:22:in `block in with_project'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/with_tmp_directory.rb:22:in `block in with_tmp_directory'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/with_directory.rb:20:in `with_directory'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/with_tmp_directory.rb:21:in `with_tmp_directory'
     # ./vendor/cache/devtools-23f0356f8bd8/lib/hanami/devtools/integration/with_project.rb:19:in `with_project'
     # ./spec/isolation/middleware/ensure_unique_stack_spec.rb:4:in `block (3 levels) in <top (required)>'

Finished in 2.34 seconds (files took 0.55033 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/isolation/middleware/ensure_unique_stack_spec.rb:3 # Hanami::MiddlewareStack#load! loads the middleware stack without duplicates
```

The extra middleware was `Hanami::Welcome`, which is being loaded when the env is not test:

```ruby
    def _load_default_welcome_page
      unless Hanami.env?(:test) || routes.defined?
        require 'hanami/welcome'
        use Hanami::Welcome
      end
    end
```

jodosha: @landongrindheim Hello 👋 `HANAMI_ENV` is meant to be used by Hanami apps, it has a meaning only there, not for the CI of Hanami. I know, it's cumbersome. 😅 

The purpose of `HANAMI_ENV` is to set behavioral changes for an Hanami app. It's not meant to be always used, and by not using it, Hanami falls back to the default value: `"development"`.

Because with the CI, we want to test how an Hanami app behaves **with and without** that env variable, **we cannot force it**.

For instance, while simulating code generators (`hanami generate model foo`), the enviroment in a real world scenario would be `"development"`, not `"test"`. By setting `HANAMI_ENV=test` we would alter this simulation.

landongrindheim: @jodosha Thanks for the explanation! Totally makes sense 🙂 

Do you have any suggestions for testing when doing local development? I was following the instructions on the README when I ran into this. Really enjoying working with Hanami 😄

jodosha: @landongrindheim If you install Docker and [Drone CLI](https://docs.drone.io/cli/install/), then you can locally reproduce the build with:

```shell
$ drone exec .drone.yml --pipeline=ruby-2-6 --timeout=2h
```

It will only test against Ruby 2.6.

---

Thanks for your kind works 😄 Do you use Hanami at work?

landongrindheim: @jodosha We don't run Hanami at work, but I _have_ been using it as an example of how to maintain good OO design as an app grows. Others have appreciated how its conventions nudge developers toward making better decisions than some of our current tooling.

I've been using Hanami on my own, as well as seeing where I can help out with the project 🙂

PR #1009, opened 2019-08-02
Excluding .env* files from git by default

wout:
A tiny contribution, but excluding sensitive data from the repo is generally considered good practice.

Even better would be encoding `.env` files with a master key, à la Rails, using the cli. If that's something you would consider merging, I'd be happy to contribute it for Hanami 2.0.0.

### DIFF
--- a/lib/hanami/cli/commands/new/gitignore.erb
+++ b/lib/hanami/cli/commands/new/gitignore.erb
@@ -1,2 +1,3 @@
 /public/assets*
 /tmp
+.env.*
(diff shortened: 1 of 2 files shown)


jodosha:
@wout Why did you close the PR?

wout:
@jodosha Good question, I honestly don't recall. :blush:  Shall I create a new PR?

jodosha:
@wout After https://github.com/hanami/hanami/pull/1004 was merged, we now generate new projects by adding `.env.local` and `.env.*.local` to `.gitignore`.

According to [`dotenv`](https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use) prescriptions regarding Git, they suggest to not ignore `.env.development` et all.

We're good to go, no need for a new PR. Thank you anyway. 👍

PR #1011, opened 2019-08-05
Use `sassc` for tests

jodosha:
Ref https://github.com/hanami/assets/pull/102

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,7 @@ if RUBY_DESCRIPTION =~ /linux/
 end
 
 # `hanami assets` integration tests
-gem 'sass',          require: false
+gem 'sassc',         require: false
 gem 'coffee-script', require: false
 
 gem 'dotenv',    '~> 2.4', require: false
--- a/spec/integration/assets/serve_spec.rb
+++ b/spec/integration/assets/serve_spec.rb
@@ -3,7 +3,7 @@ RSpec.describe "assets", type: :integration do
     it "compiles and serves assets in development mode" do
       project = "bookshelf_serve_assets"
 
-      with_project(project, gems: ['sass']) do
+      with_project(project, gems: ['sassc']) do
         generate "action web home#index --url=/"
 
         write "apps/web/assets/javascripts/application.css.sass", <<-EOF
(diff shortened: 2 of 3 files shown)


landongrindheim:
@jodosha I saw the Assets change isn't playing well with Hanami 🙁 Do you have any idea why? I've tried to do some debugging, but haven't been able to trigger any logging/`binding.irb` interaction.

I see the `pr` build is passing, but `push` isn't

Gemfile
```diff
@@ -41,7 +41,7 @@ if RUBY_DESCRIPTION =~ /linux/
 end
 
 # `hanami assets` integration tests
-gem 'sass',          require: false
+gem 'sassc',         require: false
```
landongrindheim: Is this needed so that the gem is available globally (so `bundle install` doesn't need to be run in specs)?

jodosha: @landongrindheim TL;DR: We do one `bundle install` at the beginning of the test suite. It vendors all the needed gems for all the kind of tests.

---

We generate an gem index in the `vendor/cache` directory, which is the same used by Bundler to install the gems. At this point, the CI job is self-sufficient and it doesn't need to contact Rubygems anymore.

---

Why do we do this?

Because to make integration specs more realistic, for each test, we generate a brand new app with all the settings that it needs for that specific scenario under test (e.g. enabled sessions, use minitest, don't use `hanami-model`, use Subresource Integrity, etc..).

At this point, it would be onerous to contact Rubygems for each test, so we modify the default source of the `Gemfile` of each generated app:

```diff
- source "https://rubygems.org"
+ source "file:///path/to/vendor/cache"
```

Because of the gem index, `vendor/cache` can be used as a source in `Gemfile`
Then we run `bundle install --local`, which forces Bundler to use only the local gems, which are vendored in `vendor/cache`.

---

If you want a complete picture, read the code of `lib/hanami/devtools/integration/bundler.rb` from https://github.com/hanami/devtools.

landongrindheim: **Great** explanation! Thank you!

I know this has been merged now, but would it make sense to make `sassc` a run time dependency instead of [a development dependency of Assets](https://github.com/hanami/assets/blob/master/hanami-assets.gemspec#L34)? Would that remove the need to declare it in Hanami's Gemfile?

jodosha: Nope. That would turn `sassc` into the default compressor, but that isn't the case.

PR #1014, opened 2019-09-18
Save rack.exception to log

lfidnl:
Hello!
I found following issue: lost exception info when I run application In production environment. For example, it happens when exception raises in View. In that case I see only HTTP 500 in my log without any information about exception.
This simple fix save exception info and other information about request in log.

### DIFF
--- a/lib/hanami/common_logger.rb
+++ b/lib/hanami/common_logger.rb
@@ -37,6 +37,10 @@ module Hanami
     # @api private
     RACK_ERRORS          = 'rack.errors'.freeze
 
+    # @since 1.3.3
+    # @api private
+    RACK_EXCEPTION       = 'rack.exception'.freeze
+
     # @since 1.1.0
     # @api private
     QUERY_HASH           = 'rack.request.query_hash'.freeze
@@ -69,13 +73,15 @@ module Hanami
         elapsed: now - began_at
       ]
 
+      add_exception_info!(msg, env[RACK_EXCEPTION]) if env[RACK_EXCEPTION]
+
       logger = @logger || env[RACK_ERRORS]
       # Standard library logger doesn't support write but it supports << which actually
       # calls to write on the log device without formatting
       if logger.respond_to?(:write)
         logger.write(msg)
       else
-        logger.info(msg)
+        env[RACK_EXCEPTION] ? logger.error(msg) : logger.info(msg)
       end
     end
     # rubocop:enable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


jodosha:
@lfidnl Hey, thanks for this enhancement! Do you mind to provide a test for this, or at least tell us how to manual test it? I can help with the automated test, if needed.

lfidnl:
@jodosha thanks for quick response! If my approach acceptable I'll try to write some tests :)

lfidnl:
I've added test for testing exception info in log message

lfidnl:
@jodosha ok, I'll check it

lfidnl:
@jodosha , I've completed an investigation. Duplication of exception happens because the first message is a `RACK_ERRORS` (https://github.com/hanami/controller/blob/c12e07954d56b5df5da6e5f9dd05f48634ebc38e/lib/hanami/action/throwable.rb#L160) and the second is `RACK_EXCEPTIONS` (enhancement in this PR). 
This reproduces only if we have exception in `Action`.
But when we have exception in `View` we setup only `RACK_EXCEPTIONS` (https://github.com/hanami/hanami/blob/master/lib/hanami/rendering_policy.rb#L60)
Maybe we should stop using `RACK_ERRORS` for tracking exceptions which raised during application execution? What you think about it?

jodosha:
@lfidnl Thanks for having a look at it. Did you find what's the actor that prints `RACK_ERRORS` from Rack env?

jodosha:
@lfidnl More context on `rack.errors` vs `rack.exception`.

The first one is an IO stream that is part of the [Rack SPEC](https://www.rubydoc.info/github/rack/rack/file/SPEC). This could be very likely the one that prints to stderr, so it duplicates the exception in console.

The second one is **not** part of the Rack SPEC, but historically bug reporting SaaS companies start to use it, so they created this cargo cult that nobody knows when it was generated exactly. See https://github.com/hanami/controller/pull/134

---

Given `rack.errors` is required by the Rack SPEC, and `rack.exception` is required by bug reporting companies, we shouldn't change their usage, but control how things are printed in console.

lfidnl:
@jodosha Thanks. I understood difference between `rack.errors` and `rack.exceptions`. I thought that `rack.errors` should be used for trackign service errors, such as misconfigured rack-app & etc. I think this PR is unacceptable. We should save error to `rack.errors` here https://github.com/hanami/hanami/blob/master/lib/hanami/rendering_policy.rb#L60 as well as `rack.exceptions` (like https://github.com/hanami/controller/blob/c12e07954d56b5df5da6e5f9dd05f48634ebc38e/lib/hanami/action/throwable.rb#L160) beacuse root-cause for me was lossing exception-info in view on production env.
What you think about it?

jodosha:
@lfidnl I agree. Would you like to create a different PR for fixing it?

lfidnl:
Sure, I'll complete it

PR #1019, opened 2019-11-03
New dry-system-powered application and slices structure

timriley:
Hello! Here in this PR we have a “native” dry-system integration as part of the Hanami core application structure. By “native” I mean an approach that lets us take advantage of as many dry-system features as possible, while also exposing those features to application authors for use within their apps.

The key aspect of this arrangement is one of a single _application_ and multiple _slices_. The application acts as both the boot coordinator as well as the “core” of the overall system. Each slice represents a discrete system of functionality. Slices gain access to the application’s facilities, and they’ll eventually be able to import other slices too. The end result is that we end up with the slices and application forming a _graph_ of the application’s high-level areas of functionality.

Here’s the kind of application structure that will now be possible (please excused the hurried diagram):

hanami-application-graph

This is made possible thanks to features of dry-system, which in this new Hanami arrangement, we use to the fullest extent. This kind of application structure has _also_ been proven in production over multiple applications and multiple years (within Icelab, and both @solnic and @flash-gordon 
 maintain similar apps too).

Now here’s how it all works. There’s a lot to get through here, but I’ll try hit on all the important elements:

## `Hanami::Application`

This class now represents the _application_ in the description above. It’s meant to be subclassed once (i.e. `YourApp::Application < Hanami::Application`), which is similar to Hanami v1 AFAIK.

What’s different here is that the application _class _represents a stateful, bootable app. The _instance_ of this class is just the rack application (we’ll get to that later). So hanami application exposes a few lifecycle methods:

- `.init`, which prepares the container and detects the slices
- `.init_bootable` and `.start_bootable`, which init or started individual bootable components (registered via `.register_bootable`)
- `.boot`, which finalizes the application container, boots all the slices, and registers an instance of the rack app

The application container is managed internally as a separate object. It’s defined automatically, and configured to ensure it behaves as we require. When it comes to configuration, I want to make it so that all the important settings are defined on `Hanami::Configuration` so regular users don’t need to worry about container internals. I’ve done it for some things alredy, but some more will need to be done before this is fully ready.

The container is available as `YourApp::Application.container` and is also assigned to `YourApp::Container` for convenient access.

The container’s auto-injection mixin is also assigned to `YourApp::Deps` for easy inclusion into classes.

The result is that you get all the container facilities without having to worry about setting it up yourself (or even having to see a bunch of boilerplate files inside your app).

The application also offers a small subset of the container’s methods for direct access (the ones that will most commonly be used when interacting with the app): `.keys`, `.resolve`, and `.[]`

## `Hanami::Slice`

Unlike `Application`, slices are regular classes and instantiated normally. They’re built automatically, detected from the directories inside `slices/`.

Each slice behaves in a similar way to the application. It:

- defines and configures a container
- imports the application container so all its keys are available under the “application.” namespace (“application.logger”, etc.)
- makes it accessible as `.container`
- assigns it to `SliceName::Container` for convenient access
- and assigns the auto-injection mixin to `SliceName::Deps`

Slices have the same lifecycle methods as the application (.init and .boot), support having their own bootable components, and also provide the same convenience methods that delegate to the container (`.keys`, `.resolve`, and `.[]`)

## Resulting code structure

With all of this in place, we can now write our application code as we would with dry-system normally, e.g.

```ruby
require "slice_name/some_slice_mapper"  

module SliceName
  class DoSomething 
    include Deps["article_repo"]

    def call(params)
      params = SomeSliceMapper.(params)
      article_repo.do_somethiong_with(params)
    end
  end
end
```

Some aspects of this I want to point out:

1. The auto-injection mixin is now available without any extra work, thanks to the hanami application and slice init process I described above
2. We’re relying on dry-system auto-injection to populate the containers, which means that if we’re referring to other constants within our app, we need to “require what we require” to ensure they’re available when the file is loaded.
3. Because all these files need to be requirable (both for manual requires _and_ for dry-system's auto-registration, at least in its current form), we need their paths on the file system to match their module/class hierarchy, so e.g. `SliceName::DoSomething` needs to be at `slices/slice_name/lib/slice_name/do_something.rb`

## Router enhancements

There’s an enhancement to the router to go along with the application/slices structure. It provides:

- Endpoint resolution based on container registrations (i.e. `root to: "home.show"` will load the object registered at `"actions.home.show"`)
- An enhancement to `#mount` (when called with a symbol, e.g. `mount :admin, at: "/admin"`) such that any endpoints resolved within that block are done so using that slice’s own container
- A new “context” option for the router (I’ll have a PR on hanami/router for this) so that an optional argument can be passed to the router/mount blocks. For Hanami apps, either the application object or the mounted slice object will be passed as the block arguments.
- Support for declaring middleware inline. This makes it possible to use a middleware _only_ within a specific subset of the application routes.

The end result is that this kind of thing is now possible:

```ruby
Hanami.application.routes do |application|
  use Rack::Lint

  use Rack::Static,
    urls: ["/assets"],
    root: "public"

  # Cleanly access application resources inside routes!
  use Rack::Session::Cookie,
    key: "decaf_sucks.session",
    secret: application[:settings].session_secret,
    expire_after: 10 * 365 * 24 * 60 * 60 # 10 years

  mount :main, at: "/" do
    root to: "home"
  end

  mount :admin, at: "/admin" do |slice|
    # Only authenticate /admin!
    use Rack::Auth::Basic, "Decaf Sucks" do |username, password|
      password == slice[:settings].admin_password
    end

    root to: "home"
  end
end
```

## On naming

It wouldn't be a typical PR from me if I didn't talk about naming ;)

I renamed:

- `Hanami.application_class` back to `Hanami.application`, and
- `Hanami.application` back to `Hanami.app`

IIRC these were the names Luca started with at the beginning of the v2 rewrites. They make a lot more sense, I think, now that `Hanami.application` (i.e. `YourApp::Application`) _is_ the main application (i.e. the application object with its own boot lifecycle). I'm not 100% sure about the `.app` name, since it's really more of a "rack-mountable web app" rather than just an "app", but OTOH at least it's short.

Also, regarding the auto-injection mixin, I decided to call it `Deps` instead our historically used `Import` for a few reasons:

- The `Import` module is actually _exporting_ (the dependencies) rather than importing them, if you think about it hard enough (@flash-gordon convinced me of this). In the end, I think it would be good to avoid the possibility of confusion here entirely) 
- An "Import" is actually a pretty commonly used concept within applications, and it is nice not to have the naming clash (it's certainly something we've had to deal with in our apps)
- In comparison to this, `Deps` is nice and short and easy to type, and _not_ a verb, so there's no confusion there, and it still reads nice and succinctly. It's also gets people thinking about "dependencies" while they're building out their classes.

## Other small bits of note

I upgraded `hanami/boot.rb` to keep it working like it was before, i.e. just a simple `require "hamami/boot"` will find and boot the Hanami application. Right now I’m doing this naively, simply by looking for `config/application.rb` to exist relative to the current working directory. Later we could consider making this a bit more sophisticated like how bundler searches for a relevant Gemfile.

Speaking of Bundler though, I actually extracted that application detection into an `hanami/setup.rb` file, with a name to mirror the `bundler/setup` we see so often in Ruby code. This actually works nicely given that in certain cases you may _not_ want to fully boot your Hanami application. So now you can just init it only, e.g.

```ruby
require "hanami/setup"
Hanami.init
```

## Demos

You can see this whole structure in action here:

- [This PR to Luca’s soundeck](https://github.com/jodosha/soundeck/pull/1), a very minimal native hanami v2 app
- [My decaf_sucks app](https://github.com/icelab/decaf_sucks/tree/hanami-application-and-slices), which started off as a hybrid dry-system/hanami/rom-rb app, but I’ve updated various parts as required to work with this hanami v2 application/slices structure

## TODO

I’ve gone far enough with this work that I think it’s ready for an early review. Before we can merge this PR, however, I still want to do the following:

- [x] Get the existing tests passing (of course!)
- [x] Make application `root` configurable
- [x] Make more the application/slice containers configurable through `Hanami::Configuration` settings.
- [x] Along with this, also provide a place in the configuration where a user can just pass a block that can be eval’ed within the various containers, as the ultimate escape hatch.
- [x] Add support for slices to import other slices (I’ve already built this in other work, I just need to get it enabled here)
- [x] Get the PRs to hanami-rotuer and hanami-controller merged (https://github.com/hanami/router/pull/192 & https://github.com/hanami/controller/pull/299)
- [x] ~Enhance hanami-controller so that it provides a way to have the Hanami application configuration applied automatically to a base class used for all other actions in the app (e.g. `YourApp::Action < Hanami::Action[:application]`)~ – **Update:** discussed this with @jodosha and doing it well will require some significant changes in hanami-controller. For the sake of merging this PR we're going to ignore this and come back to it.
- [x] Ensure dry-system will lazy load components when asked for them via `.key?`

I think that’s about it.

Then once this is merged I’d like to consider the following:

- [ ] Work out a formal way of handling application-level settings (maybe it's as simple as keeping Dotenv.load, maybe it's something more sophisticated, this is TBD)
- [ ] Reconcile the rack logging/filtering so it fits properly with the filtering in Hanami's own logger
- [ ] Looking at a way for a more streamlined use of auto-injection, via class-level macros instead of module inclusion. I think this would just “feel” better and probably make this idea more palatable to a wider range of people.
- [ ] Ensure it’s at least possible to build non-web Hanami v2 apps, i.e. they should run fine without routes, and we should avoid requiring web-specific files or libraries unless they’re really going to be used
- [ ] Make it possible to declare which slices should/should not be booted in a given circumstance (e.g. it's not uncommon to have subsystems that need to exist for rake tasks or background jobs _only_, and should not be booted when running the web-facing part of an app)
- [ ] Make sure slice customisation in application configuration is as flexible as it should be
- [ ] Customise dry-system importer to not load redundant `*.application.*` registrations on slices imported by other slices

## Next steps

There's a lot to digest here, but what I'd really like in terms of feedback right now is whether people are happy with the high-level structures here. If we're good with that, I think it'd be good to merge this into unstable and then have smaller iterative loops in the individual details (as @jodosha has suggested to me in a review of a previous iteration of this work).


lib/hanami/configuration.rb
```diff
@@ -61,6 +62,53 @@ def env
...
+    DEFAULT_LOGGING_FILTER_PARAMS = %w[_csrf password password_confirmation].freeze
+
+    def logging_filter_params=(params)
+      settings[:logging_filter_params] = params
+    end
+
+    def logging_filter_params
+      settings[:logging_filter_params]
+    end
+
```
timriley: n.b. I still need to settle on some better names (and placement/organisation of the methods) for these settings.

lib/hanami/configuration.rb
```diff
@@ -77,6 +114,14 @@ def logger
...
     end
 
+    def rack_logger_filter_params=(params)
+      settings[:rack_logger_filter_params] = params
+    end
+
+    def rack_logger_filter_params
+      settings[:rack_logger_filter_params]
+    end
+
```
timriley: I'm aware these (and the filtering behaviour in `RackLogger`) duplicate existing functionality in `Hanami::Logger` with its (seemingly not documented?) `filters:` option. However, for the sake of expedience, I'd like it if we could just accept these for now and then look at reconciling these different loggers post-merge.

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
+      def start_bootable(*args)
+        container.start(*args)
+      end
+
+      def keys
+        container.keys
+      end
+
+      def [](*args)
+        container.[](*args)
```
solnic: why not just `container[*args]`??

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
+      def resolve(*args)
+        container.resolve(*args)
+      end
+
+      def boot(&block)
+        return self if booted?
+
+        init
+
+        container.configure do; end # force after configure hook
```
solnic: I think we should introduce proper setup callbacks (like in rom) instead of relying on `configure`. I wanted to do this anyway so I wonder what your thoughts are.

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
+      end
+
+      def define_container
+        begin
+          require "#{application_name}/container"
+          application_module.const_get :Container
+        rescue LoadError, NameError
+          Class.new(Dry::System::Container).tap do |container|
+            application_module.const_set :Container, container
+          end
```
solnic: How about just doing `application_module.const_set(:Container, Class.new(Dry::System::Container))`??

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
           end
         end
       end
+
+      MODULE_DELIMITER = "::"
+      private_constant :MODULE_DELIMITER
+
+      def application_module
+        inflector.constantize(name.split(MODULE_DELIMITER)[0..-2].join(MODULE_DELIMITER))
+      end
```
solnic: I wonder if `application_namespace` would be a better name. `module` is less clear.

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
+
+      def application_name
+        inflector.underscore(application_module.to_s)
+      end
```
solnic: WDYT about coercing this to a symbol? I think using symbols for identifiers is more idiomatic. Also, `to_s` is not needed here.

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
+
+      def root
+        # TODO: do we need anything more sophisticated than this? This is how
+        # Dry::System::Container determines its root by default, anyway.
+        Dir.pwd
+      end
```
solnic: This must be configurable. There are cases where `Dir.pwd` won't work ie when you have an app or slice shipped by a gem.

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
+
+      def inflector
+        # TODO: might be good if we provided access as `config.inflector` too
+        config.inflections
+      end
```
solnic: This is actually pretty crucial to make it configurable right from the start due to custom inflections, like `HTTP` or `JSON` etc. This will be VERY common.

lib/hanami/application.rb
```diff
@@ -1,64 +1,282 @@
...
+        container.load_paths! "lib"
+
+        unless container.key?(:inflector)
+          container.register :inflector, inflector
+        end
+
+        unless container.key?(:logger)
+          require "hanami/logger"
+          container.register :logger, Hanami::Logger.new(configuration.logger)
+        end
```
solnic: Would checking `key?(:logger)` result in lazy-booting `:logger`? IIRC I had an issue with it, I had a logger bootable component yet the default logger would kick in regardless.

lib/hanami/cli/application/cli.rb
```diff
@@ -0,0 +1,32 @@
+require "hanami/cli"
+# require "hamami" # OK?
```
solnic: Some post-debug leftover 😄

lib/hanami/cli/application/commands/console.rb
```diff
@@ -0,0 +1,55 @@
...
+            IRB
+          end
+
+          desc "Open interactive console"
+
+          def call(**)
+            measure "#{prompt_prefix} booted in" do
+              out.puts "=> starting #{prompt_prefix} console"
+              application.boot
+            end
```
solnic: At work, I built an app console (also using Hanami::CLI) that does not boot the container and I simply rely on lazy-booting of everything. It starts under 1s and I didn't have any issues with it. I'd encourage you to try the same approach because people will love the fast booting aspect of it.

lib/hanami/components.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require "dry/system"
+require "pathname"
+
+# FIXME: dry-system wants this to be a Pathname but it really shoudn't need to be
+Dry::System.register_provider :hamami, boot_path: Pathname(File.join(__dir__, "components"))
```
solnic:
```suggestion
Dry::System.register_provider :hanami, boot_path: Pathname(File.join(__dir__, "components"))
```

lib/hanami/application.rb
timriley: Hah. Why not indeed! I'll change it.

lib/hanami/application.rb
timriley: Agreed, having to do this clearly indicates that some real lifecycle hooks are missing.

lib/hanami/application.rb
timriley: Yep, much more concise!

lib/hanami/application.rb
timriley: I like it! Plus it is more consistent with the `namespace` that we're passing to `Slice` objects

lib/hanami/application.rb
timriley: `:good_idea_sir`

lib/hanami/application.rb
timriley: Good points you raise. I'll add this to the list of TODOs in my PR description to make sure it doesn't get lost.

lib/hanami/application.rb
timriley: @solnic Are you suggesting we need to change the approach here? Right now it already _is_ configurable via that `config.inflections` setting (though I'd like to add an `inflector` alias since it feels like a more natural name).

lib/hanami/application.rb
timriley: @solnic Hmmm... if dry-system isn't already lazy-loading a component when checked by `key?`, then it _should_ be. I'll a TODO to make sure we follow up on this.

lib/hanami/cli/application/commands/console.rb
timriley: Oh yes! Thank you for the reminder about this. I've been meaning to do it with my work apps, and in the meantime I'll just go ahead and make the change directly here.

timriley: @solnic crazy idea: I wonder if we could rely on zeitwerk additionally in the console (and _only_ in the console, mind you) to auto-load any missing constants. That way you have the benefit of lazy loading for resolving container registrations _and_ any direct class lookups...

lib/hanami/application.rb
solnic: @timriley oh sorry I got confused, yeah my point was to make sure this is easily configurable, so I think we're good.

lib/hanami/cli/application/commands/console.rb
solnic: @timriley hmm yeah, this could be useful for global constants like struct classes or schemas, so I guess it's at least worth a shot. I don't think it will improve performance in any significant way though. The only benefit is going to be that you won't have to manually require some files, which is definitely good UX in a console.

lib/hanami/console/plugins/slice_readers.rb
```diff
@@ -0,0 +1,33 @@
...
+      class SliceReaders < Module
+        def initialize(ctx)
+          ctx.application.slices.each do |slice|
+            define_method(slice.name) do
+              SliceDelegator.new(slice)
+            end
+          end
+        end
+
+        class SliceDelegator < SimpleDelegator
```
jodosha: @timriley I'm not sure if this is valid for this case, but for proxy objects it's often advisable to use `BasicObject` as superclass to reduce to the minimum the clash between methods available to `Object` subclasses vs `BasicObject`.

If you think it's the case, please consider to use `Hanami::Utils::BasicObject`, which has some few extra features that are worth to have. On top of all: `#inspect` and `"pp"` compat. Given we're talking about an interactive console, those features are really valuable.

lib/hanami/setup.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "bundler/setup"
+require "hanami"
+
+begin
+  root = Dir.pwd
```
jodosha: As @solnic pointed above, is there the need here as well to make this configurable?

lib/hanami/cli/application/commands/console.rb
timriley: @jodosha I'll leave it like it is for now, and we can have a future discussion about benefits of the approach @solnic mentioned. This will mean the difference of the console being ready in 1 second versus 5+ seconds for larger applications, which is a big deal IMO.

lib/hanami/components/.keep
```diff
@@ -0,0 +1 @@
+# TODO: put actual components here
```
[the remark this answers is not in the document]
timriley: @jodosha This was me planning ahead for Hanami to vend its own dry-system bootable components, like these: https://github.com/icelab/snowpack/blob/master/lib/snowpack/components/persistence.rb

In this new dry-system-powered world, this is how we'll want Hanami to provide configurable bootable components for various things (managing the rom-rb boot definitely comes to mind, I'm not sure what else there might be for now).

What I might do for now, though, is just remove this (it was a hangover from when I copied over the Snowpack structures) and we can reintroduce it later when we need it.

lib/hanami/console/plugins/slice_readers.rb
timriley: @jodosha Thanks for the heads up about `Hanami::Utils::BasicObject`, that sounds like a handy convenience.

However, in this case, I think we actually want the `SimpleDelegator`-provided behaviour, passing through all known method calls to the underlying slice object, just with that one added enhancement via the `#method_missing` implementation (though I'm sure there are a few other shortcuts we can add in addition to this, later on).

lib/hanami/setup.rb
timriley: @jodosha We can't make this configurable because at this point, we're yet to even have loaded the application class! All we can do at this point is make a _best guess_ as to where that file may reside, based on established conventions (i.e. "try load config/application.rb").

Like I mentioned in my original comment, this could be made more sophisticated by copying Bundler's `require "bundler/setup"` behaviour (where it tries recursing up directories until it finds the file that it wants). However, I think that should definitely be a after-this-PR thing.

I will make _one_ change to this file now, though, where it looks up the location of the routes file from the application config, once it's loaded that.

timriley: FWIW, the alternative to `require "hanami/setup"`, of course, is something like `require_relative "config/application"` (or whichever relative path is appropriate for the current file), but I think the `require "hanami/setup"` file is a much nicer looking thing, and a much easier to remember thing, if we can get the behaviour right :)

lib/hanami/console/context.rb
```diff
@@ -0,0 +1,33 @@
...
+
+module Hanami
+  module Console
+    class Context < SimpleDelegator
+      attr_reader :application
+
+      def self.new(*args)
+        ctx = super
+
+        # if ctx.rom
```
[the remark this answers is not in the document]
timriley: @jodosha 100%. As soon as we work out how rom/hanami-model boots inside this new application structure (which, tbh, we can start on as soon as this is merged), we can have a whole range of nice rom conveniences within the console :)

That said, your comment here is a reminder I should just pull out this commented code, for now :)

lib/hanami/application.rb
timriley: Noted here: https://github.com/dry-rb/dry-system/issues/129

lib/hanami/application.rb
timriley: I've made sure the configured root is used everywhere we need it now. See https://github.com/hanami/hanami/pull/1019/commits/f05fafdbab2925aa0a511b6fdee6fb0b65d82ecd & https://github.com/hanami/hanami/pull/1019/commits/ab6c0b3e64ed3576299f08379b373f0dd0174e70.

I've not missed anything, have I, @jodosha?

lib/hanami/configuration.rb
timriley: Yep, this is noted 👍

lib/hanami/cli/base_command.rb
```diff
@@ -0,0 +1,51 @@
...
+    class BaseCommand < Hanami::CLI::Command
+      attr_reader :out
+      attr_reader :inflector
+      attr_reader :files
+
+      # TODO: consider injecting `out, files` from level of Snowpack::CLI?
+      def initialize(
+        command_name:,
+        out: $stdout,
+        inflector: Dry::Inflector.new,
```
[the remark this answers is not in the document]
timriley: @jodosha Yep, absolutely we should be passing the same inflector as configured for the app.

What you're seeing here is just the default param for initialising command classes _before_ they've been provided the app object.

In actual usage, this won't be used, because with the way I've set up the CLI, the `#with_application` method is called to "prepare" each already-initialized command to be used with the application. And I've just pushed https://github.com/hanami/hanami/pull/1019/commits/b075fddf25cf5a075fb17fa65c3cb2efffeec9f8 which ensures the application's inflector is passed through at that point too.

timriley:
Folks, thank you all very much for your thoughts on this and also for your patience as I fleshed out the overall structure these last few weeks.

Over the last few days I've added a range of integration specs (fast ones too, since they don't need to run in full isolation) covering the major new features that we've added with this application/slices structure, so I'm happy to merge this into `unstable` now.

From here, we can look at fully polishing this new arrangement in a tighter, more iterative fashion (with some ideas already in that list of post-merge to-dos I included in my original PR description).

Let's do this! 🎉

lib/hanami/application/router.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      alias call_router call
+      private :call_router # rubocop:disable Style/AccessModifierDeclarations
+
+      def call(env)
+        # FIXME: The to_a is here to ensure we return a proper rack-conformant
+        # 3-element array (which our rack_monitor middleware needs), because
+        # sometimes we get an Hanami::Action::Response object instead. This
+        # should not really be necessary.
+        @rack_app.call(env).to_a
```
davydovanton: WDYT if we change it to `Array(@rack_app.call(env))`?

lib/hanami/application/router.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+
+      private
+
+      def build_rack_app
+        middlewares = @middlewares
```
davydovanton: The idea for the future: can we make specific order for middlewares? It will be useful when we need to setup auth before all but after `request id` middleware or something like this. WDYT about it?

PR #1021, opened 2019-12-09
hanami/cli -> dry/cli for master branch

IvanShamatov:
This is the way how to switch to the usage of dry-cli (former hanami-cli) gem

Sqlite3 for different rubies is different, so it behaves differently too, but I finally made specs feel happy :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ which provides the glue that ties all the parts together:
 * [**Hanami::Helpers**](https://github.com/hanami/helpers) - View helpers for Ruby applications
 * [**Hanami::Mailer**](https://github.com/hanami/mailer) - Mail for Ruby applications
 * [**Hanami::Assets**](https://github.com/hanami/assets) - Assets management for Ruby
-* [**Hanami::CLI**](https://github.com/hanami/cli) - Ruby command line interface
 * [**Hanami::Utils**](https://github.com/hanami/utils) - Ruby core extensions and class utilities
 
 These components are designed to be used independently or together in a Hanami application.
--- a/spec/integration/cli/db/apply_spec.rb
+++ b/spec/integration/cli/db/apply_spec.rb
@@ -1,6 +1,6 @@
 RSpec.describe "hanami db", type: :integration do
   describe "apply" do
-    it "migrates, dumps structure, deletes migrations" do
+    it "migrates, dumps structure, deletes migrations", if: RUBY_VERSION < '2.4' do
       with_project do
         versions = generate_migrations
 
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -18,8 +18,8 @@ which provides the glue that ties all the parts together:
 * [**Hanami::Helpers**](https://github.com/hanami/helpers) - View helpers for Ruby applications
 * [**Hanami::Mailer**](https://github.com/hanami/mailer) - Mail for Ruby applications
 * [**Hanami::Assets**](https://github.com/hanami/assets) - Assets management for Ruby
-* [**Hanami::CLI**](https://github.com/hanami/cli) - Ruby command line interface
 * [**Hanami::Utils**](https://github.com/hanami/utils) - Ruby core extensions and class utilities
+* [**Dry::CLI**](https://github.com/dry-rb/dry-cli) - Ruby command line interface
```
jodosha: This isn't part of the Hanami family, it can be removed from `README`

spec/integration/cli/new/hanami_head_spec.rb
```diff
@@ -16,9 +16,9 @@
         expect('Gemfile').to have_file_content(%r{gem 'hanami-view',        require: false, git: 'https://github.com/hanami/view.git',        branch: 'develop'})
         expect('Gemfile').to have_file_content(%r{gem 'hanami-helpers',     require: false, git: 'https://github.com/hanami/helpers.git',     branch: 'develop'})
         expect('Gemfile').to have_file_content(%r{gem 'hanami-mailer',      require: false, git: 'https://github.com/hanami/mailer.git',      branch: 'develop'})
-        expect('Gemfile').to have_file_content(%r{gem 'hanami-cli',         require: false, git: 'https://github.com/hanami/cli.git',         branch: 'develop'})
         expect('Gemfile').to have_file_content(%r{gem 'hanami-assets',      require: false, git: 'https://github.com/hanami/assets.git',      branch: 'develop'})
         expect('Gemfile').to have_file_content(%r{gem 'hanami-model',       require: false, git: 'https://github.com/hanami/model.git',       branch: 'develop'})
+        expect('Gemfile').to have_file_content(%r{gem 'dry-cli',            require: false, git: 'https://github.com/dry-rb/dry-cli.git',     branch: 'master'})
```
jodosha: Because since now on we'll depend on `dry-cli` versions from Rubygems, there is no need to add it to the `Gemfile`.

jodosha: It looks like you removed `hanami-cli` from `lib/hanami/cli/commands/new/Gemfile.erb`, but did not add `dry-cli`, which is **correct**. I believe that this spec will fail.

IvanShamatov: yep, I've already passed that moment. Thanks. I have another error at another place now :)

spec/integration/cli/db/apply_spec.rb
```diff
@@ -24,6 +24,31 @@
...
+        expect(out).to include(versions.last.to_s)
+
+        db         = Pathname.new('db')
+        schema     = db.join('schema.sql').to_s
+        migrations = db.join('migrations')
+
+        expect(schema).to have_file_content <<-SQL
+CREATE TABLE `schema_migrations` (`filename` varchar(255) NOT NULL PRIMARY KEY);
+CREATE TABLE `users` (`id` integer NOT NULL PRIMARY KEY AUTOINCREMENT, `name` varchar(255), `age` integer);
+CREATE TABLE sqlite_sequence(name,seq);
```
AlfonsoUceda: Why do we need this?

AlfonsoUceda: Maybe if we need this line it is better to include in the other spec instead of duplicating it. What do you think?

IvanShamatov: At first, I checked inclusion of first and second line, then last 2 lines, and found out that earlier ruby with it's sqlite3 wraps table names with `"`. 

Actually, this spec is not about checking exactly this line, but about the execution of `hanami db apply`. So at the end of the day I decided to check a full `got` schema with full `expected` one in extra spec. 

Unfortunately, it leads to almost full duplication of spec implementation. Though it is a valid spec to have, from my point of view, because rspec will ignore another one in runtime and spec can be easily removed in future, when ruby 2.3.x support stopped.

@AlfonsoUceda WDYT?

PR #1023, opened 2019-12-21
hanami/cli -> dry/cli for main branch

IvanShamatov:
This is the way how to switch to the usage of dry-cli (former hanami-cli) gem.


adam12:
This is fairly stale and seems like it might no longer be relevant. Any thoughts @IvanShamatov @davydovanton ?

solnic:
Closing as this is stale and outdated indeed. We're gonna use hanami/cli after all.

PR #1029, opened 2020-02-13
Add application settings

timriley:
_This PR was built in collaboration with @andrewcroome — thank you Andrew!_

Add "application settings" as a first-class concept within Hanami 2.0 applications.

Settings are defined using their own block DSL, which by convention will live in `config/settings.rb` and will look like this:

```ruby
Hanami.application.settings do
  setting :base_url
  setting :sessions_secret
end
```

Setting definitions also support a second argument, which should be a callable "type" object used to determine the validity of the setting value. This means it will work very nicely with dry-types:

`lib/test_app/types.rb`:
```ruby
require "dry/types"

module TestApp
  module Types
    include Dry.Types()
  end
end
```

`config/settings.rb`:

```ruby
Hanami.application.settings do
  setting :base_url, TestApp::Types::String
  setting :sessions_secret, TestApp::Types::String
  setting :some_feature_flag, TestApp::Types::Params::Bool.optional.default(true)
end
```

If any of the setting values do not match their type expectations, the settings loading will fail and return a helpful error message containing the details of all the failed settings, e.g.

```
Hanami::Application::Settings::Loader::InvalidSettingsError:
  Could not initialize settings. The following settings were invalid:

  numeric_setting: invalid value for Integer(): "hello"
  feature_flag: maybe cannot be coerced to false
```

Settings are loaded by default when the Hanami application inits.

However, they will _also_ be loaded lazily, on first access, if accessed _before_ the application inits. This makes it possible to use those settings when configuring the Hanami app itself, e.g.

```ruby
module TestApp
    config.sessions = :cookie, {secret: settings.sessions_secret}
  end
end
```

As this also demonstrates, loading the settings creates a struct object with methods for each of the defined settings.

This object is also available as `Hanami.application.settings`, and also registered in the application container as `:settings`, which makes it possible to auto-inject the settings as a dependency of other objects in the application.

The settings loading is also flexible, with the following being configurable:

- `settings_path` - defaults to "config/settings")
- `settings_loader` - the class to use to load the settings, defaults to `Hanami::Application::Settings::Loader`, which loads dotenv (if available), fetches the settings from `ENV`, and then provides the type checking behaviour
- `settings_loader_options` - arguments to pass to the `settings_loader` when it is initialized

These configs may not be used much, but they do provide a valuable escape hatch in the case of unusual settings loading requirements.

### DIFF
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -18,7 +18,7 @@ module Hanami
     @_mutex = Mutex.new
 
     class << self
-      def inherited(klass) # rubocop:disable Metrics/MethodLength
+      def inherited(klass)
         @_mutex.synchronize do
           klass.class_eval do
             @_mutex         = Mutex.new
@@ -29,11 +29,6 @@ module Hanami
           end
 
           Hanami.application = klass
-
-          # Prepare base load path and settings as early as possible, so
-          # settings can be applied to Hanami.application.configuration
-          klass.send :prepare_base_load_path
-          klass.send :load_settings
         end
       end
     end
@@ -48,10 +43,12 @@ module Hanami
 
       alias config configuration
 
-      def init
+      def init # rubocop:disable Metrics/MethodLength
         return self if inited?
 
         configuration.finalize
+
+        load_settings
 
         @container = prepare_container
         @deps_module = prepare_deps_module
@@ -144,16 +141,19 @@ module Hanami
       end
 
       def settings(&block) # rubocop:disable Metrics/MethodLength
-        @_mutex.synchronize do
-          if block.nil?
-            defined?(@_settings) and @_settings
-          else
-            @_settings = Application::Settings.build(
-              configuration.settings_loader,
-              configuration.settings_loader_options,
-              &block
-            )
-          end
+        if block
+          @_settings = Application::Settings.build(
+            configuration.settings_loader,
+            configuration.settings_loader_options,
+            &block
+          )
+        elsif instance_variable_defined?(:@_settings)
+          @_settings
+        else
+          # Load settings lazily so they can be used to configure the
+          # Hanami::Application subclass (before the application has inited)
+          load_settings
+          @_settings ||= nil
         end
       end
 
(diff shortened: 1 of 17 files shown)


lib/hanami/application.rb
```diff
@@ -137,6 +140,23 @@ def booted?
...
+          )
+        elsif instance_variable_defined?(:@_settings)
+          @_settings
+        else
+          # Load settings lazily so they can be used to configure the
+          # Hanami::Application subclass (before the application has inited)
+          load_settings
+          @_settings ||= nil
+        end
+      end
```
timriley: FWIW I couldn't follow the `@_mutex.synchronize do ... end` dance that is followed in e.g. `.routes` inside this same class, because in the `else` case above, running `load_settings` will most likely result in the `.settings` method being called _again_ when `config/settings.rb` is loaded. This results in a deadlock if the code in this method is wrapped in the synchronize block.

The particular behaviour in that else case is important for the lazy loading of settings upon first access.

And look, while I'm here, I'll be honest: I don't _really_ understand why we're doing the synchronizing of things like the `@_routes` assignment inside this class... what is it actually in aid of? And if it's really necessary, how can we establish similar protections here?

timriley:
FYI: I plan to take another pass over this. After discussing with @solnic, there's a good chance we could use dry-configurable to provide a lot of what we've implemented here manually.

timriley:
So @jodosha has suggested we can merge this in _as-is_ and then I'll loop back to it once I've done a few view integration bits and bobs.

So this is ready for merging, pending a couple of things:

1. I'd like @jodosha (and @flash-gordon) to weigh in on a mutex-related question (I'll ping you both again separately on that one)
2. Some confirmation from @jodosha that it's OK to merge this given the build is red (or suggestions on how we handle the build) — I think this is due to the rewritten hanami-router having been merged into its own `unstable` branch... so it's not technically due to any issues related to this PR.

lib/hanami/application.rb
timriley: @jodosha Are you fine with this approach?

I wouldn't mind @flash-gordon's thoughts on this too, since he's said intelligent things about mutexes whenever I've asked him about them :)

timriley:
We've agreed this is good to merge now:

- The build is red due to the pending hanami-router updates, so it's unrelated to this work
- We'll create an issue to ensure we reconsider any thread-safety issues around the loading of the application settings

timriley:
n.b. I _will_ revisit this feature and adjust the settings DSL to use dry-configurable, as well as adjust the loading to match. This work has just taken a slightly lower priority while I switch onto view integration in shorter term.

PR #1030, opened 2020-02-15
Add ci workflow

solnic:
This adds a CI workflow that uses Github Actions to run all tests. It takes ~1.5 minute to run on MRI 2.5-2.7 + JRuby vs ~5 minutes on travis where it only runs MRI 2.5-2.6 and ruby-head.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,47 @@
+name: ci
+
+on:
+  push:
+    paths:
+      - ".github/workflows/ci.yml"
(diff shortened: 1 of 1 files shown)


cllns:
Very nice. Here's what the output looks like too, from the 'Actions' tab: https://github.com/solnic/hanami/actions/runs/39994258

solnic:
Right. I used my fork to make sure it'll run outside of the hanami org. The downside is that it didn't run *here* 😆 I'm sure it'll pass though.

PR #1032, opened 2020-03-06
Integrate `hanami-router` `v2.0.0.alpha2`

jodosha:
## What

Integrate `hanami-router` `v2.0.0.alpha2`.

### Resolver

The latest `hanami-router` (the **standalone gem**) expects Rack endpoints to be passed in the route definition. It doesn't long support the syntax `"home#index"` to define an action.

That gem now has the concept of **_resolver_** that can be passed to `Hanami::Router#initialize`. Third-party gems (like `hanami`) can pass their own initializer to resolve endpoints starting from a path (e.g. `"/admin/users"`) and an endpoint identifier (e.g. `"users#index"`).

This is the role of the new class: `Hanami::Application::Routing::Resolver`.

### Main slice

Hanami 2 will support multiple slices (formerly known as "apps") in the same app. When a developer will generate a new app with `hanami new` we will generate a default slice named **Main**.

The definition of the routes of each slice must be grouped by the `Hanami::Application::Router#slice` block. Example:

```ruby
# config/routes.rb
Hanami.application.routes do
  slice :main, at "/" do
    root to: "home#index"
  end

  slice :admin, at: "/admin" do
    get "/users", to: "users#index"
  end
end
```

### Application Router

We can't use directly `Hanami::Router`, but inherit from it, because we need more features for the full stack experience.

### Rack middleware

There are two ways to add a Rack middleware to the application stack:

  1. Add it to the application configuration. This will affect all the endpoints of the app.
  1. Add it to the router for fine-grained control.

#### Application configuration

```ruby
# config/application.rb
module Soundeck
    config.use MyGlobalRackMiddleware
  end
end
```

In this case `MyGlobalRackMiddleware` will be used by all the endpoints of the app.

This is important if third-party gems want to add their own Rack middleware to the application stack.

#### Router

```ruby
# config/routes.rb
Hanami.application.routes do
  use ElapsedTimeMiddleware

  get "/login", to: "sessions#new"

  scope "api" do
    use APIAuthentication

    get "/", to: "api#index"
  end
end
```

Rack Middleware in routes is **hierarchical**, similar to variable definitions. The outer the declaration, the broader the scope.

In this example, `ElapsedTimeMiddleware` is declared in the top-level scope, so it will be applied to all the routes. This is equivalent to `MyGlobalRackMiddleware` of the previous section.

`APIAuthentication` is declared in the `"api"` scope, so it will be used only by that narrowed section of the routes. For instance, it will be applied to `"api#index"`, but not to `"sessions#create"`


lib/hanami/application/router.rb
```diff
@@ -1,74 +1,49 @@
...
-            context: slice,
-            endpoint_resolver: @endpoint_resolver.with_container(slice),
-            &block
-          )
-
-          super(slice_router, at: at, host: host)
-        else
-          super(app, at: at, host: host)
-        end
+        remove_instance_variable(:@stack)
```
timriley: Would you mind explaining the reasoning for this? It's not immediately clear to me why we'd want to remove the stack when freezing...

lib/hanami/application/routing/resolver.rb
```diff
@@ -0,0 +1,74 @@
...
+          @inflector = inflector
+          @slices = slices
+        end
+
+        def with_container(new_container)
+          self.class.new(
+            container: new_container,
+            namespace: base_namespace,
+          )
+        end
```
timriley: Seems like this entire method isn't needed anymore? It existed previously because of the way we did the nesting of the router objects to achieve the old way of slice mounting. Can remove?

lib/hanami/application.rb
```diff
@@ -322,28 +322,26 @@ def initialize(application = self.class)
 
         application.boot
 
-        resolver = application.config.router_endpoint_resolver.new(
+        resolver = application.config.router_resolver.new(
```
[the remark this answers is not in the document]
jodosha: @timriley Agreed on `config.router.resolver`.

Regarding `use`: this is a public API, and devs are familiar with the concept of adding a Rack middleware to a middleware stack. The fact that we inject the middleware inside the router is an implementation detail. I believe it should be `config.middleware.use` instead.

lib/hanami/application/routing/middleware/stack.rb
```diff
@@ -0,0 +1,95 @@
...
+      # Hanami::Applicatione::Router middleware stack
+      #
+      # @since 2.0.0
+      # @api private
+      module Middleware
+        # Middleware stack
+        #
+        # @since 2.0.0
+        # @api private
+        class Stack
```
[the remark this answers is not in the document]
jodosha: If it should be shared at all. The implementation here and the one from `hanami-api` are already diverging. We shouldn't aim to share at all costs, but only if it makes sense.

lib/hanami/application/routing/resolver.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+module Hanami
+  class Application
+    module Routing
+      # Hanami application router endpoint resolver
+      #
+      # @since 2.0.0
+      class Resolver
+        require "hanami/application/routing/resolver/trie"
```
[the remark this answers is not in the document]
jodosha: Agreed. 👍

lib/hanami/application/routing/resolver.rb
```diff
@@ -0,0 +1,74 @@
...
+        def register_slice(path, name)
+          slices.add(path, name)
+        end
+
+        private
+
+        def resolve_string_identifier(path, identifier)
+          slice_name = slices.find(path) or raise "missing slice for #{path.inspect} (#{identifier.inspect})"
+          slice = container.slices[slice_name]
+          action_key = "actions.#{identifier.gsub(/[#\/]/, '.')}"
```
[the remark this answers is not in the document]
jodosha: I want to make this centralized. Only one class should be responsible to hold those separators and provide private API to convert.

---

Making those separators configurable will increase the internal complexity (passing full-blown config), force other actors to be aware of it (e.g. CLI code generators), and grow the effort to communicate another setting.

I believe we should make a decision, and code this decision into a central place, by removing those inline values and operations (e.g. `#gsub`).

What do you think?

lib/hanami/application/router.rb
jodosha: @timriley TL;DR: consider `Stack` as a builder for `Rack::Builder` that is not needed at the runtime.

`@stack` is only useful at "compile-time". It collects the Rack middleware usage from application configuration and `Router#use`, it applies the stack, but it isn't needed at the runtime.

When we invoke `Stack#finalize`, it returns the "real" middleware stack (a `Rack::Builder` instance). See https://github.com/hanami/hanami/pull/1032/files/d76a0cb0dd2b77dda5470e75adae3e837bda8e6f#diff-00f27089d3c7dbf5783b4a35405b603aR344

lib/hanami/application.rb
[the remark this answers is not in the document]
timriley: I'd be happy with `config.middleware.use` 👍

lib/hanami/application/routing/resolver.rb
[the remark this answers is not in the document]
timriley: Ohh, I think we might be misunderstanding each other here :)

What I was thinking should be configurable is the `"actions"` _prefix_ used when generating the identifier to get the action from the container.

For example, someone might want to make it `"web.actions"` (in fact, `"web.actions"` might be a better default) or even just `""` if they want to specify the absolute identifier in full within the route definitions.

So I wasn't talking about the _separators_ (i.e. the dots) but rather the _prefix_. I believe that could be pretty easily respected by generators and the like?

lib/hanami/application/router.rb
timriley: Ahh, I see, thanks! I'll leave another comment over there.

lib/hanami/application.rb
```diff
@@ -322,28 +322,26 @@ def initialize(application = self.class)
...
-          use application[:rack_monitor]
-
-          # Apply middleware from configuration
-          application.config.for_each_middleware do |m, *args, &block|
-            use(m, *args, &block)
-          end
-
-          run router
-        end
+        @app = stack.finalize(router)
```
timriley: Now that you've explained the reason for the stack ivar being removed when the router is frozen, this whole arrangement does feel a little bit awkward... i.e. the fact that we:

1. Create a `stack`
2. Create a `router`, passing the `stack` into it
3. And then finalize the `stack`, passing the `router` back to it!

So these two things feel very tightly coupled...

I understand the reason for the stack being passed into the router, so middleware can be used at various route namespaces, but I wonder if there's a way we could clean this relationship up some more? What do you think?

FWIW, I'd be happy just to flag this as something for us to follow up on later, but I felt it was worth mentioning here while it's fresh code.

jodosha: This is for perf reasons.

If `Stack` is empty, the returning value assigned to `@app` is the router itself. Otherwise, the returning value is a `Rack::Builder` instance.

Making all the requests to go to Rack middleware stack has a perf cost that we want to **avoid** if no middleware is set up by a dev.

jodosha: One way to decouple the finalization part would be:

```ruby
@app = if stack.empty?
  router
else
  stack.build
end
```

I've seen an opportunity for a "tell don't ask" approach with `Stack#finalize`. But we can move it to `Middleware.build(router, stack)` and that `if` will be handled by it:

```ruby
@app = Middleware.build(router, stack)
```

What do you think?

timriley: Yeah, that could help tease things apart a little.

I've another idea to throw into the mix, too: is there any reason our "application router" couldn't just create it's _own_ stack internally, when it's initialized? Then the `router` object is the only thing we deal with from within `Hanami::Application#initialize`, e.g.

```ruby
# This router would have @stack = Stack.new inside it's #initialize
router = Application::Router.new(
  resolver: resolver,
  **application.configuration.router_settings,
)

# These would go straight onto the router's own internal stack
router.use application[:rack_monitor]
application.config.for_each_middleware do |m, *args, &block|
  stack.use(m, *args, &block)
end

# Instead of calling instance_eval directly, we could create a method on
# Application::Router to make this clearer
router.instance_eval(&application.routes)

# Equivalent to stack.finalize
@app = router.to_rack_app
```

How about that?

lib/hanami/application/routing/resolver.rb
jodosha: Yes, removed by cc3b063

lib/hanami/application.rb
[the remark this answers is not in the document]
jodosha: `config.router.resolver` 158a543

Regarding `config.middleware.use` it was already in place. 😉

lib/hanami/application/routing/resolver.rb
```diff
@@ -0,0 +1,74 @@
...
+
+          unless endpoint.respond_to?(:call) # rubocop:disable Style/IfUnlessModifier
+            raise NotCallableEndpointError.new(endpoint)
+          end
+
+          endpoint
+        end
+        # rubocop:enable Metrics/MethodLength
+
+        def register_slice(path, name)
```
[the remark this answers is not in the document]
jodosha: Renamed in e515075

lib/hanami/application/routing/resolver.rb
```diff
@@ -0,0 +1,74 @@
...
+          slices.add(path, name)
+        end
+
+        private
+
+        def resolve_string_identifier(path, identifier)
+          slice_name = slices.find(path) or raise "missing slice for #{path.inspect} (#{identifier.inspect})"
+          slice = container.slices[slice_name]
```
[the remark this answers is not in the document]
jodosha: @timriley please have a look at b6a6e6a

I changed `#initalize` signature to accept the slices `application.slices`, instead of the full application. In this way we're just passing what is strictly needed (the slices).

lib/hanami/application/routing/resolver.rb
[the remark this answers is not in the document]
jodosha: @timriley OK, now I got what you mean.

I couldn't find in code where actions are registered. If possible, I want to make `actions.` prefix configurable and let the registration phase and the resolving phase to use this configuration.

There is no occurrence in code of `actions.` because they are auto-registered by `dry-container`? If so, we can't really make `actions.` configurable, right?

lib/hanami/application.rb
jodosha: @timriley please have a look at 4204cf6

lib/hanami/application.rb
```diff
@@ -322,28 +322,25 @@ def initialize(application = self.class)
...
-          &application.routes
-        )
-
-        @app = Rack::Builder.new do
-          use application[:rack_monitor]
+          routes: application.routes,
+          resolver: resolver,
+          **application.configuration.router.options,
+        ) do
+          # FIXME: enabling this line raises an exception that I cannot reproduce in dry-monitor
```
jodosha, line 335:
```
Failures:

  1) Hanami web app Routing to actions based on their container identifiers
     Failure/Error:
                   Rack::Builder.new do
                     s.each do |prefix, stack|
                       s.mapped(self, prefix) do
                         stack.each do |middleware, args, blk|
                           use(middleware, *args, &blk)
                         end
                       end

                       run app
                     end

     ArgumentError:
       wrong number of arguments (given 2, expected 1)
     # /Users/luca/.gem/ruby/2.7.0/gems/dry-monitor-0.3.1/lib/dry/monitor/rack/middleware.rb:25:in `new'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.2.2/lib/rack/builder.rb:158:in `block in use'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.2.2/lib/rack/builder.rb:235:in `block in to_app'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.2.2/lib/rack/builder.rb:235:in `each'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.2.2/lib/rack/builder.rb:235:in `inject'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.2.2/lib/rack/builder.rb:235:in `to_app'
     # ./lib/hanami/application/routing/middleware/stack.rb:61:in `to_rack_app'
     # ./lib/hanami/application/router.rb:53:in `to_rack_app'
     # ./lib/hanami/application.rb:343:in `initialize'
     # ./lib/hanami.rb:68:in `new'
     # ./lib/hanami.rb:68:in `boot'
     # ./lib/hanami/boot.rb:5:in `<top (required)>'
     # ./spec/new_integration/web_app_spec.rb:68:in `require'
     # ./spec/new_integration/web_app_spec.rb:68:in `block (3 levels) in <top (required)>'
     # ./vendor/cache/devtools-8ca13ea9f946/lib/hanami/devtools/integration/with_tmp_directory.rb:22:in `block in with_tmp_directory'
     # ./vendor/cache/devtools-8ca13ea9f946/lib/hanami/devtools/integration/with_directory.rb:20:in `with_directory'
     # ./vendor/cache/devtools-8ca13ea9f946/lib/hanami/devtools/integration/with_tmp_directory.rb:21:in `with_tmp_directory'
     # ./spec/new_integration/web_app_spec.rb:11:in `block (2 levels) in <top (required)>'

Finished in 1.81 seconds (files took 1.11 seconds to load)
49 examples, 1 failure

Failed examples:

rspec ./spec/new_integration/web_app_spec.rb:10 # Hanami web app Routing to actions based on their container identifiers

Randomized with seed 9178
```

jodosha:
@timriley 

> The approach using the trie structure looks super interesting to me. Does this mean it should be possible to use slice-specific routes at different paths, right?

That was unexpected. The assumption here is: a slice is mounted at a given **single** path. What is the need to mount it on several paths?

lib/hanami/application/routing/resolver.rb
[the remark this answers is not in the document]
timriley: @jodosha Actions will be auto-registered like most other components that go into the containers.

The reason to make the "actions base dir" a configurable thing is twofold:

1. When we have generators, the generators will respect that config and put the actions in that dir
2. And here in this resolver, it would use base dir to form the prefix of the container key, so the application authors don't need to repeat "actions.this" and "actions.that" and "actions.another_one" all the way through their routes file

Does that make sense?

I'm already doing similar with the views over in https://github.com/hanami/hanami/pull/1038, where I've added the `views.base_path` setting (which defaults to `"views"`). What we're talking about here is the equivalent config for actions.

lib/hanami/application.rb
timriley: @jodosha I LOVE IT!! 🥰 

That exception you comment about is kind of strange. Do you want me to take a look?

timriley:
@jodosha 

>> The approach using the trie structure looks super interesting to me. Does this mean it should be possible to use slice-specific routes at different paths, right?
>
> That was unexpected. The assumption here is: a slice is mounted at a given single path. What is the need to mount it on several paths?

Well, all the "mounting" is in this case is providing: (1) a base URL path, (2) a corresponding scope in which you can add extra middleware, if you wanted, and (3) the container from which to resolve the endpoint/action objects for each route within that base URL path.

There's nothing really in that set of things would make it _nonsensical_ to want to have a subset of routes come from another container.

```ruby
slice :admin, at: "/admin" do
```

lib/hanami/application/routing/resolver.rb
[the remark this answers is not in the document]
jodosha: @timriley So the registration key depends on the directory. OK now it's clear. I'll adjust the code accordingly.

lib/hanami/application.rb
jodosha: @timriley Yes, please. I've tried to reproduce the problem in `dry-monitor` master, but this test passes. Is it possible that `v0.3.1` the latest stable release diverges in behavior from master?

```diff
diff --git a/spec/integration/rack_middleware_spec.rb b/spec/integration/rack_middleware_spec.rb
index d32a0bd..4ca632a 100644
--- a/spec/integration/rack_middleware_spec.rb
+++ b/spec/integration/rack_middleware_spec.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'rack/builder'
+
 RSpec.describe Dry::Monitor::Rack::Middleware do
   subject(:middleware) { Dry::Monitor::Rack::Middleware.new(notifications).new(rack_app) }
 
@@ -40,6 +42,25 @@ RSpec.describe Dry::Monitor::Rack::Middleware do
     ].join('&')
   end
 
+  describe '#new' do
+    let(:builder) do
+      middleware = described_class.new(notifications)
+      app = rack_app
+
+      Rack::Builder.new do
+        use middleware
+        run app
+      end
+    end
+
+    it 'is compatible with Rack::Builder' do
+      expect(rack_app).to receive(:call).with(env).and_return([200, :total_success])
+
+      app = builder.to_app
+      app.call(env)
+    end
+  end
+
   describe '#call' do
     before do
       File.open(log_file_path, 'w').close
```

jodosha:
Mr @timriley & I decided to not support the [nested slices syntax](https://github.com/hanami/hanami/pull/1032#issuecomment-608327933) ☝️.

lib/hanami/application.rb
jodosha: Fixed by https://github.com/dry-rb/dry-monitor/pull/37

lib/hanami/application.rb
jodosha: https://github.com/dry-rb/dry-monitor/pull/37

lib/hanami/application/routing/resolver.rb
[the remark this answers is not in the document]
jodosha: @timriley I prefer to do this clean up when we'll have both the views and actions integrated.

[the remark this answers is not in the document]
timriley: No worries, I can help with that. I’ll be revisiting the whole configuration anyway when I port it to dry-configurable

PR #1040, opened 2020-04-07
Add application integration extensions to Hanami::View

timriley:
This PR adds:

1. Several new `Hanami.config.views.*` application configuration options
2. An extension to `Hanami::View so that its subclasses can nicely integrate with an Hanami application.

Specifically, it adds the `Hanami::View.[]` method to support usage like this:

```ruby
module Main
  class View < Hanami::View[:main]
  end

  module Views
    class Articles
      class Index < Main::View
      end
    end
  end
end
```

In this example, the `Main::View` serves as the "base" view class for the slice, and it will automatically be configured with:

- `paths` based off the slice's root dir, appended with the application's `config.views.templates_path` value (e.g. "slices/main/web/templates")
- `layouts_dir` using the application's `config.views.layouts_dir` (defaults to `"layouts"`)
- a `layout` using the application's `config.views.default_layout` (defaults to `"application"`)

The `Views::Articles::Index` view will inherit the configuration from the base `Main::View`, and also set a `template` config matching the class name, relative to the application's `config.views.base_path`. In this case, it would be configured as `"articles/index"`.

The net result of all this?

1. Boilerplate-free views
2. View classes that work in harmony with Hanami's dry-system-powered containers, in that they're "self-contained", i.e. they can be both auto-registered and lazily resolved and pull in everything they require from their class definition being evaluated.

In the process of building this (as Luca would attest to), I've yet to convince myself as to the best place to put this integration code: either here in hanami or in hanami-view itself. However, given there are a few more steps before we have _complete_ view integration, I'm going to defer any further anguishing about it until we have completed the overall picture and can consider things holistically :)

As such, this PR is ready for review/merge, because it represents a discrete aspect of the integration that we can then build upon in subsequent PRs, which will be:

1. Ensuring action classes are also container-auto-registerable, following the same overall approach as we've done here for views
2. Adding a helpful default view context for the application, along with an action helper method for rendering views using the view context object extended with request-specific data
3. Re-considering the best place to keep the view integration code
4. Building a range of view helpers to match hanami v1 as much as possible
5. Port the application-level config to use dry-configurable, and then make it so that `application.config.views` also imports _all_ configurable elements from `Hanami::View`, making any view setting possible to configure application-wide

@jodosha This is ready for review now.

I want to acknowledge that the `views` application configuration code is kind of messy, but I expect that will be all sorted out once we do step 5 above :)

I'll also be sure to squash everything when merging this, so you can largely ignore the meandering commit history here.


hanami.gemspec
```diff
@@ -22,15 +22,16 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
...
+  spec.add_dependency "dry-inflector",     "~> 0.1", ">= 0.1.2"
   spec.add_dependency "dry-monitor"
   spec.add_dependency "dry-system",        "~> 0.10"
-  spec.add_dependency "dry-inflector",     "~> 0.1", ">= 0.1.2"
-  spec.add_dependency "bundler",           ">= 1.16", "< 3"
+  spec.add_dependency "hanami-cli",        "~> 1.0.alpha"
+  spec.add_dependency "hanami-controller", "~> 2.0.alpha"
+  spec.add_dependency "hanami-router",     "~> 2.0.alpha"
+  spec.add_dependency "hanami-utils",      "~> 2.0.alpha"
+  spec.add_dependency "hanami-view",       "~> 2.0.alpha"
```
timriley, line 34: ^ I took the opportunity to alphabetize

lib/hanami/application.rb
```diff
@@ -174,12 +175,16 @@ def routes(&block)
       MODULE_DELIMITER = "::"
       private_constant :MODULE_DELIMITER
 
-      def application_namespace
+      def namespace
```
timriley, line 178: Renamed this (and added `namespace_path` below) to make it so that the application and slices can offer a consistent interface as much as possible. This is what enables the "hidden" feature of being able to create application-level views using `Hanami::View[Hanami.application]`.

This is _not_ something we will offer via generators or anything, but given that we were already 95% of the way there, I felt it was worth bringing these in line at the same time.

lib/hanami/slice.rb
```diff
@@ -14,7 +14,7 @@ def initialize(application, name:, namespace: nil, root: nil, container: nil)
       @application = application
       @name = name.to_sym
       @namespace = namespace
-      @root = root
+      @root = root ? Pathname(root) : root
```
timriley, line 17: I made it so that both the application and slices returned Pathnames for their roots. It's a useful thing to be able to rely on, particularly since we build on the root path so often.

(In this particular case, we do need to handle roots potentially being nil, since I designed slices so that – unlike applications – they don't necessarily _have_ to belong to a given directory)

PR #1043, opened 2020-04-14
Integrate view rendering through actions

timriley:


timriley:
Closing this since we're taking a different approach now in https://github.com/hanami/hanami/pull/1049

PR #1045, opened 2020-04-23
CLI application env handling and dotenv improvements

timriley:
We had a problem with our Hanami 2 app where the HANAMI_ENV, when passed by the `-e` flag to the CLI, was being set too late for some processing of application settings based on the env.

This PR defers the init of the app as late as possible, which means the correct env is in place for the entirety of the application boot (which fixes our issue and should also help with ensuring more predictable behavior overall).

Along with this, I've expanded the number of files loaded by Dotenv by default. This brings it inline with the behavior of dotenv-rails, which I think is a good model to follow. This change is bundled in here because it means that e.g. `.env.test` can finally be respected when using the `-e test` flag to CLI commands (we had to hack this into our app until I could fix it properly here).

### DIFF
--- a/lib/hanami/application/settings/loader.rb
+++ b/lib/hanami/application/settings/loader.rb
@@ -53,8 +53,17 @@ module Hanami
 
         def load_dotenv
           require "dotenv"
-          Dotenv.load if defined?(Dotenv)
+          Dotenv.load(*dotenv_files) if defined?(Dotenv)
         rescue LoadError # rubocop:disable Lint/SuppressedException
+        end
+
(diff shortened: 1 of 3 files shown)


timriley:
@jodosha Hopefully these should be fairly uncontroversial changes. I'm basically fixing up a bug that snuck in when we landed the massive new application structure PR, plus making dotenv _actually useful_ by default. I've also tested these with our production Hanami 2 app.

timriley:
@jodosha Are you happy for me to merge this?

PR #1047, opened 2020-04-24
Fix scope of used middleware in slice routes

timriley:
@jodosha Here's a failing spec demonstrating a couple of different problems with middleware being `use`d inside slice routes.

I've started poking around but haven't yet found a fix. I'd imagine you might be a lot quicker with this given you recently built all of this – would you mind having a look?

### DIFF
--- a/spec/new_integration/router/using_middleware_spec.rb
+++ b/spec/new_integration/router/using_middleware_spec.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+
+require "rack/test"
+
+RSpec.describe "Router with middleware", :application_integration do
+  include Rack::Test::Methods
+
+  let(:app) { Hanami.app }
+
+  specify "Middleware blah" do
+    with_tmp_directory(Dir.mktmpdir) do
+      write "config/application.rb", <<~RUBY
+        require "hanami"
+
+        module TestApp
+          class Application < Hanami::Application
+            config.logger = {stream: File.new("/dev/null", "w")}
+          end
+        end
+      RUBY
+
+      write "config/routes.rb", <<~RUBY
+        module TestApp
+          class Middleware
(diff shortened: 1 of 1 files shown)


spec/new_integration/router/using_middleware_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+              self.class.called_as[@tag] << env["PATH_INFO"]
+              @app.call(env)
+            end
+          end
+        end
+
+        Hanami.application.routes do
+          slice :main, at: "/" do
+            # With this here, it ends up applying to _all_ routes, including admin
+            use TestApp::Middleware, :main
```
timriley, line 44: One aspect of the problem: this shouldn't run on admin routes

spec/new_integration/router/using_middleware_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+            root to: "home#show"
+          end
+
+          slice :admin, at: "/admin" do
+            # With this here, all admin routes return 404
+            use TestApp::Middleware, :admin
```
timriley, line 51: And obviously this shouldn't make all the admin routes become 404s!

tiev: I encountered this recently. admin routes become nested once more.
* `/admin/admin` => admin's home#show
* `/admin/admin/test` => admin's test#show

And main routes are nested in too, causing
* `/admin` => main's home#show

I found it's because the `Hanami::Application::Routing::Middleware::Stack` creates a map for '/admin' but run the whole app under it.

solnic:
Closing this as it's outdated and I pushed a branch with the updated repro spec here https://github.com/hanami/hanami/tree/fix-slice-middlewares

PR #1048, opened 2020-04-27
Introduce hanami version CLI command

davydovanton:
Hey

I added `hanami version` CLI command and also fix specs. We used `run_command` but it conflicted with aruba helpers (you can find revert commit from @jodosha  here: https://github.com/hanami/devtools/commit/afc2341578408af0581e5003888bdc59f4d476db)


davydovanton:
also, we have problems with `with_project` helper that's why half of specs for version command are red:

Screenshot 2020-04-27 at 17 59 42

PR #1052, opened 2020-05-13
Remove direct view integration; replace with hook for views to integrate themselves

timriley:
Our first pass at having hanami views integrate with the application involved us having to ask users to inherit from Hanami::View in a slightly strange way:

```ruby
module Main
  class View < Hanami::View[:main] # <- this bit
  end
end
```

This worked well enough, but having to specify you want a view for the `:main` slice when you're already _inside_ the Main slice's own namespace felt a little redundant.

Further, the extra functionality for these "application views" was inserted by the hanami gem itself, by monkey-patching the `Hanami::View` class, which didn't feel clean either.

I've found a way to address both these issues:

- Move the application integration logic into hanami-view itself, and finding a way to cleanly organise both application and standalone views
- Add an `Hanami::Application.component_provider` method, which hanami-view uses to determine if a given class (`Main::View` in the example above) is being provided by the application and/or one of its slices, and if it is, it can use the details of that "provider" (the app app and/or slice) to configure the view to fit within the application.
- Add an `Hanami.application?` method to make it clear to other gems whether an application _really_ exists before they try and integration

This PR introduces the Hanami side of those changes:

- Removing the existing view integration logic
- Adding `Hanami::Application.component_provider`
- Adding `Hanami.application?`

The hanami-view side of the changes is in https://github.com/hanami/view/pull/173.


timriley:
@jodosha Ready for review!

hanami.gemspec
```diff
@@ -31,7 +31,6 @@ Gem::Specification.new do |spec| # rubocop:disable Metrics/BlockLength
   spec.add_dependency "hanami-controller", "~> 2.0.alpha"
   spec.add_dependency "hanami-router",     "~> 2.0.alpha"
   spec.add_dependency "hanami-utils",      "~> 2.0.alpha"
-  spec.add_dependency "hanami-view",       "~> 2.0.alpha"
```
jodosha, line 34: @timriley 🤔 ???

spec/unit/hanami/application/component_provider_spec.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require "hanami/application"
+
+RSpec.describe Hanami::Application, "#component_provider", :application_integration do
```
jodosha, line 5: @timriley shouldn't this test go in `spec/isolation`? There the specs are run one by one by a separated Ruby process (`bundle exec rspec spec/isolation/foo_spec.rb`).

Because we're initializing Hanami global state (e.g. `Hanami.application`), it would make the test suite brittle tests like this are sharing the same Ruby process, like unit tests do.

hanami.gemspec
timriley: Yep, we don't need it as a hard dependency anymore (n.b. this line wasn't here either _before_ I added that first pass of view integration).

_At this point,_ I think the "Application view" code does sit better inside hanami-view, because it makes it a lot easier to maintain both the standalone view/application view behaviour in the same place (especially since I had to do all that module trickery to make it happen).

spec/unit/hanami/application/component_provider_spec.rb
timriley: @jodosha The `:application_integration` metadata takes care of this, you can see how that works in `spec/support/application_integration.rb`. It's what I've used for all the specs in `spec/new_integration`, and it's working well there so far: specs run reliably and they're much, much faster than the full isolation tests.

If you want, I could move this file into that `spec/new_integration` dir? However, this actually felt like a proper unit test to me, because it's focused on testing that one specific method; it just requires the application setup to exist first.

What do you think?

hanami.gemspec
jodosha: It wasn't there because we didn't have the integration yet, as `hanami-view` 1.x wasn't compatible.

Removing the gem from here will make it optional, but we haven't planned yet for this scenario.

spec/unit/hanami/application/component_provider_spec.rb
jodosha: @timriley 
> The `:application_integration` metadata takes care of this, you can see how that works in spec/support/application_integration.rb.

My only concern here is: what if a unit test requires the state to not being initialized?

---

> However, this actually felt like a proper unit test to me

I agree, but because it depends on Hanami global state, IMO the better home for it is an isolation test. Isolation tests are like unit tests, but they need to be executed in their own Ruby process because they depend or modify the global state.

timriley: Can we compromise and just make this an “integration” test and move it alongside the others I’ve made? It seems silly to make this an isolation test and slow down our suite when we’ve already done everything that’s necessary for this test to run authentically, fast, and reliably.

hanami.gemspec
timriley: We’ve discussed this scenario previously, and at that point when the view integration code was living inside hanami, we _had_ to make it a gemspec dependency.

That’s not a hard requirement at this point, so I figured we should only bring in hard dependencies when we will _really_ need them. Can we just leave it like this and see how the remaining aspects of the action/view/framework integration play out before making a decision?

timriley: Ok, I’ll put it back for now. We can revisit once we see the full integration picture play out.

timriley: @jodosha This is back in place now.

spec/unit/hanami/application/component_provider_spec.rb
timriley: @jodosha I've moved the test into the `new_integration/` folder now along with the rest of the new-style integration specs.

(FWIW, my plans is that this will subsume/merge with the existing `integration/` folder before we're done with the 2.0 effort)

timriley: > My only concern here is: what if a unit test requires the state to not being initialized?

The before/after hooks enabled by using the `application_integration` example metadata do not actually do anything regarding application initialization. They simply set up some machinery to ensure that any state created by an application being created/booted over the course of the example is fully cleaned up again afterwards.

It doesn't boot an app by default or anything like that. In fact, it doesn't force you to write any particular style of test at all. It's just there to make sure you _can_ freely create/boot an app if that's what's necessary for your test to run.

Like I said, it works really nicely, and I've used those hooks to manage quite a range of tests now, so I don't think we need to resort to isolation testing at this point. In fact, my goal would be to reduce those to a minimum, since they run much more slowly and don't fit well with the rest of the test suite.

Does this allay your concerns?

timriley:
@jodosha Addressed your comments above. Let me know if you're happy for me to merge, thanks 😄

PR #1055, opened 2020-05-14
Fix Pry prompt deprecation in pry >= 0.13.0

adam12:
Passing an array of procs to `Pry.start` is deprecated as of 0.13.0. The Pry changelog says it will be removed on the next release tho it doesn't indicate at which release level.

No longer causes a warning by Pry.

### DIFF
--- a/lib/hanami/cli/application/commands/console.rb
+++ b/lib/hanami/cli/application/commands/console.rb
@@ -34,7 +34,31 @@ module Hanami
 
           def start_repl
             context = Hanami::Console::Context.new(application)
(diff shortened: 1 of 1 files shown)


timriley:
Thanks @adam12! I think this is a good fix for now. We might want to consider whether we have that pry version check in the future (i.e. maybe we should just require a minimum version of pry and be done with it) but I think this is good to merge now to avoid the warnings.

We'll certainly be putting more attention onto the CLI in future (and personally I want to really boost the console experience) so I'm confident we'll be able to make sure this is fully resolved before we hit 2.0.

PR #1061, opened 2020-06-20
Fix now-evaluation at Hanami::CommonLogger

wuarmin:
This commit at Rack (rack/rack@743f29d) changed the evaluation of `began_at` at `Rack::CommonLogger`. With that change Hanami's `CommonLogger` cannot calculate `elapsed` properly.

The result are logging-entries like that: 
```txt
[bookshelf] [INFO] [2020-06-20 19:11:23 +0200] HTTP/1.1 GET 200 127.0.0.1 / 14915 {} 2020-06-20 16:10:40 +0200
``` 
with this fix, `elapsed` is calculated properly
```txt
[bookshelf] [INFO] [2020-06-20 19:11:23 +0200] HTTP/1.1 GET 200 127.0.0.1 / 14915 {} 0.0457000
``` 
WDYT?
best regards

### DIFF
--- a/lib/hanami/common_logger.rb
+++ b/lib/hanami/common_logger.rb
@@ -55,7 +55,7 @@ module Hanami
     # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
     def log(env, status, header, began_at)
-      now    = Time.now
+      now    = Rack::Utils.clock_time
       length = extract_content_length(header)
 
       msg = Hash[
--- a/lib/hanami/common_logger.rb
+++ b/lib/hanami/common_logger.rb
@@ -55,7 +55,7 @@ module Hanami
     # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
     def log(env, status, header, began_at)
-      now    = Rack::Utils.clock_time
+      now    = ElapsedTime.call
       length = extract_content_length(header)
 
       msg = Hash[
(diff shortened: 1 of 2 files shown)


wuarmin:
@jodosha Thank you for your hints. I changed the implementation a bit, because of the fact, that `Rack::Utils` responds since Rack Version `2.0.0` to `clock_time`, but it was only used with version `2.1.0` for `began_at`.

wuarmin:
@jodosha this fix is very important for me. WDYT?
Thanks

wuarmin:
@jodosha this fix is very important for me. WDYT?
Thanks

jodosha (approved):
@wuarmin Thank you, this looks good now.

lib/hanami/common_logger.rb
```diff
@@ -89,5 +89,21 @@ def extract_params(env)
...
+    class ElapsedTime
+      @clock = if Gem::Version.new(Rack::RELEASE) >= Gem::Version.new('2.1.0')
+                 -> { Rack::Utils.clock_time }
+               else
+                 -> { Time.now }
+               end.freeze
+
+      def self.call
+        @clock.call
+      end
```
flash-gordon, line 106:
```ruby
if Gem::Version.new(Rack::RELEASE) >= Gem::Version.new('2.1.0')
  def self.call
    Rack::Utils.clock_time
  end
else
  def self.call
    Time.now
  end
end
```
what about?

jodosha: That could work as well. 👍

wenderjean:
Just to keep track, this PR handles the issue reported here in https://github.com/hanami/hanami/issues/1088

lib/hanami/common_logger.rb
wuarmin: Yes, that would work too :+1:  I would still stick to the first variant.

PR #1063, opened 2020-07-01
Fix isolation specs for recent "Application action" changes

timriley:
These broke when inheriting from Hanami::Action resulted in a call to Hanami.application.component_provider, which wants the slices to be registered already.

Explicitly registering the slices earlier and init’ing the app is the fix here, but I’d also like to explore making single file Hanami apps a little more ergonomic in the future.

FWIW, I still plan on converting all of the isolation specs into the "new style" integration tests, which can be run as part of a single test suite.

### DIFF
--- a/spec/isolation/hanami/application/routes/configured_spec.rb
+++ b/spec/isolation/hanami/application/routes/configured_spec.rb
@@ -8,6 +8,12 @@ module Bookshelf
 end
 
 module Web
+end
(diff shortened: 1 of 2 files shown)


timriley:
This is a minor test suite fix and is green apart from Travis (due to Rubocop not being present, which is a separate issue). I'm going to merge this in to allow us to continue unstable branch development. FYI @jodosha.

PR #1074, opened 2020-09-29
Pass base config to actions; finalize all configs

timriley:
This allows the actions configuration to adapt to various top-level configs, e.g. to enable CSRF protections automatically in the case of `config.sessions.enabled?` being true.

We handle this by passing the base application config to the nested actions config, and then call `finalize!` on the nested actions config when the base config is finalized.

This means the nested actions config can adapt its own configuration at the last moment, which ensures we support the user specifying their configs in whatever order they like prior to this.

This PR enables the automatic enabling of action CSRF protection in https://github.com/hanami/controller/pull/329

### DIFF
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -53,7 +53,7 @@ module Hanami
       @actions = begin
         require_path = "hanami/action/application_configuration"
         require require_path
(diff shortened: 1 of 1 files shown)


timriley:
Closing in favour of #1078

PR #1078, opened 2020-09-30
Finalize nested configs, move sessions config to hanami-controller

timriley:
This PR:

- Allows nested config objects to adjust their own configuration at the last moment, which the `actions` config will need to do in order to determine whether `actions.csrf_protection` should be set to true or false (relies on checking the value of another setting). This enables the automatic enabling of action CSRF protection in https://github.com/hanami/controller/pull/332
- Removes the `sessions` config, which as of https://github.com/hanami/controller/pull/332 will now be located in hanami-controller, closer to the code that actually uses it


timriley:
[Restore CSRF protection behavior](https://trello.com/c/nPpmQPRd/56-restore-csrf-protection-behavior)

jodosha (approved):

PR #1081, opened 2020-10-18
Use Hanami::View::ApplicationConfiguration provided by hanami-view

timriley:
Take out the custom view config handling we had previously, and instead rely on `Hanami::View::ApplicationConfiguration` as provided by the hanami-view gem (in https://github.com/hanami/view/pull/180). Expose this as `Application.config.views` as usual, which leaves the config handling for both actions _and_ views in a consistent arrangement.

This PR will be merged alongside https://github.com/hanami/view/pull/180.


timriley:
[Configure application views with application inflector](https://trello.com/c/jJqElBmm/77-configure-application-views-with-application-inflector)

jodosha (approved):

PR #1085, opened 2020-10-31
Remove config.security settings

timriley:
These have been replaced by a new default value for Action’s `config.default_headers` when an Action is used within a full Hanami app.

See https://github.com/hanami/controller/pull/336


timriley:
[Work out where/how we should be using the `cookies` and `security` application settings](https://trello.com/c/LcC64ra5/89-work-out-where-how-we-should-be-using-the-cookies-and-security-application-settings)

jodosha (approved):

PR #1086, opened 2020-10-31
Remove config.cookies settings

timriley:
These have been moved to `config.actions.cookies`. See https://github.com/hanami/controller/pull/337.


timriley:
[Work out where/how we should be using the `cookies` and `security` application settings](https://trello.com/c/LcC64ra5/89-work-out-where-how-we-should-be-using-the-cookies-and-security-application-settings)

jodosha (approved):
@timriley Thank you 🙏

PR #1089, opened 2021-01-04
Allow `hanami` to be started from non-root directory

ojab:
I've tried to use hanami app as a dummy for another gem testing & run it via `Rack::Builder.parse_file('./dummy/config.ru').first`, but it fails to find `config/environment` in that case. 
This change allows application to be started and actually serve some pages. More complex usage still fails due to `Hanami::Environment#root` assuming that we're running from app root, but at least it works for my usecase without models/mailers.

### DIFF
--- a/lib/hanami/cli/commands/new/config.ru.erb
+++ b/lib/hanami/cli/commands/new/config.ru.erb
@@ -1,3 +1,3 @@
-require './config/environment'
+require_relative 'config/environment'
 
 run Hanami.app


ojab:
Not sure if PR should be made against `master` or `develop`, failure on travis looks unrelated.

adam12:
The change itself seems fine, but the commit message is somewhat noisy. Are you familiar with how to tighten it up?

I think something simple such as `Load environment relative to config.ru` or similar would be sufficient.

ojab:
Oh, I was too lazy to do proper fork/clone/commit/push & done the edit via GH web interface, didn't expected that it'll put the whole comment into the commit message. Shortened/pushed.

adam12 (approved):

adam12:
Thanks for the patch!

Issue #1088, opened 2021-01-04
Hanammi::CommonLogger#log should use Process.clock_gettime(Process::CLOCK_MONOTONIC) instead of Time.now


heragu:
In the `Hanami::CommonLogger` the method `log` uses `Time.now` without considering if this value is coherent with the `began_at` given by rack.

```
module Hanami
  class CommonLogger < Rack::CommonLogger
    ....
    def log(env, status, header, began_at)
      now    = Time.now
      length = extract_content_length(header)

      msg = Hash[
        http:    env[HTTP_VERSION],
        verb:    env[REQUEST_METHOD],
        status:  status.to_s[0..3],
        ip:      env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR],
        path:    env[SCRIPT_NAME] + env[PATH_INFO].to_s,
        length:  length,
        params:  extract_params(env),
        elapsed: now - began_at
      ]
    ...
  end
end
```

`began_at` is set by `rack` gem using `Rack::Utils.clock_time` which is defined in the version 2.2.3 of the gem as follows:

```
if defined?(Process::CLOCK_MONOTONIC)
  def clock_time
    Process.clock_gettime(Process::CLOCK_MONOTONIC)
  end
else
  # :nocov:
  def clock_time
    Time.now.to_f
  end
  # :nocov:
end
```

In the most of the systems, the time picked will be `Process.clock_gettime(Process::CLOCK_MONOTONIC)` which is not necessarily `Time.now`. This causes that the elapsed time appears as a timestamp and not as the time passed:

```
{"app":"my_app","severity":"INFO","time":"2020-09-25T14:30:00Z","http":null,"verb":"GET","status":"200","ip":"127.0.0.1","path":"/api/my_path","length":"710","params":{},"elapsed":"2020-09-20 13:42:56 +0200"}
```
```

wenderjean:
@heragu I'll take a look at this issue.

jodosha:
Fixed by #1061

PR #1091, opened 2021-01-31
Use monotonic time in common_logger

wenderjean:
Relates to: https://github.com/hanami/hanami/issues/1088

### Changes

- Use monotonic time from `concurrent-ruby` to keep consistency with Rack gem
- Fix the milliseconds/seconds in the logline

### Rational

In order to keep consistency between the Rack timestamp and the Hanami one, and based on the point suggested in the same issue #1088 I decided to take a look at the Rack implementation of the elapsed time and why we were using `Time.now` instead of the monotonic version. So, I couldn't find the reasons behind that decision but It seems to make sense to me to use a monotonic time to compare with that one coming from Rack (considering it follows that time counter).

### References

The `rack` implementation:
https://github.com/rack/rack/blob/03b4b9708f375db46ee214b219f709d08ed6eeb0/lib/rack/utils.rb#L82

The `concurrent-ruby` implemetation:
https://github.com/ruby-concurrency/concurrent-ruby/blob/a5076bf9e977b290cce5a65b7b72810562bca295/lib/concurrent-ruby/concurrent/utility/monotonic_time.rb#L11

### Update

The changes this PR contains are also available at [this PR](https://github.com/hanami/hanami/pull/1061) opened 8 months ago, there are slight differences among them but feel free to close it if you guys prefer.

### DIFF
--- a/lib/hanami/common_logger.rb
+++ b/lib/hanami/common_logger.rb
@@ -55,7 +55,7 @@ module Hanami
     # rubocop:disable Metrics/AbcSize
     # rubocop:disable Metrics/MethodLength
     def log(env, status, header, began_at)
-      now    = Time.now
+      now    = Concurrent.monotonic_time
       length = extract_content_length(header)
 
       msg = Hash[
@@ -66,7 +66,7 @@ module Hanami
         path:    env[SCRIPT_NAME] + env[PATH_INFO].to_s,
         length:  length,
         params:  extract_params(env),
-        elapsed: now - began_at
+        elapsed: '%.4f' % (now - began_at)
       ]
 
       logger = @logger || env[RACK_ERRORS]
--- a/spec/unit/hanami/common_logger_spec.rb
+++ b/spec/unit/hanami/common_logger_spec.rb
@@ -26,5 +26,36 @@ RSpec.describe Hanami::CommonLogger do
         expect(device.read).to include(%(:path=>"logo.png"))
       end
     end
+
+    context "when ELAPSED time is present" do
+      it "returns the elapsed time" do
+        freeze_clock_time_at(Time.at(0), Time.at(1)) do
+          get "/"
+        end
+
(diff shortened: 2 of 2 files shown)


lib/hanami/common_logger.rb
```diff
@@ -66,7 +66,7 @@ def log(env, status, header, began_at)
         path:    env[SCRIPT_NAME] + env[PATH_INFO].to_s,
         length:  length,
         params:  extract_params(env),
-        elapsed: now - began_at
+        elapsed: '%.4f' % (now - began_at)
```
wenderjean, line 69: I may be wrong but the documentation in [this page](https://guides.hanamirb.org/projects/logging/) suggests that the logline shows the milliseconds/seconds at the end of it but performing some tests with a fresh new application I noticed that what we actually had was the following structure:

`[APP] [INFO] [2021-01-31 11:17:40 -0300] HTTP/1.1 GET 200 ::1 /users 175 {} 2021-01-27 11:39:43 -0300`
`We actualy had a DateTime information.`

Using the monotonic time I ended with a huge decimal value representing the elapsed time in milliseconds, that's the reason I decided to format, capping it to four decimals only. I supposed it is enough to make decisions based on the response time or even grab it and plot it in a dashboard but let me know any other different opinions, please.

`[APP] [INFO] [2021-01-31 11:17:40 -0300] HTTP/1.1 GET 200 ::1 /users 175 {} 0.0018`
`[APP] [INFO] [2021-01-31 11:17:40 -0300] HTTP/1.1 GET 200 ::1 /users 175 {} 0.0218`

jodosha:
@wenderjean Thanks for this PR, but I believe that https://github.com/hanami/hanami/pull/1091 is handling the fix better because it doesn't require the dep with `concurrent` gem and because it handles well pre Rack 2.1 cases.

PR #1093, opened 2021-02-09
Configure new dry-system component_dirs

timriley:
This PR updates the dry-system container setup to use the new `component_dirs` setting introduced in https://github.com/dry-rb/dry-system/pull/155 and https://github.com/dry-rb/dry-system/pull/157

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -99,17 +99,20 @@ module Hanami
       container = Class.new(Dry::System::Container)
       container.use :env
 
-      container.config.name = name
-      container.config.inflector = application.configuration.inflector
+      container.configure do |config|
+        config.name = name
+        config.inflector = application.configuration.inflector
 
-      if root && File.directory?(root)
-        container.config.root = root
-        container.config.bootable_dirs = ["config/boot"]
+        if root&.directory?
+          config.root = root
+          config.bootable_dirs = ["config/boot"]
 
-        container.config.auto_register = ["lib/#{namespace_path}"]
-        container.config.default_namespace = namespace_path.tr("/", ".")
-
-        container.load_paths! "lib"
+          if root.join("lib").directory?
+            config.component_dirs.add "lib" do |dir|
+              dir.default_namespace = namespace_path.tr(File::SEPARATOR, config.namespace_separator)
+            end
+          end
+        end
       end
 
       # Force after configure hook to run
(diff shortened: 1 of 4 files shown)


lib/hanami/slice.rb
```diff
@@ -99,17 +99,20 @@ def define_container
...
-
-        container.config.auto_register = ["lib/#{namespace_path}"]
-        container.config.default_namespace = namespace_path.tr("/", ".")
-
-        container.load_paths! "lib"
+      container.configure do |config|
+        config.name = name
+        config.inflector = application.configuration.inflector
+
+        if root && File.directory?(root)
```
jodosha, line 106: Just a minor thing, shouldn't we check `root.directory?` as we do [here](https://github.com/hanami/hanami/pull/1093/files#diff-821c48c88c392a3a4fb873393e0c3c8401ca598885b47e1cfe8c0d4c57932293R253)?

lib/hanami/slice.rb
```diff
@@ -99,17 +99,20 @@ def define_container
...
+          config.root = root
+          config.bootable_dirs = ["config/boot"]
+
+          if File.directory?(root.join("lib"))
```
jodosha, line 110: Same here.

lib/hanami/slice.rb
```diff
@@ -99,17 +99,20 @@ def define_container
...
+            config.component_dirs.add "lib" do |dir|
+              dir.default_namespace = namespace_path.tr("/", ".")
```
jodosha, line 112: This line assumes UNIX like systems, should we use `::FIle::SEPARATOR` instead? This to guarantee cross platform compatibility.

Regarding the dot, do we have a central place where this namespace separator is defined and it can be referenced from here? This is both for maintenance perspective and also code clarity.

jodosha (approved):

lib/hanami/slice.rb
timriley: Thanks @jodosha — I'm updating this now.

As for the dot, I've changed it to use `config.namespace_separator` (so it'll use the default value provided by dry-system), and added this card to make it possible for this config (and others) to be propagated to slices if the user configures something particular on the application container: https://trello.com/c/nQBwnUUX

timriley:
This is now ready to go (thanks for the feedback, @jodosha!), but I'll wait for the next dry-system release (which I plan to make in the next day or two) before I merge, so I can have it depend on a proper release rather than dry-system master.

PR #1094, opened 2021-02-15
Fix spec for `hanami new`

adam12:
Broken by e8aad95 (#1089)


jodosha (approved):
@adam12 Thanks for taking care of this 🙏

PR #1095, opened 2021-02-17
Fix middleware config so blocks are accepted without error

dabarrell:
Hey, came across this issue when trying to use middleware with a block. 

```
module Service
    # ...
    config.middleware.use OmniAuth::Builder do
      provider :developer unless Hanami.env?("production")
    end
    # ...
  end
end

```

The supplied block was being treated as an argument when the middleware was being initialised, leading to a mismatch in the number of required/supplied arguments.

The main issue seems to be the `&` in `application.config.for_each_middleware do |m, *args, &block|`. In that, `&block` won't match the `Proc` in the stack. Instead, the `Proc` will be the last value in `args`. 

I also removed the splat for the args to match how the middleware stack is handled in the router.

I've added a case to the integration spec that would test this path, but I haven't been able to get the integration specs to run successfully, so can't guarantee that it passes.

### DIFF
--- a/lib/hanami/configuration/middleware.rb
+++ b/lib/hanami/configuration/middleware.rb
@@ -11,7 +11,7 @@ module Hanami
       end
 
       def use(middleware, *args, &block)
-        stack.push([middleware, *args, block].compact)
+        stack.push([middleware, args, block].compact)
       end
 
       attr_reader :stack
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -331,7 +331,7 @@ module Hanami
         ) do
           use application[:rack_monitor]
 
-          application.config.for_each_middleware do |m, *args, &block|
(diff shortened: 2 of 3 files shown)


lib/hanami/configuration/middleware.rb
```diff
@@ -11,7 +11,7 @@ def initialize
       end
 
       def use(middleware, *args, &block)
-        stack.push([middleware, *args, block].compact)
+        stack.push([middleware, args, block].compact)
```
dabarrell, line 14: @timriley wasn't certain what the purpose behind the `.compact` was here so left it in for the moment, but I think it can be removed.

solnic:
Thanks and I'm sorry we haven't managed to deal with this PR. Related code has been reworked a bit recently and it resulted in fixing block handling (amongst other things). I closed this PR via a commit which simply adds a repro spec which now passes 😄

PR #1097, opened 2021-02-28
Remove ruby 2.4 on CI

waghanza:
Hi @jodosha,

As **ruby** `2.4` os reached end of life on march 2020, I think it's meaningless to keep running CI for it.

Regards,

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,6 @@ jobs:
           - "2.7"
           - "2.6"
           - "2.5"
-          - "2.4"
     steps:
       - uses: actions/checkout@v1
       - name: Install package dependencies


adam12:
I'm not 100% sure on this. We're still setting minimum Ruby version as 2.3, and Luca just setup the CI and I'm presuming added 2.4 for a reason.

Maybe the overall fix is sweeping changes to both gem spec and CI across all projects for the 1.x lineage of Hanami gems.

waghanza:
Of course, it's up to `hanami`'s core team to decide, but I'm not sure supporting ruby older than 2.5 has a benefit tho :-)

adam12:
> Of course, it's up to `hanami`'s core team to decide

I'm going to ask Luca and we'll go from there :)

jodosha:
@waghanza My short answer is no. 😃 

Let me elaborate on this. We care about long term maintenance. Hanami 1 started with support with Ruby 2.3+ and we're going to do until Hanami 1 series will be supported.

The reason is: you're running a Hanami 1.3 app on Ruby 2.3, you upgrade to Hanami 1.4 and your app doesn't start anymore. Or maybe Bundler prevents you to install the gem in the first place. That IS a breaking change in a minor version. In this case it isn't a change in the public API, but the ability to run the app at all.

I'm going to close this PR for this reason.

---

With that given said, why do you wanted to remove Ruby 2.4 from CI? Just because of saving CI time?

waghanza:
> With that given said, why do you wanted to remove Ruby 2.4 from CI? Just because of saving CI time?
yep

PR #1100, opened 2021-04-14
Add simple Zeitwerk integration

timriley:
Set up autoloading by default for Hanami 2 applications.

This uses Zeitwerk and dry-system's autoloading support.

To support this, we do the following:

- Set up a Zeitwerk loader by default at `Hanami.application.configuration.autoloader`
- When the application and slice containers are prepared:
  - Add their directories to the Zeitwerk loader
  - Enable the dry-system `Loader::Autoloading` and disable the containers' component dirs from being added to the load path
- Then later in the Hanami app init process
  - Configure the autoloader with the app's inflector (via an adapter, since the expected API is slightly different)
  - Then call `#setup` on the loader, effectively finalizing its config

We also support opting out of autoloading entirely by setting `Hanami.application.configuration.autoloader` to `nil` or `false`.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ gem "hanami-view", "~> 2.0.alpha", git: "https://github.com/hanami/view.git", br
 
 gem "hanami-devtools", require: false, git: "https://github.com/hanami/devtools.git", branch: "unstable"
 
+gem "dry-system", git: "https://github.com/dry-rb/dry-system.git", branch: "default-settings-for-component-dirs"
 gem "dry-configurable", git: "https://github.com/dry-rb/dry-configurable.git", branch: "master"
 
 group :test do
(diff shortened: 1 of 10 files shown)


timriley:
[Integrate with Zeitwerk for code loading](https://trello.com/c/t2Mcs3Cw/63-integrate-with-zeitwerk-for-code-loading)

fxn:
That is all? Man, I do not know the details but this patch tells me the work behind this integration is really good 💯.

timriley:
@fxn Thanks for taking a look! ❤️ 

As my latest commit shows (https://github.com/hanami/hanami/pull/1100/commits/c35c7aedc6fb6071a82334b00fc58c113039d10d), there's still a little to come, but I hope the integration picture overall won't need to become much more complex than it is now.

If there's any necessary part of integration that you think we might be overlooking, would definitely appreciate any pointers!

One thing I'm keeping in mind is enabling reloading when the application is in development mode, but I think I might consider that in another PR, after all the basic integration is done and shown to be working smoothly.

jodosha (approved):
@timriley Great to see this to happen! 👏

solnic (approved):
Things are getting real 💪🏻

flash-gordon (approved):

fxn:
🎉

PR #1102, opened 2021-05-03
Make web-specific gems optional

timriley:
Remove hanami-controller, hanami-router, and hanami-view from the Gemfile, instead allowing these gems to be included in the Gemfiles of generated Hanami applications.

This allows Hanami to be used more gracefully as a non-web application framework, since it will no longer be pulling in unnecessary dependencies for that case.


solnic (approved):
instaprove

timriley:
[Remove web-specific gems from gemspec](https://trello.com/c/abf7mmHq/116-remove-web-specific-gems-from-gemspec)

jodosha (approved):

PR #1105, opened 2021-05-10
Use dry-configurable for hanami settings

waiting-for-dev:
This is a baby step towards the to-be-defined level of integration we
want between hanami settings and dry-configurable.

This commit introduces the minimal changes to keep the very same
features we're supporting on the hanami settings side while still
delegating to dry-configurable what it's already capable of handling:

- dry-configurable is responsible for the setter/getter methods.
- Hanami settings resolve the value through a type constructor, when
  present, before delegating it to dry-configurable.
- Hanami settings wrap the definition of all the settings to detect all
  errors at once instead of failing on a one-by-one basis.

Further integration can include delegating the type constructor to a
processor keyword argument (we could already delegate it to the block,
but probably we want to do the former) and provide an error handler to
the whole definition.

### DIFF
--- a/lib/hanami/application/settings.rb
+++ b/lib/hanami/application/settings.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "dry/core/constants"
+require "dry/configurable"
 require_relative "settings/definition"
 require_relative "settings/struct"
 
@@ -10,6 +11,8 @@ module Hanami
     #
     # @since 2.0.0
     module Settings
+      extend Dry::Configurable
+
       Undefined = Dry::Core::Constants::Undefined
 
       def self.build(loader, loader_options, &definition_block)
--- a/lib/hanami/application/settings/loader.rb
+++ b/lib/hanami/application/settings/loader.rb
@@ -42,11 +42,11 @@ module Hanami
         def call(defined_settings)
           load_dotenv
 
-          settings, errors = load_settings(defined_settings)
+          errors = load_settings(defined_settings)
 
           raise InvalidSettingsError, errors if errors.any?
 
-          settings
+          Settings.config.values
         end
 
         private
@@ -67,14 +67,16 @@ module Hanami
         end
 
         def load_settings(defined_settings) # rubocop:disable Metrics/MethodLength
-          defined_settings.each_with_object([{}, {}]) { |(name, args), (settings, errors)|
+          defined_settings.reduce({}) { |errors, (name, args)|
             begin
-              settings[name] = resolve_setting(name, args)
+              value = resolve_setting(name, args)
+              Settings.setting(name, value)
+              errors
             rescue => exception # rubocop:disable Style/RescueStandardError
               if exception.is_a?(UnsupportedSettingArgumentError) # rubocop: disable Style/GuardClause
                 raise exception
               else
-                errors[name] = exception
+                errors.merge(name => exception)
               end
             end
           }
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -153,18 +153,17 @@ module Hanami
 
       def settings(&block) # rubocop:disable Metrics/MethodLength
         if block
-          @_settings = Application::Settings.build(
+          Application::Settings.build(
             configuration.settings_loader,
             configuration.settings_loader_options,
             &block
           )
-        elsif instance_variable_defined?(:@_settings)
-          @_settings
+        elsif Settings.config.values.any?
+          Settings.config
         else
           # Load settings lazily so they can be used to configure the
           # Hanami::Application subclass (before the application has inited)
           load_settings
-          @_settings ||= nil
         end
       end
 


lib/hanami/application/settings.rb
```diff
@@ -10,6 +11,8 @@ class Application
     #
     # @since 2.0.0
     module Settings
+      extend Dry::Configurable
```
timriley, line 14: I feel a bit uncomfortable making this `Settings` module itself the configurable thing, making it basically global state. I know we have _some_ of that in Hanami, but if we can avoid adding more, I feel that's a positive.

In the current settings structure, we already have `Definition` that is initialized and then becomes the thing that's configurable. Could we do similar here? In this case it could be putting `include Dry::Configurable` on `Definition` and then dynamically creating a subclass to define the settings on. This would be the place where we eval the `&definition_block` in `Settings.build`, which would allow us to use the dry-configurable `setting` method directly.

What do you think?

lib/hanami/application/settings/loader.rb
```diff
@@ -42,11 +42,11 @@ def initialize(*)
         def call(defined_settings)
           load_dotenv
 
-          settings, errors = load_settings(defined_settings)
+          errors = load_settings(defined_settings)
 
           raise InvalidSettingsError, errors if errors.any?
 
-          settings
+          Settings.config.values
```
timriley, line 49: Continuing from my comment above, having `load_settings` here actually return the configured and populated configurable object would mean we don't have this reference to a constant here appearing out of nowhere, which to me feels like a bit of a design smell.

timriley (commented):
Looking like a good start, thanks @waiting-for-dev!

I left some thoughts about how we take your work so far and rearrange it just a little — let me know what you think!

lib/hanami/application/settings.rb
solnic: I wanted to point out the same thing. BTW it would be really lovely if we could end up with a class called `Settings` defined within the Application namespace and have its *instance* registered under `:settings` key.

waiting-for-dev: I agree that using the Settings module for configuration feels hacky. However, I don't feel confident trying to extend the current architecture as I'm not sure how much of it will be necessary once we complete the integration. In other words, I'm afraid of doing iterations that will undo previous iterations, and I think a bit of upfront design will pay off here.

At this point, the most significant gap between what is supported by hanami settings and dry-configurable is that the former can catch all type errors at once. Would we be fine dropping this feature for now and re-adding it later to the proper layer (dry-configurable) in the appropriate way (maybe providing an error handler to dry-configurable or dry-types)?

solnic: Yes, this should be supported by dry-configurable IMO

timriley: At this point, I'm not sure if we need to rush to add the ability to configure multiple settings and bundle up their errors into dry-configurable, at least not until we've seen several other use cases where it makes sense. Right now it doesn't seem to be used that way.

The reason I added this to Hanami is because it's a frustrating cycle in teams with frequently changing settings files to have to make multiple passes at updating it because you're only learning about errors once at a time. Same for if there are configuration errors in a deployed environment: you're better off learning about all the errors in one go rather than in dribs and drabs.

If we did want to add that to dry-configurable, I'd say it should be a post-1.0 feature and something we should look at a fair bit later.

_Anyway, I digress!_ Getting back to the point I raised initially here, about how it might make for a better design to have individual settings instances that are configured rather than the global `Settings` module, I had a go at it on top of your branch:

https://github.com/hanami/hanami/compare/unstable...enhancement/unstable/settings-to-dry-configurable--tim

See the last commit for my changes. This has the `spec/new_integration/application_settings_spec.rb` integration spec passing, but I haven't touched unit tests.

This adjustment allowed us to get rid of _ton_ of code! And the only surface-level change is that we're using the dry-configurable `setting` API directly inside the `Hanami.application.settings(&block)`, which I believe was our end goal here anyway, especially given we have plans to tweak the dry-configurable `setting` arguments to make them clearer.

What do you think, @waiting-for-dev? Worth continuing in this direction?

lib/hanami/application/settings.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "dry/core/constants"
+require "dry/configurable"
```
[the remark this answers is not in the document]
cllns: Seems like there's not, surprisingly https://github.com/rubocop/rubocop/issues/3741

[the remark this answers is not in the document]
solnic: Too bad, well, another thing to add to my list of tasks 😆

lib/hanami/application/settings.rb
waiting-for-dev: Hey @timriley, it looks neat!

I added a commit to your branch fixing the unit tests. Please, take a look when you have a second :slightly_smiling_face: 

Moving on, I wonder if we really need the loader to be configurable. It doesn't harm, but now that the DSL is forwarded to dry-configurable, probably there's not much to tweak. I guess that what we do want to allow is using something different than `Dotenv`, but we should decouple it to another class/module besides the loader. It shouldn't be a high priority, though.

waiting-for-dev: Related to the previous comment, if we decide to provide a custom loader, maybe we could revisit the expected interface. Are `loader_options` needed? Can't we rename `load` to `call` and therefore allow any callable object?

timriley: Thanks for fixing the tests! That commit looks good.

As for the configurable loader, "allow using something different than Dotenv" is definitely the main thing I had in mind when separating it out. So I'd definitely be up for massaging the code a bit now that we have a clearer picture of this new settings structure, but I do think it'd be useful to keep the loader distinct, just so it can stay swappable. Perhaps we could even rename the default loader to `DotenvLoader` or something, to make its purpose clearer?

The loader class actually used `#call` as its main method before the commit that I pushed here. What changed is that the loader was no longer returning a newly created object, but instead mutating the one passed into it, so "call" no longer felt right given the style of functional, callable object that we're advocating for with users of Hanami 2 (among other things, "don't mutate what you don't own"). This is why I tried out "load," since it avoided muddying the "callable object" ideal. What do you think?

timriley: As for `loader_options`, one way we could avoid it is to expect an already initialized object as the loader rather than a class name and a set of options. Feels like that's a worthwhile simplification at this point!

waiting-for-dev: Hey @timriley,

> What changed is that the loader was no longer returning a newly created object, but instead mutating the one passed into it, so "call" no longer felt right given the style of functional, callable object that we're advocating for with users of Hanami 2

I think it's a good point, and I completely agree with it.

Please, take a look at the last commit in your branch: https://github.com/hanami/hanami/commit/107c8be20c1743cce89bb4d5668321a8a94d60b3

I have broken apart what is the storage for the settings (ENV with dotenv enabled in our case) from what we actually do to load the settings in the config class. Both the store and the loader are configurable. Users could provide their own store as long as it implements the `#fetch` method. As for the loader, we can discuss whether it makes any sense to be injectable, as its behavior is very private, relying on dry-configurable dependency.

If you like how it looks, please tell me, and I'll improve yard docs and the commit message.

waiting-for-dev:
Replaced by #1110

PR #1106, opened 2021-05-11
Expose url_helpers from router

cllns:
(Note: This is for Hanami 2, which is on the `unstable` branch)

I don't think this is the right place for it, considering it would be the _only_ public method available on classes that inherit from `Hanami::Application`. But just wanted to open this PR to show the simplest way it could be done.

I wasn't able to find any other publicly accesible way to access the router's `url_helpers` from within an app.

I tried to register `router.url_helpers` on the application's container, but it's already finalized by this point so it can't be added. I wonder if the router will have to be created earlier, if that's possible, then we can put a `router.url_helpers` in the container registry at that time.

(The way I was able to work around this was `Hanami.app.instance_variable_get(:@app).app.instance_variable_get(:@app).url_helpers`, which is obviously not an acceptable solution, pulling the router out from private instance variables)

### DIFF
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -331,6 +331,8 @@ module Hanami
 
     # Application instance interface
     module InstanceMethods
+      attr_reader :url_helpers
+
       # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       def initialize(application = self.class)
         require_relative "application/router"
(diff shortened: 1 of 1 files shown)


jodosha (commented):
@cllns Where do you want to access your URL helpers from?

cllns:
I want to access them from a view. So, in [`hanami-2-application-template`'s `View::Base`](https://github.com/hanami/hanami-2-application-template/blob/067b4aea8b13be1b5be08c89abca751d958ef132/slices/main/lib/main/view/base.rb), adding:

```ruby
expose(:routes) do
  Hanami.app.url_helpers # With this change
  # Without this change
  # Hanami.app.instance_variable_get(:@app).app.instance_variable_get(:@app).url_helpers.path(:root)
end
```

So that I can access `routes` in every template, then do `routes.path(:books)` to get `/books`.

If the path helpers will be handled in some other way, then feel free to close this PR :)

cllns:
Closing in favor of #1119 🌸

PR #1107, opened 2021-05-18
Port configuration to dry-configurable

timriley:
In an effort to have consistent configuration objects across all the core hanami gems, this PR ports `Hanami::Configuration` to dry-configurable. With this in place, dry-configurable-backed objects are used across all three of hanami, hanami-controller, and hanami-view.

The resulting `Hanami::Configuration` object certainly feels much more manageable, with settings and their defaults now kept together instead sprawling across multiple places in the class.

Along with this, it makes several improvements:

- The "missing configuration" objects for `actions` and `views` are now replaced with a proper `NullConfiguration` instance, which responds to `finalize!` even in the case of the action and view gems not being loaded (allowing us to remove a `respond_to?` hack in `Configuration#finalize!`
- The handling of per-environment configuration is tightened up, with configuration for the current environment applied immediately (if applicable) instead of waiting for the configuration to be finalized
- The logger configuration is pulled into a dedicated object, giving us a namespace for any router-related settings of behaviour in future. Right now it supports configuring the logger class and options separately (as before), as well as providing a pre-initialized router object. The `rack_logger_filter_params` setting has also moved into this namespace (for a future improvement, it would be nice to have this respected outside of the web apps, too)
- Allows a preinitialized inflector to be provided via `config.inflector = my_inflector`, while still preserving the previous behaviour passing custom inflection rules only via `config.inflections(&block)` 

This change does add dry-configurable as a runtime dependency of hanami, but given we're requiring it transitively already via dry-system, this feels fine to me.


lib/hanami/configuration.rb
```diff
@@ -3,268 +3,176 @@
...
       end
 
+      @middleware = Middleware.new
+
+      @router = Router.new(self)
+
       @views = begin
         require_path = "hanami/view/application_configuration"
         require require_path
         Hanami::View::ApplicationConfiguration.new
```
timriley, line 65: I actually wonder whether we should make these `setting` declarations like most of the others... by having a mix of both settings as well as directly-assigned attributes/ivars, I feel we lose a bit of consistency (and these latter attributes don't show up when inspecting the configuration's `settings` etc.)

That said, in these cases, there's no need for e.g. the `middleware` or `router` objects to be wholly assigned or _replaced_ like the other settings, so perhaps just leaving them as directly-assigned attributes makes the most since, since we can avoid having writer methods for them.

lib/hanami/configuration.rb
```diff
@@ -3,268 +3,176 @@
...
+      apply_env_config
 
-    def slices
-      settings[:slices]
-    end
+      # Finalize nested configurations
+      actions.finalize!
+      views.finalize!
+      logger.finalize!
+      router.finalize!
```
timriley, line 89: It's becoming quite a list of things we're finalizing now! At some point we might need to take a more dynamic approach to doing this, but I think we can defer this until the point where we let plugins register their own configs, etc., which will be a whole other kettle of fish.

lib/hanami/configuration.rb
```diff
@@ -3,268 +3,176 @@
...
       rescue LoadError => e
         raise e unless e.path == require_path
-        Object.new
+        require_relative "configuration/null_configuration"
+        NullConfiguration.new
       end
 
+      @middleware = Middleware.new
+
+      @router = Router.new(self)
```
jodosha, line 60: @timriley We should assume that router is bundled as it's an app dependency, not an `hanami` dep. Should we rescue `LoadError` here as well?

lib/hanami/application/container/boot/logger.rb
```diff
@@ -2,7 +2,6 @@
 
 Hanami.application.register_bootable :logger do
   start do
-    require "hanami/logger"
-    register :logger, Hanami::Logger.new(**Hanami.application.configuration.logger)
+    register :logger, Hanami.application.configuration.logger.instance
```
jodosha, line 5: @timriley It seems a SRP violation to have a config object (logger settings) to also build the target (the logger itself). Can't we keep this LOC as it was?

lib/hanami/configuration/router.rb
```diff
@@ -1,50 +1,50 @@
...
 
-      # @api private
-      # @since 2.0.0
-      def resolver
-        @resolver ||= begin
-                        require_relative "../application/routing/resolver"
-                        Application::Routing::Resolver
-                      end
-      end
+      setting :routes_path, "config/routes"
```
jodosha, line 25: @timriley Shouldn't this respect `File::SEPARATOR`?

```ruby
setting :routes_path, Hanami.root.join("config", "routes")
```

Not sure if we need to append `.rb` suffix there.

lib/hanami/configuration/router.rb
```diff
@@ -1,50 +1,50 @@
...
+
+      def method_missing(name, *args, &block)
+        if config.respond_to?(name)
+          config.public_send(name, *args, &block)
+        else
+          super
+        end
+      end
+
+      def respond_to_missing?(name, _incude_all = false)
```
jodosha, line 45:
```suggestion
      def respond_to_missing?(name, _include_all = false)
```

jodosha (commented):
@timriley Thanks for taking this step towards the configuration unification. It's the right approach.

I left a few comments for your consideration. Thanks. 🙂

lib/hanami/configuration/router.rb
solnic: @jodosha oh, this is a good reminder! I sometimes write `root.join("foo/bar")` too - totally forgot about the separator thing.

lib/hanami/configuration.rb
solnic: @timriley I'm ok with this being hardcoded but eventually I think we want a feature-detection mechanism and everything should become a registered feature. Then this code will become something like `features.each(&:finalize)` or even `features.finalize!`.

solnic: ps. I think I'll move logger out into a component btw, like I mentioned on multiple occasions 🙂

lib/hanami/configuration.rb
timriley: @jodosha if we took the approach of having the hanami-router gem hold its own configuration classes, then yep, that's definitely what we should be doing. Right now, however, the router configuration class is still housed within the hanami gem itself, which is why we don't need the LoadError protections.

I'm really glad you brought this up, though, because I think we should definitely move the router configuration and other integration bits and pieces into the hanami-router gem. I'll make a card for it.

In the meantime, this PR doesn't change much regarding the overall state of the router configuration - it's just an intermediary bit of tidying of what was already there.

lib/hanami/configuration/router.rb
timriley: Happy to convert this one to join using `File::SEPARATOR`. I think we may need to do a serious sweep across all gems to make sure we're doing it properly across the board.

I wish Ruby's file path joining respected this out of the box 😩

lib/hanami/application/container/boot/logger.rb
timriley: Yeah, on reflection I think you're right. Thanks for raising this.

I agonised over this one for a while. There's a couple of in-tension criteria I wanted to satisfy with this `configuration.logger` object:

- Allow particular aspects of the standard logger to be configured without the user having to pre-initialize a whole replacement object (e.g. the user who just wants to change the log level, perhaps also adjusting it in different Hanami environments, but leave everything else the same)
- But also allow the user to supply their own pre-initialized logger object of their choosing, and have that override those finer-grained `logger_class`/`options` settings

This is what led me to the current implementation of `Configuration#logger=` (which sets the pre-initialized logger object) as well as the `instance` attribute on which that pre-initialized logger is stored. Then from there, given the job of _constructing_ the logger is now non-trivial and contingent on several different settings, which should be checked in a particular order, I thought it would be good to encapsulate that behaviour alongside those settings, so they can all evolve together when needed.

If we removed this, it's true that it would make `Configuration::Logger` "purer" and more single-responsibility, but it does mean that this logger initialisation logic needs to go elsewhere and stay in sync.

I suppose if it _only_ ends up here in this bootable component file, and doesn't have to be repeated in other places, then I think we should be fine. I'll change it back now.

While we're discussing this, do you have any other ideas about about how we could satisfy those two different criteria I listed above, or do you think the way I've done it is reasonable enough? Is the second of those two criteria even worth supporting? (I think so, if possible, since I'd like to make Hanami as flexible as possible for those you need to go off the beaten track, so long as it doesn't compromise the experience for the other 99% of users)

lib/hanami/configuration/router.rb
timriley: Fixed this, and another instance 👍🏼

lib/hanami/application/container/boot/logger.rb
timriley: Another option here could be to move the logger options back into the top-level configuration class, but from my perspective, as soon as there are >2 settings at the top level with the same underscored "namespace", e.g. `logger_this`, `logger_that`, and `logger_another`, it's a clear sign that we could make things clearer to understand and maintain by grouping them together under a nested configuration object.

timriley: I've just pushed [`9226d1a`](https://github.com/hanami/hanami/pull/1107/commits/9226d1a20596b70c74292c58f29d359a88badd1d), which removes the building of the instance from the `Configuration::Logger` class.

lib/hanami/configuration.rb
timriley: @solnic Forgive me for asking given you've likely mentioned this in a place where I've already seen it (😅), but what do you mean by moving it into a component, exactly?

We do already have `lib/hanami/application/container/boot/logger.rb` which is the framework-provided bootable component for initialising and registering the logger in the Hanami application container. As of this PR, it currently looks like this:

```ruby
# frozen_string_literal: true

Hanami.application.register_bootable :logger do
  start do
    logger_config = Hanami.application.configuration.logger

    logger =
      if (logger_instance = logger_config.instance)
        logger_instance
      else
        logger_config.logger_class.new(**logger_config.options)
      end

    register :logger, logger
  end
end

```

lib/hanami/application/container/boot/logger.rb
timriley: On the topic of allowing a user to supply their own pre-initialized logger, I did just write this comment in the other thread here (https://github.com/hanami/hanami/pull/1107#discussion_r634323665), which describes how you can override the whole `logger` bootable component to register your own logger instance in the Hanami application container. This is already a viable path for a user who wants to do this, so perhaps we _don't_ need to account for it in this logger configuration object? That said, that's a pretty advanced technique, and we already e.g. support providing an inflector object via configuration, so perhaps allowing it for the logger does make sense too?

timriley:
[Port Hanami::Configuration to dry-configurable](https://trello.com/c/yNWhVUPn/133-port-hanamiconfiguration-to-dry-configurable)

timriley:
[Allow application inflector object to be wholly replaced via `config.inflector = new_inflector`](https://trello.com/c/jxD0gBAG/120-allow-application-inflector-object-to-be-wholly-replaced-via-configinflector-newinflector)

lib/hanami/application/container/boot/logger.rb
jodosha: @timriley Can't the responsibility of instantiating the logger be delegated to `Hanami::Configuration`?

Pseudo code to explain myself.

In the bootable component:

```ruby
register :logger, Hanami.application.configuration.application_logger
```

In configuration:

```ruby
setting :logger, Configuration::Logger.new, cloneable: true

def logger=(logger_instance)
  config.logger = Configuration::Logger.new(logger_instance)
end

# @api private
def application_logger
  manually_assigned_logger || Hanami::Logger.new(*logger_settings)
end
```

lib/hanami/configuration/router.rb
```diff
@@ -1,50 +1,50 @@
...
+        {base_url: base_configuration.base_url}
       end
 
-      # @api private
-      # @since 2.0.0
-      DEFAULT_ROUTES = File.join("config", "routes")
-      private_constant :DEFAULT_ROUTES
+      private
+
+      def method_missing(name, *args, &block)
```
waiting-for-dev, line 37: Can't we get rid of it if we make the settings with `reader: true` option?

timriley: We need it for the writers still. I'd love to have dry-configurable support this kind of use case out of the box, since I've written this particular method_missing implementation about 10 or more times in my Hanami work so far.

lib/hanami/configuration.rb
solnic: @timriley oh I missed your comment sorry! I thought we still use the logger plugin from dry-system to set up a logger, that's what I was referring to.

solnic:
@timriley could you rebase? I've improved CI setup in the unstable branch

timriley:
@solnic Yep, I can rebase once I get back onto this work properly, thanks for the heads up 👍🏼

lib/hanami/application/container/boot/logger.rb
timriley: @jodosha Thanks for the great ideas :) I've just incorporated this in [`25db2b9` (#1107)](https://github.com/hanami/hanami/pull/1107/commits/25db2b94b2e67246042d794d7a2216c597375d1e).

If you take a look at that commit, there are a couple of subtle differences I wanted to note:

1. I've called the method `#logger_instance` instead of `#application_logger`. In this case I feel it better describes what it actually returns. The application prefix felt a little redundant, and the _true_ "application logger" is the one that we register in `lib/hanami/application/container/boot/logger.rb`.
2. I also changed it so it doesn't memoize a logger built from the `logger` settings. The reason for this is that we want to make sure settings changes made to `config.logger.*` are _always_ respected when it comes time to get the full logger object. If we memoize, then as soon as we access the `logger_instance`, any subsequent changes to `configuration.logger.*` will just be ignored, which would be unexpected and potentially lead to bugs or unwanted behaviour.

Does that look OK to you?

All I think I have left here is just to make sure there's a test to cover this adjusted behaviour, then I think this'll be ready to merge.

timriley:
@solnic I just rebased, but it seems CI is still failing here when it tries to build Ruby:

```
Error: The process '/opt/hostedtoolcache/Ruby/3.0.1/x64/bin/bundle' failed with exit code 11
```

solnic:
> @solnic I just rebased, but it seems CI is still failing here when it tries to build Ruby

@timriley Gemfile here points to a dry-configurable branch that is gone https://github.com/hanami/hanami/pull/1107/checks?check_run_id=2763979440#step:3:29

PR #1110, opened 2021-05-24
Settings: use dry-configurable

waiting-for-dev:
Replaces #1105 

Uses dry-configurable for the hanami settings.

We have broken apart what is the storage for the settings (ENV with dotenv enabled in our case) from what we actually do to load the settings in the config class. Both the store and the loader are configurable. Users could provide their own store as long as it implements the `#fetch` method. As for the loader, we can discuss whether it makes any sense to be injectable, as its behavior is very private, relying on dry-configurable dependency.


solnic:
@waiting-for-dev could you rebase this? I've improved CI setup in the unstable branch.

waiting-for-dev:
> @waiting-for-dev could you rebase this? I've improved CI setup in the unstable branch.

Done @solnic

waiting-for-dev:
Closing in favor of #1114

PR #1111, opened 2021-05-25
Make a types module available to Hanami settings

waiting-for-dev:
It binds a `Types` module to the settings definition block so that a user doesn't need to rely on the more generic, if present, application types.


lib/hanami/application/settings.rb
```diff
@@ -14,8 +15,9 @@ def self.build(store, loader, &definition_block)
         # other specialised behaviour) we could turn the `Settings` module here into a
         # class that includes Dry::Configurable and then subclass _it_ instead of just
         # making a new class without any defined superclass
-        Class.new { include Dry::Configurable }
-          .instance_eval(&definition_block)
+        Class.new
+          .include(Dry::Configurable)
+          .instance_exec(Types, &definition_block)
```
solnic, line 20: I think it would be nicer to expose `Types` as a constant in this block when it's defined by the application and auto-define the default one when it's missing. With the approach now we'd introduce a different way of using types which is not strictly necessary.

waiting-for-dev: But I think that the original reason for this change is the opposite: isolate the application settings from the application-level types. From the [trello card](https://trello.com/c/qExli5cy/105-make-a-types-module-automatically-available-if-possible-inside-the-hanamiapplicationsettings-block):

> I think there's a benefit in having these types isolated from the application's own types module (if any), since it ensures that the settings load in isolation, without any other application concerns creeping in (especially since that in some cases, some of those application concerns may rely on the settings, which creates a circular dependency and will cause the app to crash at boot)

If someone wants to use the application types, they only need to access them directly and ignore the argument bound to the definition block.

solnic: Right, thanks for the reminder. So, I think we want to auto-define `Types` constant that's accessible within the settings class and provide a conventional way of defining your own types there. The reason for this is simple - you will want custom setting types in many cases and you shouldn't use your main types for that because of the potential circular-dep issue.

timriley: Certainly it's possible to make the types module available as `types` here (e.g. by defining a method or setting an ivar on the class, etc.), but I don't think it's possible to make it available as a proper `Types`-like constant, due to many Ruby reasons, mostly because of the way the user-provided settings block is evaluated, with constants looked up from the top-level, not from the scope of the class. The best I could get working was `self::Types` or `self.Types()` or `Types()`, but they're all pretty gross.

Given this, I think that makes it even harder to provide a nice way of constructing custom types within this context, because doing that with dry-types typically means reopening one of its generated modules, and defining _new_ constants building off the constants nested within the types module. But in this case if we can't refer to the types module directly, we're running out of options.

Again, the best I could come up with some noodling was:

```
types::MyBool = types::Params::Bool

setting :blah, constructor: types::MyBool
```

Are we happy with this kind of outcome? Doesn't feel super great to me... and I wonder if it means we should be doing a bit of a rethink of this in general.

timriley: One rethink-the-approach idea could be that we just use a class definition here:

```ruby
module MyApp
  class Settings < Hanami::Application::Settings
    Types = Dry.Types()

    module Types
      MyBool = Params::Bool
    end

    setting :numeric_setting, constructor: Types::Params::Integer
    setting :feature_flag, constructor: MyBool
  end
end
```

And maybe we don't even implicitly provide types, we could just generate the `Types = Dry.Types()` and related lines into the class.

This is a vastly different approach to currently, however, and would make the settings inconsistent with other similar definitions of application details, like the routes. So we'd need to think pretty hard about this (and assess other options too, this one just came to me after thinking on it for a minute).

solnic: @timriley I like the idea of just using a class. It doesn't look like there's any benefit in having a block-based DSL, right?

timriley: @solnic Yeah, at this point I think the only reason to have the block DSL is to make things feel "nice." However, as we're seeing here, it comes with some drawbacks. Using a class would also be particularly nice since we're actually exposing this as a registered object within the application container, so the class<->instance relationship should be pretty clear. Plus it provides a place where the app author can easily add extra behaviour to their settings, which is a real bonus.

@solnic What do you think about such a change effectively reserving the `Settings` name within an app's namespaces?

My other question here is around consistency. If we make this change and turn the settings into a class definition, should we do similar for the routes definition? This is the one other thing that looks like the current settings definition (it was what I copied it from originally):

```ruby
Hanami.application.routes do
  # ...
end
```

Or are we OK with these being different? Or is there a valid case for this to be made into an e.g. `MyApp::Routes` class too? It could be helpful if the resulting object became the thing that offered the helper methods to return various routes?

@jodosha What do you think about all of this?

timriley: On naming, I suppose we could call it `ApplicationSettings` if we wanted to make it less likely to conflict with a common generic term. Which could actually make our job easier when it comes to referring to it on docs too, since "application settings" is far more specific and less likely to be confused for other "settings" in general.

solnic: @timriley I'm OK with changing routes too. The less DSL the better in general. I'm not sure if there's any advantage of having a block-based routing DSL, maybe there is. If it's just "look nice" then I'm not sure if that's enough of a reason 🙂 

Re `ApplicationSettings` - perfect. In `dry-rails` I added stuff like `ApplicationContract` etc. so this is consistent. In Rails people are also used to this prefix so that's another plus.

waiting-for-dev: Hey! I think it's a good idea. I agree it will make things less complex and more flexible :tada: 

If you want, I can keep going forward with this one. However, if possible, I'd like to review & merge https://github.com/hanami/hanami/pull/1110 before, as it makes some other abstractions on the settings API, and it will avoid having to deal with conflicts.

jodosha: Can't we just define `Hanami::Types` and reference it from the block?

👉 This is more explicit than just `Types`, **especially if we want to separate framework types** (`Hanami::Types`) **from application types** (`MyApp::Types`). This can generate confusion.

I don't mind the block syntax, as it also avoids name clashing (e.g. `MyApp::Settings`), and new class definition that we expose as public API. With the proposed solution we just expose a method.

```ruby
require "dry/types"

module Hanami
  Types = Dry.Types()

  class Application
    def self.inherited(app)
      super
      app.extend(ClassMethods)
    end

    module ClassMethods
      def settings(&blk)
        if block_given?
          @settings = Class.new
            .include(Dry::Configurable)
            .instance_exec(&blk)
        else
          @settings
        end
      end
    end
  end
end

module MyApp
  end
end

MyApp::Application.settings do
  setting :numeric_setting, constructor: Hanami::Types::Params::Integer
end
```

timriley: Thanks for the input, @jodosha :)

I really think we'd be moving in a positive direction with the idea of the discrete class definition, and I hope we can try it for this. Let me address your points one by one.

> Can't we just define `Hanami::Types` and reference it from the block?

This would be possible, yes. There are still considerable downsides if you wanted to define customisations of those types for use with the settings.

You'd have to reopen the full `Hanami::Types` module and make your new types effectively global:

```ruby
MyApp::Application.settings do
  module Hanami::Types
    PositiveInteger = Hanami::Types::Params::Integer.constrained(gt?: 0)
  end

  setting :numeric_setting, constructor: Hanami::Types::PositiveInteger
end
```

This makes `Hanami::Types::PositiveInteger` available everywhere `Hanami::Types` is accessible, which is leaking a particular usage of the types into a module that (given its name) looks intended to be available everywhere. This is not a practice I'd want to encourage our users to be doing, since it goes against all the ideas of proper scoping.

An alternative could be to open a new module inside the settings block and define the new settings  there:

 ```ruby
MyApp::Application.settings do
  module Types
    PositiveInteger = Hanami::Types::Params::Integer.constrained(gt?: 0)
  end

  setting :numeric_setting, constructor: Hanami::Types::PositiveInteger
end
```

However, because of the lexical scoping of constant usage inside Ruby blocks, this will make the `Types` module there top level, i.e. `::Types`, which is not desirable. You could work around this by giving your module a more specific name, e.g. `AppSettingsTypes`, but by that point we're getting overly verbose and we're only putting a bandaid on the issue of having that top-level module defined.

Another option could be to define new types as local variables:

 ```ruby
MyApp::Application.settings do
  positive_int = Hanami::Types::Params::Integer.constrained(gt?: 0)

  setting :numeric_setting, constructor: positive_int
end
```

But that's getting quite far from idiomatic use of dry-types, which, if we're making available automatically if the gem is loaded, is what we'd expect our users to understand here.

Further, the last two code examples above are moving away from good idiomatic use of Ruby itself, particularly the case of defining a module inside a block.

This is why I'm in favour of moving to an ordinary class definition. It makes all of the above complete non-issues. We can define a properly scoped types module within any class that inherits from an "application settings" superclass and everything about interacting with it becomes completely ordinary Ruby again.

And like I described in previous comments, it gives us a number of other benefits on top of this:

- It becomes much easier for users to understand where the app settings behaviour is coming from, which is particularly important given we're registering an instance of these settings in the application container (i.e. `MyApp::Application["settings"]`), and we're expecting users to auto-inject that instance and interact with it from within their own application components
- It makes it possible for users to define their own behavior on that class, which is a fair thing to do for a class we're already expecting them to add behavior via those `setting` definitions. An immediately obvious example would be to add new methods that combine related settings into a single output. This is not something we can easily support via the existing `setting` definitions and something that is very naturally solved by just _adding another method_.
- It further reinforces the Hanami guiding principle of having clearly bounded classes representing each discrete piece of behaviour in an application. We're encouraging our users to do this via creating their own dedicated action/view/command classes, etc., so I think it would be a good thing for the framework itself to do it wherever possible, and _especially_ wherever these objects are things we expect the user to control and interact with directly.

> I don't mind the block syntax, as it also avoids name clashing (e.g. `MyApp::Settings`), and new class definition that we expose as public API. With the proposed solution we just expose a method.

Given the benefits I outlined above, I think they outweigh the cost of "reserving" a name within the app's namespace in this case.

We're already effectively reserving a name with a block-based installation anyway, since the settings instance is registered in the app container with the `settings` identifier - this would block the user from defining their own `Settings` class and have it appear in the container with the expected identifier. In this sense, we're actually taking a truer approach by exposing a class with the same name, because it makes the class<->registration correspondence much clearer.

I'd be happy to toy with some different names, but at this point I think "settings" is probably the clearest and least awkward in actual use.

> 👉 This is more explicit than just Types, especially if we want to separate framework types (Hanami::Types) from application types (MyApp::Types). This can generate confusion.

As for the name of the generated types module, we can explore ways to make it clearer, but the way I've suggested we do it here is consistent with how we suggest referring to other modules throughout an Hanami application.

For example, we have `MyApp::Deps` and `SomeSlice::Deps` modules, but inside the classes in each of those context, we've been so far encouraging it to be used simply as `Deps`, e.g.

```ruby
module MyApp
  class SomeClass
    # This is actually MyApp::Deps, but thanks to the open MyApp module, we can just use Deps
    include Deps["logger"]
  end
end
```

This is the same as the `Types` module I'm suggesting we generate (if dry-types is loaded) inside an app settings class. It would actually be `MyApp::Settings::Types`, but because we're referring to it inside that class, we can just use `Types`.

I'm happy to explore other options here though.

For instance, we could choose _not_ to autogenerate a types module _at all_, and instead generate some commented-out boilerplate into the class that the user could choose to enable, e.g.

```ruby
module MyApp
  class Settings < Hanami::Application::Settings
    # If you want type safe settings and your app uses dry-types,
    # uncomment the following and provide type objects as your 
    # setting `constructor:` option, etc. etc.
    #
    # require "dry/types"
    # Types = Dry.Types()

    setting :foo
  end
end
```

I'm sure there are other avenues here too. Because we'd be working within an ordinary Ruby class, we have far more options than we could otherwise explore than having setting definitions live entirely inside a block.

In general, I feel really positive about this direction and I'd like for us to be able to give it a try. What do you think?

jodosha: @waiting-for-dev @timriley **I'm fine with the class** (`MyApp::ApplicationSettings` or `MyApp::Application::Settings`), as it would eliminate scoping problems. 👍 

---

I'm still not 100% convinced about application vs settings types.

**Trying to put myself in dev shoes here.** I've got my types defined in `lib/my_app/types.rb` and I try to use them in `config/settings.rb`, but they don't work. There are other settings in the class that use `Types::String`, but if I try to use my own `Types::PositiveInteger` (defined in `lib/my_app/types.rb`), it won't work. Why is that? Is it a bug?

We can explain why is that, but **there is a friction**. I'm wondering how can we solved this.

@waiting-for-dev @timriley Given `lib/my_app/types.rb` doesn't need any particular boot sequence, why can't we require it and reuse in the settings?

I assume Marc's [comment](https://github.com/hanami/hanami/pull/1111#discussion_r639495331) about circular deps is outdated now that we have opted for a class definition.

To summarize, `config/settings.rb` could be:

```ruby
require "lib/my_app/types"

module MyApp
  class ApplicationSettings < Hanami::Application::Settings
    setting :foo, MyApp::Types::PositiveInteger
  end
end
```

WDYT?

solnic: > @waiting-for-dev @timriley Given lib/my_app/types.rb doesn't need any particular boot sequence, why can't we require it and reuse in the settings?

wow great point and it's literally my current use case - I've got a `MyApp::Types::PrivateKey` that is used in my settings

timriley: Thanks so much @jodosha for your open-mindedness here and for giving the full class definition a go! ❤️ 

### Regarding the name

As for the name, I think we'd have to think about how it maps to a container registration. The simplest thing would be to keep this as `MyApp::Settings`, because it would allow for a simple direct correspondence to a `"settings"` registration in the app container. It'd also allow us to keep it a `config/settings.rb` and have it match the other files there, like `application.rb`, which define same-named classes in the application's main namespace.

As for alternatives, if we stick with the standard mapping from class name to registration:

- `MyApp::ApplicationSettings` should probably reside in `config/application_settings.rb` and would result in a `"application_settings"` registration
- `MyApp::Application::Settings` should probably reside in `config/application/settings.rb` and would result in a `"application.settings"` registration

Both of these cases are quite awkward, especially if you consider the current behaviour of importing the application container into slice containers under the `"application.*"` namespace, which would lead to registrations like `"application.application_settings"`, or even worse, `"application.application.settings"`.

We'd also probably need to work out some special casing for the file name since neither `application_settings.rb` nor `application/settings.rb` feel particularly nice.

**Given all of this, I think I'd be inclined to leave it just as `Settings`. What do you think?** It feels like the better outcome on balance. We could perhaps create a new application `config.settings_class_name` setting as an escape valve for users who really need that name for their own usage inside their app, and want to name it something differently.

### Regarding types access

One of the wonderful things about this change to using an ordinary class is that it opens up so many options for us regarding access to types!

Given the feedback from @jodosha and @solnic on this point, I'd now be inclined _not_ to auto-generate any kind of types module inside the settings class. We can just allow users to access whatever types modules they need (and if need be, they can define their own local one by hand right there in the class too).

FWIW, the idea of defining a separate types module just for the settings was mine, and it arose from a real use case. As described in [the Trello card](https://trello.com/c/qExli5cy/105-make-a-types-module-automatically-available-if-possible-inside-the-hanamiapplicationsettings-block):

> I think there's a benefit in having these types isolated from the application's own types module (if any), since it ensures that the settings load in isolation, without any other application concerns creeping in (especially since that in some cases, some of those application concerns may rely on the settings, which creates a circular dependency and will cause the app to crash at boot)

I had an application where its types at one point depended on something from the app settings, which caused this circular dependency. Given that the types needed to load settings are usually very simple and specific to that use case, I felt like having a separate types module just for that case would ensure we avoid any problems with app boot.

However, now that it's straightforward to use either an outside types module or define a new one within your own settings class, these options should be more obvious to app authors and they'll be able to help themselves in the case of any boot issues.

**So let's not do anything special with types inside the app settings at all.** At most, perhaps we can generate some examples of using a types module so that users can make their own choice. Are you all OK with that?

timriley: @waiting-for-dev sounds like there's now consensus on the overall direction, so hopefully this will allow you to retool the approach here in this PR!

waiting-for-dev:
I think we can close it in favor of #1114

PR #1112, opened 2021-06-04
Give visibility to the endpoint namespace key

waiting-for-dev:
Instead of having it buried within the implementation of a private method, we give it the prominence it deserves by having it as a class constant.
    
We took the occasion to rename `slices_registry` to `slice_registry` to make the naming (singular) consistent with the new constant name.

This is not intended to be user-configurable and will not be marked as public API.

We're structuring this like so to make an important (but internal) detail of the resolver much clearer than just an incidental string deep inside a private method.

### DIFF
--- a/lib/hanami/application/routing/resolver.rb
+++ b/lib/hanami/application/routing/resolver.rb
@@ -7,6 +7,8 @@ module Hanami
       #
       # @since 2.0.0
       class Resolver
+        ENDPOINT_KEY_NAMESPACE = "actions"
+
         require_relative "resolver/trie"
 
         # @since 2.0.0
@@ -21,7 +23,7 @@ module Hanami
         def initialize(slices:, inflector:)
           @slices = slices
           @inflector = inflector
-          @slices_registry = Trie.new
+          @slice_registry = Trie.new
         end
 
         # @api private
@@ -50,7 +52,7 @@ module Hanami
         # @api private
         # @since 2.0.0
         def register_slice_at_path(name, path)
-          slices_registry.add(path, name)
+          slice_registry.add(path, name)
         end
 
         private
@@ -65,16 +67,16 @@ module Hanami
 
         # @api private
         # @since 2.0.0
-        attr_reader :slices_registry
+        attr_reader :slice_registry
 
         # @api private
         # @since 2.0.0
         def resolve_string_identifier(path, identifier)
-          slice_name = slices_registry.find(path) or raise "missing slice for #{path.inspect} (#{identifier.inspect})"
+          slice_name = slice_registry.find(path) or raise "missing slice for #{path.inspect} (#{identifier.inspect})"
           slice = slices[slice_name]
-          action_key = "actions.#{identifier}"
+          endpoint_key = "#{ENDPOINT_KEY_NAMESPACE}.#{identifier}"
 
-          slice[action_key]
+          slice[endpoint_key]
         end
       end
     end


waiting-for-dev:
> @waiting-for-dev why this feature and why into the `router` namespace? Thanks

Hey @jodosha , yeah, sorry not being more clear :slightly_smiling_face: 

This feature is discussed here: https://trello.com/c/AUOLglyi/54-make-actions-container-identifier-prefix-configurable-for-router-endpoint-resolution

I put it into the `router` namespace because it's used by the router's resolver, which is already configurable from within the `router` namespace.

jodosha:
@waiting-for-dev Can't that be passed as a configuration of the resolver itself?

waiting-for-dev:
> @waiting-for-dev Can't that be passed as a configuration of the resolver itself?

We talked about it with @timriley . He didn't feel like the users should need to know what a resolver is to change that. Eventually, we will provide the option for components to register settings, so the current approach will play well if we register from the resolver to the router.

lib/hanami/application.rb
```diff
@@ -339,7 +339,8 @@ def initialize(application = self.class)
 
         resolver = application.config.router.resolver.new(
           slices: application.slices,
-          inflector: application.inflector
+          inflector: application.inflector,
+          endpoints_namespace: application.config.router.endpoints_namespace
```
solnic, line 343: @waiting-for-dev wouldn't it make more sense to call this `action_namespace`? (or `actions_namespace` I never know if plural is better in such cases, @timriley?)

waiting-for-dev: Yeah, I always have existential thoughts about singular/plural naming :rofl: I went with plural because, in the same class, we have a `slices_registry` setting, so it's more consistent. But we can rename both. In general, I prefer singular because they are shorter.

About `actions_namespace` vs. `endpoints_namespace`, I felt that the latter is less opinionated. I mean, we can say that, by definition, that's an endpoint, as a router routes from something to a destination or point. On the other side, calling that an `action` is more of a convention. By I'm 100% fine with either way :slightly_smiling_face:

timriley: Regarding naming, one thing to consider would be whether there's anything else inside the framework in general that might want to know about our "actions namespace". Would we want to be able to configure this such that generators etc. respect it?

If that was the case, then I think it might make sense as a top-level `config.actions_namespace` settings.

However, if we don't have a clear case for that kind of usage, then I think nesting it inside `config.router` makes sense, and at that point, so does using the same terminology as the router itself, so that'd be `config.router.endpoints_namespace`.

Perhaps we should start with the latter, and then if we discover the need for something more framework-wide, we can consider moving things at that point.

Regarding singular/plural, I think singular _probably_ feels better? I'd be happy for @waiting-for-dev to do that and update the `slices_registry` to make it consistent :)

solnic: @timriley I'd already make it a top level setting, so `config.actions_namespace`, because it's not just a router concern. Generators will use this already.

waiting-for-dev: @solnic @timriley @jodosha finally I updated the PR:

- Rebased from `main`
- Made it a top level setting
- Singularized from `endpoints_namespace` => `endpoint_namespace`
- Renamed `slices_registry` => `slice_registry` for consistency

I also had to change `Hanami::Configuration.components_dir_path` to be an instance method for now, as it needs access to the new setting. My first attempt was also to add two new settings, `repository_namespace` and `view_namespace`, but I realized it makes no sense at this point. Unless I'm wrong, the view namespace is now controlled by `hanami/action`, and repositories are somehow more optional in the stack. Probably we should try to gain some consistency here :)

jodosha:
@timriley I'm gonna ask the question to you this time, as you originated the [feature](https://trello.com/c/AUOLglyi/54-make-actions-container-identifier-prefix-configurable-for-router-endpoint-resolution).

---

Why is it needed?

If we want to make this configurable **for internal purposes** (private API), I'm fine. As we would gain in code flexibility.
IF we want to make this for final users (public API), I believe this isn't necessary.

From Hanami 1 experience, people have zero interest to this level of flexibility.

---

One [spec](https://github.com/hanami/hanami/pull/1112/files#diff-0dd4c5c06883693e7f2b6f1aa8ae7baa9927d08520dd044527ee208d44eeae0dR35) of this PR is testing a scenario where an action is defined under: `slices/main/endpoints/home/index.rb`.

I believe we shouldn't support such cases for two reasons:

* Increased overhead in OSS support
* Increased complexity (or impossibility) to support that structure in CLI code generators

The return of investment is zero: is it a blocker for users being forced to put actions in `slices/main/actions` instead of `slices/main/endpoints`?

---

To summarize: we should be rigid in the app structure for certain directories, and liberal for other ones.

Where to put actions falls into the first category, how to organize `my_app/lib/` falls into the second.

timriley:
> Why is it needed?
> 
> If we want to make this configurable **for internal purposes** (private API), I'm fine. As we would gain in code flexibility. If we want to make this **for final users** (public API), I believe this isn't necessary.
> 
> 👉 From Hanami 1 experience, people have zero interest to this level of flexibility. 👈

Thanks for asking the question, @jodosha 😄 

The biggest motivator for me is to make this configurable for internal purposes. I'm alergic to seeing important details like this hard-coded deep inside some implementation file (in this case `Hanami::Application::Routing::Resolver`):

```ruby
action_key = "actions.#{identifier}"
```

That's kind of a crucial aspect of how the application's router should work, and this tucks it down somewhere way out of sight. By making it a config key, 


> 
> One [spec](https://github.com/hanami/hanami/pull/1112/files#diff-0dd4c5c06883693e7f2b6f1aa8ae7baa9927d08520dd044527ee208d44eeae0dR35) of this PR is testing a scenario where an action is defined under: `slices/main/endpoints/home/index.rb`.
> 
> I believe we shouldn't support such cases for two reasons:
> 
> * Increased overhead in OSS support
> * Increased complexity (or impossibility) to support that structure in CLI code generators
> 
> The return of investment is zero: is it a blocker for users being forced to put actions in `slices/main/actions` instead of `slices/main/endpoints`?
> 
> To summarize: we should be rigid in the app structure for certain directories, and liberal for other ones.
> 
> Where to put actions falls into the first category, how to organize `my_app/lib/` falls into the second.

jodosha:
@timriley If it's gonna be `@api private`, I'm totally fine. Let's remove the code example from the description, as we don't want final users to utilize it.

timriley:
@jodosha Updated the PR description :)

lib/hanami/configuration.rb
```diff
@@ -129,9 +129,13 @@ def logger_instance
...
     setting :slices, default: {}, constructor: :dup.to_proc
 
+    setting :endpoint_namespace, default: "actions"
+
     # TODO: turn this into a richer "source dirs" setting that can support enabling
     # of container component loading as an opt in behvior
-    setting :component_dir_paths, default: %w[actions repositories views]
+    def component_dir_paths
+      [endpoint_namespace, "repositories", "views"]
+    end
```
timriley, line 138: Ugh, it's always awkward when settings depend on each other, but I think this is fine for now. I need to work on this part of the config next anyway. I'll see what we can do.

jodosha:
@timriley @waiting-for-dev Can I suggest to rename it: `endpoint_namespace` => `action_namespace`.

I know it's technically an endpoint, but usually speaking we have actions as endpoints. This will make obvious the link between the concepts.

lib/hanami/configuration.rb
jodosha: Are we going thru premature DRY?

`endpoint_namespace` is a key prefix for a container.
`"actions"` here is a file system path fragment.

They happen to have the same value, and (given it's Ruby) we use strings for everything. Would we have done the same change if we had `Pathname` in this array=

The values are conceptually linked, but I would leave the raw string here, to have the implementation to refer to the same level of abstraction: an array of strings to represent FS fragments.

timriley: Yeah, I was actually just about to revisit this and say something similar. I think this should remain a hard-coded list of string paths in which to look for code to load.

_Especially_ given we don't have motivations for the `endpoint_namespace` to be a central config that drives other application behaviours.

timriley:
Given the challenge in finding the "right" name for this setting, maybe we could actually sidestep this whole topic for now and just take a 3rd approach: not having a setting provide the endpoint key namespace, and just doing something like this instead:

```ruby
module Hanami
  class Application
    module Routing
      class Resolver
        DEFAULT_ENDPOINT_KEY_NAMESPACE = "actions"

        def initialize(slices:, inflector:, endpoint_namespace: DEFAULT_ENDPOINT_KEY_NAMESPACE)
          # ...
```

On reflection, this probably feels like the more appropriate path given @jodosha's points earlier making it clear that we should not be "blessing" this level of configurability for end users.

What do you think?

timriley:
After sleeping on this, I still think this is a reasonable interim solution. I think eventually we'll want to bring the string `"actions"` out into some shared, central setting or constant that our various tools can all look at, so that we're not hard-coding `"actions"` (and others) across too many places (e.g. generators, router, etc.), but I think we can use that moment (i.e. when we build the _next_ thing that depends on that string) to decide where to put that string, rather than now, since we'll have more information available to guide our decisions.

jodosha:
@timriley Fine, up to you and @waiting-for-dev to wrap this up.
We're on the same page. That's what matters. 🌸

waiting-for-dev:
@jodosha @timriley I agree with the decision taken. I've updated the PR title/description and the code accordingly. Some considerations:

- As the only reason for this change is to give visibility to the string, we're not having it accepted as a parameter. There's no point in the application class to call the resolver and give it a different namespace key. So, after this change, this is just the string's extraction from the private method's body to a class constant.
- The constant name is not `DEFAULT_ENDPOINT_KEY_NAMESPACE` but `ENDPOINT_KEY_NAMESPACE`. The reason is that it's not a default, but the only one accepted at this point. Having the abstraction removed from the settings `component_dirs_path` means that if we only update the `ENDPOINT_KEY_NAMESPACE` to something else, it won't work. So, it's not a default; it has a single valid value.
- Because of that, I removed the integration test configuring the namespace, and because it's not behavior for end-users, I removed the integration test configuring the namespace.

timriley (approved):

timriley:
LGTM!

waiting-for-dev:
> Thanks for this, @waiting-for-dev! I'm sorry for the runaround here, but I think this has led to a really valuable discussion, so I want to thank you for your part in making that happen smile

Thank you both for your insights! I agree we took the best decision here, and I'm happy the discussion here helped to have a clearer vision of where we are and where we want to go :)

PR #1114, opened 2021-06-15
Use a class for the application settings

waiting-for-dev:
We switch from a block-based definition for settings to a regular Ruby    
  class so that users can leverage all the flexibility that it brings,    
  like defining custom methods or using whatever Types module they want.    
  We also get rid of custom configuration code and depend on dry-configurable.    
      
  By default, the settings class is expected to be defined in    
  config/settings.rb file, and its name should be MyApp::Settings,    
  where MyApp represents the application namespace.    
      
  The user settings class inherits from a Hanami::Application::Settings    
  class, a dry-configurable-enabled class that delegates any unknown    
  method to its Config instance so that users don't need to define    
  reader: true every single time.    
      
  We're also breaking apart a settings store abstraction, which was previously    
  coupled with the load process. By default, we use a store that leverages Dotenv to    
  load settings from .env files, or directly from ENV when not available.    
  The loading of settings happens when the instance is initialized.    
      
  We leverage the parent class Hanami::Application::Settings to    
  use it as a null object when users don't define their own class


lib/hanami/application.rb
```diff
@@ -151,21 +151,8 @@ def booted?
...
-          @_settings
-        else
-          # Load settings lazily so they can be used to configure the
-          # Hanami::Application subclass (before the application has inited)
-          load_settings
-          @_settings ||= nil
-        end
+      def settings
+        @_settings ||= load_settings
       end
```
timriley, line 156: This reduction in complexity is very nice!

lib/hanami/application/settings/dotenv_store.rb
```diff
@@ -0,0 +1,39 @@
...
+  class Application
+    class Settings
+      # Default application settings store.
+      #
+      # Uses dotenv (if available) to load .env files and then loads settings
+      # from ENV.
+      #
+      # @since 2.0.0
+      # @api private
+      class DotenvStore
```
timriley, line 13: I like this abstraction!

timriley (changes requested):
Looking good, @waiting-for-dev!

The last thing I'd like to see before we merge this is removing the idea of the "loader" and putting that logic into `Hanami::Application::Settings` itself.

Now that this class exists, and now we have the well-defied "store" abstraction, I don't see why we need to have the "loader" be a separate thing at all anymore. It doesn't feel like something that would need to be switched out (switching out the store is enough!), and with `Hanami::Application::Settings` now existing, it feels like the right place to put the logic for interfacing itself with the store.

What do you think about this? I did share some ideas about this with you in chat earlier; would you be up for giving this a go?

solnic:
@waiting-for-dev please update this branch on top of main because CI won't pass until you do so

lib/hanami/application.rb
```diff
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 require "dry/system/container"
+require "hanami/application/settings"
 require "hanami/configuration"
 require "pathname"
 require "rack"
 require_relative "slice"
 require_relative "application/autoloader/inflector_adapter"
-require_relative "application/settings"
```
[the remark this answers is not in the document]
waiting-for-dev: No particular reason besides that I tend to use `require_relative` only when `require` doesn't work :slightly_smiling_face: Do you prefer to switch it back?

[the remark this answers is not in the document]
waiting-for-dev: For me, I feel nicer if we don't expose that we are relying on the filesystem structure, even if it's the same under the hood.

lib/hanami/application.rb
```diff
@@ -322,9 +313,14 @@ def load_routes
       end
 
       def load_settings
-        prepare_base_load_path
```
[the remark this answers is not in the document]
waiting-for-dev: Readded!

lib/hanami/configuration.rb
```diff
@@ -143,25 +142,36 @@ def settings_path
       settings.fetch(:settings_path)
     end
 
+    def settings_class_name=(builder)
+      settings[:settings_class_name] = builder
+    end
+
+    def settings_class_name
+      settings.fetch(:settings_class_name, ->(namespace_name) { "#{namespace_name}::Settings" })
+    end
```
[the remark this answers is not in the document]
waiting-for-dev: Fair enough. Done!

lib/hanami/application/settings/dotenv_store.rb
```diff
@@ -0,0 +1,39 @@
...
+        def fetch(name, &block)
+          ENV.fetch(name.to_s.upcase) { block.call }
```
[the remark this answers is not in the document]
waiting-for-dev: Good point. Done!

waiting-for-dev:
> @waiting-for-dev please update this branch on top of main because CI won't pass until you do so

Done!

waiting-for-dev:
> The last thing I'd like to see before we merge this is removing the idea of the "loader" and putting that logic into Hanami::Application::Settings itself.

It's done now, @timriley!

I also updated `DotenvStore#fetch` to be fully compatible with `Hash#new` signature and added docs for the `Settings` class.

lib/hanami/application.rb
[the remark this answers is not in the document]
timriley: @waiting-for-dev I've taken to the pattern of using require_relative for requires to neighbouring files in the gem. It's a signal about which requires are external vs internal, and it's also a minor performance boost (see e.g. [this PR](https://github.com/rails/rails/pull/29638) where the entire Rails codebase was ported over.

In a repo like this one it's somewhat inconsistent since it's only in place for the areas I've worked on recently, but I hope to port everything over with time.

timriley (approved):
Lovely! Thank you @waiting-for-dev!

Left you one last note about switching the require back to a require_relative. After that's done this is good to merge from my perspective :)

As per usual, I'll yield to @jodosha for any comments he might like to make before we merge.

lib/hanami/application.rb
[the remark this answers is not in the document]
waiting-for-dev: So let's keep it consistent! Switched back! :slightly_smiling_face: 

In fact, it's not that I wanted to replace your `require_relative`, but I didn't notice it, and I added my `require` on top. Later I saw the `require_relative`, and I thought that I had added it myself before :sweat_smile:

waiting-for-dev:
Thanks, @timriley! I just switched the `require_relative` back (see my comment above).

I also fixed a flaky test. As settings are stored as a `Set` in `dry-configurable`, there's no warranty of the order in which they are being processed, so the expectation on the error message was sometimes failing because of the settings being processed in the reverse order. I have updated the regular expression to catch both scenarios. Even if it would be more user-friendly, for now, I don't think we need to complicate things and ensure the errors are ordered. WDYT?

lib/hanami/application/settings.rb
```diff
@@ -1,22 +1,94 @@
...
+    #
+    #   module MyApp
+    #     class Settings < Hanami::Application::Settings
+    #       setting :database_url
+    #       setting :feature_flag, default: false, constructor: Types::Params::Bool
+    #     end
+    #   end
+    #
+    # Settings are defined with
+    # [dry-configurable](https://dry-rb.org/gems/dry-configurable/), so you can
```
jodosha, line 28: @waiting-for-dev Should we link the exact version that we support?

lib/hanami/application/settings.rb
```diff
@@ -1,22 +1,94 @@
...
+    # take a look there to see the supported syntax.
+    # 
+    # Users work with an instance of this class made available within the
+    # `settings` key in the container. The instance gets its settings populated
+    # from a configurable store, which defaults to
+    # {Hanami::Application::Settings::DotenvStore}. It uses
+    # [dotenv](https://github.com/bkeepers/dotenv) under the hood. The usual
```
jodosha, line 35: I'd not explain which `.env` files are supported, but just link via YARD tag `@see Hanami::Application::Settings::DotenvStore`. As a reader, there I expect a fully explanation of the `.env` files, including their priority.

lib/hanami/application/settings/dotenv_store.rb
```diff
@@ -0,0 +1,46 @@
...
+      # @since 2.0.0
+      # @api private
+      class DotenvStore
+        Undefined = Dry::Core::Constants::Undefined
+
+        def fetch(name, default_value = Undefined, &block)
+          name = name.to_s.upcase
+          args = (default_value == Undefined) ? [name] : [name, default_value]
+
+          ENV.fetch(*args, &block)
```
jodosha, line 22: @waiting-for-dev It's fine for this PR, but we should not use `ENV` directly. This had problems with Hanami 1, that we should avoid with 2.0. Please have a look at this [article](https://lucaguidi.com/2016/12/27/isolate-global-state/).

Again, not for this PR, but could you follow up with another one fixing this problem? We could either reuse the same Hanami 1 approach, or use `dry-effects`. Thanks 😉

lib/hanami/application/settings/dotenv_store.rb
```diff
@@ -0,0 +1,46 @@
...
+require "dry/core/constants"
+
+module Hanami
+  class Application
+    class Settings
+      # Default application settings store.
+      #
+      # Uses dotenv (if available) to load .env files and then loads settings
+      # from ENV.
+      #
```
jodosha, line 12: We should document which files are loaded (according to the current env), and what's their priority.

spec/unit/hanami/application/settings/dotenv_store_spec.rb
```diff
@@ -0,0 +1,116 @@
+# frozen_string_literal: true
```
jodosha, line 1: @waiting-for-dev Given this spec messes with the global env (`ENV`), it should be an "isolation" spec. It should go into `spec/isolation`. 🙂

jodosha (changes requested):
@waiting-for-dev You nailed it! The feature looks great.

I rejected the approval for a couple of reasons:

  1. The build is failing. That is due to the fact that `unstable` branches were removed, so CI fails to fetch the deps. To fix it, just rebase from `main`, or fix the `Gemfile` of this PR. At the end you should reference `main` branches of other Hanami gems.
  2. The DotenvStore test, should be an isolation test, not a unit test, because it changes the global state.

lib/hanami/application/settings.rb
waiting-for-dev: I thought about it, but I reckoned that it would be something with many possibilities to get outdated at some point. Users can look at the gemspec to see the version we are using, which could be a range BTW, and it should be our responsibility to keep dependencies updated anyway. However, if you still prefer it, I'll add it here. Just confirm me your decision :slightly_smiling_face:

lib/hanami/application/settings.rb
waiting-for-dev: Yeah, I put everything here because `DotenvStore` is a kind of private API, but probably it makes more sense to do it in that way. I'll update it and let you know.

waiting-for-dev: Done @jodosha!

lib/hanami/application/settings/dotenv_store.rb
waiting-for-dev: Done!

lib/hanami/application/settings/dotenv_store.rb
waiting-for-dev: Hey @jodosha, I completely agree with you. I ended up doing it in this PR as the change is minimal. Please, take a look, and if you still prefer, I can revert it and do it on another one. I followed the approach in your article so that we don't add `dry-effects` as a dependency until the benefit is greater.

We still depend directly on `ENV` to decide with `.env` files are needed, so at this point, our injection is a bit artificial. We need to update how `Hanami.env` is fetched first, and that definitely belongs to another PR :slightly_smiling_face:

waiting-for-dev:
> The build is failing. That is due to the fact that unstable branches were removed, so CI fails to fetch the deps. To fix it, just rebase from main, or fix the Gemfile of this PR. At the end you should reference main branches of other Hanami gems.

Thanks for pointing out @jodosha! It's fixed now.

spec/unit/hanami/application/settings/dotenv_store_spec.rb
solnic: I would actually recommend porting this entire spec to `spec/isolation` and removing mocking completely. Dotenv should be treated as an implementation detail and there should be no need to mention it in the specs.

spec/unit/hanami/application/settings/dotenv_store_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+      context "HANAMI_ENV is 'test'" do
+        before do
+          @hanami_env = ENV["HANAMI_ENV"]
+          ENV["HANAMI_ENV"] = "test"
+        end
+
+        after do
+          ENV["HANAMI_ENV"] = @hanami_env
+        end
```
solnic, line 42: this should be an around block with ensure:

```ruby
        around do |example|
          @hanami_env = ENV["HANAMI_ENV"]
          ENV["HANAMI_ENV"] = "test"
          example.run
        ensure
          ENV["HANAMI_ENV"] = @hanami_env
        end
```

solnic (commented):
Great job on this one Marc ❤️ Let's improve the spec before merging though and let me know if you need some help with this.

spec/unit/hanami/application/settings/dotenv_store_spec.rb
timriley: Among other things, this file is unit testing the behaviour of of this class when dotenv is or isn't around. I think it's important to have that covered. It's not an implementation detail for sake of this class, it's actually the _whole purpose_ of the class.

spec/unit/hanami/application/settings/dotenv_store_spec.rb
waiting-for-dev: @solnic with the new implementation where the environment is injected that's no longer needed :slightly_smiling_face:

jodosha (approved):
@waiting-for-dev Thanks for making this to happen. 👏

spec/unit/hanami/application/settings_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      }.freeze
+
+      expect { config_class.new(store) }.to raise_error(
+        described_class::InvalidSettingsError,
+        /(database_url: nope to database).*(redis_url: nope to redis)/m,
+      )
+    end
+  end
+
+  it 'delegates unknown methods to config' do
```
solnic, line 46:
```suggestion
  it "delegates unknown methods to config" do
```

spec/unit/hanami/application/settings_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    config_class = Class.new(described_class) do
+      setting :foo, default: 'bar'
```
solnic, line 48:
```suggestion
      setting :foo, default: "bar"
```

spec/unit/hanami/application/settings_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+    store = {}.freeze
+
+    instance = config_class.new(store)
+
+    expect(instance.foo).to eq('bar')
```
solnic, line 54:
```suggestion
    expect(instance.foo).to eq("bar")
```

solnic (approved):
approving but I actually want to get back to env handling as a follow-up, let's get this merged first 💥

waiting-for-dev:
> approving but I actually want to get back to env handling as a follow-up, let's get this merged first

Thanks, @solnic. I squashed all the commits and updated the message :+1:

solnic:
@waiting-for-dev cool thanks, are you planning to merge this?

waiting-for-dev:
sure, @solnic, done! :rocket:

PR #1116, opened 2021-06-25
Use a class for the application routes

waiting-for-dev:
We move the definition of routes as a block provided to
`Hanami::Application.routes` to its own class
`Hanami::Application::Routes`.

Users need to create a subclass and define the routes through a block
applied to the `.routes` class method. As before, the hanami application
captures the context of the block and forwards it to
`Hanami::Application::Router` for its evaluation.

E.g.:

```ruby

require "hanami/application/routes"

module MyApp
  class Routes < Hanami::Application::Routes
    routes do
      slice :main, at: "/" do
        root to: "home.show"
      end
    end
  end
end
```

For consistency with settings, we move the routes path configuration to
the main `Hanami::Configuration` class.

### DIFF
--- a/lib/hanami/application/routes.rb
+++ b/lib/hanami/application/routes.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module Hanami
+  class Application
+    # Application routes
+    #
+    # Users are expected to inherit from this class to define their application
+    # routes.
+    #
+    # @example
+    #   # config/routes.rb
+    #   # frozen_string_literal: true
+    #
+    #   require "hanami/application/routes"
+    #
+    #   module MyApp
+    #     class Routes < Hanami::Application::Routes
+    #       define do
+    #         slice :main, at: "/" do
+    #           root to: "home.show"
+    #         end
+    #       end
+    #     end
+    #   end
+    #
+    #   See {Hanami::Application::Router} for the syntax allowed within the
+    #   `define` block.
+    #
+    # @see Hanami::Application::Router
+    # @since 2.0.0
+    class Routes
+      # Defines application routes
+      #
+      # @yield DSL syntax to define application routes executed in the context
+      # of {Hanami::Application::Router}
+      # @returns [Proc]
+      def self.define(&block)
+        @_routes = block
+      end
+
+      # @api private
+      def self.routes
+        @_routes || raise(<<~MSG)
+          Routes need to be defined before being able to fetch them. E.g.,
+            define do
+              slice :main, at: "/" do
+                root to: "home.show"
+              end
+            end
+        MSG
+      end
+    end
+  end
+end
--- a/spec/new_integration/web_app_spec.rb
+++ b/spec/new_integration/web_app_spec.rb
@@ -20,13 +20,17 @@ RSpec.describe "Hanami web app", :application_integration do
       RUBY
 
       write "config/routes.rb", <<~RUBY
-        Hanami.application.routes do
-          slice :main, at: "/" do
-            root to: "home#index"
-          end
+        module TestApp
+          class Routes < Hanami::Application::Routes
+            define do
+              slice :main, at: "/" do
+                root to: "home#index"
+              end
 
-          slice :admin, at: "/admin" do
-            get "/dashboard", to: "dashboard#show"
+              slice :admin, at: "/admin" do
+                get "/dashboard", to: "dashboard#show"
+              end
+            end
           end
         end
       RUBY
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -6,6 +6,7 @@ require "pathname"
 require "rack"
 require_relative "slice"
 require_relative "application/autoloader/inflector_adapter"
+require_relative "application/routes"
 require_relative "application/settings"
 
 module Hanami
@@ -155,15 +156,9 @@ module Hanami
         @_settings ||= load_settings
       end
 
-      def routes(&block)
+      def routes
         @_mutex.synchronize do
-          if block.nil?
-            raise "Hanami.application.routes not configured" unless defined?(@_routes)
-
-            @_routes
-          else
-            @_routes = block
-          end
+          @_routes ||= load_routes
         end
       end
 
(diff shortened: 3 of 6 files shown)


jodosha (approved):
@waiting-for-dev LGTM 👍

solnic (commented):
@waiting-for-dev could we have the DSL at the class level so that there's no need to use `routes` block within the class?

waiting-for-dev:
> @waiting-for-dev could we have the DSL at the class level so that there's no need to use `routes` block within the class?

I agree that it would look nicer. In that case, probably the simplest solution involves letting the user inherit from `Hanami::Application::Router` and moving those `scope`, `slice`, etc., methods to the class level. So the capture of the block of routes and the router wouldn't be in separated classes anymore. Are we ok with that?

waiting-for-dev:
> @waiting-for-dev could we have the DSL at the class level so that there's no need to use `routes` block within the class?

@solnic, I've been looking into it, and it's not that easy. We are now just capturing the block and forwarding it to a `Hanami::Application::Router` instance, which executes it in its context. We need to provide a block to a method for that, as we can't just capture the class definition context.

As to using `Hanami::Application::Router` directly, it's not easy if we want users to interact with it at the class level (in contrast to an instance of it). Take the `scope` method, [it ends up calling `super`](https://github.com/hanami/hanami/blob/main/lib/hanami/application/router.rb#L38), which belongs to `hanami/router`, so we can't just move those methods at the class level. A workaround could redefine those methods (`scope`, `slice`, etc.) at the class level to build an AST so that later they are executed in the same order at the instance level. We must be sure that it's worth the effort, though. Thoughts?

solnic:
@waiting-for-dev isn't it possible to create a router instance at a class level and forward DSL methods to it?

waiting-for-dev:
> @waiting-for-dev isn't it possible to create a router instance at a class level and forward DSL methods to it?

Yup, you're right. We can do it. Please, take a look at the last commit.

However, I feel a bit uneasy moving such a thing to the global state. WDYT?

If you think it's a better idea, I'll polish everything and add tests to it.

lib/hanami/application/routes.rb
```diff
@@ -0,0 +1,61 @@
...
+    class Routes
+      def self.router
+        @router ||= Router.new(
+          resolver: Hanami.application.config.router.resolver.new(
+            slices: Hanami.application.slices,
+            inflector: Hanami.application.inflector
+          ),
+          **Hanami.application.configuration.router.options,
+        ) do
+          use Hanami.application[:rack_monitor]
```
jodosha, line 40: @waiting-for-dev Please remove those middleware from here and add them to the Rack middleware stack of the application.

In depth explanation. We can have three nested layers of Rack middleware stacks. Here's them listed from the outermost to the innermost:

* `config.ru`
* Application
* Router

The less stack we setup the better, for the app performance. When adding default middleware to the stack, we should target Application.

lib/hanami/application/routes.rb
```diff
@@ -0,0 +1,61 @@
...
+    #   end
+    #
+    #   See {Hanami::Application::Router} for the syntax allowed at the class
+    #   level.
+    #
+    # @see Hanami::Application::Router
+    # @since 2.0.0
+    class Routes
+      def self.router
+        @router ||= Router.new(
```
jodosha, line 33: @waiting-for-dev `||=` isn't thread-safe, is this code executed within a context that guarantees thread-safety?

This configuration heavily depends on global state of `Hanami.application`. How do we ensure that all these components are ready when we're instantiating the router? Shouldn't this be the role of the caller instead?

jodosha (changes requested):
@waiting-for-dev The code has been changed since my last review. 🙂 

I'm concerned that we're complicating the loading process to save us from the ugly `routes` block.

---

Remember that we're doing this change not because it's needed but because of syntax consistency with application settings.

It's fine to have consistency, but if that has to come at the cost of complexity, we should stop and reflect.

I agree that `routes` is ugly, but better having it, than complicating our lives. 😃 

/cc @solnic

waiting-for-dev:
> @waiting-for-dev The code has been changed since my last review.

Yeah, @jodosha, the last commit is more like a POC so that we can see how it'd look if we take it that way. I agree with your concerns, so I think that something like a `routes` method (we could rename it) is the less worst option here.

solnic:
Can we then rename it to `define`? The combination of `Routes` class name and then `routes` block method name is not 🧊 🙂

waiting-for-dev:
> Can we then rename it to `define`? The combination of `Routes` class name and then `routes` block method name is not

Hey @solnic, I agree with you. It's done now. I'm still leaving the `.routes` method for the reading operation, though, so we no call it `.define` :upside_down_face: 

@jodosha, could you please review it again? @timriley I'd also like to have your eyes on it :slightly_smiling_face:

lib/hanami/application/routes.rb
waiting-for-dev: Not a concern in the new implementation

lib/hanami/application/routes.rb
waiting-for-dev: I think it's food for another PR :slightly_smiling_face:

spec/new_integration/web_app_spec.rb
```diff
@@ -20,13 +20,17 @@ class Application < Hanami::Application
...
+          class Routes < Hanami::Application::Routes
+            define do
+              slice :main, at: "/" do
+                root to: "home#index"
+              end
 
-          slice :admin, at: "/admin" do
-            get "/dashboard", to: "dashboard#show"
+              slice :admin, at: "/admin" do
+                get "/dashboard", to: "dashboard#show"
```
jodosha, line 31: @waiting-for-dev How this could work if the new separator is a `.` instead of the old `#`? 😅 
Old: `"dashboard#show"`, new `"dashboard.show"`.

waiting-for-dev: That's something that hasn't changed with this PR. Here we're using the same routes definition example as before. The only difference is that they're defined in a class. Of course, we must study it, but it's not related to this PR :slightly_smiling_face:

jodosha (approved):
@waiting-for-dev Thank you very much! LGTM 👍

spec/new_integration/web_app_spec.rb
timriley: The separators `#` and `/` are accepted here (as well as the standard `.`). The built-in application endpoint resolver converts them back to dots as it resolves the object from the container: https://github.com/hanami/hanami/blob/f93c4d2618171c6527cd18d1d394b61ba3f87d9c/lib/hanami/application/routing/resolver.rb#L75

We can probably tweak that a bit, but I did this purposefully to preserve existing behaviour, and I figured the `#` actually helped provide a little more meaning to the endpoint names.

Happy to discuss elsewhere :)

timriley (approved):
Looks great! Can't wait to see what we can do with this object now that we have an actual class backing it.

PR #1117, opened 2021-06-25
Fix warning on Hash#fetch

waiting-for-dev:
Calling a block with a default argument and a block warns with:

> warning: block supersedes default value argument

There's no need to test Ruby default behavior.

The warning was introduced in #1114


jodosha (approved):
👍

PR #1118, opened 2021-07-04
Introduce `Hanami.shutdown`

jodosha:
## Proposal

It's a single entry point to be used to shutdown an application. The ideal context is Puma's `before_fork` block.

## Problem

Right now all the app components that need to be shutdown must be manually referenced to trigger custom shutdown logic.

This is inconvenient and it leaks details of the single components outside of their definition.

## Solution

Given `dry-system` supports the definition of the shutdown logic via `stop` block, we can suggest users to utilize this API.

`dry-system` also exposes a private API to iterate all the registered components and trigger their shutdown logic (`stop` blocks).

The idea is to expose a new public API for Hanami which can safely, conveniently shutdown an application: `Hanami.shutdown`. This new method will hook into the `dry-system` mechanism described above.

## Example

Imagine to have a component that needs to be cleaned up when the application shuts down. In our example, we want to close the connection with Redis database.

```ruby
# config/boot/redis.rb

Hanami.application.register_bootable :redis do
  init do
    require "redis"
  end

  start do
    register(:redis, Redis.new)
  end

  stop do |container|
    container[:redis].close
  end
end
```

Puma server has a set of hooks to allow application developers to plug in their own cleanup logic:

```ruby
# config/puma.rb

## BEFORE THIS FEATURE
before_fork do
  Hanami.application[:redis].close
  # repeat for each component that needs to be shutdown
end

## AFTER THIS FEATURE
before_fork do
  Hanami.shutdown
end
```

---

https://trello.com/c/vJ4AQMlV
https://discourse.hanamirb.org/t/hanami-v2-0-0-alpha2-app-template-feedback/606/37?u=jodosha

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -73,6 +73,10 @@ module Hanami
     end
   end
 
+  def self.shutdown
+    application.shutdown
+  end
+
   def self.bundler_groups
     [:plugins]
   end
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -151,6 +151,10 @@ module Hanami
         @booted
       end
 
+      def shutdown
+        container.shutdown!
+      end
+
       def settings
         @_settings ||= load_settings
       end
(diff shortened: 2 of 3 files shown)


timriley (approved):
Lovely!

PR #1119, opened 2021-07-05
Make a routes helper class available in the container

waiting-for-dev:
We add a new `Hanami::Application::RoutesHelper` class which delegates
`path` & `url` methods into the router's URL helpers. The intention is
for it to be made available to whichever view system in use (i.e.,
dry-view).

We have slightly rearranged the boot process. Instead of explicitly
creating the router when the application instance is initialized, we add
it to the `boot` phase of the application (which is effectively the
same). This way, we can register the helpers in the container just
afterward, still on the boot phase, when the container is finalized.
There's no need to load the routes beforehand as the router will already
do that.

### DIFF
--- a/lib/hanami/application/routes_helper.rb
+++ b/lib/hanami/application/routes_helper.rb
@@ -9,7 +9,7 @@ module Hanami
     # it to get the route helpers for your application.
     #
     # @example
-    #   MyApp::Aplication["routes_helper"].path(:root) # => "/"
+    #   MyApp::Application["routes_helper"].path(:root) # => "/"
     #
     # @see Hanami::Router::UrlHelpers
     # @since 2.0.0
--- a/lib/hanami/application/routes_helper.rb
+++ b/lib/hanami/application/routes_helper.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Hanami
+  class Application
+    # Hanami application routes helpers
+    #
+    # An instance of this class gets registered in the container
+    # (`routes_helper` key) once the Hanami application is booted. You can use
+    # it to get the route helpers for your application.
+    #
+    # @example
+    #   MyApp::Aplication["routes_helper"].path(:root) # => "/"
+    #
+    # @see Hanami::Router::UrlHelpers
+    # @since 2.0.0
+    class RoutesHelper
+      # @since 2.0.0
+      # @api private
+      def initialize(router)
+        @router = router
+      end
+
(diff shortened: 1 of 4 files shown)


timriley:
Thanks for your continued work here, @waiting-for-dev!

I'd be good with this as a first step forward, but I think we'd do well to further experiment with options here.

For example: it'd be good if we could make it so that application authors could define their own additional route helper methods somewhere. There's no place for that in this current structure (yet).

waiting-for-dev:
I rebased from master and set the PR as ready for review. Looking forward to reading your thoughts on this!

lib/hanami/application/routes_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+module Hanami
+  class Application
+    # Hanami application routes helpers
+    #
+    # An instance of this class gets registered in the container
+    # (`routes_helper` key) once the Hanami application is booted. You can use
+    # it to get the route helpers for your application.
+    #
+    # @example
+    #   MyApp::Aplication["routes_helper"].path(:root) # => "/"
```
jodosha, line 12:
```suggestion
    #   MyApp::Application["routes_helper"].path(:root) # => "/"
```

jodosha (commented):
@waiting-for-dev Thanks! This LGTM. 👍 

Could you please on a subsequent PR to make `MyApp::Aplication["routes_helper"]` available as `routes` in actions, views, and templates? Thanks! 🙂

waiting-for-dev:
> Could you please on a subsequent PR to make MyApp::Aplication["routes_helper"] available as routes in actions, views, and templates?

Sure! I guess that should be done in the app template, right?

jodosha:
@waiting-for-dev Nope, in the framework code.

For instance, one way to get those helpers in the actions, it to inject this new dependency. It can be either `#initialize` or via `include Deps`.

Also, for views/templates, we have a default view scope/context that can automatically include those routes.

Pseudo code:

In framework:
```ruby
class Hanami::View::ApplicationContext
  include Deps["routes_helper"]
end
```

In application:

```ruby
class MyApp::Views::ApplicationContext < Hanami::View::ApplicationContext
end
```

By inheriting from `Hanami::View::ApplicationContext`, we get the routes.
Please talk to @timriley about which `hanami-view` concept to pick: scope or context.

jodosha (approved):

waiting-for-dev:
Oh, I see. I think those parts are extracted into `hanami/view` & `hanami/controller`, though.

PR #1122, opened 2021-07-16
Fix dry-inflector to 0.2.1 or greater

waiting-for-dev:
It allows us to detect `CSRF` acronym on
`Hanami::Action::CSRFProtection`, which is needed in the application
template.

See hanami/hanami-2-application-template#33


timriley (approved):
Thanks!

PR #1131, opened 2021-11-11
Content Security Policy

jodosha:
## Feature

Handle Content Security Policy (CSP) defaults.
CSP defaults are handled by `hanami-controller` gem, but it needs to take into account `hanami-assets` settings (the Webpack server URL).

## Implementation details

This PR introduces code that will be moved later on in `hanami-assets` gem.
For the sake of not porting now that gem into 2.0 ready state, I preferred to put this new code here first.

Application configuration now exposes `assets`, which is an instance of `Hanami::Assets::ApplicationConfiguration`.
This follow the same pattern already established with actions and views configurations.

### Content Security Policy

```ruby
module MyApp
    config.actions.content_security_policy
  end
end
```

For a list of examples, please refer to: https://github.com/hanami/controller/pull/353

### Assets Configuration

```ruby
module MyApp
    config.assets
  end
end
```

For now it only exposes `server_url`.

## FIXME

`server_url` for now it's hardcoded (`http://localhost:8080`).

When we will work on `hanami-assets` we need to take into account:

* The default value
* A custom value that a user wants to assign
* Pass exactly the same URL to Webpack

### DIFF
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -28,7 +28,7 @@ module Hanami
     attr_reader :actions
     attr_reader :middleware
     attr_reader :router
-    attr_reader :views
+    attr_reader :views, :assets
 
     attr_reader :environments
     private :environments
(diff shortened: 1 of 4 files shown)


mereghost (approved):

lib/hanami/configuration.rb
```diff
@@ -42,12 +42,22 @@ def initialize(env:)
...
+        NullConfiguration.new
+      end
+
       # Config for actions (same for views, below) may not be available if the gem isn't
       # loaded; fall back to a null config object if it's missing
       @actions = begin
         require_path = "hanami/action/application_configuration"
         require require_path
-        Hanami::Action::ApplicationConfiguration.new
+        Hanami::Action::ApplicationConfiguration.new(assets_server_url: assets.server_url)
```
timriley, line 60: @jodosha I think passing the `server_url` value direct here in `#initialize` will be problematic if a user (either manually or via some plugin eventually) ever _changes_ the value of this `server_url` setting away from its default. In that case, the `Action::AppConfiguration` will have a _stale value_ for `server_url`.

I think we'll need to support this as part of `actions.finalize!` (at which point the actions config will know the `server_url` shouldn't change again) or find some other approach.

Are you OK to have a look at this?

jodosha: @timriley Will do, thanks for the good catch. 👍

timriley: I suspect there'll be other externally-provided config objects that end up with a similar requirement. We may want to consider updating those `finalize!` methods to receive `self` or something like that...

PR #1133, opened 2021-12-01
Add source_dirs configuration

timriley:
This PR improves upon the temporary `component_dir_paths` setting that was introduced in #1123 with the revamp of our component and source directory structure.

Here we replace `component_dir_paths` with a unified `source_dirs` config, which itself contains both `component_dirs` and `autoload_paths`:

- `component_dirs` is an instance of `Dry::System::Config::ComponentDirs`, which in conjunction with https://github.com/dry-rb/dry-system/pull/195 gives the user full access to the component dir config, as well as the ability to remove default dirs and configure their own entirely separately.
- `autoload_paths` is a simple array of paths

During the Hanami application setup, the component dirs are used to configure the slice containers as well as the autoloader, and the autoload paths are used to configure the autoloader only (those directories are intended to hold source files that do not need to become container components).

See `spec/unit/hanami/configuration/source_dirs_spec.rb` for how a user may interact with both `component_dirs` and `autoload_paths` to customise their config.

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -112,29 +112,30 @@ module Hanami
           config.bootable_dirs = ["config/boot"]
 
           # Add component dirs for each configured component path
-          application.configuration.source_dirs.component_dirs.each do |common_component_dir|
-            next unless root.join(common_component_dir.path).directory?
+          application.configuration.source_dirs.component_dirs.each do |component_dir|
+            next unless root.join(component_dir.path).directory?
 
-            component_dir = common_component_dir.dup
+            component_dir = component_dir.dup
 
             # TODO: this `== "lib"` check should be codified into a method somewhere
             if component_dir.path == "lib"
-              # Expect component files in the root of the lib
-              # component dir to define classes inside the slice's namespace.
+              # Expect component files in the root of the lib/ component dir to define
+              # classes inside the slice's namespace.
               #
-              # e.g. "lib/foo.rb" should define SliceNamespace::Foo, and will be
-              # registered as "foo"
-              component_dir.namespaces.root(key: nil, const: namespace_path)
-
-              application.autoloader.push_dir(root.join("lib"), namespace: namespace)
+              # e.g. "lib/foo.rb" should define SliceNamespace::Foo, to be registered as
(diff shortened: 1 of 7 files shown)


timriley:
Ah, I realise there's also a bit gap here we need to address: right now there's no way for the "lib" component dir to be configured by the user. We should probably bring that into the list of default component dirs so the user can apply their own config there too.

timriley:
I'm not fully sold yet on reusing dry-system's `Config::ComponentDirs`like I've done so far, though I do think it does bring in a lot of benefits, especially since we're mapping these directly to component dirs on each slice's container anyway. Open to feedback on this front.

timriley:
Pushed up some changes here. For the latter two items on the TODO list, i.e.

> - Allow users to provide additional configuration to particular dirs within the defaults (e.g. a user wants to set a custom `auto_register` policy on the `"repositories"` component dir)
> - Allow users to _remove_ entries from the default component dirs (e.g. user doesn't want `"repositories"` to be included at all; they'll be providing something different)

I think we can satisfy these at the dry-system level by making it's `Config::ComponentDirs` API more complete. See the work ongoing at https://github.com/dry-rb/dry-system/pull/195 for detail.

lib/hanami/slice.rb
```diff
@@ -111,47 +111,69 @@ def define_container
...
+              inflector.constantize(inflector.camelize(dir_namespace_path))
+            rescue NameError
+              namespace.const_set(inflector.camelize(autoload_path), Module.new)
+            end
+
+            application.autoloader.push_dir(
+              container.root.join(autoload_path),
+              namespace: autoloader_namespace
+            )
+          end
```
timriley, line 176: The amount of setup code we have here is getting pretty out of hand, and I'd like to find ways to better structure it, but I'd like to that as a separate task in the future, so for simplicity's sake I've just added this extra stanza at the bottom of all the existing setup code.

solnic (approved):

solnic:
@timriley looks like specs are flaky?

timriley:
@solnic you're right. I'll check it out, thanks.

jodosha (approved):

PR #1135, opened 2021-12-19
Initialize router lazily

timriley:
This change makes it possible to run a non-booted Hanami application as a Rack app. It does this by making the application router lazy loadable — it will be set up and memoized on first access.

This makes the rack app setup consistent with the rest of the Hanami 2 application setup principles: that it should be possible to access any part of a non-booted application and have the minimal set of required components lazily loaded for you.

After this change, it should now be possible to deploy a large Hanami 2 application to e.g. one or more serverless functions (e.g. AWS Lambda) in non-booted mode, and have only the subset of the application load for whichever routes are exposed to those functions. In this way, Hanami 2 better fulfils its promise of an application that can scale all the way _down_ as well as up.

The key user-facing changes here are:

- `Hanami::Application.rack_app` sets up and memoizes the rack application, and is also available as `Hanami.rack_app`
- `Hanami.boot` no longer calls `Hanami::Application.new`, which was the previous place in which the router was set up. Then, given we're no longer having to work with both a global application class as well a single global application _instance_, we no longer need `Hanami.app` as a way to access the latter.
- Finally, because the router is no longer a mandatory part of the application boot process (`Hanami::Application.boot` does not set it up; it is only accessed on demand via `Hanami::Application.rack_app`), we no longer need the `web_app:` keyword param for `Hanami::Application.boot`)

The last change above is another nice outcome, because it takes away the one place we had to add special-casing for non-web applications, and in turn better fulfilling one of our other promises for Hanami 2: that it should work well for all kinds of applications, not just web apps. With this change, users can now remove "hamami-router" from their `Gemfile` and everything will continue to work just fine, so long as they don't call `Hanami::Application.rack_app`.

I could also see this opening up the possibility in the future of hanami-router shipping its own application integration hooks that add the rack app, and having all of that code move out of the core gem entirely.


timriley:
[Initialize router lazily](https://trello.com/c/ly7ex0d4/169-initialize-router-lazily)

solnic:
@timriley what's preventing us from using system's booting and lazy-loading for accessing the router, ie `Hanami.app[:router]`?

timriley:
@solnic Good question :) In my mind, it doesn't belong in the container because it is acting as an entrypoint to the _overall application_, via the containers belonging to each of the slices, which we pass to the router's endpoint resolver.

In this way, it feels like something that wraps the container rather than the other way around.

Playing it through, if we put it in the core application container, then right now it would also end up being imported into each of the slice containers, meaning there is then an object in each slice container that knows about every other slice's containers too, which doesn't feel right.

I do think we want some router related concerns put into the container, like an object that provides methods for generating URLs, etc. I think this can be separate to the router itself, i.e. something that doesn't carry those references to all the containers. We've already done some work in this direction, which I hope we can further refine with time (e.g. I'd like to make the route definitions themselves a separate abstraction rather than being something coupled to the router).

Does this seem reasonable to you?

solnic:
@timriley yes this makes sense and it got me thinking now - we're missing an abstraction! It looks like the router is "a front-end" to the application. It would be good to come up with an abstraction for this eventually. Then we could unify things like a CLI front-end, HTTP front-end etc. There are some common concerns here, especially when it comes to handling "env" value, loading correct application settings and so on.

timriley:
@solnic Absolutely! Some extra attention on that area could definitely be helpful 😄

jodosha (approved):
@timriley Thank you for this change. 🙏

PR #1137, opened 2022-01-05
Logger defaults

jodosha:
# Feature

Apply Zero-Config ™️ strategy to logger configuration.

Print the application name in the log prefix.

```
[my_app] [INFO] [2022-01-05 16:33:27 +0100] ...
```

### Development

* level: debug
* stream: stdout
* format: single line
* color: yes

### Test

* level: debug
* stream: `log/test.log`
* format: single line
* color: no

### Production

* level: info
* stream: stdout
* format: json
* color: no

### Brand new app

```ruby
# config/application.rb

require "hanami"

module MyApp
  end
end
```

### Assign custom settings

```ruby
# config/application.rb

require "hanami"

module MyApp
    config.logger.level = :info

    config.logger.stream = $stdout
    config.logger.stream = "/path/to/file"
    config.logger.stream = StringIO.new

    config.logger.format = :json
    config.logger.format = MyCustomFormatter.new

    config.logger.color = false # disable coloring
    config.logger.color = MyCustomColorizer.new

    config.logger.filters << "secret" # add
    config.logger.filters += ["yet", "another"] # add
    config.logger.filters = ["foo"] # replace

    # See https://ruby-doc.org/stdlib/libdoc/logger/rdoc/Logger.html
    config.logger.options = ["daily"] # time based log rotation
    config.logger.options = [0, 1048576] # size based log rotation
  end
end
```

### Assign per env defaults

```ruby
# config/application.rb

require "hanami"

module MyApp
    config.environment(:staging) do
      config.logger.level = :info
    end
  end
end
```

### Assign custom logger

```ruby
# config/application.rb

require "hanami"

module MyApp
    config.logger = MyLogger.new
  end
end
```

# Breaking Changes (from 2.0.alpha5)

* `config.logger.options` used to be a `Hash` to hold all the settings, now it's an `Array` to hold Ruby `Logger#initialize` args.
* `config.logger.options[:stream]` was removed in favor of `config.logger.stream` etc..

---

Ref https://trello.com/c/OXxer9oU

### DIFF
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -36,11 +36,8 @@ module Hanami
     attr_reader :environments
     private :environments
 
-    attr_reader :application_name
-
     def initialize(application_name:, env:)
       @namespace = application_name.split(MODULE_DELIMITER)[0..-2].join(MODULE_DELIMITER)
-      @application_name = inflector.underscore(@namespace).to_sym
 
       @environments = DEFAULT_ENVIRONMENTS.clone
       config.env = env
@@ -50,7 +47,7 @@ module Hanami
       self.root = Dir.pwd
       self.settings_store = Application::Settings::DotenvStore.new.with_dotenv_loaded
 
-      config.logger = Configuration::Logger.new(env: env, application_name: self.application_name)
+      config.logger = Configuration::Logger.new(env: env, application_name: method(:application_name))
 
       @assets = begin
         require_path = "hanami/assets/application_configuration"
@@ -113,6 +110,10 @@ module Hanami
 
     def namespace
       inflector.constantize(@namespace)
+    end
+
+    def application_name
+      inflector.underscore(@namespace).to_sym
     end
 
     setting :env
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -137,7 +137,7 @@ module Hanami
     end
 
     def logger_instance
-      @logger_instance || logger.logger_class.new(**logger.options)
+      @logger_instance || logger.instance
     end
 
     setting :settings_path, default: File.join("config", "settings")
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -50,7 +50,7 @@ module Hanami
       self.root = Dir.pwd
       self.settings_store = Application::Settings::DotenvStore.new.with_dotenv_loaded
 
-      config.logger = Configuration::Logger.new(application_name: self.application_name)
+      config.logger = Configuration::Logger.new(env: env, application_name: self.application_name)
 
       @assets = begin
         require_path = "hanami/assets/application_configuration"
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -25,6 +25,9 @@ module Hanami
     DEFAULT_ENVIRONMENTS = Concurrent::Hash.new { |h, k| h[k] = Concurrent::Array.new }
     private_constant :DEFAULT_ENVIRONMENTS
 
+    MODULE_DELIMITER = "::"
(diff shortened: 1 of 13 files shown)


jodosha:
Converted to draft, as I need to add another feature.

timriley (approved):
Great to see this configuration expanded and formalised, @jodosha 😄 

I've left some notes about potential issues re: ordering of configs being applied; I'd be happy for us either to consider and address those here, or in a follow-up PR (provided we make a note about it on Trello!).

lib/hanami/configuration.rb
```diff
@@ -33,7 +36,12 @@ class Configuration
     attr_reader :environments
     private :environments
 
-    def initialize(env:)
+    attr_reader :application_name
+
+    def initialize(application_name:, env:)
```
timriley, line 41: Might be better to have this `application_name` param actually called `application_class_name` or something, since that's what it is at this point? i.e. it's not yet transformed into the actual concept of the "application name" that we otherwise expect throughout the system.

lib/hanami/configuration.rb
```diff
@@ -33,7 +36,12 @@ class Configuration
...
+      @namespace = application_name.split(MODULE_DELIMITER)[0..-2].join(MODULE_DELIMITER)
```
timriley, line 42: Thinking it through further, it actually feels kind of odd to have the determination of `namespace` as part of the `Configuration` class. This isn't something that's going to be configurable at all, and it should really be the responsibility of the application class itself to tell the rest of the system about its namespace. Do you reckon there's a way we can keep it back in `Application`?

It's kind of the same thing for the `application_name`, too, really. It's a value that's _useful_ for parts of the configuration to know, as we've seen, but it's not really the _job_ of the `Configuration` to determine the name in the first place (unless, of course, we want to make it a truly configurable setting...)

lib/hanami/configuration/logger.rb
```diff
@@ -13,14 +13,54 @@ class Logger
...
-      # Currently used for logging of Rack requests only.
-      #
-      # TODO: incorporate this into the standard logging some way or another
-      setting :filter_params, default: %w[_csrf password password_confirmation].freeze
+        config.level = case env
+                       when :production
+                         :info
+                       else
+                         :debug
+                       end
```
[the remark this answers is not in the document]
jodosha: After a quick Slack discussion with @timriley, we established that this isn't an issue.

We assume that once the Ruby process starts, the Hanami env won't change.

lib/hanami/configuration.rb
jodosha: @timriley Nope, this is already the `application_name`.

For a given app defined by `MyApp::Application`, we receive as argument here the underscored version of the namespace: `my_app`.

lib/hanami/configuration.rb
jodosha: @timriley I refactored this under the need of accessing the `application_name`, without thinking about the responsibilities between application and configuration.

To avoid cyclic dependencies between the two, I moved the namespace logic from application to configuration.

I consider namespace in the configuration, a **private utility for us**, NOT something that we need to configure.

Of course, if you have a suggestion, we can apply it. 🙂

lib/hanami/configuration.rb
```diff
@@ -33,7 +36,12 @@ class Configuration
...
+      @application_name = inflector.underscore(@namespace).to_sym
```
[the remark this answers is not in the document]
jodosha: @timriley Good point, we may need to resolve custom inflections before to use the inflector.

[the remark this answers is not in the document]
jodosha: Fixed by https://github.com/hanami/hanami/pull/1137/commits/0b25db7c4baaa4cd38f15e914aae3ef7710f6c6c

PR #1139, opened 2022-01-10
Simplify `Hanami::Web::RackLogger`

jodosha:
## Enhancements

Removed responsibilities from `Hanami::Web::RackLogger` to simplify its implementation and delegate those responsibilities to proper collaborator objects.

### Parsing Params

`Hanami::Web::RackLogger` should NOT be responsible of **parsing Rack env**.
This is is a Router responsibility. See https://github.com/hanami/router/pull/217

Logging params that aren't parsed by router can lead to bugs and confusion.

### Filtering Params

`Hanami::Web::RackLogger` should NOT be responsible of **filtering params** that contain sensitive data.
This is a logger responsibility of the logger (`Hanami::Loggger` for now, then `dry-logger`).

### Log Formatting

`Hanami::Web::RackLogger` should NOT be responsible of **formatting logs**.
This is a logger responsibility of the logger (`Hanami::Loggger` for now, then `dry-logger`).

## Technical notes

I removed HTTP version from the logs, because it's deprecated in Rack. https://github.com/rack/rack/pull/1691

### DIFF
--- a/spec/unit/hanami/web/rack_logger_spec.rb
+++ b/spec/unit/hanami/web/rack_logger_spec.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require "hanami/web/rack_logger"
+require "hanami/logger"
+require "stringio"
+require "rack/mock"
+
+RSpec.describe Hanami::Web::RackLogger do
+  subject { described_class.new(logger) }
+  let(:logger) { Hanami::Logger.new(application_name, stream: stream, level: Hanami::Logger::DEBUG, filter: filters) }
+  let(:stream) { StringIO.new }
+  let(:filters) { ["user.password"] }
+  let(:application_name) { "my_app" }
+
+  describe "#initialize" do
+    it "returns an instance of #{described_class}" do
+      expect(subject).to be_kind_of(described_class)
+    end
+  end
+
+  describe "#log_request" do
+    it "logs current request" do
+      path = "/users"
+      ip = "127.0.0.1"
+      status = 200
+      time = 0.0001
+      content_length = 23
+      verb = "POST"
+
+      env = Rack::MockRequest.env_for(path, method: verb)
+      env["Content-Length"] = content_length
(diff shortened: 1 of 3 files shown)


cllns (approved):

cllns (approved):
Couple quick spec suggestions/questions (meant to include them before)

spec/unit/hanami/web/rack_logger_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      params = {"user" => {"password" => "secret"}}
+      env["router.params"] = params
+
+      subject.log_request(env, status, time)
+
+      stream.rewind
+      actual = stream.read
+
+      expect(actual).to include(%([#{application_name}] [INFO] [#{time}] #{verb} #{status} #{ip} #{path} #{content_length} {"user"=>{"password"=>"[FILTERED]"}}))
```
cllns, line 42:
```suggestion
      expect(actual).to include(%([my_app] [INFO] [0.0001] POST 200 127.0.0.1 /users 23 {"user"=>{"password"=>"[FILTERED]"}}))
```
Thoughts on this? In my experience it's a little more robust, since renaming a variable above could make one of these into `nil`, and `#{nil}` is an empty string and we wouldn't know

spec/unit/hanami/web/rack_logger_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      it "logs full referenced relative path" do
+        env = Rack::MockRequest.env_for(path = "/users")
+        env["SCRIPT_NAME"] = script_name = "/v1"
+
+        subject.log_request(env, 200, 0.1)
+
+        stream.rewind
+        actual = stream.read
+
+        expect(actual).to include(script_name + path)
```
cllns, line 71:
```suggestion
        expect(actual).to include("/v1/users")
```
Same as above

timriley:
Fantastic change, Luca! 👏🏼

PR #1142, opened 2022-01-15
Don't declare inflector value as cloneable

timriley:
This is not a mutable object, since it has all its inflection rules received at initialization time only, so we can treat it as a non-cloneable value. This fixes the assumptions in our test cases that were no longer true with cloneable objects as of dry-configurable 0.14.0


timriley:
Tests are failing on 2.6 and 2.7 due to latest dry-configurable requiring a minimum Ruby version of 3.0.

Merging this anyway and will address the Ruby version in a separate PR.

PR #1146, opened 2022-02-05
Use dry-system zeitwerk plugin

timriley:
In this change we use dry-system's zeitwerk plugin, and allow the plugin to take care of most of the work of setting up Zeitwerk.

This removes a bunch of looping over component_dirs and manually pushing those dirs into Zeitwerk (along with dynamic module generation as required). The more we can keep this kind of code in one place only, the better we can look after it. And dry-system is the best place for it given it is the home for the component dir class and all its possible configuration and behaviour.

Along with this, we're updating the container setup so it no longer uses the block based `configure do |config|` approach, and instead use the explicit `container.configured!` event. This allows the container configuration code to be one level flatter, and in general easier to follow, since it is clearer every time we're altering container-specific config.

This change required a tweak in dry-system's zetiwerk plugin (done across https://github.com/dry-rb/dry-system/pull/222, https://github.com/dry-rb/dry-system/commit/13f8c8715701b9d52829913cc24eea04407245f4, and https://github.com/dry-rb/dry-system/pull/223), which have all been merged now.


lib/hanami/application.rb
```diff
@@ -222,54 +219,48 @@ def prepare_base_load_path
...
-        namespace.const_get :Container
-      rescue LoadError, NameError
-        namespace.const_set :Container, Class.new(Dry::System::Container)
-      end
+        container =
+          begin
+            require "#{application_name}/container"
+            namespace.const_get :Container
+          rescue LoadError, NameError
+            namespace.const_set :Container, Class.new(Dry::System::Container)
```
solnic, line 229: Could we use dry-core's class builder so that the anonymouse class is a bit nicer to work with in the IRB? ie:

```ruby
irb(main):010:0> Class.new(Dry::System::Container)
=> #<Class:0x00007faae3b7c7b8>
irb(main):011:0> Dry::Core::ClassBuilder.new(parent: Dry::System::Container, name: "MyContainer").()
=> MyContainer
```

timriley: Happy to - I plan to polish all the container setup for the alpha after this one :)

timriley: FWIW I think it already gets a nice name after it's assigned the the const, anyway (but I'll double check all of this later in this month).

PR #1150, opened 2022-02-14
Use concrete classes for slices, and various app/slice tidy-ups

timriley:
## Enhancements

- Introduce `Hanami::ApplicationLoadError` and `Hanami::SliceLoadError` exceptions to represent errors encountered during application and slice loading.

## Changes

- Slices are now represented as concrete classes (such as `Main::Slice`) inheriting from `Hanami::Slice`, as opposed to _instances_ of `Hanami::Slice`. You may create your own definitions for these slices in `config/slices/[slice_name].rb`, which you can then use for customising per-slice config and behavior, e.g.

    ```ruby
    # config/slices/main.rb:

    module Main
      class Slice < Hanami::Slice
        # slice config here
      end
    end
    ```
- Application-level `config.slice(slice_name, &block)` setting has been removed in favour of slice configuration within concrete slice class definitions
- You can configure your slice imports inside your slice classes, e.g.

    ```ruby
    # config/slices/main.rb:

    module Main
      class Slice < Hanami::Slice
        # Import all exported components from "search" slice
        import from: :search
      end
    end
    ```
- You can configure your slice exports inside your slice classes, e.g.

    ```ruby
    # config/slices/search.rb:

    module Search
      class Slice < Hanami::Slice
        # Export the "index_entity" component only
        export ["index_entity"]
      end
    end
    ```
- For advanced cases, you can configure your slice's container via a `prepare_container` block:

    ```ruby
    # config/slices/search.rb:

    module Search
      class Slice < Hanami::Slice
        prepare_container do |container|
          # `container` object is available here, with 
          # slice-specific configuration already applied
        end
      end
    end
    ```

---

## Implementation notes

Aside from the user-facing changes noted above, there's a few internal implementation details worth highlighting here:

- All responsibility for slice loading, class building, and registration has moved from `Application` to a new `SliceRegistrar`
- The `.prepare` methods in both `Application` and `Slice` are now roughly identical in structure, with all of their specific setup steps moved into individual methods called from a `.prepare_all` private method. This should make it clearer at a higher-level what exactly is being prepared, and could also pave the way for pluggable Application/Slice lifecycle hooks in a post-2.0 world.
- I removed the `config.slices_namespace` setting from `Application`. This existed in theory to allow slices to also live inside the application's own module namespace if a user desired (e.g. `MyApp::Main` instead of just `Main`), but I think this will never really be used, and the our slices always being representative of single top-level modules will be important for our documentation (and generators, and many other things I'm sure) to be clearer.
- I also removed the `config.slices_dir` setting from `Application`, for all the same reasons as above. Hanami applications should always load their slices from the same directory, this doesn't make sense to be configurable.
- I've introduced a `lib/hanami/constants.rb` for sharing common constants
- In both `Application` and `Slice` I've initialized a few extra things as early as possible in their respective `.inherited` hooks, such as the `autoloader` (for the application) and the `container` (for the application and slice), since it's helpful to have these objects available _before_ the user even runs `prepare`. In the case of the container, it's helpful because it allows us to call `import` on it for any `import` statements the user makes inside their own slice classes.

### DIFF
--- a/lib/hanami/application/slice_registrar.rb
+++ b/lib/hanami/application/slice_registrar.rb
@@ -18,8 +18,10 @@ module Hanami
       end
 
       def register(name, slice_class = nil, &block)
-        # TODO: real error class
-        raise "Slice +#{name}+ already registered" if slices.key?(name.to_sym)
+        if slices.key?(name.to_sym)
+          raise SliceLoadError, "Slice '#{name.to_s}' is already registered"
+        end
+
         # TODO: raise error unless name meets format (i.e. single level depth only)
 
         slices[name.to_sym] = slice_class || Slice.build_slice(name, &block)
@@ -27,7 +29,7 @@ module Hanami
 
       def [](name)
         slices.fetch(name) do
-          raise "Slice #{name} not found"
+          raise SliceLoadError, "Slice '#{name}' not found"
         end
       end
 
--- a/lib/hanami/application/slice_registrar.rb
+++ b/lib/hanami/application/slice_registrar.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "../constants"
+require_relative "../slice"
 
 module Hanami
   class Application
@@ -24,7 +25,7 @@ module Hanami
 
         # TODO: raise error unless name meets format (i.e. single level depth only)
 
-        slices[name.to_sym] = slice_class || Slice.build_slice(name, &block)
+        slices[name.to_sym] = slice_class || build_slice(name, &block)
       end
 
       def [](name)
@@ -83,6 +84,18 @@ module Hanami
           end
 
         register(slice_name, slice_class)
+      end
+
+      def build_slice(slice_name, &block)
+        slice_module =
+          begin
+            slice_module_name = inflector.camelize(slice_name.to_s)
+            inflector.constantize(slice_module_name)
+          rescue NameError
+            Object.const_set(inflector.camelize(slice_module_name), Module.new)
+          end
+
+        slice_module.const_set(:Slice, Class.new(Hanami::Slice, &block))
       end
 
       def root
--- a/lib/hanami/application/slice_registrar.rb
+++ b/lib/hanami/application/slice_registrar.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Hanami
+  class Application
+    # @api private
+    class SliceRegistrar
+      attr_reader :slices
+      private :slices
+
+      def initialize
+        @slices = {}
+      end
+
+      def register(name, slice_class = nil, &block)
+        # TODO: real error class
+        raise "Slice +#{name}+ already registered" if slices.key?(name.to_sym)
+        # TODO: raise error unless name meets format (i.e. single level depth only)
+
+        slices[name.to_sym] = slice_class || Slice.build_slice(name, &block)
+      end
+
+      def [](name)
+        slices.fetch(name) do
+          raise "Slice #{name} not found"
+        end
+      end
+
+      def each(&block)
+        slices.each_value(&block)
+      end
+
+      def to_a
+        slices.values
+      end
+    end
+  end
+end
--- a/lib/hanami/application/slice_registrar.rb
+++ b/lib/hanami/application/slice_registrar.rb
@@ -22,13 +22,7 @@ module Hanami
 
         # TODO: raise error unless name meets format (i.e. single level depth only)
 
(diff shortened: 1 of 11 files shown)


timriley:
[Make container setup a lot nicer](https://trello.com/c/HMWp4TlU/167-make-container-setup-a-lot-nicer)

timriley:
[Provide better centralized slice configuration](https://trello.com/c/hLjFHzyD/68-provide-better-centralized-slice-configuration)

lib/hanami/slice.rb
```diff
@@ -1,148 +1,201 @@
...
+        # Create a (to be configured later) container as early as possible, since code in
+        # the slice subclass may want to start referring to it right away (e.g. `.import`)
+        klass.instance_variable_set(:@container, Class.new(Dry::System::Container))
+      end
 
-    def prepare(provider_name = nil)
-      if provider_name
-        container.prepare(provider_name)
-        return self
+      def application
```
jodosha, line 24: @timriley Shouldn't this be injected as dependency?

jodosha (approved):
@timriley Wonderful work! 👏

lib/hanami/slice.rb
```diff
@@ -1,148 +1,201 @@
...
+          loader: application.autoloader,
+          run_setup: false,
+          eager_load: false
+        )
+      end
+
+      def prepare_container_base_config
+        container.config.name = slice_name
         container.config.root = root
         container.config.provider_dirs = ["config/providers"]
```
jodosha, line 176:
```suggestion
        container.config.provider_dirs = ::File.join("config", "providers")
```

timriley: Fixed, thanks :)

lib/hanami/slice.rb
timriley: @jodosha Ah yes, thanks for raising this. I hard-coded it while doing all this refactoring, as a matter of expediency, but it does make sense to allow this to be configured.

Since this is class-level state in the new slice structure, we can't "inject" the application in the typical sense, but I've have made it a class-level ivar and changed how we pass it through in https://github.com/hanami/hanami/pull/1150/commits/7eade29833fc2db90d579732457a462a98bde138:

1. It's set as a class-level ivar in the `.inherited` hook like the container
2. It can be explicitly reassigned via `.application(new_application)` if needed (this capacity will not be user-visible for now, but if we ever reach a future where we want to support multiple distinct Hanami applications in memory, this is how we can do it for slices)
3. The slice registrar uses the above method to assign the application when building slice classes dynamically
4. And finally, the slice registrar will raise an exception if it ever receives a slice that has a _different_ application assigned

Together I feel this makes for a pretty robust approach. Let me know what you think :)

timriley: For the record, I discussed this with @jodosha and this original question was based on a misunderstanding. I've just pushed https://github.com/hanami/hanami/pull/1150/commits/a8d67df0db441cfa73d8e8e498f630e0bdc76a8a which goes back to slices all having a fixed reference to the single `Hanami.application`.

This does keep a couple of structural changes from the first commit that I liked:

- Keeping most of the `Slice` behaviour in a `ClassMethods` module that is extended with `Slice` is inherited.
- Providing access to the application by assigning it to a class-level ivar when Slice is inherited (rather than having a method with a hard-coded reference). This at least makes it clearer that the `@application` is some externally provided state, which is a useful signal should we decide we want to do anything different with it in future.

lib/hanami/application/slice_registrar.rb
```diff
@@ -0,0 +1,107 @@
...
+        (slice_dirs + slice_configs).uniq.sort.each do |slice_name|
+          load_slice(slice_name)
+        end
+
+        self
+      end
+
+      def each(&block)
+        slices.each_value(&block)
+      end
```
vieteh, line 56: I'm upgrading to alpha7 for my project and now cannot run `hanami console`.
I found this method behaves differently, therefore, breaks hanami-cli. https://github.com/hanami/cli/blob/v2.0.0.alpha6.1/lib/hanami/console/plugins/slice_readers.rb#L15
I'm not sure how to best fix this. Can you help resolve this?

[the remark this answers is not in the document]
timriley: @tiev This is fixed and now released in hanami-cli 2.0.0.alpha7, thanks for the notice about this!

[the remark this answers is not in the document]
tiev: Thanks a lot! I should have said it was not urgent 😅

PR #1151, opened 2022-02-23
Actions/View structure

jodosha:
## Enhancements

  * Register View provider: `view.context` key to hold `[AppName]::View::Context` instance.

---

Ref https://github.com/hanami/view/pull/203
Ref https://github.com/hanami/hanami-2-application-template/pull/76

### DIFF
--- a/lib/hanami/application/container/providers/view.rb
+++ b/lib/hanami/application/container/providers/view.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+Hanami.application.register_provider :view do
+  prepare do
+    require "hanami-view"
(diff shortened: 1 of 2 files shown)


timriley (commented):
@jodosha I'm not sure if we even need this change at all...

This is registering the `"view.context"` component in the _application_ container, but since we expect all our concrete views to be in the slices, which have auto-registration enabled for their `lib/` dirs, into which we'll be generating `lib/view/context.rb`, everything should _just work_ without any special application-level provider at all.

So, could we get away with just closing this PR?

jodosha:
@timriley I just verified with https://github.com/hanami/hanami-2-application-template/pull/76, you're right!

PR #1153, opened 2022-03-01
Clarify `env` handling in configuration by removing writer method

timriley:
In actual practice, we expect env to be set once when the configuration is initially loaded (typically via the `HANAMI_ENV` env var), and not changed afterwards.

This change codifies this expectation by removing the `#env=` writer method from `Hanami::Configuration`.

---

## Changes

- The `env` provided to `Hanami::Configuration` (which is based on the value in the `HANAMI_ENV` env var) can no longer be changed after the the configuration is initialized, ensuring consistent behavior for any default configs that are dependent on the env, such as the `config.logger` defaults.

### DIFF
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -13,6 +13,7 @@ require_relative "configuration/middleware"
 require_relative "configuration/router"
 require_relative "configuration/sessions"
 require_relative "configuration/source_dirs"
+require_relative "constants"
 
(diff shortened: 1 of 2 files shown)


jodosha (approved):

timriley:
[Remove `env=` setter from Configuration](https://trello.com/c/KdKSHAXQ/171-remove-env-setter-from-configuration)

PR #1154, opened 2022-03-01
Shutdown slices as well as application

timriley:
This extends upon the work you did with `Application#shutdown`, @jodosha, and ensures that slices are shutdown at the same time (since they can and in many cases will have their own slice-specific providers).

Changeloggy bits:

---

## Additions

- `Hanami::Slice.shutdown` can be used to stop all the providers in a slice

## Changes

- `Hanami::Application.shutdown` will now also shutdown all registered slices

### DIFF
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -81,7 +81,9 @@ module Hanami
       end
 
       def shutdown
+        slices.each(&:shutdown)
         container.shutdown!
+        self
       end
 
       def prepared?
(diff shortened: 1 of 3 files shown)


timriley:
A thought I had while doing this: with the "prepare/boot/shutdown" naming we have now, I wonder if there's some merit in changing this to "prepare/start/stop", which'd provide symmetry at the application/slice level with the lifecycle we also have for our providers?

jodosha (approved):
@timriley Nice one!

timriley:
[Hanami.shutdown should also shutdown slice containers](https://trello.com/c/sXFNIecH/159-hanamishutdown-should-also-shutdown-slice-containers)

PR #1156, opened 2022-03-05
Centralize action & view integrations

timriley:
### Overview

This PR moves the application integration code from the hanami-controller and hanami-view gems into a central place within the hanami gem.

This change serves two purposes:

1. It lets us bring together code (and tests!) that have tended to change together over the last couple years of Hanami 2 development. From here, we can better manage any changes to the application core or these action/view integration, ensuring we provide a reliable experience for our users.
2. It allows us to create distinct base classes for actions and views providing this application integration (`Hanami::Application::Action` and `Hanami::Application::View`), allowing users to opt into that integration explicitly, rather than having it provided automatically (with no opt out) like we had when these existed as part of the base `Hanami::Action` and `Hanami::View` behavior.

This is a pragmatic move above all else — given we're a small team, we want to give ourselves the best chance at confidently and effectively moving towards the 2.0 release, and co-locating this code will help with this.

Longer term, we’ll want to provide a first-class integration story for third party components, and I believe we can lead the way in how we deliver that via our actions and views, but that’s now firmly a post-2.0 concern in my mind.

However, even within this change, I've started to feel my way a little closer to that goal, particularly through the introduction of the `Hanami::SliceConfigurable` module, which provides behaviour that could dropped into any class with configuration that may need to be derived from the current Hanami slice/application for it to integrate properly.

#### Added

- Introduced `Hanami::Application::Action` and `Hanami::Application::View` as base classes for actions and views that integrate with Hanami applications. Base action and view classes in Hanami applications should now inherit from these.

### Implementation notes

This is _mostly_ a mechanical movement of code from hanami-{controller,view} into this repo, but there's a few notable details:

- As described above, we're now using classes to provide the entrypoint to the integrated behaviour, so e.g. `Hanami::Action::ApplicationAction` has been ported to `Hanami::Application::Action`
- These new classes use the new `Hanami::SliceConfigurable` module to provide a hook for slice-specific configuration, once they are subclassed within a slice's namespace (this was previously built-in for any subclass of e.g. `Hanami::Action`)
- `SliceConfigurable` now only calls the `.configure_for_slice` hook _exactly once_ for each new slice-associated module that it encounters. So for a class hierarchy like `MyApp::Actions::Base < Hanami::ApplicationAction; Main::Action::Base < MyApp::Actions::Base; Main::Actions::Home::Show < Main::Action::Base`, `.configure_for_slice` will only be called for the first two classes, not for the third, since in the case of the third class, its still in the same slice as its superclass. This should make inheritance of these classes somewhat more efficient.
- I've introduced a private `#load_dependent_config` helper into `Hanami::Configuration` to make it so that any nested configuration objects (such as `config.actions` and `config.views`) are not loaded unless their parent gem is available.
- Renamed `Hanami::Application::Action#render?` private method to `#auto_render?` to better reflect its behavior
- I did some tidying up of `ViewNameInferrer`, improving on the parameter and method names a bit

See also the corresponding PRs in the other two gems:

- https://github.com/hanami/controller/pull/370
- https://github.com/hanami/view/pull/207


lib/hanami/application.rb
```diff
@@ -19,10 +19,11 @@ class Application
     class << self
       def inherited(klass)
         super
+
         @_mutex.synchronize do
           klass.class_eval do
             @_mutex = Mutex.new
-            @_configuration = Hanami::Configuration.new(application_name: name, env: Hanami.env)
+            @_configuration = Hanami::Configuration.new(application_name: klass.name, env: Hanami.env)
```
timriley, line 26: AFAICT this should've always been like this, since we want to pass in the name of the subclass, not just `Hanami::Application` itself.

lib/hanami/application.rb
```diff
@@ -48,6 +49,10 @@ def self.extended(klass)
         end
       end
 
+      def application
+        self
+      end
```
timriley, line 54: After I'm done with this PR, one of the next things I want to look at is making it so the application is just a specialised form of slice. This is a tiny step towards this, because it means we can allow `SliceConfigurable` and it users just to expect a `slice` to be given, not "a `slice` or an `application`", which is what led me to use the ill-fitting term "provider" to represent this union in previous iterations of this code.

lib/hanami/application/view/context.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        # @see SliceConfiguredContext#define_new
+        def initialize(**kwargs)
+          defaults = {content: {}}
+
+          super(**kwargs, **defaults)
+        end
+
+        def inflector
+          _options.fetch(:inflector)
```
jodosha, line 32: @timriley Where `_options` are defined?

timriley: @jodosha It's from `Hanami::View::Context` itself: https://github.com/hanami/view/blob/ec90f452facf349d15b536ea8b8f1c46cfd1c3a1/lib/hanami/view/context.rb#L57-L60

The reason why we have to capture the options like this is to allow the context to be reinitialized for each render_env. This is something that I'd like to address before hanami-view 2.0: it should be possible to keep a singular context object across the whole rendering life cycle if we use a dry-effects reader effect to give it access to the render_env, [which I wrote about here](https://timriley.info/writing/2020/10/06/open-source-status-update-september-2020/).

PR #1159, opened 2022-04-19
Introduce Hanami::SliceName

timriley:
This gives us a single object that can do the work of determining a slice/application's name and namespace, which means we can now set this up early and then pass it to any places that need it, like the configuration (since it wants to provide the app name as part of logger configuration).

This is an improvement on the (temporary) solution we had for the logger configuration, which required us to entirely relocate the work of determining the namespace etc. to the configuration object itself, which isn't really its job.

### DIFF
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -47,6 +47,8 @@ module Hanami
     module ClassMethods
       attr_reader :application_name, :configuration, :autoloader, :container
 
+      alias_method :slice_name, :application_name
+
       alias_method :config, :configuration
 
       def application
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -149,11 +149,6 @@ module Hanami
         application_name.namespace
       end
 
-      # TODO: Remove this after https://github.com/hanami/hanami/pull/1156 is merged
-      def namespace_path
-        application_name.name
-      end
-
       def root
         configuration.root
       end
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -7,6 +7,7 @@ require "rack"
 require "zeitwerk"
 require_relative "constants"
 require_relative "slice"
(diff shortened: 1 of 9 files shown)


lib/hanami/application.rb
```diff
@@ -147,19 +141,12 @@ def settings
...
-
-      def namespace_name
-        namespace.name
+        application_name.namespace
       end
 
+      # TODO: Remove this after https://github.com/hanami/hanami/pull/1156 is merged
       def namespace_path
-        inflector.underscore(namespace)
-      end
```
timriley, line 159: Thanks to https://github.com/hanami/hanami/pull/1156/commits/097b6226f8c98e79cb0a2a1b5b43b50a9ed8ada1 in https://github.com/hanami/hanami/pull/1156, there's no longer the need for this to be public API.

jodosha (approved):
@timriley Introducing this concept brings more clarity into implementation details. Bravo! 👏 

One consideration for the future: having `@application_name` to be a `SliceName` can be confusing.

timriley:
> @timriley Introducing this concept brings more clarity into implementation details. Bravo! 👏

Glad you like it, @jodosha! 😊 

> One consideration for the future: having `@application_name` to be a `SliceName` can be confusing.

Yeah, I agree here. This will hopefully get better, though: in my next round of work I want to unify a lot of the slice/application behavior and put it into a class or module with "slice" somewhere in its name. At this point, the "application" should both be and feel like a regular slice with a small number of extra specialisations. At that point, hopefully the name being represented by a `SliceName` will make more sense. (FWIW, this parallels the work I did with the new `SliceConfigurable` module, which actually works with both the application and slices).

solnic:
> This will hopefully get better, though: in my next round of work I want to unify a lot of the slice/application behavior and put it into a class or module with "slice" somewhere in its name

@timriley this is amazing! ❤️

PR #1162, opened 2022-04-23
Combine application and slice

timriley:
In this PR we move most of the core framework organisational responsibilities from `Application` to `Slice`, and then make `Application` a subclass of `Slice`.

It conveys many benefits, mainly around turning slices into a full-powered, independent, portable abstraction that can work not only as the key organising facility within Hanami applications, but also provide an avenue for functionality to be distributed _between_ Hanami apps.

I also think it makes the code easier to follow and maintain 😄 

### History

Before going deeper, let's recap the history here: the application/slices pairing was introduced back in late 2019 as the first broad stroke of the Hanami 2.0 redevelopment effort. It was pulled from Icelab's in-house [snowpack](https://github.com/icelab/snowpack) framework, and has been left largely untouched within Hanami since then (we had plenty of other work to do in the meantime). So with this PR, we give it the attention required to bring it into the state where we can confidently release it as part of 2.0.

## Changes

The broad changes here are:

- `Hanami::Slice` has been beefed up, with it now handling most of the responsibilities that were previously within `Hanami::Application`. This makes slices much more powerful as independent units.
- `Hanami::Application` has been turned into a subclass of `Hanami::Slice`, with its distinct responsibilities being the setup of the (very few) components that we want to be truly application-wide.

This means that every individual slice can now have its own:

- Configuration
- Settings
- Routes, router, and rack app
- Further nested slices

### Configuration

Every Slice's configuration is now `dup`'ed from the Application's configuration at the time of first access.

This means that we can now have users define their own slices classes and customise any aspect of configuration for _that slice only_, while still having the rest of the application config apply everywhere else, e.g.

```ruby
# config/slices/admin.rb

module Admin
  class Slice < Hanami::Slice
    # Applies to actions within Admin slice only
    config.actions.default_request_format = :json
    config.actions.default_response_format = :json
  end
end
```

One caveat to this approach is that it does requires any application-level configuration be made _before_ any slices are loaded. In the ordinary course of application booting, that will not be a problem, since app-level configuration is done within the body of the application class, which is loaded before slices.

### Settings, routes, etc.

These special framework-level components were previously only available on the singular application, but now they're available to _every slice_. However, every one of these is _entirely opt-in_ at the per-slice level.

This means that for simpler applications, we can still choose to keep all of these on the application slice only (i.e. only a single settings file in `/config/settings.rb`, only a single set of routes in `/config/routes.rb`, and only a single collection of slices at `/slices/*`).

However, now we can also gradually introduce per-slice instances of these wherever it makes sense. A good (and most likely to be exercised) example of this is settings: by introducing per-slice settings, we no longer need to have a single settings class that has to hold all the settings required for _every_ slice, and have all those settings unnecessarily available to every slice; instead, we can keep settings local to each slice, which means the components within the slice only have access to what they need, and no more.

How do we achieve this?

- Conditionality in the code to load settings, routes, etc. within `Hanami::Slice`. The `.load_settings` and `.load_routes` methods will now return nil if their respective require files or classes are missing.
- From there, conditionality in the setup of providers for each slice. See `Slice.prepare_container_providers`: if `routes` or `settings` are `nil`, then the provider for each is _not_ registered, which means there will be no such component registered in the container.
- We've added a new `slices.shared_component_keys` to `Hanami::Configuration`, which defaults to a small collection of components we expect should be shared across all slices. Leveraging https://github.com/dry-rb/dry-system/pull/236 and https://github.com/dry-rb/dry-system/pull/241, we now import these components from the application container to the root level of each slice (i.e. `Application["settings"]` will become `Admin::Slice["settings"]`), but _only_ if there isn't already a slice-local version of this component already available. This is what makes these components shared by default, but also possible to override per-slice.

As a bonus, it's also possible to "mount" a slice with its own routes into your top-level routes by using the `slice` DSL method without a block, e.g.

```ruby
module MyApp
  class Routes < Hanani::Routes
    define do
      # local routes here

      # Admin slice has its own set of routes,
      # which we make available in full under `/admin`
      slice :admin, at: "/admin"
    end
  end
end
```

### Nested slices (Yo dawg...)

Every slice can now have its own further nested slices (via `Slice.slices` and `Slice.register_slice`). In ordinary applications, this will still typically only be a single set of slices all registered on the application, but this flexibility does open up some interesting possibilities for the future.

For example, with slices become more powerful (see everything above), they could be distributed via gems providing high-level units of functionality to be plugged into Hanami apps. Depending on their complexity, those external slices may themselves be composed of a number of other slices.

This isn't a pathway we're particularly focusing on at the moment, but this change made sense in the broader context, and certainly doesn't impede in the ordinary construction of Hanami apps.

On a technical note, there are a few changes that follow from having arbitrarily nested slices:

- Every slice now has a `parent`, which is the slice into which it was registered (e.g. for `ParentSlice.register_slice(:child, ChildSlice)`, `ParentSlice` is the parent and `ChildSlice` is the child)
- For our existing `Slice.import` functionality (through which a slice may import components from another), now the only slices that will be visible for import as _siblings_, i.e. those slices which have been registered with the same parent

### Other details

- Given slices are now much more flexible, I had to make some decisions about our minimum expectations for them. One of these is that we now expect every slice to have a `root`. This is because `Dry::System::Container` also expects a root (which defaults to `Dir.pwd`), and if we don't pass a distinct root for each slice, it will result in them trying to load their components from the root of the overall application, which will lead to unexpected results. See `Slice.ensure_root` for how we handle this.
  - We do provide a sensible default for each slice's root, so users do not have to worry about that detail unless they want to. The default root is `slices/[slice_name]/ underneath the _parent's_ own `root` dir. See `SliceRegistrar#configure_slice` for details.
- Since `Hanami::Configuration` is now `dup`ed as part of Slice loading, I've added a bunch of `#initialize_copy` implementations to it and all its nested configuration classes. Revisiting these classes has reminded me there's probably a bunch of tidy-up we could do to make them more consistent/easier to understand, but at least for the time being we should have rational and deep `dup` behaviour.
- I merged the `rack_monitor` and `rack_logger` providers into a single `rack` provider, since it felt much better from a naming perspective. It also now only registers `"rack.monitor"` into the container, since there didn't seem to break any need to make `"rack.logger"` available.
- All the internal classes that were previously under the `Hanami::Application` namespace are now under the `Hanami::Slice` namespace (e.g. `Hanami::Slice::Router`). This makes the most sense given that those classes relate to facilities that are now managed through `Hanami::Slice`, and in addition it helps as avoid the awkwardness of having them namespace under a class that requires a superclass every time you use it as a namespace (e.g. `class Application < Slice; class Foo`)

This PR depends on a number of changes to dry-system and dry-container:

- https://github.com/dry-rb/dry-system/pull/236
- https://github.com/dry-rb/dry-system/pull/240
- https://github.com/dry-rb/dry-system/pull/241
- https://github.com/dry-rb/dry-container/pull/83

I'll need to merge and release the before we merge this PR.


solnic:
I’m really excited to see you tackling this again 🙂 One “exercise” that would prove that the design is really good would be to see if you can nest slices more than one level deep. Not that it would be a common thing to do, but it would mean that there are no unexpected assumptions when it comes to location of things.

lib/hanami/slice.rb
```diff
@@ -205,27 +304,114 @@ def prepare_container_component_dirs # rubocop:disable Metrics/AbcSize
...
+        if routes
+          require_relative "providers/routes"
+          register_provider(:routes, source: Hanami::Providers::Routes.for_slice(self))
+        end
+
+        if settings
+          require_relative "providers/settings"
+          register_provider(:settings, source: Hanami::Providers::Settings.for_slice(self))
+        end
+      end
```
timriley, line 334: I just want to point out this out because I'm particularly chuffed about this outcome: if we hadn't added support for provider sources as concrete classes to dry-system, there's no way we could've set up this kind of conditional provider loading against multiple arbitrary containers! ✨

lib/hanami/providers/rack.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "dry/system/provider/source"
+
+module Hanami
+  module Providers
+    class Rack < Dry::System::Provider::Source
+      def prepare
+        require "dry/monitor/rack/middleware" # FIXME: this _used_ to be just "dry/monitor"
```
timriley, line 9: This was changed to just `require "dry/monitor"` by @solnic in https://github.com/hanami/hanami/pull/1171, but then changed back by @jodosha in https://github.com/hanami/hanami/commit/c816219bcd38024c48b6b16bcd36a2cde365917f.

Now that dry-monitor has zeitwerk-enabled autoloading, I think we should leave it as `require "dry/monitor"`. @jodosha: was there any reason you changed it back?

timriley: n.b. this does mean we'll need to make a dry-monitor release soon.

solnic (approved):
🚀 ❤️ 👏🏻

lib/hanami/providers/rack.rb
jodosha: @timriley Because with just `require "dry/monitor"` it was crashing. 🤷‍♂️

jodosha (approved):
@timriley I really like how you split different responsibilities between Slice and Application.

jodosha (changes requested):
@timriley I tested this PR manually with a new Hanami app, and it won't start.

Steps to reproduce:

```
git clone https://github.com/hanami/cli.git
cd cli
bundle install && bundle exec rake install
cd # get back to your home dir
hanami new bookshelf
cd bookshelf
nvim Gemfile # Edit Gemfile to point to this PR branch
bundle install
bundle exec hanami server # or console
```

It crashes: https://gist.github.com/jodosha/bb1f069df5a1c968b22d0711965b9154

jodosha (approved):
@timriley Sorry for the false alarm, in my manual test I haven't used the temporary code from dry-rb gems.

Everything works fine. Thanks for the pointer. 🙂

lib/hanami/application.rb
```diff
@@ -1,203 +1,84 @@
...
-        container.config.root = configuration.root
-        container.config.inflector = configuration.inflector
+        # Ensure all basic slice preparation is complete before we make adjustments below
+        # (which rely on the basic prepare steps having already run)
+        super
+
+        # Run specific prepare steps for the application slice. Note also that some
+        # standard steps have been skipped via the empty method overrides below.
+        prepare_application_component_dirs
+        prepare_application_providers
```
timriley, line 67: I just want to put on the record that I'm not particularly enamoured with this solution here, but I think it's good enough for now.

Longer term I could see the slice having a registered list of setup steps (perhaps one that can be added to by extension gems), and at that point, the app class could explicitly delete the unwanted steps, rather than blanking them out like this.

timriley:
[Provide better centralized slice configuration](https://trello.com/c/hLjFHzyD/68-provide-better-centralized-slice-configuration)

timriley:
[Unify Application and Slices](https://trello.com/c/av9oD1hh/186-unify-application-and-slices)

PR #1164, opened 2022-05-16
Use Hanami::Utils.env and Hanami::Utils.env?

tak1n:
Relates to https://github.com/hanami/utils/pull/399

Creating a draft PR as I'm going to change the git source once the related PR is merged.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ unless ENV["CI"]
   gem "yard",   require: false
 end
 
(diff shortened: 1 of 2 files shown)


tak1n:
[Add Hanami.env and Hanami.env? to hanami-utils, so we can rely on them in gems like hanami-controller when they are used standalone](https://trello.com/c/dpt5iSkY/108-add-hanamienv-and-hanamienv-to-hanami-utils-so-we-can-rely-on-them-in-gems-like-hanami-controller-when-they-are-used-standalone)

PR #1166, opened 2022-06-01
Rescue expected LoadErrors only

timriley:
This way we don't swallow unexpected LoadErrors and give users poor feedback about the real reason for a particular failure.


katafrakt:
Ohhh, I was just bitten by this today and spent like 40 minutes figuring out what's happening. So I'm definitely for this change :bowing_man:

PR #1168, opened 2022-06-01
Respect RACK_ENV if HANAMI_ENV is not set

timriley:
Given RACK_ENV is an established, commonly used env var for signifying the application environment, it would be useful for us to respect it in the case of HANAMI_ENV not being present.

This change was inspired by a friend of mine working on porting a previous dry-rb-based application to Hanami 2, and having his development database wiped clean because his spec_helper.rb was still only setting RACK_ENV, not HANAMI_ENV.

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -44,7 +44,7 @@ module Hanami
   end
 
   def self.env
-    (ENV["HANAMI_ENV"] || "development").to_sym
+    ENV.fetch("HANAMI_ENV") { ENV.fetch("RACK_ENV", "development") }.to_sym
   end
 
   def self.env?(*names)
--- a/spec/unit/hanami/env_spec.rb
+++ b/spec/unit/hanami/env_spec.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+RSpec.describe Hanami, ".env" do
+  subject(:env) { described_class.env }
+
+  before do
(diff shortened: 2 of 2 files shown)


solnic:
I'm curious why we couldn't (could we?) use dry-system's env plugin for this? My intention with this plugin was to make it generic so that any framework could use it.

timriley:
@solnic The reason why we can't rely exclusively on dry-system's env plugin is that we access the `Hanami.env` in several places _before_ the application containers are even set up, e.g. [when we set up the configuration object](https://github.com/hanami/hanami/blob/6fa761e9044877747f889064125ccf7114929d52/lib/hanami/application.rb#L28). In addition, `Hanami.env` is public API and something we want to allow people to call even before an application is loaded for the first time.

We still do use the dry-system env plugin, but we just pass in `Hanami.env` as [the value for its env inferrer](https://github.com/hanami/hanami/blob/6fa761e9044877747f889064125ccf7114929d52/lib/hanami/application.rb#L182).

IMO this feels like the arrangement for these details. We'll always want `Hanami.env` to be available in all situations, and having the logic for determining the env available within that method (rather than tucked deeply away inside some part of our container setup) makes it easier to discover for users and developers.

jodosha (approved):
@timriley Good, Hanami 1 has the same behavior. 👍

PR #1169, opened 2022-06-11
Fix Content-Security-Policy

katafrakt:
CSP was not working for two reasons:

1. `assets-server-url` was not passed to `Actions` initialization, so it wasn't added to `scritp-src` and `style-src`.
2. For some reason, the CSP header was parsed incorrectly by both Firefox and Chromium when it contained values joined by a new line. This seems to be in line with [RFC 7230 section 3.2.4](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4), however adding a space after a `\n` didn't help - only having the header value as a single line works in these browsers.

### DIFF
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -56,7 +56,7 @@ module Hanami
 
       @actions = load_dependent_config("hanami/action") {
         require_relative "configuration/actions"
-        Actions.new(assets_server_url: assets.server_url)
+        Actions.new(assets: assets)
       }
 
       @middleware = Middleware.new
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -56,7 +56,7 @@ module Hanami
 
       @actions = load_dependent_config("hanami/action") {
         require_relative "configuration/actions"
-        Actions.new
+        Actions.new(assets_server_url: assets.server_url)
       }
 
       @middleware = Middleware.new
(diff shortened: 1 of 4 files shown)


lib/hanami/configuration.rb
```diff
@@ -56,7 +56,7 @@ def initialize(application_name:, env:)
 
       @actions = load_dependent_config("hanami/action") {
         require_relative "configuration/actions"
-        Actions.new
+        Actions.new(assets_server_url: assets.server_url)
```
timriley, line 59: There's a couple of problems with this line:

1. If hanami-asets isn't available, then `assets` will be a `NullConfiguration` instance, and will raise a `NoMethodError` for `#server_url`.
2. By referencing the server_url directly, if the user ends up ever changing this, we will end up using a stale value of the setting in the actions config.

We probably want to pass the whole `assets` object into `Actions.new` directly, and then refer to the `assets.server_url` in the `Actions#finalize!` method.

@katafrakt — would you mind having a go at that?

timriley (changes requested):
Thanks for sorting all of this out, @katafrakt!

I've left one note with you about a change we'll need to see about how we pass assets config into the actions config — would you mind taking a look at this?

lib/hanami/configuration.rb
katafrakt: @timriley I changed it to pass whole assets configuration, but it gets ugly, because `NullConfiguration` might be not loaded at this point (I assume this it by design?). Also not sure what you meant by referencing in `finalize!` - should the whole CSP-related logic move there?

jodosha:
@katafrakt Thank you. Closing in favor of https://github.com/hanami/hanami/pull/1268

katafrakt:
@jodosha #1268 addresses only half of what this PR is about. Assets still would not work because `assets_server_url`  setting is not passed to `Actions` initializer. Are there any plans to fix that as well?

PR #1171, opened 2022-06-15
Rely on autoloading in rack_monitor provider

solnic:
Refs dry-rb/dry-monitor#47

### DIFF
--- a/lib/hanami/application/container/providers/rack_monitor.rb
+++ b/lib/hanami/application/container/providers/rack_monitor.rb
@@ -1,12 +1,11 @@
 # frozen_string_literal: true
 
 Hanami.application.register_provider :rack_monitor do
-  start do
+  prepare do
     require "dry/monitor"
-    require "dry/monitor/rack/middleware"
+  end
 
-    middleware = Dry::Monitor::Rack::Middleware.new(target[:notifications])
-
-    register :rack_monitor, middleware
+  start do
+    register :rack_monitor, Dry::Monitor::Rack::Middleware.new(target[:notifications])
   end
 end
--- a/spec/new_integration/rack_app/rack_app_spec.rb
+++ b/spec/new_integration/rack_app/rack_app_spec.rb
@@ -7,6 +7,32 @@ RSpec.describe "Hanami web app", :application_integration do
 
   let(:app) { Hanami.rack_app }
 
+  specify "has rack monitor preconfigured with default request logging" do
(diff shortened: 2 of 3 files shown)


spec/new_integration/rack_app/rack_app_spec.rb
```diff
@@ -7,6 +7,30 @@
 
   let(:app) { Hanami.rack_app }
 
+  specify "has rack monitor preconfigured" do
```
timriley, line 10: Would it be worth expanding on this description a little bit? Right now it's a bit of a confusing leap from "has rack monitor preconfigured" to "there are logs from web requests". Is this test actually more about asserting the web requests are logged, with rack monitor being more of an implementation detail?

timriley (approved):
One small step for dry-monitor, one giant leap for dry-rb kind!

Left you a couple of q's, mostly minor, but still keen to hear your thoughts. Otherwise happy to see this go in whenever you're ready :)

lib/hanami/application/container/providers/rack_monitor.rb
```diff
@@ -1,12 +1,11 @@
...
     require "dry/monitor"
-    require "dry/monitor/rack/middleware"
-
-    middleware = Dry::Monitor::Rack::Middleware.new(target[:notifications])
+  end
 
-    register :rack_monitor, middleware
+  start do
+    register :rack_monitor, Dry::Monitor::Rack::Middleware.new(target[:notifications])
   end
```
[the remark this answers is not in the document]
solnic: Pragmatic approach is like you said - just have it all in a single `start` block but I think the built-in defaults should serve as an example of a clear separation between making something available (which potentially changes app's state) vs making something do what it's suppose to be doing (which may change the state even more + have additional side-effects, like here you end up with a logger listener).

In this case, `app.prepare(:rack_monitor)` would give you access to the `Dry::Monitor` API. Maybe you have a rake task that wants to use it somehow without attaching the listener and could leverage a more granular state management.

It's of course up to the end user to decide, but if we don't make it clear that this is not "like Rails initializers" and there's actually an explicit API for managing life-cycle of an external dependency, then people may not be able to fully leverage it.

spec/new_integration/rack_app/rack_app_spec.rb
solnic: Good point. I'll make it more clear and explicit.

PR #1172, opened 2022-06-20
Remove `::Application::` namespace in public API class names.

jodosha:


lib/hanami.rb
```diff
@@ -67,7 +63,23 @@ def self.shutdown
...
+  end
+
   def self.bundler_groups
     [:plugins]
   end
+
+  require_relative "hanami/version"
+  require_relative "hanami/errors"
+  require_relative "hanami/extensions"
+  require_relative "hanami/application"
```
timriley, line 84: @jodosha Is there a reason you keep these nested inside the `module Hanami` rather than just at the bottom of the file?

I think I'd prefer having them right at the bottom, because in reading this, having them nested seems to signify that that nesting is critical in some way, when it's actually not (I moved them to the bottom when playing around with this code and everything still worked).

lib/hanami/extensions.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+require_relative "./extensions/action" if Hanami.bundled?("hanami-controller")
```
timriley, line 3: Rubocop will probably complain but given the code that we have below, this structure would make this file easier to read, IMO:
 
```suggestion
if Hanami.bundled?("hanami-controller")
  require_relative "./extensions/action"
end
```

timriley (approved):
Left you a couple minor notes, but otherwise let's get this merged, thanks for doing it!

lib/hanami.rb
jodosha: @timriley "extensions" depend on Hanami.bundled?, which is defined above in this file.

timriley: @jodosha Right! So would we have just that one at the end of the file, and the rest at the top? That would give a clearer signal that "extensions" is the special case.

PR #1173, opened 2022-06-21
Depend on dry-files' main branch

waiting-for-dev:


timriley:
Is this just for consistency with the lines above, @waiting-for-dev? My understanding for git sources is that just uses whatever the remote's default branch is, so this would already be working with the main branch. Or am I wrong?

waiting-for-dev:
> Is this just for consistency with the lines above, @waiting-for-dev? My understanding for git sources is that just uses whatever the remote's default branch is, so this would already be working with the main branch. Or am I wrong?

TBH, not sure. I tried yesterday and `bundle` complaint about `master` branch not being present. But I've tried again and it worked, so maybe it was something else :shrug: Feel free to ignore the PR. But, yeah, we can also merge it for consistency.

PR #1174, opened 2022-06-22
Add `app/` as source dir for the application container; unify code loading rules for `app/` and slice dirs

timriley:
Here we add code loading to a new `app/` directory in the root of the Hanami project.

This will populate the application container (available via `Hanami.application` or the e.g. `Hanami::Application` subclass that you define in `config/application.rb`).

Adding code into `app/` is going to be our default approach for new Hanami apps, as our way to streamline onboarding for new users. You can, of course, opt into using slices at any time.

In this PR, we also make the transition between `app/` and slices smoother by unifying the code loading behaviour across both. Now:

- Any files or directories can be added to the root of either `app/` or slice dir will be autoloaded _and_ auto-registered into their respective containers. We expect files in the root of these directory to match the respective Ruby namespace of the app or slice, e.g. `app/foo.rb` should define `MyApp::Foo`, and `slices/main/foo.rb` should define `Main::Foo`.
- A special `lib/` subdirectory will also be treated as holding files within the app/slice root namespace. You can use this directory if you want to reserve the root of `app/` or your slice dir for a certain subset of your files; the rest can go in `lib/`. For example of this, `app/lib/bar.rb` should define `MyApp::Bar` and `slices/main/lib/bar.rb` should define `Main::Bar`.
- Via a new `config.no_auto_register_paths` setting (which defaults to `%w[entities]`), any root dirs under `app/` or a slice dir will be skipped from auto-registration (they will still be autoloaded). Dirs like these are the ones that contain classes you want to refer to and instantiate directly, not access as pre-built instances through the container.
  - Later on, our handling of this setting could be expanded to support glob strings, but I've kept it simple for now.
- `config/` under the slice directories will be ignored from code loading completely, since that dir is specially handled by the framework, and it will not contain classes we expect to be autoloaded or auto-registered.

With this change, we've also been able to remove the entire complex `config.source_dirs` object and greatly simplify our internal setup code.


solnic (approved):
The flexibility of this system is 🤯. This is like having the best of both worlds - auto-loadable constants with DI baked in + classic auto-loadable constants that you can refer to and handle manually if that's your thing (or requirement, which can happen).

This is really great, clearly shows how versatile Hanami 2.0 is, while "having opinions", it can also serve people who don't share them.

👏🏻 👏🏻 👏🏻

lib/hanami/application.rb
```diff
@@ -195,13 +195,23 @@ def prepare_container_base_config
...
 
-        # Expexct component files in the root of the app/ component dir to define classes
-        # inside the application's namespace
-        container.config.component_dirs.add(APP_DIR) do |dir|
-          dir.namespaces.add_root(key: nil, const: application_name.name)
+        # Component files in both `app/` and `app/lib/` define classes in the
+        # application's namespace
+
+        if root.join(APP_DIR, LIB_DIR).directory?
+          container.config.component_dirs.add(File.join(APP_DIR, LIB_DIR)) do |dir|
```
jodosha, line 205: @timriley any specific reason why we use `root.join` for the `if` statement and then `File.join` for adding the dir to the config?

Is that because we need the absolute dir to check its existence and then the relative dir to pass to dry-system?

spec/new_integration/code_loading/loading_from_slice_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+
+        module TestApp
+          class Application < Hanami::Application
+          end
+        end
+      RUBY
+
+      write "slices/main/config/settings.rb", <<~'RUBY'
+        module Main
+          class Settings < Hanami::Application::Settings
```
jodosha, line 17: @timriley shouldn't this be just `Hanami::Settings`? as per https://github.com/hanami/hanami/pull/1172

jodosha (commented):
@timriley The direction that we're taking is great!

A question about this:
> Via a new `config.no_auto_register_paths setting`

Does this mean that the end users will need to know:

* Everything is auto-registered by default
* Opt-out an entire directory, via `config.no_auto_register_paths`
* Opt-out a single file, via `# auto_register: false` magic comment

Anything else that they need to know?

lib/hanami/application.rb
timriley: @jodosha 100% correct 😄

timriley:
> A question about this:
> 
> > Via a new `config.no_auto_register_paths setting`
> 
> Does this mean that the end users will need to know:
> 
> * Everything is auto-registered by default
> * Opt-out an entire directory, via `config.no_auto_register_paths`
> * Opt-out a single file, via `# auto_register: false` magic comment

@jodosha, yep, that's exactly right! Hopefully these will be fairly straightforward to explain: one standard behaviour, and two ways to opt out (one for whole groups of files, and the other for individual files).

Nothing else you need to know besides those :)

spec/new_integration/code_loading/loading_from_slice_spec.rb
timriley: @jodosha Just rebased now. And I think I used the old class purely from habit; the updated version of this spec now has this inside `config/settings.rb`:

```ruby
write "slices/main/config/settings.rb", <<~'RUBY'
  raise "This file should never be loaded"
RUBY
```

This should hopefully express the intent of this test much more clearly — it's not really about the contents of this file, it's about the fact that it should _never be autoloaded_ by virtue of its location. Hopefully that `raise` makes it a lot clearer :)

timriley:
@jodosha All comments addressed if you wanted to give this a final check :)

jodosha (approved):

PR #1175, opened 2022-06-24
Remove routes and settings file configs

timriley:
We’ll better support our users by having these files found in the same place and with the same name in all cases.


jodosha (approved):
@timriley Nice one! 👍

PR #1178, opened 2022-07-06
Allow injecting file system

waiting-for-dev:
It's interesting to allow a hanami application to run in an in-memory
filesystem for testing purposes. We need to make explicit the transitive
dependency on `dry-files` and reorganize the `setup/prepare/boot`
commands to accept a filesystem argument.


waiting-for-dev:
We don't need it for now.

PR #1179, opened 2022-07-06
Fix use of session middleware

timriley:
With the `nil` as the final option (intended to represent a nil block) in the array of middleware args, we received this error when trying to use it:

```
ArgumentError: wrong number of arguments (given 3, expected 1..2)
```

### DIFF
--- a/lib/hanami/configuration/sessions.rb
+++ b/lib/hanami/configuration/sessions.rb
@@ -30,7 +30,7 @@ module Hanami
       end
 
       def middleware
-        [storage_middleware, options, nil]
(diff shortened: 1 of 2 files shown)


timriley:
@solnic FYI, since this was related to your recent change here.

Without this in place, updating the app template to the latest hanami (https://github.com/hanami/hanami-2-application-template/pull/88) was failing to `use` the session middleware, giving the error message I noted above.

solnic:
Sorry about that. I assumed we have that covered by specs. I'll fill in this gap!

PR #1181, opened 2022-07-10
Update rubocop and address issues

solnic:
For now I just updated rubocop.yml manually but once we have beta1 out, I'll update action-hero and finally enable it for hanami org so that we can keep all repos in sync automatically 😄


jodosha (changes requested):
@solnic Removing the reference to `https://raw.githubusercontent.com/hanami/devtools/main/.rubocop.yml`, this PR makes `hanami/hanami` source code different from the other `hanami/*` repositories. 👎

PR #1185, opened 2022-07-17
Ensure prepare_container works as expected in App

timriley:
We want to ensure that the `prepare_container` block is run at the _last moment_ before the container is marked as `configured!`. With this change, this is now true for both `App` and `Slice`.

Before this change, the `prepare_container` block for `App` was being run before the app-specific configuration was applied.

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -219,6 +219,7 @@ module Hanami
 
         prepare_all
 
+        instance_exec(container, &@prepare_container_block) if @prepare_container_block
         container.configured!
 
         @prepared = true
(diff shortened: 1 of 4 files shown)


lib/hanami/slice.rb
```diff
@@ -219,6 +219,7 @@ def prepare_slice
 
         prepare_all
 
+        instance_exec(container, &@prepare_container_block) if @prepare_container_block
```
timriley, line 222: This is basically the change, moving this instance_exec to a different spot. Everything else in this PR is ensuring we now have appropriate test coverage for this behaviour :)

timriley:
[Allow application author to configure application container](https://trello.com/c/KwVnA48c/102-allow-application-author-to-configure-application-container)

timriley:
@jodosha this is mostly a PR to ensure completeness/consistency in the existing `prepare_container` behaviour, which until #1162 existed on `Slice` only. Now that it's on `App` and `Slice`, it needed a little tweak. As such, I'm going to merge it in now, but happy to discuss it post-merge if need be 😄

PR #1187, opened 2022-07-21
Specify beta versions of hanami-cli and hanami-utils.

andrewcroome:
This bumps the hanami-cli and hanami-utils dependencies to their beta releases.

Background: when installing Hanami v2.0.0.beta1 via `gem install hanami --pre` we did not get the corresponding beta release of hanami-cli, which meant we did not have the `hanami new` command.


timriley:
@jodosha I think it'd probably be worth doing a quick beta1.1 release after we merge this, what do you reckon?

timriley (approved):

timriley:
I'm going to merge this and release as beta1.1. Thanks @andrewcroome!

PR #1188, opened 2022-07-21
Improve LOAD_PATH handling

timriley:
With this PR we have robust and complete handling (as well as better tests!) around the setup of $LOAD_PATH as part of app booting.

There are two key changes here:

- We now re-run `App.prepare_load_path` as part of the `App.prepare` step to ensure the load path is updated in the case of a re-configured app root.
- We also make `App.prepare_load_path` public API for people who have modified the `config.root` and want to ensure the new root’s `lib/` subdir is immediately added to the load path.

### DIFF
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -33,7 +33,10 @@ module Hanami
           @configuration = Hanami::Configuration.new(app_name: slice_name, env: Hanami.env)
           @autoloader = Zeitwerk::Loader.new
 
-          prepare_base_load_path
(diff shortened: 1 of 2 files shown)


timriley:
[Add all (app and slice) lib/ dirs to LOAD_PATH as early as possible](https://trello.com/c/XoduxZxZ/101-add-all-app-and-slice-lib-dirs-to-loadpath-as-early-as-possible)

jodosha (approved):

PR #1189, opened 2022-07-22
Conditional slice loading

timriley:
Enable conditional slice loading via a new `config.slices` setting.

For example, given this app class:

```ruby
module TestApp
  class App < Hanami::App
    config.slices = %w[admin]
  end
end
```

And the presence of both `slices/admin/` and `slices/main/`, when the app is prepared (or booted), only the admin slice will be loaded:

```ruby
require "hanami/prepare"

Hanami.app.slices.keys # => [:admin]

Admin::Slice # exists, as expected
Main         # raises NameError, since it was never loaded
```

When slices are skipped from loading, they're _completely ignored_ — their slice classes are not loaded, nor are any of their source files. This is valuable because it can keep the app's memory profile as low as possible for any configured set of slices.

Since loading/skipping particular slices is likely to be common adjustment for different deployment modes (e.g. deploying just the admin slice to a different set of containers, or running just a subset of slices for a background worker), we make it easy to specify this config via an `HANAMI_SLICES` env var. When using this, specify the slices as comma-separated:

```
HANAMI_SLICES="main,shop"
```

Lastly, the `config.slices` setting also supports arbitrarily nested slices. For example, if have an "admin" slice with its own child "shop" and "editorial" slices, you can specify `config.slices.load_slices = ["admin.shop"]` to load only its nested shop slice.

---

One caveat about this feature is that it's not "smart" — if one slice e.g. imports another (or expects another slice to be available for any other reason), and if that other slice is skipped, you'll receive errors. It's entirely the responsibility of the user to make sure that the appropriate set of slices is loaded when configuring `load_slices`/`skip_slices`.

I took this approach in order to keep the implementation simple and the feature easy to understand. If I tried to make it smarter I think we'd over-complicate the implementation and probably still end up with a range of cases where the system just can't automatically infer what it needs.

Does this trade off seem reasonable to you?

---

Implementation-wise, this was a fairly well-contained change.

The conditional loading logic is kept entirely within `SliceRegistrar`. There is a new private `#filter_slice_names` method that is given an array of slice names, and (based on the parent slice's `config.slices`) returns the subset of those slices that is permissible to load. Then the rest of the slice registration continues as always, just on this subset of slices only.

This slice name filtering is done in both `SliceRegistrar`'s explicit `#register` method (which is called when slices are registered explicitly using `Slice.register_slice`) as well as its `#load_slices` method (which is used when running `Slice.prepare` and detecting slices kept in a `slices/` directory), so it works consistently across both cases.

The only other change is in `SliceRegistrar#configure_slice`, which is called just before a slice is registered on a parent. Here we look for dot-delimited nested slice names in  `config.slices` and adapt them for the about-to-be-registered slice. So if we had a `config.slices` of `["admin.shop"]` and we were registering the `"admin"` slice, then _its_ `config.slices` would become `["shop"]`. This is how the nested slice support works, by making sure the list is relevant for each slice.

Lastly, to support the new env vars, we have a new private `#load_from_env` method on `Hanami::Configuration`, which is called during its `#initialize`, and simply does this:

```ruby
def load_from_env
  self.slices = ENV["HANAMI_SLICES"]&.split(",")&.map(&:strip)
end
```

This may be a useful hook for other settings we want to have loaded from corresponding env vars in the future.

---

_This change was originally built with separate (and mutually exclusive) `load_slices` and `skip_slices` settings, which we've reduced to a single `slices` setting for simplicity. The original description is below, for reference._

<details>
<summary>Original description using both <code>load_slices</code> and <code>skip_slices</code></summary>

Enable conditional slice loading via two new settings: `slices.load_slices` and `slices.skip_slices`.

For example, given this app class with `config.slices.load_slices`:

```ruby
module TestApp
  class App < Hanami::App
    config.slices.load_slices = %w[admin]
  end
end
```

And the presence of both `slices/admin/` and `slices/main/`, when the app is prepared (or booted), only the admin slice will be loaded:

```ruby
require "hanami/prepare"

Hanami.app.slices.keys # => [:admin]

Admin::Slice # exists, as expected
Main         # raises NameError, since it was never loaded
```

These `load_slices` and `skip_slices` are settings are mutually exclusive: if `load_slices` is configured, then only those slices will be loaded, and `skip_slices` will be ignored. If `load_slices` is not configured, however, then `skip_slices` will be checked and the configured slices will be skipped from loading.

When slices are skipped via either of these two settings, they're _completely_ ignored — their slice classes are not loaded, nor are any of their source files. This is valuable because it can keep the app's memory profile as low as possible for any configured set of slices.

Since loading or skipping certain slices is likely to be common adjustment made for different deployment modes (e.g. deploying just the admin slice to a different set of containers, or running just a subset of slices for a background worker), we make it easy to specify this config via `HANAMI_LOAD_SLICES` and `HANAMI_SKIP_SLICES` env vars. When using these, specify the slices as comma-separated:

```
HANAMI_LOAD_SLICES="main,shop"
# or
HANAMI_SKIP_SLICES="admin"
```

Lastly, the `load_slices` and `skip_slices` settings also support arbitrarily nested slices. For example, if have an "admin" slice with its own child "shop" and "editorial" slices, you can specify `config.slices.load_slices = ["admin.shop"]` to load only its nested shop slice.

---

One caveat about this feature is that it's not "smart" — if one slice e.g. imports another (or expects another slice to be available for any other reason), and if that other slice is skipped, you'll receive errors. It's entirely the responsibility of the user to make sure that the appropriate set of slices is loaded when configuring `load_slices`/`skip_slices`.

I took this approach in order to keep the implementation simple and the feature easy to understand. If I tried to make it smarter I think we'd over-complicate the implementation and probably still end up with a range of cases where the system just can't automatically infer what it needs.

Does this trade off seem reasonable to you?

---

Implementation-wise, this was a fairly well-contained change.

The conditional loading logic is kept entirely within `SliceRegistrar`. There is a new private `#filter_slice_names` method that is given an array of slice names, and (based on the parent slice's `load_slices` or `skip_slices` config) returns the subset of those slices that is permissible to load. Then the rest of the slice registration continues as always, just on this subset of slices only.

This slice name filtering is done in both `SliceRegistrar`'s explicit `#register` method (which is called when slices are registered explicitly using `Slice.register_slice`) as well as its `#load_slices` method (which is used when running `Slice.prepare` and detecting slices kept in a `slices/` directory), so it works consistently across both cases.

The only other change is in `SliceRegistrar#configure_slice`, which is called just before a slice is registered on a parent. Here we look for dot-delimited nested slice names in the `load_slices` or `skip_slices` config and adapt them for the about-to-be-registered slice. So if we had a `load_slices` of `["admin.shop"]` and we were registering the `"admin"` slice, then _its_ `load_slices` would simply become `["shop"]`. This is how the nested slice support works, by making sure the list is relevant for each slice.

Lastly, to support the new env vars, we have a new private `#load_from_env` method on `Hanami::Configuration`, which is called during its `#initialize`, and simply does this:

```ruby
    def load_from_env
      slices.load_slices = ENV["HANAMI_LOAD_SLICES"]&.split(",")&.map(&:strip)
      slices.skip_slices = ENV["HANAMI_SKIP_SLICES"]&.split(",")&.map(&:strip)
    end
```

This may be a useful hook for other settings we want to have loaded from corresponding env vars in the future.
</details>

### DIFF
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -29,18 +29,18 @@ module Hanami
 
     setting :settings_store, default: Hanami::Settings::EnvStore.new
 
+    setting :shared_app_component_keys, default: %w[
+      inflector
+      logger
+      notifications
+      rack.monitor
+      routes
+      settings
+    ]
+
     setting :slices do
       setting :load_slices
       setting :skip_slices
-
-      setting :shared_app_component_keys, default: %w[
-        inflector
-        logger
-        notifications
-        rack.monitor
-        routes
-        settings
-      ]
     end
 
     setting :base_url, default: "http://0.0.0.0:2300", constructor: ->(url) { URI(url) }
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -38,10 +38,7 @@ module Hanami
       settings
     ]
 
-    setting :slices do
-      setting :load_slices
-      setting :skip_slices
-    end
+    setting :slices
 
     setting :base_url, default: "http://0.0.0.0:2300", constructor: ->(url) { URI(url) }
 
@@ -202,8 +199,7 @@ module Hanami
     private
 
     def load_from_env
-      slices.load_slices = ENV["HANAMI_LOAD_SLICES"]&.split(",")&.map(&:strip)
-      slices.skip_slices = ENV["HANAMI_SKIP_SLICES"]&.split(",")&.map(&:strip)
+      self.slices = ENV["HANAMI_SLICES"]&.split(",")&.map(&:strip)
     end
 
     def apply_env_config(env = self.env)
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -30,7 +30,7 @@ module Hanami
     setting :settings_store, default: Hanami::Settings::EnvStore.new
 
     setting :slices do
-      setting :shared_component_keys, default: %w[
+      setting :shared_app_component_keys, default: %w[
         inflector
         logger
         notifications
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -30,6 +30,9 @@ module Hanami
     setting :settings_store, default: Hanami::Settings::EnvStore.new
 
     setting :slices do
+      setting :load_slices
+      setting :skip_slices
+
       setting :shared_app_component_keys, default: %w[
         inflector
         logger
--- a/lib/hanami/configuration.rb
+++ b/lib/hanami/configuration.rb
@@ -103,9 +103,9 @@ module Hanami
       @environments = DEFAULT_ENVIRONMENTS.clone
       @env = env
 
-      # Some default setting values must be assigned at initialize-time to ensure they
-      # have appropriate values for the current app
+      # Apply default values that are only knowable at initialize-time (vs require-time)
       self.root = Dir.pwd
+      load_from_env
 
       config.logger = Configuration::Logger.new(env: env, app_name: app_name)
 
@@ -201,14 +201,17 @@ module Hanami
 
     private
 
-    # @api private
+    def load_from_env
+      slices.load_slices = ENV["HANAMI_LOAD_SLICES"]&.split(",")&.map(&:strip)
+      slices.skip_slices = ENV["HANAMI_SKIP_SLICES"]&.split(",")&.map(&:strip)
+    end
+
     def apply_env_config(env = self.env)
       environments[env].each do |block|
         instance_eval(&block)
       end
     end
 
-    # @api private
     def load_dependent_config(require_path, &block)
       require require_path
       yield
@@ -219,7 +222,6 @@ module Hanami
       NullConfiguration.new
     end
 
-    # @api private
     def method_missing(name, *args, &block)
       if config.respond_to?(name)
         config.public_send(name, *args, &block)
@@ -228,7 +230,6 @@ module Hanami
       end
     end
 
-    # @api private
     def respond_to_missing?(name, _incude_all = false)
       config.respond_to?(name) || super
     end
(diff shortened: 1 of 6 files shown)


timriley:
[Properly support conditional loading of slices](https://trello.com/c/HDuJeazH/107-properly-support-conditional-loading-of-slices)

jodosha (commented):
@timriley Before to dive into the implementation, I'd like to hear the rationale behind the design.

IMO _load_ vs _skip_ and _setting_ vs _env var_ creates a complex matrix of combinations, hard to maintain and difficult to digest.

---

What if we simplify with: _load_ (eliminate skip), via env var?
That would be `HANAMI_SLICES=comma,separated,slice,names`

Reasons:

* We're solving one important monolith problem: the impossibility to scale up only portions of an app. E.g. you have an HTTP API that has zillion of requests, whereas the web UI is less busy. So you have let's say 5 servers with `HANAMI_SLICES=api` and 1 server with `HANAMI_SLICES=web,signup`.
* Having a setting to load the slices, most of the time, will end up `config.load_slices = ENV["SOME_ENV_VAR"]`, because which slice to load depends on the deployment env. Limiting to special env var `HANAMI_SLICES`, makes the code less verbose.
* We already have other special env vars e.g. `HANAMI_ENV`, and soon `SERVE_STATIC_ASSETS` (for Ruby hosting purposes).

timriley:
@jodosha and I quickly discussed this offline. We will:

- change this to a single slice allow list only (remove the "skip" setting)
- rename the setting to `config.slices`
- rename the existing `config.slices.shared_app_component_keys` to a top-level setting instead
- change the ENV var to `HANAMI_SLICES`
- the setting has to stay because of the way it works with nested slices, see `SliceRegistrar#configure_slice` in the PR diff

solnic (approved):
Wonderful 👏🏻

PR #1190, opened 2022-07-22
Make env file loading a clearer step when subclassing app

timriley:
This makes more sense than loading it as a side effect of preparing the settings store.

Yes, the settings are popualted from env vars, but that’s not the _only_ thing in an app that may rely upon them, so loading dotenv as early as possible makes more sense.

To address this, we:

1. Move dotenv loading into `Hanami::App` (in particular the private `.load_dotenv`) method
2. Call `.load_dotenv` as soon as `App` is subclassed, making the populated `ENV` available as early as possible (this would be useful for e.g. referring to `ENV` vars inside the `App` subclass body)
3. Rename `Hanami::Settings::DotenvStore` to `EnvStore`
4. Create a new set of integration tests to cover the various `.env` files that can be loaded


spec/new_integration/dotenv_loading_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    ENV.replace(@orig_env)
+  end
+
+  context "dotenv gem is available" do
+    before do
+      require "dotenv"
+    end
+
+    context "hanami env is development" do
+      it "loads .env.development.local, .env.local, .env.development and .env (in this order) into ENV", :aggregate_failures do
```
timriley, line 20: FWIW, a lot of these tests now overlap with the ones I added in the (also open) https://github.com/hanami/hanami/pull/1186. Once these are both merged, I'll remove any redundant tests (leaving this file as the canonical set of tests exercising the `.env` file loading behaviour).

In the meantime, I figured it made sense to let each PR stand on its own.

solnic:
Why was this needed? It doesn't sound like a good idea to load file contents in an inherited hook. The only thing that should depend on dotenv behavior is the settings object. I'm having trouble understanding why we need to load it so early (which makes the app coupled to dotenv).

solnic:
@timriley thank you, this makes sense, and you're awesome 😄 I'm just a bit reluctant whenever I see complex logic added to inherited hook as it gave us a lot of pain in rom-rb in its early days. One idea I had to avoid it here was to make it very explicit by having an extra line in the generated class skeleton, like "load_dotenv" or something, just to make it clear that it's happening. We can of course stick to what you did and see how it goes. It's reassuring that this is how dotenv-rails works. I appreciate your thoughtful explanation ❤️

lib/hanami/app.rb
```diff
@@ -53,6 +54,32 @@ def prepare_base_load_path
...
+        hanami_env = Hanami.env
+        dotenv_files = [
+          ".env.#{hanami_env}.local",
+          (".env.local" unless hanami_env == :test),
+          ".env.#{hanami_env}",
+          ".env"
+        ].compact
+
+        require "dotenv"
+        Dotenv.load(*dotenv_files) if defined?(Dotenv)
```
jodosha, line 78: @timriley I want to double check this with you: what's the Public API to access env variables?

In H1 we had `Hanami::Env`, reasons are written here: https://lucaguidi.com/2016/12/27/isolate-global-state/

---

I'm asking because we could populate `Hanami::Env`, instead of `ENV`, by using `Dotenv.parse(*dotenv_files)`, which returns a `Hash`.

jodosha (approved):
@timriley LGTM. 👍
Left a couple of comments for potential enhancements.

lib/hanami/app.rb
```diff
@@ -53,6 +54,32 @@ def prepare_base_load_path
...
+      rescue LoadError => e
```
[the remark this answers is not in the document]
timriley: Oh yeah, thanks for the reminder, @jodosha! I've used this here now too, see https://github.com/hanami/hanami/pull/1190/commits/03a50e9ab5a48b49a8f77f80ebbb7d4147ebbee8

lib/hanami/app.rb
timriley: @jodosha It's a good question. This PR certainly doesn't have the ambition of providing such an interface — its job is just to populate `ENV` as we've always done, but just in a clearer way.

As for whether we should provide a first party, "safer" interface to ENV, my current feeling is that it's probably not worthwhile it at this point: we should be encouraging our users to take advantage of our settings components to access relevant values from the env. Cases where that is not possible (like some of the scenarios I described in my comment above) are hopefully relatively minor and not the kind of thing that would be close to the actual working business logic of Hanami apps.

What do you think?

timriley:
Thanks again @solnic for the initial prompt and @jodosha for checking all of this out 😄

@solnic as for your idea for having `load_dotenv` or similar be baked out into our generated `config/app.rb` class, I think that's a reasonable possibility to keep alive, but in the spirit of keeping as much as possible _not_ baked out, I do think it's worth the experiment of the automatic approach as we have here (especially since it's mostly just a rearrangement of behaviour that was already in place).

A related thing that I would like to consider in the post-2.0 world is some kind of well-defined, extensible structure for the overall framework app boot process. In this world, we could have "load dotenv" as one of the pre-defined steps in the boot, which a user could just explicitly disable if they didn't want it.

Anyway, I'll merge this one in now and we can collectively keep an eye out for any issues. That said, my feel is that we'll be fine to bring this through to the 2.0 releases :)

solnic:
> A related thing that I would like to consider in the post-2.0 world is some kind of well-defined, extensible structure for the overall framework app boot process. In this world, we could have "load dotenv" as one of the pre-defined steps in the boot, which a user could just explicitly disable if they didn't want it.

Yes I've had that idea for a long time but I thought about making it a feature in dry-system. It'd allow you to define a Unit of Work for the booting process where steps can have dependencies and then it's resolved, sorted and executed in the right order based on your configuration. When this process is hardcoded, you lose a lot of flexibility.

timriley:
[Load Dotenv separately from settings store](https://trello.com/c/c3KEXhVt/189-load-dotenv-separately-from-settings-store)

PR #1191, opened 2022-07-22
Make middleware stack enumerable and add docs

waiting-for-dev:
We already defined the `#each` method to delegate to the underlying data structure. We go all the way down to be able to use any Enumerable method.

We also add some docs to make it more easily understandable.

### DIFF
--- a/lib/hanami/slice/routing/middleware/stack.rb
+++ b/lib/hanami/slice/routing/middleware/stack.rb
@@ -15,6 +15,8 @@ module Hanami
         # @since 2.0.0
         # @api private
         class Stack
+          include Enumerable
+
           # @since 2.0.0
           # @api private
           ROOT_PREFIX = "/"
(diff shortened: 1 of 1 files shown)


solnic:
@waiting-for-dev I'm OK with merging this in although the stack could use some more love. ie I don't like the internal hash representation and that `each` yields key/value pairs. This looks like primitive obsession to me and we'd probably benefit from having a simple representation of a middleware "item" that could just be a struct instead of dealing with paths and these arrays that are difficult to operate.

waiting-for-dev:
Thanks for your review, @solnic. Yeah, I agree with everything. This is only a PR to make consuming what we have now from `hanami/cli`. Merging it :slightly_smiling_face:

timriley:
["`hanami middlewares" command](https://trello.com/c/Pbd1BM6U/190-hanami-middlewares-command)

PR #1193, opened 2022-08-02
Respect customized config in slice configured classes

timriley:
Update `SliceConfigurable` as well as `SliceConfiguredAction` and `SliceConfiguredView` to ensure action and view config is applied only once per slice, and that custom config set in a superclass isn't overwritten by default values from a slice that have already been applied from the app.

This ensures that you can now use an app-level base action class with custom config. So given this base class:

```ruby
module MyApp
  class View < Hanami::View
    config.default_response_format = :json
  end
end
```

And this subclass (which could reside either in the app or a slice):

```ruby
module MySlice
  module Actions
    class SomeAction < MyApp::Action
    end
  end
end
```

Then after this change, the `config.default_response_format` of the subclass would remain as `:json`, which given the code above, would be the expected outcome.

The way we achieve this is to skip setting any `config.actions` or `config.views` values from a slice if they are the same as the config values on the app.


timriley:
@andrewcroome @pat This should address both of your issues with config set inside action/view classes being overridden with the default values from the app/slice. Would you mind giving it a go within your apps to confirm it now works like you expect?

timriley:
[Fix issue with slice actions/views config overriding config set explicitly in superclass](https://trello.com/c/voYJzBko/203-fix-issue-with-slice-actions-views-config-overriding-config-set-explicitly-in-superclass)

lib/hanami/extensions/view/slice_configured_view.rb
```diff
@@ -32,12 +32,52 @@ def inspect
         # rubocop:disable Metrics/AbcSize
         def configure_view(view_class)
           view_class.settings.each do |setting|
-            if slice.config.views.respond_to?(:"#{setting}")
-              view_class.config.public_send(
-                :"#{setting}=",
-                slice.config.views.public_send(:"#{setting}")
-              )
-            end
+            next unless slice.config.views.respond_to?(:"#{setting}")
```
solnic, line 35: @timriley if you put a `!slice.parent` guard here too, the logic below could be simpler

lib/hanami/extensions/view/slice_configured_view.rb
```diff
@@ -32,12 +32,52 @@ def inspect
...
+            # If we blindly re-configure all view settings each time `SliceConfiguredView` is
+            # extended, then at the point of (2) above, we'd end up overwriting the custom
+            # `config.layout` explicitly configured in the `MyApp::View` base class, leaving
+            # `MySlice::Views::SomeView` with `config.layout` of `"app"` (the default as specified
+            # at `Hanami.app.config.views.layout`), and not the `"custom_layout"` value configured
+            # in its immediate superclass.
+            #
+            # This would be surprising behavior, and we want to avoid it.
+            slice_value = slice.config.views.public_send(:"#{setting}")
+            parent_value = slice.parent.config.views.public_send(:"#{setting}") if slice.parent
```
solnic, line 76: Can’t we rely on `Config#[]` here?

solnic (approved):
Left a couple of minor comments/suggestions, otherwise LGTM

andrewcroome:
> @andrewcroome @pat This should address both of your issues with config set inside action/view classes being overridden with the default values from the app/slice. Would you mind giving it a go within your apps to confirm it now works like you expect?

Thank you @timriley. I can confirm that our action configs inherit again with this change 🙏  

So with this type of base action class for an API slice:

```ruby
module API
  class Action < Hanami::Action
    config.handle_exception StandardError => :handle_standard_error

    private

    def handle_standard_error(req, res, exception)
      res.status = 500
      res.format = format(:json)
      res.body = {error: "Something went wrong processing your request"}.to_json
    end
  end
end
```

We see the above error handling behaviour when inheriting from that action:

```ruby
module API
  module Actions
    class TestException < Action
      def handle(req, res)
        raise "Oops!"
      end
    end
  end
end
```

lib/hanami/extensions/view/slice_configured_view.rb
timriley: @solnic Thanks for the suggestion :)

I've just taken a look at this now... and I'm not sure I could move the parent check up here. We still _do_ want to apply config for slices with or without parents; we just don't want to _reapply_ the config if the slice has a parent _and_ its values match everything the parent would have already applied.

I'll leave things as they are for now, but you'd be very welcome to take a pass at this as a follow up :)

(I'll readily admit that all the work we're having to do here is more than I would like. It'd be good if we could provide some dry-configurable API to help with this kind of case in the future, i.e. knowing whether a config has been assigned or mutated)

lib/hanami/extensions/view/slice_configured_view.rb
timriley: Unfortunately not, since `config.views` here is and instance of `Hanami::Configuration::Views`, which delegates to an inner `base_configuration` object for most of its settings. I could probably extend it to add support for delegating to that object for `#[]`, but for right now I might leave it as is (that kind of change feels a little outside of the scope of this bug fix PR).

timriley:
@andrewcroome Thank you very much for checking this! Glad it sorted you out 😄

PR #1197, opened 2022-08-12
Make setup (finding and loading the app) more robust

waiting-for-dev:
This PR introduces a new `Hanami.setup` method, which is now called when requiring `"hanami/setup"`.

This method will look for the app file (`config/app.rb`) in the current directory, as well as search upwards through parent directories until the app file can be found.

This means that the `hanami` CLI will work as expected even when executed from a deeply nested directory within your Hanami app.

When an app file cannot be found, an `Hanami::AppLoadError` method is raised. This raising behaviour can be disabled by calling `Hanami.setup(raise_exception: false)`.

A secondary `Hanami.app_path` method is also introduced, which is the method that does the actual searching for the app file path. This method either returns the absolute path of the found file, or `nil` if none could be found.


waiting-for-dev:
[Make app detection in hanami/setup.rb more robust](https://trello.com/c/B9JeaqrL/78-make-app-detection-in-hanami-setuprb-more-robust)

PR #1201, opened 2022-08-16
Move require for rack inside to_rack_app

timriley:
Now that the Hanami::Routing::Middleware::Stack is loaded inside Hanami::Configuration, we need to be tolerant of apps that do not have rack bundled.

### DIFF
--- a/lib/hanami/slice/routing/middleware/stack.rb
+++ b/lib/hanami/slice/routing/middleware/stack.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "rack/builder"
-
 module Hanami
   class Slice
     module Routing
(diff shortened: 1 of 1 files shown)


jodosha (commented):
@timriley I don't understand this change 🙂 .

When we activate `lib/hanami/slice/routing/middleware/stack.rb` we should be already sure that Hanami web stack is activated and so `rack` is available.

---

If we're requiring `lib/hanami/slice/routing/middleware/stack.rb`, and `rack` isn't available, it's our fault, not user's. We should have been detected `hanami-router` before to require it.

---

We're giving our users misleading (IMO) instruction to bundle `rack` gem. But would the rest work as expected? We're in the `Routing` module, so we expect to work with `hanami-router`.

timriley:
@jodosha Agreed! Let's take another look at this, but after beta2 is out.

PR #1204, opened 2022-08-26
Make router configuration optional and have the config system raise when a component is missing

waiting-for-dev:
We're now configuring `Hanami::Configuration::Router` only when `hanami/router` is available.

Up to this point, we were returning a `NullConfiguration` instance when trying to configure an unavailable component. That was too implicit. We want to tell users what is missing.

That also lets us remove the check for rack when creating a web application and detect the wrong assumption earlier.


waiting-for-dev:
[Do not load full `config.middlewares` if rack is not bundled (or similar)](https://trello.com/c/AsMLLUOV/205-do-not-load-full-configmiddlewares-if-rack-is-not-bundled-or-similar)

lib/hanami/configuration.rb
```diff
@@ -215,8 +219,7 @@ def load_dependent_config(require_path)
     rescue LoadError => e
       raise e unless e.path == require_path
 
-      require_relative "configuration/null_configuration"
-      NullConfiguration.new
+      raise ComponentNotAvailable, "`#{require_path}` is not available"
```
timriley, line 222: Thanks for making this change! I feel much more comfortable now with this kind of error versus surprising NoMethodErrors on that NullConfiguration.

While we're making this change, however, could we expand on this message here to leave no room for doubt as to the reason for the error?

e.g. for this case:

```
ComponentNotAvailable: you must "hanami-router" the gem to configure `config.router` settings
```

timriley (changes requested):
Great stuff, thanks @waiting-for-dev! Just left one little note about improving the friendliness of our error message here — would you mind giving it a tweak?

lib/hanami/configuration.rb
timriley: To make this message easier to build from within load_dependent_config, it might actually be easier now to pass it a `gem_name` instead of `require_path`, and then have it use `Hanami.bundled?` internally instead of the LoadError rescue.

waiting-for-dev: Makes sense, @timriley, I like it. Please, take a look at my last commit :pray:

timriley: Thanks @waiting-for-dev! Just left a note for you on that new commit :)

lib/hanami/configuration.rb
```diff
@@ -209,14 +213,12 @@ def apply_env_config(env = self.env)
...
 
     # @api private
-    def load_dependent_config(require_path)
-      require require_path
-      yield
-    rescue LoadError => e
-      raise e unless e.path == require_path
+    def load_dependent_config(gem_name)
+      raise ComponentNotAvailable, <<~MSG unless Hanami.bundled?(gem_name)
+        You must add "#{gem_name}" to the Gemfile to configure its component
```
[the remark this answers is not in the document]
waiting-for-dev: Passing the extra argument bothered me, but if you think it's a needed trade-off, let's do it! I've just updated the commit with it. :slightly_smiling_face:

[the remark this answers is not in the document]
timriley: @waiting-for-dev Thanks for that! ❤️

Yeah, I know what you're saying about the extra argument, but given this is entirely internal API, it's worth it to prioritise the user experience over streamlining method arguments :)

timriley (approved):
Let's get this merged! :shipit:

waiting-for-dev:
Thanks for your support on this, @timriley!

timriley:
@waiting-for-dev I've just reverted this since I've realised it doesn't work at all in real app situations:

- if e.g. hanami-view is not bundled, it'll raise an exception as soon as you initialise `Configuration`, which will prevent the app from event starting
- We also need to do something about the fact that we finalize! on all the nested configs

PR #1207, opened 2022-08-30
Remove old integration tests

waiting-for-dev:
They're too far away from the current setup, so we ditch them and rename
"new_integration" to "integration".


waiting-for-dev:
[Port/remove old integration tests](https://trello.com/c/5ipfXuvG/182-port-remove-old-integration-tests)

timriley:
Thanks for doing this, @waiting-for-dev! I still think there might be some value in combing through these legacy integration tests and making sure we have sufficient coverage for our 1.0->2.0 upgrades, but I think that can still reasonably be done by looking at the legacy tests from the 1.x branches, rather than continuing to keep them here in the main branch.

Happy for you to merge whenever.

timriley (approved):

PR #1208, opened 2022-08-31
Add support for defining routes without `define` block

solnic:
Before:

```ruby
module TestApp
  class Routes < Hanami::Routes
    define do
      get "home", to: "home.show", as: :home
    end
  end
end
```

After:

```ruby
module TestApp
  class Routes < Hanami::Routes
    get "home", to: "home.show", as: :home
  end
end
```


solnic:
Actually, this class could be utilized better. It feels like the complexity of defining the router could be moved to `Hanami::Routes`. We could have something like `Hanami::Routes#to_router` and an instance of the routes that we'd create would know what to do. I also wonder if this class wouldn't be a good place to store all kinds of information about the router and routes for things like helpers etc.

solnic:
[Remove Hanami::Routes.define in favor of direct routing DSL](https://trello.com/c/O8iuFKUM/196-remove-hanamiroutesdefine-in-favor-of-direct-routing-dsl)

lib/hanami/routes.rb
```diff
@@ -38,7 +40,7 @@ def self.define(&block)
 
     # @api private
     def self.routes
-      @_routes || raise(<<~MSG)
+      (@_routes || build_routes) || raise(<<~MSG)
         Routes need to be defined before being able to fetch them. E.g.,
           define do
```
jodosha, line 45:
```suggestion
```

lib/hanami/routes.rb
```diff
@@ -47,5 +49,50 @@ def self.routes
...
+        supported_methods.include?(name) || super
+      end
+
+      # Capture all method calls that are supported by the router DSL
+      # so that it can be evaluated lazily during configuration/boot
+      # process
+      #
+      # @api private
+      def method_missing(name, *args, **kwargs, &block)
+        return super unless supported_methods.include?(name)
```
jodosha, line 92:
```suggestion
        return super unless respond_to?(name)
```

jodosha (approved):
@solnic Great to see this to happen! 💯

lib/hanami/routes.rb
```diff
@@ -47,5 +49,50 @@ def self.routes
...
+
+      # @api private
+      def definitions
+        @definitions ||= []
+      end
+
+      # @api private
+      def supported_methods
+        @supported_methods ||= Slice::Router.public_instance_methods
+      end
```
timriley, line 77: Could this also be moved below `private`?

timriley: (Alternatively, could possibly turn into a private constant, making it clearer that this is intended to be static?)

lib/hanami/routes.rb
```diff
@@ -38,7 +40,7 @@ def self.define(&block)
 
     # @api private
     def self.routes
-      @_routes || raise(<<~MSG)
+      (@_routes || build_routes) || raise(<<~MSG)
```
timriley, line 43: Should we also remove the `.define` method (above) as well as this `@_routes` ivar?

Unless there's potential drawbacks to the new class-level approach of routes definition, it feels like we'd be smart to provide just one way to define routes in this class.

timriley (approved):
Nice and simple change in the end, kudos @solnic! Couple q's but otherwise happy to see this go in :)

timriley:
> Actually, this class could be utilized better. It feels like the complexity of defining the router could be moved to `Hanami::Routes`. We could have something like `Hanami::Routes#to_router` and an instance of the routes that we'd create would know what to do. I also wonder if this class wouldn't be a good place to store all kinds of information about the router and routes for things like helpers etc.

I'm in fierce agreement with you here!

What I aways had in mind for this class was that it should be the thing that canonically  holds the _route definitions_ and provides helpful facilities around them. For example, it could be the thing that is registered in our containers and holds the routing helpers, such that it could be auto-injected into any component that needs them (this IMO is done sub sub-optimally right now, see `lib/hanami/providers/routes.rb` and the `RoutesHelper` class). It could also be a place where users can define any custom routing helpers of their own — things like this is the benefit of it being a concrete class in the first place!

As for whether this can actually vend a fully fledged router, I'm definitely open to see some experiments here, though if it were to be widely auto-injected, perhaps having a `#to_router` on it might be inadvisable... I think I'd rather that be completely separated in this kind of arrangement.

Keen to continue the discussion here, this is one of the last rough bits that would be good to get right in the next month or two.

solnic:
@timriley yes exactly 🙂 I'd give this a shot and initially just move Router initialization to Routes instance and make routes instance available via container. This would be a good start.

PR #1209, opened 2022-09-08
Make router configuration optional

waiting-for-dev:
It restores #1204, which was reverted in https://github.com/hanami/hanami/commit/f627f2101475fb348d348bd5a2b4d5cf70005067, without the raising part. I.e., we're making the router configuration optional, but we keep using the null configuration pattern.

We're also adding integration tests to ensure that `Hanami.setup` works when the optional gems are unavailable.


timriley (approved):
Looks great, thank you @waiting-for-dev!

PR #1212, opened 2022-09-30
Fix not taking nested slices for Hanami::SliceConfigurable

waiting-for-dev:


waiting-for-dev:
[SliceConfigurable needs to find slices in a way that can work with nested slices](https://trello.com/c/FfWsL9dO/202-sliceconfigurable-needs-to-find-slices-in-a-way-that-can-work-with-nested-slices)

solnic:
Why do we need both `slices` and `slices_registrar`? Can't we just have `slices` returning registrar instance?

timriley:
Yeah, I'm in agreement with @solnic here on the naming. 

I think @waiting-for-dev is going to be overhauling this PR, however, now that he and I got on the same page about its intended purpose, so I wouldn't consider its current state intended for merge. Is that a fair summary, @waiting-for-dev?

waiting-for-dev:
Yeah, what @timriley said is correct.

waiting-for-dev:
Hey, @timriley and @solnic, this is the minimal change for the card. I still think there's room for better naming & encapsulation, but it can be done later if you feel it at that point.

lib/hanami/slice_registrar.rb
```diff
@@ -75,6 +75,12 @@ def to_a
       slices.values
     end
 
+    def with_nested
+      to_a.map do |slice|
+        [slice] + slice.slices.with_nested
+      end.flatten
```
timriley, line 81: Mind making this change, @waiting-for-dev? Just keeps the brace usage consistent with the rest of the file, and `flat_map` lets us remove the trailing `flatten`.

```suggestion
      to_a.flat_map { |slice|
        [slice] + slice.slices.with_nested
      }
```

spec/unit/hanami/slice_configurable_spec.rb
```diff
@@ -28,15 +28,15 @@ def self.inherited(subclass)
...
-    Hanami.app.prepare
   end
 
   context "subclass inside slice namespace" do
     before do
       module Main
         class MySubclass < TestApp::BaseClass; end
       end
+
+      Hanami.app.prepare
```
timriley, line 39: I think we can actually remove all the `prepare` calls from the test altogether. I've just double checked this without them an everything passes. Would you mind taking care of that?

timriley (approved):
Wonderful stuff, @waiting-for-dev! I love how simple this change was in the end :)

Left you a couple of suggestions for changes here. Would be happy for you to merge this as soon as you've taken care of those.

lib/hanami/slice_registrar.rb
waiting-for-dev: Good point. Done!

spec/unit/hanami/slice_configurable_spec.rb
waiting-for-dev: 👌 I realized that, but I thought it was there to reflect the actual setup. Removed!

PR #1213, opened 2022-09-30
Make settings available in app/slice class bodies

timriley:
This change restores class-level access to `settings` within app and slice classes. This is important because it allows user-defined settings to be used to configure the app itself. For example:

```ruby
module MyApp
  class App < Hanami::App
    config.sessions = :cookie, {
      key: "my_app.session",
      secret: settings.session_secret, # <=== here
      expire_after: 60*60*24*365
    }
  end
end
```

To support this, we:

- No longer use a container provider to load the settings, since providers can only be used after the slice `prepare` step, which is too late for immediate access inside the class body.
- Remove the expectation that autoloadable constants are available inside the settings classes, since autoloading is also only available after the slice `prepare` step.
- Assist with easy use of type objects for coercing/validating settings by auto-generating a nested `Types` constant within `Hanami::Settings` subclasses. This is intended to replace the use of the `MyApp::Types` module defined in `lib/my_app/types.rb` or `app/types.rb`.

A note about the types module:

Since types for settings tend to be simple, this kind of inline types definition should be sufficient for the majority of cases, and is also the easiest for us to describe in documentation and other supporting material.

Users requiring more advanced usage of types (such as sharing of types between the settings classes and elsewhere) can do so by adding their own manual `require` statements to the class as needed. By doing this and eschewing the provided nested types module, this is a clear signal to the user that they're going off the "golden path", and as such, their expectations around support via documentation etc. should be adjusted accordingly.


timriley:
[Add safer access to ENV at the application class level](https://trello.com/c/dQpVV2vn/236-add-safer-access-to-env-at-the-application-class-level)

solnic:
I agree that option 2 sounds like a better and simpler solution

solnic (approved):
❤️ 🚀

lib/hanami/slice.rb
```diff
@@ -270,6 +281,15 @@ def prepare_all
         prepare_container_providers
       end
```
solnic, line 282: I really think this should be made abstract at some point post 2.0. We could have something like boot/preparation stages and a public API to add your own. It should also be possible to define dependencies between stages so that we can be sure things are run in the corrected order. Proper exception handling with meaningful messages could be achieved too as a nice side-effect, so that IF your app crashes during boot, you don't get an obscure stack trace and instead you clearly see what happened (with a trace of stage execution and the final error message).

Personally, I think Hanami boot/preparation process shouldn't be hardcoded and/or hidden as some kind of a special magical process that happens behind the scenes. It should be public and documented as it's a fundamental piece of the stack. This is another thing that would make DX just nicer.

I can write an issue about it and add to trello if you think this would make sense.

timriley: @solnic 100% agree. I've always considered this manually-managed app boot to be a compromise we have to make in order to get something shipped for 2.0.

For a subsequent release I'd love to turn it into something more generic, and something that provides hooks for users or third-party gem providers to add their own boot logic into.

PR #1215, opened 2022-10-04
Improved middleware setup

solnic:
This improves how loading middlewares work:

- You can pass a symbol now and it'll be resolved automatically ie `config.middleware.use(:body_parser, :json)`
- You can configure custom namespaces that define middlewares ie `config.middleware.namespaces = [Hanami::Middleware, MyStuff::Middlewares]` - the first one with a given constant wins
- We try to require `hanami/middleware/#{middleware_identifier}` automatically for better DX

Overall, this helps because you don't have to deal with requires, you don't need to know where hanami defines its middlewares etc.

### DIFF
--- a/lib/hanami/errors.rb
+++ b/lib/hanami/errors.rb
@@ -12,4 +12,7 @@ module Hanami
 
   # @since 2.0.0
   ComponentLoadError = Class.new(Error)
+
+  # @since 2.0.0
+  UnsupportedMiddlewareSpecError = Class.new(Error)
 end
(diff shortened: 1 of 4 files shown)


solnic:
[Body parsing should be built into hanami-controller, not be (solely) a standalone middleware](https://trello.com/c/dh4Wtjea/110-body-parsing-should-be-built-into-hanami-controller-not-be-solely-a-standalone-middleware)

timriley (approved):

PR #1216, opened 2022-10-06
Use new class-level Hanami::Action settings/config

timriley:
This is an update to work with the latest dry-configurable and the hanami-controller changes in https://github.com/hanami/controller/pull/392.

I'll merge this after the hanami-controller PR is merged.


timriley:
[Update Hanami::Action to use class-level config and latest dry-configurable (reduce memory usage)](https://trello.com/c/ExtpYdCm/208-update-hanamiaction-to-use-class-level-config-and-latest-dry-configurable-reduce-memory-usage)

PR #1217, opened 2022-10-07
Hide values when inspecting Hanami settings

waiting-for-dev:
Users can explicitly ask for them to be rendered through
`#inspect(show_values: true)`.


waiting-for-dev:
[Application settings object should not leak its values through `#inspect`](https://trello.com/c/kI0s6e9m/127-application-settings-object-should-not-leak-its-values-through-inspect)

spec/unit/hanami/settings_spec.rb
```diff
@@ -43,6 +43,30 @@
...
 
+  describe "#inspect" do
+    it "hides values" do
+      config_class = Class.new(described_class) do
+        setting :password, default: "dont_tell_anybody"
+        setting :passphrase, default: "shhhh"
+      end
+
+      expect(config_class.new.inspect).to eq(
+        "#<#{config_class.to_s} password: [:hidden:], passphrase: [:hidden:]>"
```
timriley, line 54: The repetition of `[:hidden:]` here feels like a lot of visual noise, and the use of the square braces may (momentarily at least) confuse users that these are actually array values.

I wonder if there's a more concise display mode we could use for when we're hiding values?

Just remove the values entirely?

```
#<MyApp::Settings password:, passphrase:>
```

I could see this being confusing too, however.

Perhaps we could use a different way of denoting elided values?

```
#<MyApp::Settings password: ..., passphrase: ...>
```

Or perhaps take another approach altogether when values are hidden, instead condensing it down to a list of setting names only?

```ruby
#<MyApp::Settings[:password, :passphrase]>
```

Or:

```ruby
#<MyApp::Settings settings=[:password, :passphrase]>
```

What do you think? I'd definitely like us to find an alternative to the repeated `[:hidden:]` strings, since I think that's going to feel ugly in practice.

timriley (changes requested):
Thanks for taking this on, @waiting-for-dev!

I think it'd be good if we could explore a different way of handling the hidden values... left a note for you inline. Keen to hear your thoughts :)

spec/unit/hanami/settings_spec.rb
waiting-for-dev: Thanks for your review, @timriley. I think your concerns are very valid.

I think I'd prefer not to change the structure of the output, as it'll require less effort to adapt the eyes if `show_values: true` is given just afterward.

My option would therefore be:

```ruby
#<MyApp::Settings password: ..., passphrase: ...>
```

I also thought about helping users with the following:

```ruby
#<MyApp::Settings password: ..., passphrase: ...> (run `#inspect(show_values: true)` to render elided values)
```

I know that's not the best place for documentation, but it could significantly help while debugging. WDYT?

timriley: Yeah, I think your last example is a good compromise! My only suggestion would be to word it "to show elided values" ("show" not "render") since it is simpler language and actually fits better with the `show_values` option name.

spec/unit/hanami/settings_spec.rb
```diff
@@ -43,6 +43,30 @@
     end
   end
 
+  describe "#inspect" do
+    it "hides values" do
+      config_class = Class.new(described_class) do
```
timriley, line 48: Minor nitpick, this is actually a `settings_class`

spec/unit/hanami/settings_spec.rb
waiting-for-dev: Makes sense.

I'm also not particularly happy with the boolean parameter. What do you think about adding another `#inspect_raw` instead?

solnic: So, my suggestions:

1) `Settings#inspect` shows `#<MyApp::Settings[foo, bar, baz]>`
2) `Settings#inspect_values` shows `#<MyApp::Settings foo="value1" bar="value2">`

I don't think adding info about how to display values is a good idea as it would clutter the output.

solnic (changes requested):
Thanks for tackling this! I left a couple of suggestions.

spec/unit/hanami/settings_spec.rb
waiting-for-dev: Updated.

spec/unit/hanami/settings_spec.rb
waiting-for-dev: @timriley, @solnic, after thinking about it, I feel that @solnic proposal is the least surprising and more semantic. I've done just that. Let me know what you think.

waiting-for-dev:
I don't have the time to look into the CI failure now, but it looks unrelated 🤔

timriley:
@waiting-for-dev If you rebase this branch, the spec failures should go away

spec/unit/hanami/settings_spec.rb
timriley: I'm very happy with this approach!

timriley (approved):
Looking good now, let's get this merged!

PR #1221, opened 2022-10-13
Add a spec for configurable body parser

solnic:
This only adds a spec covering new functionality of the router which looks like this:

```ruby
module MyApp
  class App < Hanami::App
    config.actions.formats["application/json+scim"] = :json

    # This thing right here is new - you can now pass an array with format/mime-type
    # map that a given parser should handle in addition to what it specifies by default
    config.middleware.use :body_parser, [json: "application/json+scim"]
  end
end
```

# TODO

- [x] go back to router from main once hanami/router#230 is merged


solnic:
[Support activating built-in body parser middlewares for custom mime types](https://trello.com/c/6ousYBxH/247-support-activating-built-in-body-parser-middlewares-for-custom-mime-types)

PR #1222, opened 2022-10-13
Raise errors for missing settings

timriley:
In https://github.com/hanami/hanami/pull/1216 (specifically, https://github.com/hanami/hanami/pull/1216/commits/d4c036b32a4471a13d70a15bc7f8cef9b5c4509f), we made a small tweak to our settings loading to adapt to changed handling of `Undefined` values in dry-configurable 0.16.0.

The change in dry-configurable is that `Undefined` values sent to config no longer implicitly fall back to using their default value; `Undefined` instead is accepted as the value itself (though not run through the constructor).

To work around this, we stopped setting values on our Settings config object if no such key existed in the store (i.e. if no ENV var existed at all).

However, this turned out to have the undesirable side effect of missing ENV vars no longer triggering exceptions from their constructors. That is, if a setting is required, but not defined in ENV at all, it would no longer halt the boot of the Hanami app and provide early feedback to the user.

This is critical behavior for Hanami, so here we restore it through attempting to _read_ the config value for any keys missing from ENV. This ensures the consrtuctor is run (as it is done for all config values read for the first time) and errors are appropriately raised.

Add a new unit test for this specific scenario at the same time.

### DIFF
--- a/lib/hanami/settings.rb
+++ b/lib/hanami/settings.rb
@@ -139,6 +139,8 @@ module Hanami
       end
 
       raise InvalidSettingsError, errors if errors.any?
+
+      config.finalize!
     end
 
     def inspect
--- a/lib/hanami/settings.rb
+++ b/lib/hanami/settings.rb
@@ -125,9 +125,15 @@ module Hanami
     def initialize(store = EMPTY_STORE)
       errors = config._settings.map(&:name).each_with_object({}) do |name, errs|
         value = store.fetch(name, Undefined)
-        next if value.eql?(Undefined)
 
-        public_send("#{name}=", value)
+        if value.eql?(Undefined)
+          # When a key is missing entirely from the store, _read_ its value from the config instead,
+          # which ensures its setting constructor runs (with a `nil` argument given) and raises any
+          # necessary errors.
+          public_send(name)
+        else
+          public_send("#{name}=", value)
+        end
       rescue => e # rubocop:disable Style/RescueStandardError
         errs[name] = e
       end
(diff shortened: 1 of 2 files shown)


timriley:
@solnic This fixes the issue you raised with me last week about the settings no longer halting app boot in certain circumstances.

@andrewcroome This is also the same bug you raised with me last night :)

Since you both observed this in the wild, is there any chance you could run a quick test against this branch to ensure you're happy things are working again as you expect?

timriley:
[Fix issue where Hanami::Settings did not raise errors for missing settings](https://trello.com/c/dusOXMAF/257-fix-issue-where-hanamisettings-did-not-raise-errors-for-missing-settings)

lib/hanami/settings.rb
```diff
@@ -125,9 +125,15 @@ def to_s
...
 
-        public_send("#{name}=", value)
+        if value.eql?(Undefined)
+          # When a key is missing entirely from the store, _read_ its value from the config instead,
+          # which ensures its setting constructor runs (with a `nil` argument given) and raises any
+          # necessary errors.
+          public_send(name)
+        else
+          public_send("#{name}=", value)
+        end
```
timriley, line 136: Actually, another way to do this might be just to assign the values only still, and then run `config.finalize!` at the end of this method, which would make it so the config cannot be subsequently written to, which is a positive side effect. I might try that in a few hours.

solnic (approved):
I tested it out and it worked as expected

solnic:
@timriley yes this makes perfect sense

spec/unit/hanami/settings_spec.rb
```diff
@@ -41,6 +44,32 @@
...
+      settings_class = Class.new(described_class) do
+        setting :database_url
+      end
+
+      store = {database_url: "postgres://localhost/test_app_development"}
+
+      settings = settings_class.new(store)
+
+      expect(settings.config).to be_frozen
+      expect { settings.database_url = "new" }.to raise_error(Dry::Configurable::FrozenConfig)
```
timriley, line 71: Jus a heads up: I'll need to change this to FrozenConfigError once we make the next dry-configurable release.

PR #1224, opened 2022-10-15
Rely on auto-loadable dry-system

solnic:
# TODO

- [x] `./spec/unit/hanami/config/router_spec.rb` this spec is now unstable due to how `bundled?` is implemented (I think) so it's something to address

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -113,8 +113,7 @@ module Hanami
   def self.bundled?(gem_name)
     @_mutex.synchronize do
       @_bundled[gem_name] ||= begin
-        gem(gem_name)
-        true
+        Bundler.gem(gem_name)
       rescue Gem::LoadError
         false
       end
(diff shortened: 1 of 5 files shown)


timriley (approved):
Let's get this in! We want the main branch to... work :)

lib/hanami.rb
```diff
@@ -113,8 +113,7 @@ def self.shutdown
   def self.bundled?(gem_name)
     @_mutex.synchronize do
       @_bundled[gem_name] ||= begin
-        gem(gem_name)
-        true
+        Bundler.gem(gem_name)
```
timriley, line 116: @solnic I was fine for us to merge this PR wholesale to get the main branch working again, but I'm interested in learning why you had to make this change?

solnic: @timriley looks like it was not needed. I changed it back in main.

PR #1226, opened 2022-10-22
Use Zeitwerk to auto-load hanami

solnic:


timriley:
@solnic FYI, I've rebased this over `main` and force pushed.

timriley (approved):
Tests are green and we'll be actively testing Hanami over the next few weeks before release. I reckon we should get this in 🚀

PR #1228, opened 2022-10-25
Add Slice.stop forwarding to container

timriley:
This is needed to round out the .prepare and .start methods already forwarding to the container.

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -153,6 +153,10 @@ module Hanami
         container.start(...)
       end
 
+      def stop(...)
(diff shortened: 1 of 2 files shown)


solnic (approved):
I understand that this is trivial but we need a spec for it regardless 😬

timriley:
Yeah, I'll take care of this before merging :) 

Thanks for being my conscience, haha

timriley:
[Add missing `Slice.stop` method](https://trello.com/c/V5dksxzA/270-add-missing-slicestop-method)

PR #1229, opened 2022-10-27
Update API docs for 2.0.0

timriley:
Add Yard API docs to everything such that this command returns (almost) no undocumented methods:

```
yard list --query 'o.docstring == "" && o.tag(:api).text != "private" && o.tags(:overload).empty? && o.tags(:see).empty?'
```

(This particular incantation is necessary because yard treats `@since` as transitive, and as such a single `@since` tag on a class will mark all its methods as "documented")

The two files still to be documented are `lib/hanami/config/sessions.rb` and `lib/hanami/config/actions/sessions.rb`. These are duplicates of each other, and I plan to update their docs in another PR that removes this duplication.

I also re-wrapped any doc strings at 100 characters.

While making these changes, make a few minor code tweaks at the same time:

- Made `Config#logger` a standard ivar and reader method instead of a dry-configurable setting, just for consistency with the rest of our "nested" configs
- Changed `config.logger.colors` to have a simple `default: false`, rather than `false` in production and test and `nil` in all other environments (these are effectively equivalent and `default: false` is a much clearer implementation)
- Changed `Hanami::Config::Actions::ContentSecurityPolicy#to_str` to `#to_s`

### DIFF
--- a/lib/hanami/assets/app_config.rb
+++ b/lib/hanami/assets/app_config.rb
@@ -5,20 +5,19 @@ require_relative "config"
 
 module Hanami
   module Assets
-    # @since 2.0.0
-    # @api public
+    # App config for assets.
+    #
+    # This is not released as of 2.0.0.
+    #
+    # @api private
     class AppConfig
       include Dry::Configurable
 
-      # @since 2.0.0
-      # @api private
       attr_reader :base_config
       protected :base_config
 
       setting :server_url, default: "http://localhost:8080"
 
-      # @since 2.0.0
-      # @api private
       def initialize(*)
         super
 
@@ -30,8 +29,6 @@ module Hanami
         @base_config = source.base_config.dup
       end
 
-      # @since 2.0.0
-      # @api private
       def finalize!
       end
 
@@ -39,7 +36,6 @@ module Hanami
       #
       # @return [Set]
       #
-      # @since 2.0.0
       # @api private
       def settings
         base_config.settings + self.class.settings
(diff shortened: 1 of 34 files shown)


lib/hanami/slice.rb
```diff
@@ -89,23 +130,101 @@ def root
...
+      # @see Config#inflector
+      # @see Config#inflections
+      #
+      # @api public
+      # @since 2.0.0
       def inflector
         config.inflector
       end
 
+      # @overload preapre
```
solnic, line 148:
```suggestion
      # @overload prepare
```

timriley: Thanks :)

lib/hanami/assets/app_config.rb
```diff
@@ -4,21 +4,21 @@
...
 module Hanami
+  # @api private
   module Assets
-    # @since 2.0.0
-    # @api public
+    # App config for assets.
+    #
+    # This is NOT RELEASED as of 2.0.0.
+    #
+    # @api private
```
timriley, line 13: Leaving this note as well as `@api private` for any code inside hanami that's not officially part of 2.0.0 (e.g. views, assets, etc.)

timriley:
@jodosha @solnic Since this is such an expansive diff, it's going to be difficult to keep it cleanly mergeable into main. As such, I'm going to merge it straight away (after all, _some_ docs is unequivocally better than _no_ docs, right!), but I'm definitely happy to do some post-merge discussion here in this PR, which we can feed into future tweaks to the docs.

timriley:
[Prepare API docs for hanami](https://trello.com/c/YE0SbRwI/263-prepare-api-docs-for-hanami)

timriley:
[Hanami::Config::Actions::ContentSecurityPolicy#to_str should really just be #to_s](https://trello.com/c/yxab15AB/280-hanamiconfigactionscontentsecuritypolicytostr-should-really-just-be-tos)

PR #1230, opened 2022-10-28
Remove check for slices root

waiting-for-dev:
We're no longer allowing slices with no associated root, so we were
checking for a condition that is never met.


timriley (approved):
🚀

PR #1232, opened 2022-10-30
Improve router missing action errors

### LINKED ISSUE #1231: Partial name in action registry
I have the following route:

```ruby
module TestApp
  class Routes < Hanami::Routes
    root to: "home.show"
    # ....

    slice :admin, at: "/admin" do
      root to: "home.show"
    end
  end
end
```

But WITHOUT the corresponding action.
When I start the app via `hanami server` and try to reach that endpoint, the app crashes (expected) with the following message:

```shell
⚡ curl -I http://0.0.0.0:2300/admin
HTTP/1.1 500 Internal Server Error
Content-Length: 2977
```

```shell
2022-10-29 13:57:23 +0200 Rack app ("HEAD /" - (127.0.0.1)): #<Hanami::Slice::Routing::UnknownActionError: unknown action referenced in router: `"actions.home.show"'>
```

---

I see a few problems with this exception:

* It's hard to correlate which action is missing. I have two actions with the same identifier `"home.show"`, at the first sight I don't know if it's the one at the root, or the one inside `admin` slice.
* It doesn't tell me how to fix the problem.

---

https://trello.com/c/NwYUAL8n

timriley:
When a route contains a key for a missing action, raise a (renamed) `Hanami::Routes::MissingActionError` with a full message like the following:

```
Could not find action with key "actions.missing.action" in Admin::Slice

To fix this, define the action class Admin::Actions::Missing::Action in /full/path/to/slices/admin/actions/missing/action.rb
```

Resolves #1231

### DIFF
--- a/lib/hanami/routes.rb
+++ b/lib/hanami/routes.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-require "hanami/slice/router"
+require_relative "constants"
+require_relative "slice/router"
 
 module Hanami
   # App routes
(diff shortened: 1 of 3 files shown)


timriley:
[Partial name in action registry](https://trello.com/c/NwYUAL8n/272-partial-name-in-action-registry)

jodosha (approved):
@timriley Thanks for jumping quickly into this!

PR #1233, opened 2022-10-30
Fix flaky tests for config with non-bundled gems

timriley:
In the cases of these gems not being bundled, we can’t count on their paritcular config classes being loaded.


timriley:
[Fix flaky tests in hanami/hanami](https://trello.com/c/CwQSFoKx/273-fix-flaky-tests-in-hanami-hanami)

PR #1234, opened 2022-10-31
Fix slice Rack middleware

jodosha:
# Fix

Ensure to properly mount Rack middleware in routing `scope` and `slice`.

---

https://trello.com/c/4otUMpaN


spec/integration/rack_app/middleware_spec.rb
```diff
@@ -235,4 +235,424 @@ def handle(req, res)
...
+                def handle(req, res)
+                  res.body = "Hello from admin (User ID " + req.env['AUTH_USER_ID'] + ")"
+                end
+              end
+            end
+          end
+        end
+      RUBY
+
+      # write "slices/api_v1/actions/home/show.rb", <<~RUBY
```
jodosha, line 569: @timriley If I uncomment this file, I get the following error:

```shell
⚡ bundle exec rspec ./spec/integration/rack_app/middleware_spec.rb:343
Failures:

  1) Hanami web app with complex app uses Rack middleware
     Failure/Error: config.public_send(name, *args, &block)

     Dry::Configurable::FrozenConfig:
       Cannot modify frozen config
# ...
```

I don't understand why the other action above doesn't cause this problem, but this one does.

solnic (approved):
I'm approving because the specs passed...otherwise I'm unable to review because I've no idea how things work "on the other side" 😅

timriley (approved):
Masterful work, @jodosha! I'm delighted with the end result here — this is such a powerful feature for Hanami!

From a quick glance at https://github.com/hanami/hanami/pull/1234, is it true that the other side of this work was mostly relocating code from hanami-api to hanami-router so that it could become accessible to us here in the hanami gem too?

Certainly having the `Rack::Builder` detail move outside of hanami and become a concern of `Hanami::Middleware::App` seems a win as well.

Anyway, very happy to see this go in 👍🏼

jodosha:
@timriley Yes, we had the chance to extract code from `hanami-api`, which was the first gem to have this feature.

PR #1235, opened 2022-11-01
Keep only a single sessions config class

timriley:
Remove `config.sessions` and keep `config.actions.sessions`.

Certain implementation details in the two underlying classes were slightly different (the former used a "null" sessions objected while the latter doesn't), but the external behaviour is the same and our continued passing tests are showing this.


timriley:
[DID YOU KNOW: We have **both** config.sessions _and_ config.actions.sessions — I'm going to remove config.sessions](https://trello.com/c/3M3mgSh0/279-did-you-know-we-have-both-configsessions-and-configactionssessions-im-going-to-remove-configsessions)

jodosha (approved):

PR #1236, opened 2022-11-01
Use dry-logger

solnic:
There are three changes in behavior:

- `filters` are now passed as `filters` to the logger constructor
- `logger_constructor` option was added for more flexibility
- `logger.template` is configured to include progname by default (previously known as the application formatter)

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -164,7 +164,7 @@ module Hanami
   #     end
   #   end
   #
-  # @return [Hanami::Logger]
+  # @return [Dry::Logger::Dispatcher]
   #
   # @api public
   # @since 1.0.0
--- a/lib/hanami/config/logger.rb
+++ b/lib/hanami/config/logger.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require "dry/configurable"
-require "hanami/logger"
+require "dry/logger"
 
 module Hanami
   class Config
@@ -45,14 +45,13 @@ module Hanami
       # @!attribute [rw] formatter
       #   Sets or returns the logger's formatter.
       #
-      #   This may be a name that matches a formatter registered with `Hanami::Logger`, which
-      #   includes `:default` and `:json`.
+      #   This may be a name that matches a formatter registered with `Dry::Logger`, which includes
+      #   `:string`, `:rack` and `:json`.
       #
       #   This may also be an instance of Ruby's built-in `::Logger::Formatter` or any compatible
       #   object.
       #
-      #   Defaults to `:json` for the production environment, and `nil` for all others. A `nil`
-      #   value will result in a plain `::Logger::Formatter` instance.
+      #   Defaults to `:json` for the production environment, and `:rack` for all others.
       #
       #   @return [Symbol, ::Logger::Formatter]
       #
@@ -60,8 +59,8 @@ module Hanami
       #   @since 2.0.0
       setting :formatter
 
-      # @!attribute [rw] colors
-      #   Sets or returns whether log lines should be colorized.
+      # @!attribute [rw] template
+      #   Sets or returns log entry string template
       #
       #   Defaults to `false`.
       #
@@ -69,7 +68,7 @@ module Hanami
       #
       #   @api public
       #   @since 2.0.0
-      setting :colors, default: false
+      setting :template, default: "[%<progname>s] [%<severity>s] [%<time>s] %<message>s"
 
       # @!attribute [rw] filters
       #   Sets or returns an array of attribute names to filter from logs.
@@ -83,29 +82,26 @@ module Hanami
       #   @since 2.0.0
       setting :filters, default: %w[_csrf password password_confirmation].freeze
 
-      # @!attribute [rw] logger_class
-      #   Sets or returns the class to use for the logger.
+      # @!attribute [rw] logger_constructor
+      #   Sets or returns the constructor proc to use for the logger instantiation.
       #
-      #   This should be compatible with the arguments passed to the logger class' `.new` method in
-      #   {#instance}.
-      #
-      #   Defaults to `Hanami::Logger`.
+      #   Defaults to `Dry.method(:Logger)`.
       #
       #   @api public
       #   @since 2.0.0
-      setting :logger_class, default: Hanami::Logger
+      setting :logger_constructor, default: Dry.method(:Logger)
 
       # @!attribute [rw] options
-      #   Sets or returns an array of positional arguments to pass to the {logger_class} when
-      #   initializing the logger.
+      #   Sets or returns a hash of options to pass to the {logger_constructor} when initializing
+      #   the logger.
       #
       #   Defaults to `[]`
       #
-      #   @return [Array<Object>]
+      #   @return [Hash]
       #
       #   @api public
       #   @since 2.0.0
-      setting :options, default: [], constructor: ->(value) { Array(value).flatten }
+      setting :options, default: {}
 
       # Returns a new `Logger` config.
       #
@@ -135,6 +131,8 @@ module Hanami
         config.formatter = case env
                            when :production
                              :json
+                           else
+                             :rack
                            end
       end
 
@@ -145,19 +143,22 @@ module Hanami
       # @api public
       # @since 2.0.0
       def instance
-        logger_class.new(
-          app_name.name,
-          *options,
-          stream: stream,
-          level: level,
-          formatter: formatter,
-          filter: filters,
-          colorizer: colors
-        )
+        logger_constructor.call(app_name.name, **logger_constructor_opts)
       end
 
       private
 
+      # @api private
+      def logger_constructor_opts
+        {stream: stream,
+          level: level,
+          formatter: formatter,
+          filters: filters,
+          template: template,
+          **options}
+      end
+
+      # @api private
       def method_missing(name, *args, &block)
         if config.respond_to?(name)
           config.public_send(name, *args, &block)
@@ -166,6 +167,7 @@ module Hanami
         end
       end
 
+      # @api private
       def respond_to_missing?(name, _incude_all = false)
         config.respond_to?(name) || super
       end
--- a/lib/hanami/web/rack_logger.rb
+++ b/lib/hanami/web/rack_logger.rb
@@ -57,8 +57,7 @@ module Hanami
           ip: env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR],
           path: env[SCRIPT_NAME] + env[PATH_INFO].to_s,
           length: extract_content_length(env),
-          params: env[ROUTER_PARAMS],
-          time: Time.now,
+          params: env[ROUTER_PARAMS]
         }
 
         logger.info(data)
--- a/spec/unit/hanami/config/logger_spec.rb
+++ b/spec/unit/hanami/config/logger_spec.rb
@@ -7,23 +7,15 @@ require "logger"
 require "stringio"
 
 RSpec.describe Hanami::Config::Logger do
-  subject { described_class.new(app_name: app_name, env: env) }
-  let(:app_name) { Hanami::SliceName.new(double(name: "MyApp::app"), inflector: -> { Dry::Inflector.new }) }
-  let(:env) { :development }
-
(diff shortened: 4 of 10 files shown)


solnic:
[We're still using Hanami::Logger — should this be a Dry::Logger? (Can we release dry-logger in its current state as 1.0.0 so we can use it?)](https://trello.com/c/ckcYO3Kd/277-were-still-using-hanamilogger-should-this-be-a-drylogger-can-we-release-dry-logger-in-its-current-state-as-100-so-we-can-use-it)

solnic:
Notice that I had to do [this](https://github.com/dry-rb/dry-logger/commit/116e47a640f13b8a46f9da20b064b4268e03e440) in dry-logger.

In general, I should clean things up a bit there because I originally implemented this weird `Dry::Logger.new` method just to make things compatible with `Hanami::Logger` but I don't think it makes sense now. Other than that, I'd just add API docs, maybe add a bit more specs and release dry-logger 1.0.0.rc1.

lib/hanami.rb
```diff
@@ -164,7 +164,7 @@ def self.env?(*names)
   #     end
   #   end
   #
-  # @return [Hanami::Logger]
+  # @return [Dry::Logger::Dispatcher]
```
timriley, line 167: Question about the multi-backend dispatching that we have in place with `Dry::Logger::Dispatcher`: do we actually need it? It does make dry-logger more challenging to understand (think about this case right here, a user may ask "what is a  _dispatcher_ and why am I getting one of them instead of an ordinary _logger?_") for what feels like a fairly obtuse feature. What did you have in mind when adding this in, @solnic?

timriley (commented):
Changes here look straightforward and sensible to me, though I do have a couple of existential questions and a small suggestion or two.

Let me know what you think, @solnic :)

lib/hanami.rb
jodosha: @solnic I agree.

flash-gordon: Shouldn't `Dry::Logger` and `Dry::Logger::Dispatcher` implement the same interface? So that we have `Dry::Logger` in annotations but the end user can swap a simple logger with one having multiple backends (with whatever DSL we provide them with)? The notion of multiple backends shouldn't affect the app in any way, right? @solnic @timriley

flash-gordon: for the record, I haven't seen dry-logger yet :)

solnic: The whole purpose of dry-logger is to provide a multi-backend logging, otherwise you can just use logger from stdlib. This is extremely useful as it allows you to configure different types of logging for different environments, contexts, sub-apps etc.

This gives you very powerful logging capabilities and I **really** want Hanami to have this provided by default as it will be a big advantage.

Logging/observability has been typically very primitive in the Rails world precisely due to having just one logger that has one destination. Then people would struggle adding other loggers or even other gems that provide more powerful logging features (ie see https://logger.rocketjob.io/). This was incoherent and because of this I've a strong opinion that a framework should provide this built-in, with conventions how things should be configured.

I can give you a simple example that's pretty common these days where advanced logging is needed: you run things on GCP, it's got very powerful logging/alerts/metrics features but you have to log in a format that it understands. It's *not needed* in local environments though. So, how do you do this? If you have only one logger with one destination then you need a different logger for production env. This is really cumbersome to handle though and it's easy to end up with inconsistent logging. Now, when you have a powerful logger with support for multiple destinations, you simply set up a GCP destination (aka backend) in production and you're done. Another use case is exception logger - on production you can add a destination logger that logs exceptions to ie Rollbar or Sentry *but* in development you see exceptions in a simple log file. This kind of scenarios are much simpler to handle when you have a unified logging interface!

This is a real game-changer when it comes to logging and it will be, for sure, one of the selling points of Hanami.

lib/hanami/config/logger.rb
```diff
@@ -83,29 +83,35 @@ class Logger
...
+      #   @since 2.0.0
+      setting :logger_constructor, default: Dry.method(:Logger)
+
       # @!attribute [rw] logger_class
       #   Sets or returns the class to use for the logger.
       #
-      #   This should be compatible with the arguments passed to the logger class' `.new` method in
-      #   {#instance}.
-      #
-      #   Defaults to `Hanami::Logger`.
```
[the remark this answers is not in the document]
solnic: I agree. I left it there because I wanted to maintain the original functionality, but if  we're fine with removing it then I'm happy to do it in this PR.

lib/hanami/config/logger.rb
```diff
@@ -145,19 +153,26 @@ def initialize(env:, app_name:)
...
 
       private
 
+      # @api private
+      def logger_constructor_opts
+        {stream: stream,
+          level: level,
+          formatter: formatter,
+          filters: filters,
+          colorizer: colors,
```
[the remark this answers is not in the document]
solnic: Oh good catch, I was aware of this but forgot to remove it 😅

lib/hanami/config/logger.rb
```diff
@@ -135,6 +141,8 @@ def initialize(env:, app_name:)
         config.formatter = case env
                            when :production
                              :json
+                           else
+                            :application
```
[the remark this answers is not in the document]
solnic: It's prefixed with the application name but I agree that having a dedicated formatter feels like too much. I can just add a configured string formatter for Hanami purposes *to Hanami* instead.

> Could we just default to JSON logging everywhere?

I don't like JSON logging in development and test envs. It is easier to parse string logs with your eyes. We should absolutely default to JSON in production envs though (which we do).

lib/hanami.rb
flash-gordon: @solnic I didn't question the idea of multiple backends, I can provide you with my examples if you will :) The question was how a dispatcher's interface differs from an ordinary logger for the end user.

There are two stages of working with a logger:
1. Making a logger. This may involve an arbitrary API for setting multiple backends or what have you. It returns a, say, `Dry::Logger`.
2. Working with a logger, that is calling `.info`, `debug`, and so on. In other words, a standard logger interface.

It seems like we conflate the two here.
I think in most the end user should see the second version, which is `Dry::Logger`.
Am I making sense?

solnic: `Dry::Logger` is a namespace, it's not a class. `Dry::Logger.setup` should be the primary interface for setting up a logger. The only reason why there's `Dry::Logger.new` is because originally I've made it like that to maintain compatibility with `Hanami::Logger`. Now it should be just removed. The `Dry::Logger::Dispatcher` implements the same interface as every logging backend, so `info`, `debug` etc. are available and it simply forwards to configured backends.

lib/hanami/config/logger.rb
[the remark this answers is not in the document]
timriley: > It's prefixed with the application name but I agree that having a dedicated formatter feels like too much. I can just add a configured string formatter for Hanami purposes to Hanami instead.

Yeah, I think this would be a better arrangement 👍🏼

lib/hanami.rb
timriley: Thanks for the explanation, @solnic :) Definitely sounds like you've thought about this a lot more than me, so I'm happy to trust your instincts here.

I guess this just leaves the matter of the `Dry::Logger::Dispatcher` name. In isolation I don't mind that name at all — it describes its role well. So I think if we're going to advertise the Hanami logger as possessing additional helpful capabilities like multiple backends, then we may as well just lean into this and expose the `Dispatcher` name wherever it makes sense.

I'm comfortable with this now.

timriley:
Thanks for the all the feedback here, @solnic. On the whole I feel positive about moving forward with this particular change within Hanami. We'll of course want @jodosha's input too, but I feel like we're better off getting onto a future-oriented standalone logger library now rather than waiting, especially if it's already compatible with what Hanami::Logger offered.

Naturally we'll want @jodosha's input, but for me, the biggest remaining question is: how confident are you about the current state of the dry-logger codebase? Is it ready to 1.0.0.rc1 in the next couple days, with a proper 1.0.0 within a week or so after (or even ready to go straight to 1.0.0 without the rc stop)?

lib/hanami.rb
solnic: @timriley we *could* just do `Hanami::Logger < Dry::Logger::Dispatcher` but I dunno

lib/hanami/config/logger.rb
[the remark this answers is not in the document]
solnic: @timriley done via 38e1676537c3c5d0862920902952d19ed0b2b271

solnic:
> how confident are you about the current state of the dry-logger codebase? Is it ready to 1.0.0.rc1 in the next couple days, with a proper 1.0.0 within a week or so after (or even ready to go straight to 1.0.0 without the rc stop)?

It's 💯 good enough to be released as 1.0.0.rc1 along with hanami 2.0.0.rc1. It's *mostly* based on hanami-utils' logger and the logger from stdlib. I'd only remove the `.new` thing in favor of `.setup` and the rest is in a good shape API-wise IMO.

lib/hanami.rb
timriley: @solnic That feels like it'd do more harm than good in the long run, I'm fine with Dry::Logger::Dispatcher

lib/hanami/config/logger.rb
```diff
@@ -83,29 +82,26 @@ class Logger
...
       #   @api public
       #   @since 2.0.0
-      setting :logger_class, default: Hanami::Logger
+      setting :logger_constructor, default: Dry.method(:Logger)
 
       # @!attribute [rw] options
-      #   Sets or returns an array of positional arguments to pass to the {logger_class} when
-      #   initializing the logger.
+      #   Sets or returns a hash of options to pass to the {logger_constructor}
+      #   when initializing the logger.
```
timriley, line 96:
```suggestion
      #   Sets or returns a hash of options to pass to the {logger_constructor} when initializing
      #   the logger.
```

lib/hanami/config/logger.rb
```diff
@@ -135,6 +131,8 @@ def initialize(env:, app_name:)
         config.formatter = case env
                            when :production
                              :json
+                           else
+                            :rack
```
timriley, line 135:
```suggestion
                             :rack
```

spec/unit/hanami/config/logger_spec.rb
```diff
@@ -80,8 +72,8 @@
   end
 
   describe "#formatter" do
-    it "defaults to nil" do
-      expect(subject.formatter).to eq(nil)
+    it "defaults to :application" do
```
timriley, line 75:
```suggestion
    it "defaults to :rack" do
```

timriley (approved):
Looks good! Found a bunch of minor spacing/wrapping/typo issues for you :) Otherwise let's get this in!

lib/hanami/web/rack_logger.rb
```diff
@@ -57,8 +57,7 @@ def log_request(env, status, elapsed)
           ip: env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR],
           path: env[SCRIPT_NAME] + env[PATH_INFO].to_s,
           length: extract_content_length(env),
-          params: env[ROUTER_PARAMS],
-          time: Time.now,
```
jodosha, line 61: @solnic Why? This `time` argument was passed to Rack. I can't find the Rack counterpart at this moment.

How are we measuring elapsed time? Via `dry-monitor`?

jodosha (commented):
@solnic Please clarify this one 🙏 https://github.com/hanami/hanami/pull/1236/files#r1015189183

timriley:
@solnic You also need to add dry-logger to the gemspec, btw

lib/hanami/web/rack_logger.rb
solnic: @jodosha this is just a log timestamp, it's generated on the logger side so no need to instantiate another time instance here

solnic: @jodosha oh and elapsed time is passed from dry-monitor, yes (last argument in `log_request`)

jodosha (approved):
@solnic LGTM. Please add `dry-logger` as a runtime dependency in `hanami.gemspec`.

PR #1238, opened 2022-11-02
Simplify and clarify usage of Config#environment

timriley:
Since https://github.com/hanami/hanami/pull/1153, the `env` cannot be changed for an `Hanami::Config`. This means we do not need to store the environment-specific config blocks, and can instead evaluate them immediately. This is simpler for the user to understand and makes for a much simpler internal implementation too.

Also update the docs to make it clearer that the argument yielded to the `#environment` block should be the object used to set the config. Using this yielded argument is safer than relying on the implicit `self`, since because `Config#config` exists, a user may call `config` inside the block (with implicit self) and receive the internal `Dry::Configurable::Config` when they actually expect `Hanami::Config`.

### DIFF
--- a/spec/unit/hanami/config_spec.rb
+++ b/spec/unit/hanami/config_spec.rb
@@ -11,8 +11,8 @@ RSpec.describe Hanami::Config do
     end
 
     before do
-      config.environment :production do |c|
-        c.logger.level = :info__set_for_production_env
+      config.environment :production do |env|
+        env.logger.level = :info__set_for_production_env
       end
     end
 
--- a/spec/unit/hanami/config_spec.rb
+++ b/spec/unit/hanami/config_spec.rb
@@ -11,8 +11,8 @@ RSpec.describe Hanami::Config do
     end
 
     before do
-      config.environment :production do |env|
-        env.logger.level = :info__set_for_production_env
+      config.environment :production do |c|
+        c.logger.level = :info__set_for_production_env
       end
     end
 
--- a/lib/hanami/config.rb
+++ b/lib/hanami/config.rb
@@ -287,13 +287,15 @@ module Hanami
     # If the given `env_name` matches {Hanami.env}, then the block will be evaluated in the context
     # of `self` via `instance_eval`.
     #
+    # If the env does not match, then the block is not evaluated at all.
+    #
     # @example
     #   config.environment(:test) do |env|
     #     env.logger.level = :info
     #   end
     #
     # @param env_name [Symbol] the environment name
-    # @yieldparam env [self] the config object
+    # @yieldparam c [self] the config object
     #
     # @return [self]
     #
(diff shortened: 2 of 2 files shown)


timriley:
[Clarify eval context for Config#environment](https://trello.com/c/jQw3KOqB/276-clarify-eval-context-for-configenvironment)

solnic (approved):

solnic:
@timriley FYI I rebased this and f-pushed and approved

spec/unit/hanami/config_spec.rb
```diff
@@ -11,8 +11,8 @@
     end
 
     before do
-      config.environment :production do |c|
-        c.logger.level = :info__set_for_production_env
+      config.environment :production do |env|
+        env.logger.level = :info__set_for_production_env
```
jodosha, line 15: @timriley This is confusing 😕 .

All the settings (like `logger`) are usually accessible via `config`.
Why adding more concepts like `env`, as yielded object?

Is it possible to keep `config`?

timriley: This is less a "new concept", just an explicit local variable name. We can call it env_config if you like. Or c. Calling it config causes variable shadowing so I was looking to avoid that. 

As I explained in the commit message, we want to encourage use of the yielded object (at least with the current implementation) because if you just used the bareword `config` inside the block, it would actually return the _internal_ Dry::Configurable::Config object (because this block is being instance_evaled), which would be surprising behavior for someone expecting config there to be the Hanami::Config object. To access that Hanami::Config object inside the block without using the block argument, you have to use self. 

The other option here is to create a custom evaluation context for that block that wraps the Hanami::Config instance and returns _that_ as config, but I felt that would be adding more complication than we might like, especially since we have the block argument we can leverage.

timriley: To put it another way, usage of the bareword `config` inside the block will only work for the accessors defined on the class using `setting`, because they're available as methods on the internal Dry::Configurable::Config object. Any methods declared on Hanami::Config will _not_ be available on the bareword config inside this block. That will be surprising and confusing behavior.  You can use `self` instead of course, but that doesn't feel good either. So by recommending the use of the block arg, we avoid this issue entirely.

jodosha: @timriley OK, so which type is `env`? 🙂

timriley: @jodosha its in the docs I added ;)

```
@yieldparam whatever_name_you_prefer [self]
```

So it's the Hanami::Config instance. This is just the built-in behavior of instance_eval - it yields its receiver to the block. 

Happy to adjust the specs and docs to use a different name for this block param.

jodosha: @timriley `cfg`? or simply `c`?

timriley: @jodosha I'll just call it `c`, then I'll merge this :)

PR #1240, opened 2022-11-06
Expect nested slices to use parent’s namespace

timriley:
For nested slices, define (and expect) their classes to use their parent's namespace. Given either `slices/main/slices/nested/` or `slices/main/config/slices/nested.rb` or `Main::Slice.register_slice(:nested)`, these will all expect or define a `Main::Nested::Slice` class.

It is still possible to opt out of this convention be calling register_slice with an existing slice class, e.g. `register_slice(:nested, SomeOtherNamespace::Slice)`. This kind of opt out requires explicit user intentionality, so at that time I expect them to be aware they're operating outside the framework's ordinary conventions.

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -113,7 +113,7 @@ module Hanami
       # @since 2.0.0
       def config
         @config ||= app.config.dup.tap do |slice_config|
-          # Remove specific values from app that will not apply to this slice
+          # Unset config from app that does apply to ordinary slices
           slice_config.root = nil
         end
       end
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -113,7 +113,7 @@ module Hanami
       # @since 2.0.0
       def config
         @config ||= app.config.dup.tap do |slice_config|
-          # Unset config from app that does apply to ordinary slices
+          # Unset config from app that does not apply to ordinary slices
           slice_config.root = nil
         end
       end
--- a/lib/hanami/slice_registrar.rb
+++ b/lib/hanami/slice_registrar.rb
@@ -74,7 +74,9 @@ module Hanami
 
     def with_nested
       to_a.flat_map { |slice|
-        [slice] + slice.slices.with_nested
+        # Return nested slices first so that their more specific namespaces may be picked up first
+        # by SliceConfigurable#slice_for
+        slice.slices.with_nested + [slice]
       }
     end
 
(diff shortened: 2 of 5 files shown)


timriley:
👋🏼 @waiting-for-dev I took your initial pass at this and got it ready for merge. Played around with the idea of having a `slice_namespace` config setting but it felt unnecessary in the end — it was ungainly in practice and we don't want the added complication of users changing this for now.

timriley:
[Nested slices should look for modules nested within their parent’s namespace by default](https://trello.com/c/0wnMBERY/237-nested-slices-should-look-for-modules-nested-within-their-parents-namespace-by-default)

lib/hanami/slice.rb
```diff
@@ -113,7 +113,7 @@ def app
       # @since 2.0.0
       def config
         @config ||= app.config.dup.tap do |slice_config|
-          # Remove specific values from app that will not apply to this slice
+          # Unset config from app that does apply to ordinary slices
```
waiting-for-dev, line 116:
```suggestion
          # Unset config from app that doesn't apply to ordinary slices
```

lib/hanami/slice_registrar.rb
```diff
@@ -88,24 +90,27 @@ def inflector
       parent.inflector
     end
 
+    def parent_slice_namespace
+      parent.eql?(parent.app) ? Object : parent.namespace
```
waiting-for-dev, line 94: An edge case I had in mind to confirm, should we use `::Object` to bypass a slice defining a `SliceNamespace::Object` constant?

waiting-for-dev (approved):
Thanks for taking it, @timriley! I left a suggestion for what I think it's a typo and a question about what we should do in a very edge case.

lib/hanami/slice_registrar.rb
timriley: Definitely worth doing for clarity, thanks for the suggestion! I'll add it in.

timriley: Actually, I don't think this is required. `parent_slice_namespace` is always the first part of a string passed into `inflector.camelize`, which itself does an `Object.const_get(str)`. Both `Dry::Inflector` and `Hanami::SliceRegistrar` (which run this code) exist outside of any slice namespace, so there's no risk of clashing with any constants nested inside slice namespaces themselves.

timriley: So I'll leave it until someone can demonstrate it's actually a problem :)

PR #1241, opened 2022-11-07
Expose rack monitor to actions and forward exception logging to the logger

solnic:
- Automatically resolve rack_monitor for every action
- Delegate exception logging to the logger in rack logger


solnic:
Please notice that I did not configure default exception handling. I am *not sure how to do it* 😅

timriley (approved):
@solnic Looks good!

Just checking with you: was your intention here to make it possible for users to do this manually (by making rack_monitor available, and allowing users to do what you do in the test), or were you looking to have the framework do this automatically?

PR #1244, opened 2022-11-08
Replace Config.environment with Slice.environment

timriley:
This is exactly the same behavior as before: execute a code block if the given environment name matches, but this makes for a much more natural fit: now everything at the slice-level can be accessed without an explicit receiver, so e.g. env-specific config can be set like so:

```ruby
module MyApp
  class App < Hanami::App
    environment(:test) do
      config.logger.level = :info
    end
  end
end
```

This also makes for a much more natural fit for any other behavior that needs to be invoked on an env-specific basis, such as development-only requires:

```ruby
module MyApp
  class App < Hanami::App
    environment(:development) do
      require "awesome_print"
    end
  end
end
```


solnic (approved):
Yes this makes perfect sense and looks so much better

jodosha (approved):

PR #1247, opened 2022-11-08
Fix sessions middleware spec

timriley:
This array structure will work with the new way of loading middleware.

### DIFF
--- a/lib/hanami/config/actions/sessions.rb
+++ b/lib/hanami/config/actions/sessions.rb
@@ -62,7 +62,7 @@ module Hanami
         def middleware
           return [] unless enabled?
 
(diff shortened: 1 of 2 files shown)


lib/hanami/config/actions/sessions.rb
```diff
@@ -62,7 +62,7 @@ def enabled?
         def middleware
           return [] unless enabled?
 
-          [[storage_middleware, options]]
+          [storage_middleware, options].flatten
```
solnic, line 65:
```suggestion
          [storage_middleware, options].flatten(1)
```

solnic (changes requested):
Sorry, that was my bad, so here's a suggestion

Issue #1248, opened 2022-11-08
NoMethodError: undefined method `sessions' for Hanami::Config::NullConfig


jodosha:
I have a Hanami 2 application using only `hanami-router`, and NOT `hanami-controller`.

```
2022-11-08 15:41:44 +0100 Rack app ("GET /user/0" - (127.0.0.1)): #<NoMethodError: undefined method `sessions' for #<Hanami::Config::NullConfig:0x000000011165e278 @config=#<Dry::Configurable::Config values={}>>

          use(*config.actions.sessions.middleware) if config.actions.sessions.enabled?
                                                                    ^^^^^^^^^>
```

I'm using `2.0.0.rc1`

---

https://trello.com/c/RRVtK310

PR #1249, opened 2022-11-08
`config.actions.sessions` to be accessed only when `hanami-controller` is bundled

jodosha:
Fixes https://github.com/hanami/hanami/issues/1248
https://trello.com/c/RRVtK310


timriley (approved):
Oh yes, great spotting!

PR #1251, opened 2022-11-10
Add lib/ to $LOAD_PATH before loading app file

### LINKED ISSUE #1250: Registering custom middleware in a way that's visible in `hanami middleware`
hkIn order to test out Hanami 2, I'm trying to set up a blog that uses Rodauth. I have `config/providers/sequel.rb` which sets up a Sequel database:

```rb
# config/providers/sequel.rb
Hanami.app.register_provider(:sequel) do
  prepare do
    require "sequel"
  end

  start do
    register "db", Sequel.connect(target["settings"].database_url)
  end

  stop do
    target["db"].disconnect
  end
end
```

And I have `app/rodauth_app.rb` which defines the Rodauth app:

```rb
# app/rodauth_app.rb
module Blog
  class RodauthApp < Roda
    plugin :middleware
    plugin :rodauth do
      # ...
    end
  end
end
```

The Rodauth app has to be loaded after the Sequel provider has booted, because configuring Rodauth will fail if there is no Sequel database connection. In order to help with that, I decided to create a wrapper middleware, which can be added to the stack without immediately loading the Rodauth app.

```rb
# app/rodauth_middleware.rb
module Blog
  class RodauthMiddleware
    def initialize(app)
      @app = app
    end

    def call(env)
      Blog::RodauthApp.new(@app).call(env)
    end
  end
end
```

Where is the best place to add `Blog::RodauthMiddleware` to the middleware stack? I tried `config/app.rb`, but autoloading isn't set up at that point yet:

```rb
# config/app.rb
require "hanami"

module Blog
  class App < Hanami::App
    config.middleware.use Blog::RodauthMiddleware
  end
end
```

I then tried moving the middleware definition to `lib/blog/rodauth_middleware.rb` instead, and requiring `blog/rodauth_middleware`, but that doesn't work, presumably because the load path isn't set up at that point yet.

Next I tried moving it into `config/providers/rodauth.rb`, which worked:

```rb
# config/providers/rodauth.rb
Hanami.app.register_provider(:rodauth) do
  start do
    Hanami.app.config.middleware.use Blog::RodauthMiddleware
  end
end
```

However, it doesn't show up in `hanami middleware`, presumably because that command doesn't boot the application. Is there a recommended way to do this so that it's listed in the list of middlewares?

timriley:
This makes it possible to require files from lib/ at conventional spot at the top of the file, rather than inside the body of the app class.

Fixes #1250.

### DIFF
--- a/lib/hanami.rb
+++ b/lib/hanami.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "pathname"
 require "zeitwerk"
 require_relative "hanami/constants"
 
@@ -36,7 +37,8 @@ module Hanami
     app_path = self.app_path
 
     if app_path
-      require(app_path)
+      prepare_load_path
+      require(app_path.to_s)
       app
     elsif raise_exception
       raise(
@@ -45,6 +47,19 @@ module Hanami
         "Your app file should be at `config/app.rb` in your project's root directory."
       )
     end
+  end
+
+  # @api private
+  # @since 2.0.0
+  private_class_method def self.prepare_load_path
+    lib_path = app_path&.join("..", "..", LIB_DIR)
+
+    if lib_path&.directory?
+      path = lib_path.realpath.to_s
+      $LOAD_PATH.prepend(path) unless $LOAD_PATH.include?(path)
+    end
+
+    lib_path
   end
 
   # Returns the Hamami app class.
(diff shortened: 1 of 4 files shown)


timriley:
[Add lib/ to $LOAD_PATH before loading app file](https://trello.com/c/XFLK7HVh/296-add-lib-to-loadpath-before-loading-app-file)

lib/hanami.rb
```diff
@@ -110,7 +125,7 @@ def self.app_path(dir = Dir.pwd)
     path = dir.join(APP_PATH)
 
     if path.file?
-      path.to_s
+      path
```
solnic, line 128: This changes public API. Are we sure we want this to be a pathname? (I'm always in favor of pathnames FWIW).

timriley: Yep, I did consider this, and IMO this is a low risk API change and has the benefit of consistency with the other places we return pathnames, like config.root.

timriley: FWIW I doubt anyone is calling .app_path directly yet anyway. We added it relatively recently to make it so you could call the Hanami CLI in deeply nested directories and have everything still work.

solnic: Yes I agree

solnic (approved):

jodosha (approved):

PR #1253, opened 2022-11-14
Streamline action format setup

timriley:
The PR goes along with https://github.com/hanami/controller/pull/403 and makes it much more streamlined to set up action formats within full Hanami apps.

Before, it was this (aka “yo dawg, I heard you like `:json`”):

```ruby
config.actions.accepted_formats = [:json]
config.actions.default_request_format = :json
config.actions.default_response_format = :json
config.middleware.use :body_parser, :json
```

Now it's just this:

```ruby
config.actions.use_format :json
```

The first three lines of the "before" scenario are handled by the hanami-controller PR. This PR here deals with the last one.

Here we add some behavior to `Config#finalize` such that if `config.actions.accepted_formats` is set, then we automatically use the body_parser middleware with the appropriate arguments for the configured formats.

I've made sure that this will not disturb existing setups - if anyone manually configures a body_parser middleware, then we will not do anything extra.

### DIFF
--- a/spec/integration/action/format_config_spec.rb
+++ b/spec/integration/action/format_config_spec.rb
@@ -1,0 +1,194 @@
+# frozen_string_literal: true
+
+require "json"
+require "rack/test"
+
+RSpec.describe "App action / Format config", :app_integration do
+  include Rack::Test::Methods
+
+  let(:app) { Hanami.app }
+
+  around do |example|
+    with_tmp_directory(Dir.mktmpdir, &example)
+  end
+
+  it "adds a body parser middleware for the accepted formats from the action config" do
+    write "config/app.rb", <<~RUBY
+      require "hanami"
+
+      module TestApp
+        class App < Hanami::App
+          config.logger.stream = StringIO.new
+
+          config.actions.format :json
+        end
+      end
+    RUBY
+
+    write "config/routes.rb", <<~RUBY
+      module TestApp
+        class Routes < Hanami::Routes
+          post "/users", to: "users.create"
+        end
+      end
+    RUBY
+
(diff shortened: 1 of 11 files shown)


timriley:
[Consider a convenience method to make controller setup for the case of accepting/responding with a single format](https://trello.com/c/lWNdGriJ/334-consider-a-convenience-method-to-make-controller-setup-for-the-case-of-accepting-responding-with-a-single-format)

jodosha (approved):

spec/integration/action/format_config_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+
+  specify "adds a body parser middleware configured to parse any custom content type for the accepted formats" do
+    write "config/app.rb", <<~RUBY
+      require "hanami"
+
+      module TestApp
+        class App < Hanami::App
+          config.logger.stream = StringIO.new
+
+          config.actions.formats.add :json, ["application/json+scim", "application/json"]
```
jodosha, line 79: @timriley Nitpick, why don't:

```ruby
config.actions.formats.add(json: ["application/json+scim", "application/json"])
```

timriley: @jodosha so my thinking is that `add(format_name: ["mime_types", here"])` implies that _multiple_ formats and mime types can be added in one go, because it looks like a hash, and hashes are unbounded. But that's not how this method actually works, so I thought I'd make everything clearer by just acknowledging these are two distinct parameters.

timriley: If we feel like it I'm sure we could update it to accept a single open-ended hash in a later enhancement!

PR #1254, opened 2022-11-15
Improve logger setup

solnic:
Fixes dry-rb/dry-logger#19

- [x] Configure regular string-based logger in dev/test **by default** rather than a rack-specific logger
- [x] Configure a rack-specific logger only for logging rack requests
- [x] Make sure in production mode all logs go through the default JSON logger (we want to use just a single backend in prod after all)
- [x] Figure out a nice way of marking log entries as rack so that we don't have to check its payload. Maybe use tagged logs for that?
- [x] ~Figure out why elapsed time is not provided when logging exceptions coming from rack monitor~ it was provided but it was `0` in case of <1ms response times 😆 this is now address via configured dry-monitor's clock to use microseconds by default

### DIFF
--- a/lib/hanami/config/logger.rb
+++ b/lib/hanami/config/logger.rb
@@ -15,8 +15,14 @@ module Hanami
       # @return [Hanami::SliceName]
       #
       # @api private
-      # @since 2.0.o
+      # @since 2.0.0
       attr_reader :app_name
+
+      # @return [Symbol]
+      #
+      # @api private
+      # @since 2.0.0
+      attr_reader :env
 
       # @!attribute [rw] level
       #   Sets or returns the logger level.
@@ -40,7 +46,7 @@ module Hanami
       #
       #   @api public
       #   @since 2.0.0
-      setting :stream
+      setting :stream, default: $stdout
 
       # @!attribute [rw] formatter
       #   Sets or returns the logger's formatter.
@@ -57,7 +63,7 @@ module Hanami
       #
       #   @api public
       #   @since 2.0.0
-      setting :formatter
+      setting :formatter, default: :string
 
       # @!attribute [rw] template
       #   Sets or returns log entry string template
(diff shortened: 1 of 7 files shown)


solnic:
[Wrap up and release dry-logger 1.0](https://trello.com/c/FBpFJAuO/285-wrap-up-and-release-dry-logger-10)

timriley:
[Consider a convenience method to make controller setup for the case of accepting/responding with a single format](https://trello.com/c/lWNdGriJ/334-consider-a-convenience-method-to-make-controller-setup-for-the-case-of-accepting-responding-with-a-single-format)

lib/hanami/web/rack_logger.rb
```diff
@@ -43,36 +46,43 @@ def attach(rack_monitor)
...
+      def data(env, status:, elapsed:)
+        {
           verb: env[REQUEST_METHOD],
           status: status,
           elapsed: "#{elapsed}ms",
           ip: env[HTTP_X_FORWARDED_FOR] || env[REMOTE_ADDR],
           path: env[SCRIPT_NAME] + env[PATH_INFO].to_s,
           length: extract_content_length(env),
-          params: env[ROUTER_PARAMS]
+          params: env[ROUTER_PARAMS] || EMPTY_PARAMS
```
jodosha, line 80:
```suggestion
          params: env.fetch(ROUTER_PARAMS, EMPTY_PARAMS)
```

WDYT?

solnic:
[Finish PR with improved logger setup](https://trello.com/c/XJh6mIQK/342-finish-pr-with-improved-logger-setup)

lib/hanami/config/logger.rb
```diff
@@ -113,27 +119,18 @@ class Logger
...
-        config.formatter = case env
-                           when :production
-                             :json
-                           else
-                             :rack
-                           end
+        @env = env
+
+        case env
+        when :development, :test
```
jodosha, line 125: @solnic Shouldn't development be stdout?

jodosha (changes requested):
@solnic Please answer my question 🙂

lib/hanami/config/logger.rb
solnic: @jodosha $stdout is the default stream, so no need to set it again here

jodosha (approved):

PR #1260, opened 2022-11-21
Remove default action format of :html

timriley:
For maximal success with early adopters, leave the action format unspecified, meaning we accept all format types, and users can then opt into the format of their preference when they need to.

`:html` as a default format for our 2.0 release makes little sense since the gems we'll be releasing initially make the most sense for building APIs.

If we kept this as the default format, then anyone testing a request with an `Accept: application/json` header against a vanilla Hanami app would immediately receive a HTTP 406 error, which will be a jarring experience.

### DIFF
--- a/lib/hanami/config/actions.rb
+++ b/lib/hanami/config/actions.rb
@@ -136,8 +136,6 @@ module Hanami
 
       # Apply defaults for base config
       def configure_defaults
-        base_config.format :html
-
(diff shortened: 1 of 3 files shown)


jodosha (approved):
@timriley Makes sense 👍

timriley:
Thanks @jodosha!

Issue #1264, opened 2022-11-29
Hanami server is not picking up env variables


kucho:
Hello! Today I decided to start a new project using Hanami 2.0 and started poking around a little bit. 

I stumbled across an annoying little bug. I have set env variables using `direnv` and a `.env` file that looks like this:

```
toolbox main*​​​​ 
❯ cat .env
HANAMI_PORT=3000
``` 

However, it is not getting picked up at boot:

```
toolbox main*​​​ 
❯ echo $HANAMI_PORT 
3000

toolbox main*​​​ 
❯ hanami server 
11:56:26 - INFO - Using Guardfile at /Users/kucho/Projects/toolbox/Guardfile.
11:56:26 - INFO - Puma starting on port 2300 in development environment.
11:56:26 - INFO - Guard is now watching at '/Users/kucho/Projects/toolbox'
[80911] Puma starting in cluster mode...
[80911] * Puma version: 6.0.0 (ruby 3.1.2-p20) ("Sunflower")
[80911] *  Min threads: 5
[80911] *  Max threads: 5
[80911] *  Environment: development
[80911] *   Master PID: 80911
[80911] *      Workers: 2
[80911] *     Restarts: (✔) hot (✖) phased
[80911] * Preloading application
[80911] * Listening on http://0.0.0.0:2300
[80911] Use Ctrl-C to stop
[80911] * Starting control server on http://127.0.0.1:9293
[80911] * Starting control server on http://[::1]:9293
[80911] - Worker 0 (PID: 80915) booted in 0.0s, phase: 0
[80911] - Worker 1 (PID: 80916) booted in 0.0s, phase: 0
^C[80911] - Gracefully shutting down workers...
11:56:31 - INFO - Bye bye...

toolbox main*​​​ 6s 
❯ [80911] * Stopping immediately!
[80911] === puma shutdown: 2022-11-29 11:56:31 -0500 ===
[80911] - Goodbye!

```

I tried to debug it with `ruby/debug` and `byebug` but it just hung every time. At first, I thought it was something related to my dev machine, so I run other projects with direnv and .env files but I had no problems. I am kind of lost.

kucho:
Hanami console shows the correct value:

```
hanami c
toolbox[development]> ENV["HANAMI_PORT"]
=> "3000"
toolbox[development]>
```

jodosha:
@kucho Sorry for this problem. I fixed, but I need a bit more time to write tests and open a few PRs (hanami, cli, reloader).

[...]

PR #1266, opened 2022-12-01
Extract `Hanami::Env` and `Hanami::Port`

jodosha:
Fixes https://github.com/hanami/hanami/issues/1264


jodosha:
Confirmed by @kucho

PR #1269, opened 2022-12-04
Ensure rack events are on notifications

### LINKED ISSUE #1262: Register rack events on notifications component before attaching middleware
i.e. the same as this patch, but built in:

```ruby
require "dry/system"
require "hanami/providers/rack"

module Hanami
  module Providers
    class Rack < Dry::System::Provider::Source
      module Fixes
        def start
          notifications = target[:notifications]

          notifications.register_event(Dry::Monitor::Rack::Middleware::REQUEST_START)
          notifications.register_event(Dry::Monitor::Rack::Middleware::REQUEST_STOP)
          notifications.register_event(Dry::Monitor::Rack::Middleware::REQUEST_ERROR)

          super
        end
      end

      prepend Fixes
    end
  end
end
```

Will avoid an error like this:

```
     Dry::Events::InvalidSubscriberError:
       you are trying to subscribe to an event: `rack.request.stop` that has not been registered
```

timriley:
This fixes #1262, an issue where the rack events are not available if something interacts with our `notifications` bus before the rack provider is prepared.

This can happen in apps setting up rom-rb and using its sql instrumentation plugin, e.g.

```
rom_config.plugin(:sql, relations: :instrumentation) do |plugin_config|
  plugin_config.notifications = target["notifications"]
end
```

If rom-rb is set up in another provider, and that provider is started before the rack provider, then rack logging will crash with an error like this:

```
Dry::Events::InvalidSubscriberError:
  you are trying to subscribe to an event: `rack.request.stop` that has not been registered
```

With this change here, this will no longer happen, and rack logging will run correctly at all times.

### DIFF
--- a/lib/hanami/providers/rack.rb
+++ b/lib/hanami/providers/rack.rb
@@ -15,20 +15,27 @@ module Hanami
     class Rack < Dry::System::Provider::Source
       # @api private
       def prepare
-        require "dry/monitor"
-        require "hanami/web/rack_logger"
-
         Dry::Monitor.load_extensions(:rack)
+
+        # Explicitly register the Rack middleware events on our notifications bus. The Dry::Monitor
+        # rack extension (activated above) does register these globally, but if the notifications
+        # bus has been used before this provider loads, then it will have created its own separate
+        # locally copy of all registered events as of that moment in time, which will not included
+        # the Rack events globally reigstered above.
+        notifications = target["notifications"]
+        notifications.register_event(Dry::Monitor::Rack::Middleware::REQUEST_START)
+        notifications.register_event(Dry::Monitor::Rack::Middleware::REQUEST_STOP)
+        notifications.register_event(Dry::Monitor::Rack::Middleware::REQUEST_ERROR)
       end
 
       # @api private
       def start
         target.start :logger
 
-        notifications = target[:notifications]
-
-        clock = Dry::Monitor::Clock.new(unit: :microsecond)
-        monitor_middleware = Dry::Monitor::Rack::Middleware.new(notifications, clock: clock)
+        monitor_middleware = Dry::Monitor::Rack::Middleware.new(
+          target["notifications"],
+          clock: Dry::Monitor::Clock.new(unit: :microsecond)
+        )
 
         rack_logger = Hanami::Web::RackLogger.new(target[:logger], env: target.env)
         rack_logger.attach(monitor_middleware)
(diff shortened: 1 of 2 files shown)


timriley:
If you want more detail about how this was happening, see this comment on the original issue:

https://github.com/hanami/hanami/issues/1262#issuecomment-1336327552

solnic (commented):
I was hoping we could get it fixed by making sure that we load rack monitor before we set up notifications. Is that not possible?

timriley:
I don't think that's really desirable. We set up the dry-system notifications plugin as early as possibly precisely so that it can be available for anything that wants to use it.

In theory, we _could_ rearrange the boot sequence so that the rack provider is started first, and that would fix things now, but what about when _another_ integration requires earlier access to the notifications bus? This doesn't feel like a sustainable approach.

In addition, we actually _want_ our rack support to be lazily rather than eagerly activated, because it is (a) dependent on whether rack is in the bundle, and (b) still only activated when the router is initialised, which is something that may not always happen over the lifecycle of various app processes or workloads.

Given all of this, I think it's a good arrangement to have all the rack initialisation code together as much as possible, which is why I think the current state of the provider is actually pretty good.

Part of the issue here is dry-monitor/dry-event kind of awkwardly straddling this line between events that can be registered on either a class or an instance.

Given this, I think we should be encouraging users of Hanami to treat our `app["notifications"]` object as an instance only, and expecting all new events to be registered directly on that instance rather than on some `Dry::Monitor::Notifications` class, which is actually not something that we advertise to our users at all.

In this light, I think the practice of explicitly registering the events that we need on the notifications instance inside the rack provider is an example of good behaviour, and something that we should encourage others to follow.

Could we possibly make this nicer by upstreaming some of this code? I think certainly yes. For example, we could add a `Dry::Monitor::Rack::Middleware.register_events(notifications)` method, so that we only have to call that one method here inside Hanami, and not have to worry about the specific event names or constants. I think that would be a good improvement, but something we could do as a follow-up, rather than blocking on it.

timriley:
To look at this in another way: I'd prefer to see this merged as our initial fix because this seems like the simplest, least intrusive way of addressing this bug. Plus, I'm 100% confident in the efficacy of this fix.

Anything else would require rearranging our provider registration during boot and would possibly introducing other issues. Certainly it would require a lot more testing, at very least.

I'd definitely be open to investigating other ways of solving this, but I think this change here is our best bet in the meantime, and doesn't prevent the possibility of other changes in the future 😊

solnic (approved):

solnic:
You're right, this needs more thinking / more work and honestly feels like 3.0 topic.

PR #1273, opened 2022-12-27
Add a new `Hanami::Slice::Routing::Endpoint` class to be used by the routing resolver

pauc:
Related to this thread in the hanami chat room: https://gitter.im/hanami/chat?at=63a06fbc69ce3c5338359379

The goal is to facilitate testing and debugging routes. Adding this `Endpoint` objects allows us to do this kind of tests:

```ruby
route = router.recognize("/books/23")
expect(route).to be_routable
expect(route.endpoint.slice).to be API::Slice
expect(route.endpoint.action_key).to eq "books.show"
```

### DIFF
--- a/lib/hanami/slice/routing/endpoint.rb
+++ b/lib/hanami/slice/routing/endpoint.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module Hanami
+  class Slice
+    # @api private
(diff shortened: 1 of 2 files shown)


pauc:
Hi, my first contribution here, can add documentation and unit tests if you think it's interesting.

PR #1276, opened 2023-01-05
Support registrations_dir on Slice Containers

alassek:
This was defaulting to `system/registrations` from dry-system, but it makes much more sense to belong in `config/`.

I feel like I might be the only person who uses this feature, but I think it will prove indispensable for refactoring container keys transparently, or dealing with special cases where the default key convention doesn't fit.

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -843,6 +843,7 @@ module Hanami
         container.config.name = slice_name.to_sym
         container.config.root = root
         container.config.provider_dirs = [File.join("config", "providers")]
+        container.config.registrations_dir = File.join("config", "registrations")
 
         container.config.env = config.env
         container.config.inflector = config.inflector
--- a/spec/integration/slices/slice_registrations_spec.rb
+++ b/spec/integration/slices/slice_registrations_spec.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+RSpec.describe "Slice Registrations", :app_integration do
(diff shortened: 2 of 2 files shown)


timriley:
Thanks for finding and addressing this @alassek! It'll definitely be good for the Hanami integration with dry-system to be complete and consistent.

In terms of how we implement this, however, with Hanami I'm less interested in making every little dry-system knob configurable, instead opting for sensible defaults for certain things that wouldn't make sense for users to change.

Take the dry-system `provider_dirs` setting, for example. In Hanami we don't expose that as user configurable, and instead just set it directly to "config/providers" ([here in the code](https://github.com/hanami/hanami/blob/05b99b100e32a902be2ceb5d888a1fc8760ec198/lib/hanami/slice.rb#L845)).

The reason we decided to do this is best considered through the documentation and support use case. If the provider_dirs were configurable, then we'd need to caveat every mention of `"config/providers"` in our documentation with "or whichever provider_dirs you have configured". It would also make support of user problems harder because we could no longer depend on these core directory structures being in place.

So, given all of this, I reckon we should just take the same approach for `registrations_dir`. Could you just pop it under that same line of code that I linked to above? I'm happy with `config/registrations` as the directory.

alassek:
@timriley no problem! If someone really, _really_ needs to tweak this they can always tap into `prepare_container`.

alassek:
@timriley I've removed the configuration setting

timriley:
Thanks @alassek!

PR #1279, opened 2023-01-19
Provide a fail-safe config for logger

alassek:
If `Hanami.env` is not one of: `:development`, `:test`, `:production`, `Hanami::Config#logger_instance` would raise an exception.

There shouldn't be any need to officially support more than these environments, however if someone does need to do this it shouldn't break the system.


jodosha (approved):
@alassek Thanks for this PR. LGTM 👍

PR #1281, opened 2023-01-26
Use Rack production logger by default, unless app env is `development` or `test`.

Drowze:
Always use the rack production logger UNLESS the app env is specifically 'development' or 'test'

This should allow for usage of custom envs such as `prod` or `stage` without using development logger by accident. As a caveat, using a custom test/dev env such as `dev` might result on unwanted usage of production rack logger.

---
Also related to: https://github.com/hanami/hanami/pull/1279, as that PR also adds better support of custom prod envs.

### DIFF
--- a/lib/hanami/web/rack_logger.rb
+++ b/lib/hanami/web/rack_logger.rb
@@ -71,7 +71,7 @@ module Hanami
       # @since 2.0.0
       def initialize(logger, env: :development)
         @logger = logger
-        extend(env == :production ? Production : Development)
+        extend(%i[development test].include?(env) ? Development : Production)
       end
 
       # @api private


timriley:
This feels like a reasonable adjustment to me. What do you think, @solnic, @jodosha?

That said, in general I always encourage folks just to stick to `production`/`development`/`test` for their Rails or Hanami or Rack environments. These names often _mean something_ to the framework's internal behaviour (as we've seen here), and by introducing different environment names, you run the risk of introducing (or removing) behaviour in a way you don't expect.

Instead, to help differentiate different deployment targets, I encourage folks to use something like `ENV["STAGE"]` or similar, which itself can be any name that is meaningful to your and your team/organisation, so things like `"production"`, `"preprod"`, `"staging"`, `"testing-farm-1"` etc. are all completely fair game here, and there's no change you'll encounter unexpected framework behaviour.

jodosha (approved):
@Drowze Thanks for this PR. I believe the change makes sense. Thanks!

PR #1292, opened 2023-02-25
Update app view context for view gem changes

timriley:
Prepare for the changes to `Hanami::View::Context` in https://github.com/hanami/view/pull/223:

- Keep separate ivars for each context attribute (these are no longer captured by the superclass in an `_options` hash)
- Provide a custom `#initialize_copy` to ensure objects that may be mutated over the course of a rendering are duped (in this case, the `@content_for` hash)
- Provide a `#with` method that allows the request to be provided to an already initialized context
- Stop passing the response to the context before rendering, since `#flash` is now available on the request too


timriley:
Since these are mostly compatibility changes to keep in line with the adjustments in https://github.com/hanami/view/pull/223, I'm going to merge this right away so that I can get https://github.com/hanami/view/pull/223 merged too.

PR #1293, opened 2023-02-28
Ensure to mount Rack Middleware in routes using kwargs

jodosha:
## Problem

Given the following route scope with a middleware that accepts kwargs:

```ruby
    scope "/sidekiq" do
      use Rack::Session::Cookie,
          key: "_my_app.session",
          secret: Hanami.app.settings.session_secret,
          expire_after: 60 * 60 * 24 * 365

      mount Sidekiq::Web, at: "/sidekiq"
    end
```

The app crashes with the following:

```shell
    ArgumentError:
       unknown keyword: :secret
     # ./lib/hanami/slice/routing/middleware/stack.rb:94:in `use'
```

### DIFF
--- a/spec/integration/rack_app/middleware_spec.rb
+++ b/spec/integration/rack_app/middleware_spec.rb
@@ -41,6 +41,19 @@ RSpec.describe "Hanami web app", :app_integration do
             @app.call(env)
           end
         end
+
+        class AppendSession < Core
+          def initialize(app, options = {})
+            super()
+            @app = app
+            @options = options
+          end
+
+          def call(env)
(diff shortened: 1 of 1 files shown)


timriley:
This is now fixed by #1370 (and https://github.com/hanami/router/pull/267). Both are released as part of Hanami 2.1.0.

timriley:
@wuarmin If you could have another try with 2.1.0 and let me know how you go, that'd be great, thanks!

wuarmin:
@timriley Thanks! It now works perfectly.

Issue #1294, opened 2023-02-28
Add logger configuration for backends


bkuhlmann:
## Overview

Hello. :wave: I ran into a situation where I needed to colorize logging in my development environment while *also* logging to a file and was wondering if it'd be possible to add `config.logger.add_backend` support instead of `config.logger.instance.add_backend`?

I realize this is a minor quality of life improvement in terms of intuiting matching functionality as found in [Dry Logger](https://dry-rb.org/gems/dry-logger) but would reduce a few characters of typing and _feels_ a bit nicer. Also, having to mutate the instance feels wrong too.

## Steps to Recreate

Here's how to make this work:

``` ruby
# config/app.rb
environment :development do
  config.logger.options[:colorize] = true

  config.logger = config.logger.instance.add_backend(
    colorize: false,
    stream: Hanami.app.root.join("log/development.log")
  )
end

```

## Environment

- **Ruby**: ruby 3.2.1 (2023-02-08 revision 31819e82c8) +YJIT [arm64-darwin22.3.0]
- **Hanami**: 2.0.3

solnic:
Yeah I'd be in favor of doing this even though it'd make things a bit more complicated on the config side. We could even capture all method calls that are supported by the instance and eventually forward them once the instance is available.

bkuhlmann:
Yeah, this would be great because if the Hanami configuration for the logger is a simple proxy it means you could potentially do something like this where you chain the desired behavior like you can do with the native Dry Logger (well, sort of, I'm dreaming a bit here). Example:

``` ruby
config.logger
      .colorize
      .formatter(:rack)
      .add_backend(stream: Hanami.app.root.join("log/development.log")
```

This would make configuring the logger a rather elegant but probably the most powerful aspect of this is being able to add multiple backends easily with minimal effort.

solnic:
Right. My only concern is that config object methods may potentially clash with logger methods. I wonder now if maybe using a provider for logger to tune the instance would be better? Something like:

```ruby
class MyApp < Hanami::App
  # This doesn't exist, but we could potentially have shortcuts like that
  prepare(:logger) do
    logger.add_backend(stream: ...)
  end
end
```

The reasoning here is that extending config with even more options just to translate them into method calls is actually more complex than exposing a convenient way to access components (logger in this case) in order to tweak them using their own API.

WDYT @jodosha @timriley?

timriley:
[...]

PR #1298, opened 2023-03-11
Allow rack provider to be replaced by custom provider.

andrewcroome:
To use a custom logger in several applications, we currently replace the default `logger` provider by placing our own at `config/providers/logger.rb` (use the mechanism [mentioned here](https://github.com/hanami/hanami/issues/1294#issuecomment-1451823937) by @timriley).

This works well, but as part of our logging story, we'd also like to replace Hanami's rack logger with our own custom rack logger, so that we have control over rack request logging too.

Hanami's rack logger is [initialized here](https://github.com/hanami/hanami/blob/main/lib/hanami/providers/rack.rb#L40) in the rack provider. 

This PR allows framework users to replace that rack provider with a custom one of their own making, should they so choose, by placing a custom implementation at `config/providers/rack.rb`.

This is just one way we could achieve this - there might be an alternative option that allows the rack logger Hanami will use to be decided via configuration?

### DIFF
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -151,15 +151,17 @@ module Hanami
         require_relative "providers/inflector"
         register_provider(:inflector, source: Hanami::Providers::Inflector)
 
-        # Allow logger to be replaced by users with a manual provider, for advanced cases
+        # Allow logger and rack to be replaced by users with manual providers, for advanced cases
         unless container.providers.find_and_load_provider(:logger)
           require_relative "providers/logger"
           register_provider(:logger, source: Hanami::Providers::Logger)
         end
 
         if Hanami.bundled?("rack")
-          require_relative "providers/rack"
-          register_provider(:rack, source: Hanami::Providers::Rack, namespace: true)
+          unless container.providers.find_and_load_provider(:rack)
+            require_relative "providers/rack"
+            register_provider(:rack, source: Hanami::Providers::Rack, namespace: true)
+          end
         end
       end
 
--- a/spec/integration/container/standard_providers_spec.rb
+++ b/spec/integration/container/standard_providers_spec.rb
@@ -89,10 +89,19 @@ RSpec.describe "Container / Standard providers", :app_integration do
         end
       RUBY
 
+      write "config/providers/rack.rb", <<~RUBY
+        Hanami.app.register_provider :rack do
+          start do
+            register "rack.monitor", "custom rack monitor"
+          end
+        end
+      RUBY
+
       require "hanami/setup"
       Hanami.boot
 
       expect(Hanami.app[:logger]).to eq "custom logger"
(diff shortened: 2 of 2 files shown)


timriley:
Thanks for this, @andrewcroome.

I'd like to do some deeper consideration of our options here before picking an approach. 

One of my reasons for this is that our [rack provider](https://github.com/hanami/hanami/blob/b73ad93534a2b5402d34dc365fe5fcbea52e8cbd/lib/hanami/providers/rack.rb#L15) does quite a lot. Allowing a user to replace it wholesale is creating a lot of opportunity for things to become broken.

Would you be able to share more of what you're looking to achieve? Possibly a sanitized version of your custom rack provider, or a description of your full set of requirements?

Thanks!

solnic:
Yes we should isolate the rack logger part of the rack provider and allow replacing *just that*. This would also be a power-user thing, because the rack logger is expected to behave in a specific way in order to work with the rack logging.

andrewcroome:
Thanks @timriley and @solnic.

> our [rack provider](https://github.com/hanami/hanami/blob/b73ad93534a2b5402d34dc365fe5fcbea52e8cbd/lib/hanami/providers/rack.rb#L15) does quite a lot. Allowing a user to replace it wholesale is creating a lot of opportunity for things to become broken.

Yes, that's true.

> Would you be able to share more of what you're looking to achieve?

There's a few things that custom a rack logger can achieve - one is extracting additional information from the rack request, like the value of any `X-Amzn-Trace-Id` or correlation id headers, and logging those things in a structured way using the underlying logger. In our case it also extracts a few domain-specific pieces of information that can be inferred from the request. 

I guess this boils down to (a) controlling [what data](https://github.com/hanami/hanami/blob/main/lib/hanami/web/rack_logger.rb#L99) is provided to the underlying logger, and also (b) controlling how that underlying logger is invoked (to pass additional information to it).

We're also considering the ability to skip logging certain requests - such as successful hits against a health check endpoint from a load balancer.

At the moment, we have a custom rack logger and Hanami's rack logger in play simultaneously, which means each request gets logged twice. 

I think it would be useful to be either able to:
- replace just the rack logger element of the rack provider as @solnic suggests, and/or
- have a configuration option that skips the registering of the default Hanami rack logger (folks could then register their own in a custom provider)

Shall I close this PR for now as enabling the replacement of the whole rack provider doesn't seem right? And open a thread on discourse instead?

solnic:
> Shall I close this PR for now as enabling the replacement of the whole rack provider doesn't seem right? And open a thread on discourse instead?

Yes this sounds like a good idea. It would be good to gather feedback from more people too. The subject of logger adjustments has come up in multiple places already.

andrewcroome:
Thanks @solnic. I'll move this discussion to discourse. 🙏

PR #1161, opened 2022-04-22
Hanami 2 Helpers

jodosha:

PR #1305, opened 2023-04-08
Add working Hanami::Helpers::FormHelper

timriley:
Change tack from #1161 and reintroduce `Hanami::Helpers::FormHelper`, intended to be mixed into an app's base `Part` and `Scope` classes.


timriley:
[Finish form helpers](https://trello.com/c/RYVt348g/4-finish-form-helpers)

lib/hanami/helpers/form_helper.rb
```diff
@@ -0,0 +1,213 @@
...
+      #   </form>
+      #
+      # @example Method override
+      #   <%= form_for("/books/123", method: :put) do |f|
+      #     <%= f.text_field "book.title" %>
+      #     <%= f.submit "Update" %>
+      #   <% end %>
+      #
+      #   =>
+      #   <form action="/books/123" accept-charset="utf-8" id="book-form" method="POST">
```
jodosha, line 125:
```suggestion
      #   <form action="/books/123" accept-charset="utf-8" method="POST">
```

It can't guess the `id`, without the `base_name` argument.

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -0,0 +1,1387 @@
...
+        #   # A convenience for:
+        #   # <%= f.text_field "address.street" %>
+        #   # <%= f.text_field "address.suburb" %>
+        #
+        #   =>
+        #   <input type="text" name="delivery[customer_name]" id="delivery-customer-name" value="">
+        #   <input type="text" name="delivery[address][street]" id="delivery-address-street" value="">
+        #
+        # @example Multiple levels of nesting
+        #   <% f.fields_for "address" %>
```
jodosha, line 153: @timriley To make the usage more explicit for the user, should we consider yielding a builder in the `fields_for` block?

Example:

```
<% f.fields_for "address" |fa| %>
  <%= fa.text_field "street" %>

  <% fa.fields_for "location" |fal| %>
    <%= fal.text_field "city" %>
  <% end %>
<% end %>
```

lib/hanami/helpers/form_helper/form_builder.rb
```diff
@@ -0,0 +1,1387 @@
+# frozen_string_literal: true
+
+require "dry-inflector"
```
jodosha, line 3:
```suggestion
```

`Dry::Inflector` isn't referenced in the code. It's a dependency that the consumer of this class must provide.

jodosha (approved):
@timriley Excellent and patient work. You also solved pending problems from Hanami 1 (see file multipart). Thanks!

lib/hanami/helpers/form_helper/form_builder.rb
timriley: Ah yeah, you're right, we should do this. I'll make sure I add this in before I merge this PR.

timriley:
Thank you for the careful review, @jodosha! I've incorporated all of your feedback now, and will merge in a moment :)

PR #1307, opened 2023-04-25
Seamlessly provide all first-party and user-defined helpers

timriley:
This PR builds on the simple earlier steps we've taken in hanami to make helpers available in templates and parts:

- https://github.com/hanami/hanami/pull/1303
- https://github.com/hanami/hanami/pull/1304

Here we take things to the next level, by firstly including all the helpers that we now have available:

- First party helpers from hanami-view, presently waiting to be merged in https://github.com/hanami/view/pull/229.
- First party helpers within hanami itself (which require the full app, rather than just the standalone view layer), also waiting to be merged. https://github.com/hanami/hanami/pull/1305 (this PR builds upon this branch)
- To round this out, we **also include user-defined helpers,** expecting these to be in a `Views::Helpers` module within each app or slice namespace.

At the same time, we also refactor our view extensions to make them less tangled, more directly connected to the components they are extending.

Let's walk through the core details of implementation:

- We now have new extension modules specifically for `Hanami::View::Part` and `Hanami::View::Scope`. These have the job of adding whatever behaviour we need for each of these classes, rather than it being the job of the `Hanami::View` extension. This is a better approach because it means if a user ever accesses a part or scope class _without_ first accessing a view, the class will have the appropriate behaviour applied to it already.
- These Part and Scope extensions are now the ones responsible for including all the helper modules (previously this was the the View extension, but this is no more due to the reasons explained just above).
- We include the helpers in two ways:
  - First-party helpers via a `StandardHelpers` module that we include in each. This is done just once, because these are static helpers, and there's no need to include these differently per slice.
  - User-defined helpers via a dynamic `SliceConfiguredHelpers` module. This looks for a `Views::Helpers` module inside the slice namespace, and if it is present, it includes it.

The net result is that we'll likely generate the following new file in Hanami apps:

```ruby
# For the app: app/views/helpers.rb

module MyApp
  module Views
    # Helpful explanatory comment will go here
    module Helpers
    end
  end
end

# For each slice: slices/[slice_name]/views/helpers.rb

module SliceName
  module Views
    # Helpful explanatory comment will go here
    module Helpers
    end
  end
end
```

Users can put their own helpers into these modules, and these helpers will automatically become available inside scopes, templates and parts, no further steps required.

Naturally, users may not want every custom helper to go into this one file. To take care of this, I propose our users take advantage of... ✨ Ruby! ✨ 

What this means is that they can just use standard Ruby techniques to organise their helper code. I propose we make it possible to generate new nested helper modules via generators, etc., which'd result in the following:

```ruby
# app/views/helpers/some_helper.rb

module MyApp
  module Views
    module Helpers
      module SomeHelper
        # custom helper methods can go here
      end
    end
  end
end
```

And this module could be included like so:

```ruby
# app/views/helpers.rb

module MyApp
  module Views
    module Helpers
      include SomeHelper
    end
  end
end
```

And if there are helpers that users do not want included in _everywhere_, this also becomes manageable via plain Ruby too — just don't include them in the `Helpers` module, and then directly include them where you need them, e.g. in a part class:

```ruby
# app/views/parts/some_part.rb

module MyApp
  module Views
    module Parts
      class SomePart < MyApp::Part
        include Helpers::SomeSpecialHelper
      end
    end
  end
end
```

This is a fairly light touch integration that feels to me like an appropriate balance of magic and explicitness: first party helpers are available everywhere, we have a special user-owned module that becomes available everywhere, and because that itself is just an ordinary module, users can then rely on standard Ruby techniques to work on it as they wish.

---

There's one other detail I took care of while finishing this work. In 0936435d2d3d29bb890d2e53ff197a499584b74e, I've made it so we no longer rely on a `"views.context"` container registration for the context that we provide when setting up views and actions that render views.

The issues with this approach became clear when I was setting up some of the integration tests for the work above: if I didn't concretely define a `app/views/context.rb` class, then everything would break, because I was no longer using a `Context` class with our `Extensions::View::Context` module applied to it. For this extension to apply, it needs a concrete class definition.

So now with this we take the same approach as we do for other parts of our view extensions: we look for an already-defined `Views::Context` class, and if there isn't one, _we define our own_, automatically.

On top of this, we no longer look for a container registration for the context, and instead look for the `Views::Context` class instead, and then initialize that directly when required. This is much cleaner for the purposes of our action/view extensions because it can be done without depending on the container lifecycle at all.

Since working with the context is needed from both our action and view extensions, we make this behaviour consistent by exposing a `Hanami::Extensions::View::Context.context_class(slice)` method that can be used from either place.


timriley:
[Automatically include user-defined helpers](https://trello.com/c/mwEKr7Rg/25-automatically-include-user-defined-helpers)

solnic (approved):
LGTM 👏🏻

jodosha (approved):
@timriley Excellent work.

I'm still on the fence about explicitly including helpers in Parts after a user already included a helper in `MyApp::Views::Helpers`.

Dogfooding views will help us to know this aspect better.

---

Given we took the `MyApp::Views::Helpers` namespace, we should make it a reserved word in the CLI generators.

timriley:
@jodosha @solnic Thanks for checking this out!

@jodosha I'm happy to provide an answer to this already:

> I'm still on the fence about explicitly including helpers in Parts after a user already included a helper in MyApp::Views::Helpers.

No need to worry about this: any method available inside `MyApp::Views::Helpers` **will** automatically be available to all part classes!

The reason I showed a demonstrate of an explicit `include` within a Parts class was to demonstrate how a user might choose to create certain helpers that they do not want available everywhere. It's the exception, not the rule. (Feel free to revisit that section above and ask me any more questions, if you'd like!)

I also agree that realistic usage of views and helpers will help ensure we get things into the best possible shape, so I'm very excited to be merging this now and getting ready for our next steps! 😄

PR #1308, opened 2023-04-26
Fix SliceConfiguredView config issues

timriley:
This fixes a bug where a slice-based view class would inherit the app-level `paths` config, meaning that slice views were not finding their templates properly.

It also refines our view context setup, having auto-defined slice-based contexts inherit from the app-level context where possible.

Along with this, this expands and better organises our test coverage for slice configured views.


timriley:
[Fix bug with paths not being configured correctly in slices](https://trello.com/c/YWGKskcb/33-fix-bug-with-paths-not-being-configured-correctly-in-slices)

timriley:
@jodosha FYI, this is just a fix and range of internal tidy-ups, so I'm just going to go ahead and merge this right away.

PR #1309, opened 2023-05-02
Render nice errors

timriley:
TODO:

- [ ] `render_exceptions` config to turn this on/off; default to true in production mode
- [ ] Make `rescue_responses` from `RenderableException` configurable on the app too
- [ ] update hanami-router to handle `not_allowed` like `not_found`
- [ ] Figure out if we're happy with those extra hanami-router errors being here
- [ ] Inline docs
- [ ] A better fallback if there are no matching files in public/?

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -948,23 +948,37 @@ module Hanami
 
         require_relative "slice/router"
 
+        slice = self
         config = self.config
         rack_monitor = self["rack.monitor"]
 
+        # TODO: update hanami-router to handle `not_allowed` like `not_found`
         Slice::Router.new(
           inspector: inspector,
           routes: routes,
           resolver: config.router.resolver.new(slice: self),
+          not_found: ROUTER_NOT_FOUND_HANDLER,
           **config.router.options
         ) do
           use(rack_monitor)
-          if Hanami.bundled?("hanami-controller")
-            use(*config.actions.sessions.middleware) if config.actions.sessions.enabled?
+
+          use(
+            Hanami::Middleware::RenderExceptions,
+            Hanami::Middleware::PublicExceptionsApp.new(slice.root.join("public"))
+          )
+
+          if Hanami.bundled?("hanami-controller") && config.actions.sessions.enabled?
+            use(*config.actions.sessions.middleware)
           end
 
           middleware_stack.update(config.middleware_stack)
         end
       end
+
+      ROUTER_NOT_FOUND_HANDLER = -> env {
+        raise Hanami::Router::NotFoundError.new(env)
+      }.freeze
+      private_constant :ROUTER_NOT_FOUND_HANDLER
 
       # rubocop:enable Metrics/AbcSize
     end
--- a/lib/hanami/extensions.rb
+++ b/lib/hanami/extensions.rb
@@ -10,3 +10,7 @@ if Hanami.bundled?("hanami-view")
   require_relative "extensions/view/part"
   require_relative "extensions/view/scope"
 end
+
+if Hanami.bundled?("hanami-router")
+  require_relative "extensions/router/errors"
+end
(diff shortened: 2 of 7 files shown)


timriley:
[Customizable error views](https://trello.com/c/emzWb2vR/15-customizable-error-views)

lib/hanami/config.rb
```diff
@@ -127,6 +127,45 @@ class Config
...
+    #
+    #   For HTML responses, these error pages are in `public/{404,500}.html`.
+    #
+    #   Defaults to `true` in production mode, `false` in all others.
+    #
+    #   @return [Boolean]
+    #
+    #   @api public
+    #   @since 2.1.0
+    setting :render_errors, default: false
```
timriley, line 141: These settings will be unneeded/unused when using Hanami in "non-web" mode (i.e. with hanami-router and hanami-controller removed from the app's `Gemfile`), but I decided to put them at the top-level here (rather than e.g. nest them inside `config.router` or some other namespace) because they're kind of a cross-cutting concern and I don't think we need to go nuts creating namespaces for general settings just yet.

In addition, they're not harming anyone if they exist here unused in the case of non-web apps.

For future major versions of Hanami we might consider a different approach, but this feels appropriate to me for now.

lib/hanami/slice.rb
```diff
@@ -948,24 +948,44 @@ def load_router(inspector:)
...
         ) do
           use(rack_monitor)
-          if Hanami.bundled?("hanami-controller")
-            use(*config.actions.sessions.middleware) if config.actions.sessions.enabled?
+
+          use(
+            Hanami::Middleware::RenderErrors,
+            config,
+            Hanami::Middleware::PublicErrorsApp.new(slice.root.join("public"))
+          )
```
timriley, line 969: I decided to keep this middleware inserted at all times — even when `config.render_errors` is false — because it helpers to provide a "stable" default middleware stack that our users can depend upon for inserting their middleware at the right spot.

And since this is an exception-catching middleware, it's actually quite important for user-provided to middleware to know whether they're inserting themselves before or after this one.

In terms of how this works when `config.render_errors` is false, it simply re-raises the exception in that case.

Given that this middleware is likely to be enabled for the majority of apps in production mode, having it still exist (but do nothing) in development/test modes seems like a worthwhile tradeoff of the predictable middleware stack.

spec/integration/rack_app/middleware_spec.rb
```diff
@@ -597,15 +599,15 @@ def handle(req, res)
...
-      expect(last_response.headers["X-Identifier-Root"]).to eq("true")
-      expect(last_response.headers).to have_key("X-Elapsed")
-      expect(last_response.headers).to_not have_key("X-Auth-User-ID")
-      expect(last_response.headers).to_not have_key("X-API-Rate-Limit-Quota")
-      expect(last_response.headers).to_not have_key("X-API-Version")
+      expect(last_response.headers).not_to have_key("X-Identifier-Root")
+      expect(last_response.headers).not_to have_key("X-Elapsed")
+      expect(last_response.headers).not_to have_key("X-Auth-User-ID")
+      expect(last_response.headers).not_to have_key("X-API-Rate-Limit-Quota")
+      expect(last_response.headers).not_to have_key("X-API-Version")
```
timriley, line 610: This is one behavioural change worth noting — a request for a not found URL is no longer going to have all the headers from other middleware preserved in the final response.

I think this is actually a more reasonable position — errors responses like this should remain as simple and as independent as possible.

lib/hanami/middleware/public_errors_app.rb
```diff
@@ -0,0 +1,75 @@
...
+        if content_type.to_s.start_with?("application/json")
+          require "json"
+          [JSON.generate(default_body), "application/json"]
+        else
+          [render_html_content(status, default_body), "text/html"]
+        end
+      end
+
+      def render_html_content(status, default_body)
+        path = "#{public_path}/#{status}.html"
```
timriley, line 65: I'm expecting `public/` as the directory here, whereas in Hanami 1.x we expected these HTML files to be in `apps/web/templates`, and we also seemingly supported ERB templating for these files.

In the interest of keeping error rendering as robust as possible (and for this implementation being as simple as possible), I took the path of these being static files only.

PR #1312, opened 2023-05-07
Fix double escaping of form values

timriley:
While I was demoing our latest helpers to @andrewcroome, we noticed an issue where form values were being double-escaped, i.e. value of `"<hello>"` from the params or provided struct being turned into `"&amp;&lt;hello&amp;gt;"` instead of just `"&lt;hello&gt;"` (which will correctly render in the browser's field as `"<hello>"`).

This PR fixes that, while also removing a couple of stale "TODO" comments from the form helpers.

Here's what a corresponding form looks like now:

Screenshot 2023-05-07 at 2 37 58 pm

### DIFF
--- a/lib/hanami/helpers/form_helper/form_builder.rb
+++ b/lib/hanami/helpers/form_helper/form_builder.rb
@@ -1289,7 +1289,6 @@ module Hanami
         # @api private
         # @since 2.0.0
         def _value(name)
-          # TODO: to_sym should not be necessary here
           values.get(*_split_input_name(name).map(&:to_sym))
         end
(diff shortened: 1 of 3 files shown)


timriley:
[Fix double escaping of form field values when using form helpers](https://trello.com/c/4RqIwExU/41-fix-double-escaping-of-form-field-values-when-using-form-helpers)

timriley:
Since this is a straight-up bug fix, I'm just gong to go ahead and merge it. Happy for post-merge comments if you have any, @jodosha :)

PR #1313, opened 2023-05-07
Fix action/view integration when hanami-view is not bundled

timriley:
In our action integration, we should have been checking explicitly for whether hanami-view was bundled, rather than checking for the presence of the `Extensions::Views` const, which is automatically made available via Zeitwerk, and led to a `LoadError` when hanami-view was not bundled.

The only need we had for our action extension to reach into the view extension was for the shared code for loading the view context. Instead, we now explicitly check to see if hanami-view is bundled before using that code.

In addition, we add a fallback to look for a same-named context class in the conventional location, which would be useful if a 3rd party view layer was to be used.

### DIFF
--- a/lib/hanami/extensions/action/slice_configured_action.rb
+++ b/lib/hanami/extensions/action/slice_configured_action.rb
@@ -121,9 +121,11 @@ module Hanami
             slice: slice,
           )
 
-          view_identifiers.detect do |identifier|
-            break slice[identifier] if slice.key?(identifier)
+          view_identifiers.each do |identifier|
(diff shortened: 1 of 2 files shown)


timriley:
[Fix bug with action extension trying to access hanami/view even when it is not bundled](https://trello.com/c/dc9vpQV0/43-fix-bug-with-action-extension-trying-to-access-hanami-view-even-when-it-is-not-bundled)

timriley:
FYI: Another small fix, merging this one right in. Open to post-merge comments and adjustments, of course!

PR #1314, opened 2023-05-07
Raise more helpful errors in view context extension

timriley:
Follow the existing pattern we have of raising a ComponentLoadError when `@assets` is missing, and do the same for `@request` and `@routes`.

These error and messages should be enough to help the user identify the reason for the missing value and make corrections as required.

### DIFF
--- a/lib/hanami/extensions/view/context.rb
+++ b/lib/hanami/extensions/view/context.rb
@@ -76,10 +76,6 @@ module Hanami
 
             attr_reader :settings
 
-            attr_reader :routes
-
-            attr_reader :request
-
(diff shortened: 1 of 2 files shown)


timriley:
[Raise a meaningful error if we try to access request/response on an Hanami::View::ApplicationContext and they are not present](https://trello.com/c/LX90BXVW/13-raise-a-meaningful-error-if-we-try-to-access-request-response-on-an-hanamiviewapplicationcontext-and-they-are-not-present)

timriley:
Just another small fix, merging right away.

This class needs docs, but now's not the time to do that. I'll add a card to ensure we do it before release.

PR #1316, opened 2023-05-22
Remove `frozen_string_literal: true` comments

parndt:
Since we've moved to requiring Ruby >= 3.0 we no longer need to have the
`#frozen_string_literal: true` magic comment as strings are now frozen
by default.


parndt:
OK, so apparently this got walked back and I didn't realise: https://bugs.ruby-lang.org/issues/11473#note-53 😧

PR #1319, opened 2023-06-19
Integrate with hanami-assets 2.1 and add asset helpers

jodosha:
- Depend on hanami-assets for development/tests, including the npm package
- Conditionally register a provider for assets if hanami-assets gem is bundled
- From this provider, register an `"assets"` component providing access to assets
- Use `Hanami::Assets::Config` instance as `app.config.assets`
- Introduce `Hanami::Helpers::AssetsHelper`, containing helper methods moved over from hanami-assets (with method names adjusted for clarity and consistency with helpers from other hanami and hanami-view modules)
- Include `Hanami::Helpers::AssetsHelper` in the standard helpers if hanami-assets is bundled


lib/hanami/helpers/assets_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Hanami
+  module Helpers
+    module AssetsHelper
+      def javascript(...)
+        _context.assets.javascript(...)
+      end
```
parndt, line 8: Is the intent here to also have a stylesheet helper?

```suggestion
      end
      
      def stylesheet(...)
        _context.assets.stylesheet(...)
      end
```

parndt: and maybe an image helper?

```suggestion
      end
      
      def image(...)
        _context.assets.image(...)
      end
```

timriley: We have both of these now :)

PR #1320, opened 2023-06-29
Fix body parser middleware already in use check.

nashby:
In an app that has multiple slices I noticed that `BodyParser` middleware added the same amount of times as the number of slices. After some investigation I ended up with this line that incorrectly checks for already added Body parser. 

Make sure we check for middleware's class name as a constant, not a string since that's what we have in middleware stack.

### DIFF
--- a/spec/integration/action/format_config_spec.rb
+++ b/spec/integration/action/format_config_spec.rb
@@ -136,13 +136,24 @@ RSpec.describe "App action / Format config", :app_integration do
   it "does not add a body parser middleware if one is already added" do
     write "config/app.rb", <<~RUBY
       require "hanami"
+      require 'hanami/middleware/body_parser/parser'
+
+      class CustomBodyParser < Hanami::Middleware::BodyParser::Parser
+        def self.mime_types
+          ['application/json']
+        end
+
+        def parse(body, *)
+          JSON.parse(body).transform_values { |v| v.map(&:upcase) }
+        end
+      end
 
       module TestApp
         class App < Hanami::App
           config.logger.stream = StringIO.new
 
           config.actions.format :json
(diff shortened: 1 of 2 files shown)


spec/integration/action/format_config_spec.rb
```diff
@@ -137,12 +137,26 @@ def handle(req, res)
...
+      class CustomBodyParser
+        def mime_types
+          ['application/json']
+        end
+
+        def parse(body, *)
+          JSON.parse(body).transform_values { |v| v.map(&:upcase) }
+        end
+
+        def empty?
```
nashby, line 149: I'm not sure why it keeps calling `empty?` on a custom parser. Without this method I get this:

     NoMethodError:
       undefined method `empty?' for #<CustomBodyParser:0x000000010bb4f170>

                 return DEFAULT_BODY_PARSERS if parser_specs.empty?
                                                            ^^^^^^^
     # ./lib/hanami/slice/routing/middleware/stack.rb:128:in `new'
     # ./lib/hanami/slice/routing/middleware/stack.rb:128:in `to_rack_app'

I'll try to investigate what's going on but if you have any idea please let me know!

nashby: figured that out

afurm:
The `==` check works for exact class matches, but what about subclasses of `Hanami::Middleware::BodyParser`? `Hanami::Middleware::BodyParser == SomeSubclass` returns false. If a slice registers a custom subclass directly rather than `BodyParser` itself, would this check still incorrectly detect it as a duplicate?

katafrakt:
I think this is no longer relevant after #1575. Good catch at the time though and sorry it took so long to get to this.

timriley:
Yes, thank you all, and sorry we didn't get onto this quicker. 🙇🏼

PR #1322, opened 2023-08-07
Fix an issue where the rack app was being re-initialised for every re…

DangerDawson:
…quest

This took me a while to track down, but when we switched over to using memcached based seasions using the dalli gem we noticed that we were leaking memcache connections:

The culprit was this bit of code: https://github.com/petergoldstein/dalli/blob/main/lib/rack/session/dalli.rb#L70

As Hanami was re-initalising the app on every single request:

https://github.com/hanami/hanami/blob/1.3.x/lib/hanami/middleware_stack.rb#L52

Because the call method on for the rack builder was calling `to_app` on every request:

https://github.com/rack/rack/blob/da03bfab6497ab05e9ed795f5d3c033e47927483/lib/rack/builder.rb#L276

I am not sure if this is the correct way to fix this? or it was intended that the rack app should be re-initalised on every request?

Also I am not sure if this is present in Hanami v2 either

### DIFF
--- a/lib/hanami/application.rb
+++ b/lib/hanami/application.rb
@@ -150,7 +150,7 @@ module Hanami
     # @api private
     def initialize
       @renderer   = RenderingPolicy.new(configuration)
-      @middleware = configuration.middleware
+      @middleware = configuration.middleware.to_app
     end
 
     # Process a request.
--- a/lib/hanami/middleware_stack.rb
+++ b/lib/hanami/middleware_stack.rb
@@ -39,17 +39,11 @@ module Hanami
       self
     end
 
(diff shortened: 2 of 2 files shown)


katafrakt:
Following my investigation, this is not happening on Hanami 2, where a Rack app is memoized per slice (see `Hanami::Slice#rack_app` method). Since Hanami 1.x is no longer developed, I'm going to close this. But I want to let you know this was a great investigation. If you ever have problems with Hanami 2.x and Dalli, don't hesitate to open an issue or another PR. We will try to respond in a more timely manner.

PR #1325, opened 2023-08-25
API URL pointing to rdoc.info does not work properly

swilgosz:
Fixed this by pointing to the gemdocs.org instead


timriley:
Thanks!

PR #1329, opened 2023-09-11
Update docs to match asset helpers change

parndt:
Since #1319 we have different ways of using the asset helpers, so this
updates the docs to match the current usage.


timriley:
Thanks Phil!

parndt:
> Thanks Phil!

You're welcome! Thanks for the rest of it 😄

PR #1311, opened 2023-05-06
Use hanami-webconsole to render detailed errors

timriley:

PR #1330, opened 2023-09-11
Return appropriate response statuses based on error type

### LINKED ISSUE #1327: Not found route returns 500 instead of 404
This is a bug.

```shell
⚡ curl -I http://localhost:2300/foo
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
Set-Cookie: BetterErrors-2.9.1-CSRF-Token=71b3287d-ea5a-4b9e-bdc2-75499809da59; path=/; HttpOnly; SameSite=Strict
Content-Length: 2455
```

When visiting in the browser, the web console indicates that the offending line is: https://github.com/hanami/hanami/blob/7cfdd344e1cba12bc7727cc0a69d5038b88c9cae/lib/hanami/slice.rb#L990

timriley:
Properly return 404 errors when the app is called with not found routes.

To do this, cover two scenarios:

- When hanami-webconsole **is not bundled**, or when `app.config.render_detailed_errors` is false, then do not provide our custom `not_allowed` and `not_found` procs to the slice's router instance. This means that the `Hanami::Router` default responses will be returned for these cases, which include the appropriate error codes.
- When hanami-webconsole **is bundled**, and when `app.config.render_detailed_errors` is true (which will be the default for all new Hanami 2.1 apps), then rely on the changes in https://github.com/hanami/webconsole/pull/7 to capture the error rescued by `BetterErrors::Middleware` and then re-apply an appropriate response code based on the `app.config.render_error_responses` setting that we already use to map exceptions to response types. This is a nice reuse of this existing setting (which was previously only used for the production-mode `render_errors` middleware) and provides greater consistency across both dev and production mode.

Resolves #1327.

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -952,8 +952,11 @@ module Hanami
         config = self.config
         rack_monitor = self["rack.monitor"]
 
+        render_errors = render_errors?
+        render_detailed_errors = render_detailed_errors?
+
         error_handlers = {}.tap do |hsh|
-          if render_errors?
+          if render_errors || render_detailed_errors
             hsh[:not_allowed] = ROUTER_NOT_ALLOWED_HANDLER
             hsh[:not_found] = ROUTER_NOT_FOUND_HANDLER
           end
@@ -974,7 +977,7 @@ module Hanami
             Hanami::Middleware::PublicErrorsApp.new(slice.root.join("public"))
           )
 
-          if config.render_detailed_errors && Hanami.bundled?("hanami-webconsole")
+          if render_detailed_errors
             require "hanami/webconsole"
             use(Hanami::Webconsole::Middleware, config)
           end
@@ -988,8 +991,11 @@ module Hanami
       end
 
       def render_errors?
-        config.render_errors ||
-          (config.render_detailed_errors && Hanami.bundled?("hanami-webconsole"))
+        config.render_errors
+      end
+
+      def render_detailed_errors?
+        config.render_detailed_errors && Hanami.bundled?("hanami-webconsole")
       end
 
       ROUTER_NOT_ALLOWED_HANDLER = -> env, allowed_http_methods {
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -976,7 +976,7 @@ module Hanami
 
           if config.render_detailed_errors && Hanami.bundled?("hanami-webconsole")
             require "hanami/webconsole"
-            use(Hanami::Webconsole::Middleware)
+            use(Hanami::Webconsole::Middleware, config)
           end
 
           if Hanami.bundled?("hanami-controller") && config.actions.sessions.enabled?
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -952,12 +952,16 @@ module Hanami
         config = self.config
         rack_monitor = self["rack.monitor"]
 
+        render_errors =
+          config.render_errors ||
+          (config.render_detailed_errors && Hanami.bundled?("hanami-webconsole"))
+
         Slice::Router.new(
           inspector: inspector,
           routes: routes,
           resolver: config.router.resolver.new(slice: self),
-          not_allowed: ROUTER_NOT_ALLOWED_HANDLER,
-          not_found: ROUTER_NOT_FOUND_HANDLER,
+          not_allowed: ROUTER_NOT_ALLOWED_HANDLER.curry[render_errors],
+          not_found: ROUTER_NOT_FOUND_HANDLER.curry[render_errors],
           **config.router.options
         ) do
           use(rack_monitor)
@@ -981,13 +985,21 @@ module Hanami
         end
       end
 
-      ROUTER_NOT_ALLOWED_HANDLER = -> env, allowed_http_methods {
-        raise Hanami::Router::NotAllowedError.new(env, allowed_http_methods)
+      ROUTER_NOT_ALLOWED_HANDLER = -> render_errors, env, allowed_http_methods {
+        if render_errors
+          raise Hanami::Router::NotAllowedError.new(env, allowed_http_methods)
+        end
+
+        Hanami::Router::NOT_ALLOWED.call(env, allowed_http_methods)
       }.freeze
       private_constant :ROUTER_NOT_ALLOWED_HANDLER
 
-      ROUTER_NOT_FOUND_HANDLER = -> env {
-        raise Hanami::Router::NotFoundError.new(env)
+      ROUTER_NOT_FOUND_HANDLER = -> render_errors, env {
+        if render_errors
+          raise Hanami::Router::NotFoundError.new(env)
+        end
+
+        Hanami::Router::NOT_FOUND.call(env)
       }.freeze
       private_constant :ROUTER_NOT_FOUND_HANDLER
 
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -952,16 +952,18 @@ module Hanami
         config = self.config
         rack_monitor = self["rack.monitor"]
 
-        render_errors =
-          config.render_errors ||
-          (config.render_detailed_errors && Hanami.bundled?("hanami-webconsole"))
+        error_handlers = {}.tap do |hsh|
+          if render_errors?
+            hsh[:not_allowed] = ROUTER_NOT_ALLOWED_HANDLER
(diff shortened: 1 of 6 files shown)


lib/hanami/slice.rb
```diff
@@ -981,13 +985,21 @@ def load_router(inspector:)
...
       end
 
-      ROUTER_NOT_ALLOWED_HANDLER = -> env, allowed_http_methods {
-        raise Hanami::Router::NotAllowedError.new(env, allowed_http_methods)
+      ROUTER_NOT_ALLOWED_HANDLER = -> render_errors, env, allowed_http_methods {
+        if render_errors
+          raise Hanami::Router::NotAllowedError.new(env, allowed_http_methods)
+        end
+
+        Hanami::Router::NOT_ALLOWED.call(env, allowed_http_methods)
```
timriley, line 993: Here and below we're using the default proc from `Hanami::Router` for the fallback behaviour if we're not configured for any kind of nice error rendering in the app.

These are flagged as `@api private` over in that gem, but I feel this is reasonable usage here given these two gems are both maintained by is.

timriley: No longer relevant, I took a different approach (see last commit)

lib/hanami/slice.rb
```diff
@@ -981,6 +987,11 @@ def load_router(inspector:)
         end
       end
 
+      def render_errors?
+        config.render_errors ||
+          (config.render_detailed_errors && Hanami.bundled?("hanami-webconsole"))
```
jodosha, line 992: @timriley Why do we need `hanami-webconsole` knowledge here?

Isn't webconsole a Rack middleware that is able to catch exceptions and present them in a nice page?

timriley: Yes, it is, but if it’s not bundled, then the value of render_detailed_errors here is moot: the middleware won’t be activated, it won’t catch the exceptions raised from our procs below, and your not found pages will come back as 500 errors instead of 404 responses.

This method is just mirroring the logic above from inside load_router. The duplication isn’t ideal but it seems manageable for now.

jodosha: But why our 404 responses are raising exceptions?
Given the router doesn't raise any exceptions.

timriley: I might be able to explore some other options here where we have our standard RenderErrors middleware provide some sensible fallbacks in the case of render_errors setting being false, but this here is the simplest fix for the short term, and it won't prevent us from taking a more sophisticated approach later.

timriley: @jodosha the 404s are being made into exceptions so our error rendering features (render_errors or rendered_detailed_errors) can pick them up and handle them as appropriate. 

I don't think you spent much time familiarizing yourself that work because it happened during one of those phases when you got very busy with non-hanami matters, which is why this might seem foreign to you now. 

See:

- https://github.com/hanami/hanami/pull/1309
- https://github.com/hanami/hanami/pull/1311

jodosha: @timriley I'm fine with the short-term fix, but we need to discuss simplifying the implementation.

To summarize: the router doesn't raise exceptions, but we raise it to print a nice error, and now we're handling those exceptions to not bubble up. 😉

jodosha (changes requested):
@timriley The fix seems not to be working. 😿 

Steps to reproduce:

```
⚡ bundle show hanami
/Users/jodosha/.gem/ruby/3.2.2/bundler/gems/hanami-f9489ddd2d29 # LATEST COMMIT
⚡ bundle exec hanami server
```

```bash
⚡ curl -I http://localhost:2300/foo
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
Set-Cookie: BetterErrors-2.9.1-CSRF-Token=9722597f-15f3-437c-95a2-759257783a7d; path=/; HttpOnly; SameSite=Strict
Content-Length: 2514
```

timriley:
@jodosha Ah, that's because in this first cut I focused on returning the appropriate error code only when webconsole wasn't bundled (and/or when render_detailed_errors was disabled).

Next time I have a chance, I'll come back and prepare a fix for the case when these are active.

timriley:
I've now tested this with an app using this branch as well as the branch for https://github.com/hanami/webconsole/pull/7, and here we can see a 404 is being appropriately returned for a not found error (while still incorporating all of the BetterErrors output):

```
❯ curl -v http://localhost:2300/blorp
*   Trying 127.0.0.1:2300...
* Connected to localhost (127.0.0.1) port 2300 (#0)
> GET /blorp HTTP/1.1
> Host: localhost:2300
> User-Agent: curl/8.1.2
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Content-Type: text/plain; charset=utf-8
< Set-Cookie: BetterErrors-2.9.1-CSRF-Token=3a7b7b35-0e91-4cd4-98a4-d90d289bf873; path=/; HttpOnly; SameSite=Strict
< Content-Length: 2959
<
Hanami::Router::NotFoundError at /blorp
=======================================

No route found for GET /blorp

> To access an interactive console with this error, point your browser to: /__better_errors


 () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/hanami-b8528c749671/lib/hanami/slice.rb, line 1001
--------------------------------------------------------------------------------------------------------------------------------------

``` ruby
  996           raise Hanami::Router::NotAllowedError.new(env, allowed_http_methods)
  997         }.freeze
  998         private_constant :ROUTER_NOT_ALLOWED_HANDLER
  999
  1000         ROUTER_NOT_FOUND_HANDLER = -> env {
> 1001           raise Hanami::Router::NotFoundError.new(env)
  1002         }.freeze
  1003         private_constant :ROUTER_NOT_FOUND_HANDLER
  1004
  1005         # rubocop:enable Metrics/AbcSize
  1006       end
```

App backtrace
-------------



Full backtrace
--------------

 -  () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/hanami-b8528c749671/lib/hanami/slice.rb:1001:in `block in <module:ClassMethods>'
 -  () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/router-b502a68d66f6/lib/hanami/router.rb:664:in `not_found'
 -  () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/router-b502a68d66f6/lib/hanami/router.rb:105:in `call'
 - better_errors (2.9.1) lib/better_errors/middleware.rb:87:in `protected_app_call'
 - better_errors (2.9.1) lib/better_errors/middleware.rb:82:in `better_errors_call'
 - better_errors (2.9.1) lib/better_errors/middleware.rb:60:in `call'
 -  () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/webconsole-3395783347c9/lib/hanami/webconsole/middleware.rb:56:in `call'
 -  () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/hanami-b8528c749671/lib/hanami/middleware/render_errors.rb:57:in `call'
 - dry-monitor (1.0.1) lib/dry/monitor/rack/middleware.rb:36:in `block in call'
 - dry-monitor (1.0.1) lib/dry/monitor/clock.rb:15:in `measure'
 - dry-monitor (1.0.1) lib/dry/monitor/rack/middleware.rb:36:in `call'
 -  () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/router-b502a68d66f6/lib/hanami/middleware/app.rb:40:in `call'
 -  () Users/tim/.local/share/rtx/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/hanami-b8528c749671/lib/hanami/slice.rb:758:in `call'
 - puma (6.3.1) lib/puma/configuration.rb:270:in `call'
 - puma (6.3.1) lib/puma/request.rb:100:in `block in handle_request'
 - puma (6.3.1) lib/puma/thread_pool.rb:344:in `with_force_shutdown'
 - puma (6.3.1) lib/puma/request.rb:99:in `handle_request'
 - puma (6.3.1) lib/puma/server.rb:443:in `process_client'
 - puma (6.3.1) lib/puma/server.rb:245:in `block in run'
 - puma (6.3.1) lib/puma/thread_pool.rb:151:in `block in spawn_thread'
```

jodosha (approved):

PR #1334, opened 2023-09-20
Add stub_assets test helper to generate manifest; fix tests requiring manifest

timriley:
After https://github.com/hanami/assets/pull/127, hanami-assets will now raise an error when fetching assets but no manifest file exists.

This led to failures in Hanami's tests, where we have a bunch of unit tests for our asset helpers that bypass the full `assets compile` process for the sake of simplicity and performance.

To address this, introduce a `stub_assets` test helper that can be used to manually generate a manifest file with assets for the tests.

### DIFF
--- a/spec/unit/hanami/helpers/assets_helper/javascript_tag_spec.rb
+++ b/spec/unit/hanami/helpers/assets_helper/javascript_tag_spec.rb
@@ -56,6 +56,8 @@ RSpec.describe Hanami::Helpers::AssetsHelper, "#javascript_tag", :app_integratio
         }
       CSS
 
+      stub_assets("feature-a.js")
+
       require "hanami/setup"
       before_prepare if respond_to?(:before_prepare)
       require "hanami/prepare"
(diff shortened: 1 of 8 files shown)


spec/unit/hanami/helpers/assets_helper/javascript_tag_spec.rb
```diff
@@ -76,7 +78,7 @@ class Context < Hanami::View::Context
     expect(actual).to eq(%(<script src="/assets/feature-a.js" type="text/javascript"></script>))
   end
 
-  it "renders <script> tag without appending ext after query string" do
+  xit "renders <script> tag without appending ext after query string" do
```
timriley, line 81: @jodosha This one (and the similar one in `stylesheet_link_tag_spec.rb` fails now where it passed before, because we were much looser with our asset URL generation before we required the manifest to be present.

I'm not sure what we actually expect the behaviour to be here?

Do we want to...

1. Find the right asset by taking `"feature-x?callback=init"` and looking up the asset for `"feature-x.js"`
2. Then take the resulting asset path, _remove_ the extension again, and then re-add the query string on the end?

If so, that will require some updates to the logic in our helpers. (Which I can happily do in a follow-up PR)

But if there's a different approach we should take, let me know and I can make that happen! I'm not really familiar with the use case here, so I'd appreciate your advice :)

jodosha: @timriley I am trying to remember why we have this test here. Leave it skipped for now.

jodosha (approved):

PR #1336, opened 2023-09-22
Rake task for hanami-assets: `bundle exec rake assets:precompile`

jodosha:
Closes https://github.com/hanami/hanami/issues/1335

---

⚠️ Note for the reviewers: if you try to list the Rake tasks in a Hanami app (`rake -T`), this Rake task will NOT show on purpose.

We want to encourage developers to use the `hanami` CLI. In this case to use `bundle exec hanami assets compile`.
This Rake task is here only for compatibility with the most common Ruby hosting, that expect this Rake task to be present, because of Ruby on Rails.

### DIFF
--- a/lib/hanami/config/null_config.rb
+++ b/lib/hanami/config/null_config.rb
@@ -9,6 +9,9 @@ module Hanami
     # loaded)
     class NullConfig
       include Dry::Configurable
+
+      def finalize!(*)
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


jodosha:
Manually tested with https://github.com/jodosha/hanami-210-beta2-test-app/commit/507a203a4617c93c683619845e1d77303e600a6f

timriley (approved):
Looks good, and I like the strategy of hiding the rake task from the `-T` output 👍🏼

PR #1337, opened 2023-09-29
Fix error logging on handled exceptions

### LINKED ISSUE #1338: Handled errors should not log as errors
See https://discourse.hanamirb.org/t/how-can-i-avoid-40x-error-from-logging-as-error/853

timriley:
In a forum post entitled [How can I avoid 40x error from logging as ERROR?](https://discourse.hanamirb.org/t/how-can-i-avoid-40x-error-from-logging-as-error/853), a Hanami 2 user reported that exceptions handled by `handle_exception` in their actions were still being logged as errors, _in addition to_ another log line for the expected request status.

This PR addresses this by fixing the overloaded `#_handle_exception` in our framework-level `Hanami::Action` extension. This method is called for _any_ error raised inside an action, regardless of whether it is configured to be handled by the user. So instead of instrumenting (and therefore logging) the error every time, only do so if an unhandled exception is raised from within this method. This ensures that only true exceptions get logged as errors.

Along with this, I've also expanded the rest of test coverage for request logging, and brought all logging-related tests under a single directory so we can view and work on them more easily.

_This PR is split into 2 commits: the first commit fixes the bug above, and the second commit reorganises the tests._

Resolves #1338, and fixes the behavior originally introduced in #1241.


jodosha (approved):
@timriley Thanks for this fix.

It works as expected. I manually tested it with https://github.com/jodosha/hanami-210-beta2-test-app/commit/cf220fc1231014aedbc2ee7f364558249a19eaea

PR #1344, opened 2023-10-04
Automatically mount `Rack::MethodOverride`

### LINKED ISSUE #1343: Use rack method override middleware by default

jodosha:
## Enhancement

Enable by default _method override_ by mounting `Rack::MethodOverride` middleware.

Users can opt out via the following app setting (in `config/app.rb`):

```ruby
require "hanami"

module TestApp
  class App < Hanami::App
    config.actions.method_override = false
  end
end
```

---

Closes #1343

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -982,8 +982,15 @@ module Hanami
             use(Hanami::Webconsole::Middleware, config)
           end
 
-          if Hanami.bundled?("hanami-controller") && config.actions.sessions.enabled?
-            use(*config.actions.sessions.middleware)
+          if Hanami.bundled?("hanami-controller")
+            if config.actions.method_override
+              require "rack/method_override"
+              use(Rack::MethodOverride)
+            end
+
+            if config.actions.sessions.enabled?
+              use(*config.actions.sessions.middleware)
+            end
           end
 
           if Hanami.bundled?("hanami-assets") && config.assets.serve
--- a/lib/hanami/config/actions.rb
+++ b/lib/hanami/config/actions.rb
@@ -74,6 +74,20 @@ module Hanami
       # @since 2.0.0
       attr_accessor :content_security_policy
 
+      # @!attribute [rw] method_override
+      #   Sets or returns whether HTTP method override should be enabled for action classes.
+      #
+      #   Defaults to true. You can override this by explicitly setting a
+      #   true or false value.
+      #
+      #   When true, this will mount `Rack::MethodOverride` in the Rack middleware stack of the App.
+      #
+      #   @return [Boolean]
+      #
+      #   @api public
+      #   @since 2.1.0
+      setting :method_override, default: true
+
       # The following settings are for view and assets integration with actions, and are NOT
       # publicly released as of 2.0.0. We'll make full documentation available when these become
       # public in a subsequent release.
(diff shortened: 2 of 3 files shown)


lib/hanami/slice.rb
```diff
@@ -982,6 +982,11 @@ def load_router(inspector:)
             use(Hanami::Webconsole::Middleware, config)
           end
 
+          if Hanami.bundled?("hanami-router")
+            require "rack/method_override"
+            use(Rack::MethodOverride)
+          end
```
timriley, line 988:
```suggestion
          require "rack/method_override"
          use(Rack::MethodOverride)
```

I don't think we need the hanami-router bundled check here, because we're already inside a block being passed to `Slice::Router`, which itself has already required hanami/router (at the top of `lib/hanami/slice/router.rb`), so an extra check at this point is redundant.

timriley (commented):
The changes as they are here look good, @jodosha!

I did have one code suggestion for you, hopefully a trivial one to apply.

---

Zooming back a bit, I think what we have here is just one half of what we need to ship for 2.1. While we're including this by default, I think we should also allow a way for it to be _removed_ by people who don't want it. For example, users who're building straight-up APIs and don't need this behavior, like you mentioned to me previously.

Given this, I wonder if `use`-ing this middleware right inside the block we're providing to our `Slice::Router` is the right place for it. Any middleware that goes here has no easy way to be removed, so I think we should reserve this for absolutely critical middleware only.

So, what do you think about the idea of moving this middleware to our default `config.middleware` instead?

It could look _something_ like this, inside `config.rb`:

```ruby
      @router = load_dependent_config("hanami-router") {
        require_relative "config/router"
        @middleware = Slice::Routing::Middleware::Stack.new

        require "rack/method_override"
        @middleware.use(Rack::MethodOverride)
       
        Router.new(self)
      }
```

This would see the `MethodOverride` middleware applied to our `Slice::Router` by default, thanks to the `middleware_stack.update(config.middleware_stack)` line we have inside `slice.rb`. The only difference here would be that the middleware gets added slightly later in the stack.

The benefit, however, is that we could then add a `Slice::Routing::Middleware::Stack#delete` method that will allow a user to do something like this inside their app class:

```
config.middleware.delete Rack::MethodOverride
```

This way, we're at least introducing this default while also providing a way for users to opt out.

(Later, we can look for a convenience method to easily apply defaults for APIs vs full stack apps, but I think we can wait until post-2.2 for that).

jodosha:
@timriley I changed the approach by introducing a consistent way to toggle features: `config.actions.method_override`.

This aligns with other settings and is more elegant than `config.middleware.delete Rack::MethodOverride`.

See the updated description of this PR.

timriley (approved):
Bravo, @jodosha! This solution is much cleaner while still giving us the flexibility we want. 👏🏼 

(Longer-term I still think it would be nice to add a `#delete` method to our middleware stack, but I like that we could avoid it being the only option for removing the method override middleware).

Technically, this _could_ go into `config.router`, since it's a middleware inserted into the router and doesn't otherwise depend on anything in hanami-controller. However, I think putting it in the actions config feels appropriate in spirit, so I'm good with it there 👍🏼

jodosha:
@timriley Yes I know it's technically going to the router. But this is an implementation detail.

So far we added all the HTTP related settings under `actions`. To make life easier of our users, I added this setting there as well. 🙂

timriley:
@jodosha yep, it's good with me! Thanks!

PR #1345, opened 2023-10-06
Refactor `SliceRegistrar#load_slice` to avoid an exception

parndt:
Currently we are trying to require a file that we aren't sure exists and
rescuing the `LoadError` that we get if it doesn't. 

We can completely avoid an exception per slice for the knowable case
where this file doesn't exist by first checking using `Pathname#sub_ext`
to add the `RB_EXT` extension to the path and then making use of
`Pathname#file?` on the resulting path.

### DIFF
--- a/lib/hanami/slice_registrar.rb
+++ b/lib/hanami/slice_registrar.rb
@@ -101,11 +101,12 @@ module Hanami
     # at `slices/[slice_name]`. Attempts to require the slice class, if defined, before registering
     # the slice. If a slice class is not found, registering the slice will generate the slice class.
     def load_slice(slice_name)
-      slice_require_path = root.join(CONFIG_DIR, SLICES_DIR, slice_name).to_s
+      slice_require_path = root.join(CONFIG_DIR, SLICES_DIR, slice_name)
+
       begin
-        require(slice_require_path)
+        require(slice_require_path.to_s) if slice_require_path.sub_ext(RB_EXT).file?
       rescue LoadError => e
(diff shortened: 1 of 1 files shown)


parndt:
Interesting, it seems like an exception is required to be thrown here 🤔

timriley:
So if I'm understanding this right, @parndt, we shouldn't need the begin/rescue at all, because we're directly checking for file existence?

parndt:
> So if I'm understanding this right, @parndt, we shouldn't need the begin/rescue at all, because we're directly checking for file existence before attempting to require?

@timriley technically, yes, but it doesn't protect us from errors in the file itself, I guess? I think it would be good to expose those though rather than trying to be defensive. It doesn't break any tests so I've rebased from main and pushed it.

timriley:
Thanks @parndt! Just FYI, that `rescue`—by design—didn't protect from LoadErrors raised from _within_ the file because of the re-raising that happened after:

```
raise e unless e.path == slice_require_path
```

To put this into words: this would only swallow a `LoadError` if the error itself was for the _exact slice file_ that we were trying to load. Any other errors would be re-raised.

So given you're now protecting us from requiring non-existent slice files, we no longer need this rescue at all 😄

parndt:
Ah of course, that's why it was being re-raised. 👍

timriley (approved):

timriley:
Happy for you to merge this, @parndt, thanks!

PR #1346, opened 2023-10-06
Extract `slice_module_name` method to avoid duplication

parndt:
We have identical code in two adjacent methods which is ripe for an
extract method refactor to avoid duplication and the potential for
drift.


timriley:
Nice little improvement, thanks @parndt!

PR #1347, opened 2023-10-07
Allow slice definitions to also exist in `slices/[slice_name]/config/slice.rb`

parndt:
Currently the slice has no way to provide its own `Slice` class but when
working with slices it is very helpful to be able to define one so that
configuration can be set without expecting the app to do it for the
slice.

This change allows a slice to have its class defined inside the slice
at a predefined location of `slices/[slice_name]/config/slice.rb`. This
works for slices nested under the main app as well as slices nested
inside other slices. Each layer can be overridden by the layer above it
so that everything can be re-configured by its containing slice or app.


lib/hanami/slice_registrar.rb
```diff
@@ -85,6 +82,16 @@ def with_nested
...
 
+    def ancestors
+      list = []
+      ancestor = parent
+      while ancestor
+        list.unshift(ancestor)
+        ancestor = ancestor.parent
+      end
+      list
+    end
```
parndt, line 93: keen for any input on a better way to do this?

lib/hanami/slice_registrar.rb
```diff
@@ -46,19 +46,16 @@ def freeze
     end
 
     def load_slices
-      slice_configs = Dir[root.join(CONFIG_DIR, SLICES_DIR, "*#{RB_EXT}")]
-        .map { |file| File.basename(file, RB_EXT) }
+      slice_configs = root.join(CONFIG_DIR, SLICES_DIR).glob("*#{RB_EXT}")
```
timriley, line 49: TIL `Pathname#glob`. Seriously, it has to be one of the best Ruby classes out there!

lib/hanami/slice_registrar.rb
```diff
@@ -46,19 +46,16 @@ def freeze
...
+        .map { _1.basename(RB_EXT) }
```
timriley, line 50: This is a good use of `_1` IMO, especially when combined with `Pathname#basename`. Appreciate you.

timriley: (General appreciation extends to the similar refactoring in the stanza below)

lib/hanami/slice_registrar.rb
```diff
@@ -97,16 +104,13 @@ def parent_slice_namespace
       parent.eql?(parent.app) ? Object : parent.namespace
     end
 
-    # Runs when a slice file has been found at `config/slices/[slice_name].rb`, or a slice directory
-    # at `slices/[slice_name]`. Attempts to require the slice class, if defined, before registering
-    # the slice. If a slice class is not found, registering the slice will generate the slice class.
+    # Runs when a slice file has been found at `config/slices/[slice_name].rb`, or else inside
+    # the slice at `slices/[slice_name]/config/slice.rb`, or a slice directory at `slices/[slice_name]`.
```
timriley, line 108: Technically, "or else inside the slice at `slices/[slice_name]/config/slice.rb`" doesn't really apply here, since we only get to the point of running `load_slice` if only one of these two conditions are met?

1. slice file has been found at `config/slices/[slice_name].rb`, OR
2. a slice directory has been found at `slices/[slice_name]`

The job of looking for the slice file inside `slices/[slice_name]/config/slice.rb` happens _within_ this method rather than _before_ it, right? So perhaps we could tighten up the comment here to make sure it's not misleading at all?

lib/hanami/slice_registrar.rb
```diff
@@ -85,6 +82,16 @@ def with_nested
 
     private
 
+    def ancestors
```
timriley, line 85: Could we please shift this method so it's below `find_slice_require_path`? It's only used by that one method and I generally find classes easier to follow when related methods like that stick together.

timriley: Also, could you please put a little bit of a documentation comment on top of this one? It's a very abstract kind of method so it'd be useful to to have a one-liner explaining roughly what it does and why we use it.

(I know nothing else in this class has doc comments, but now's always a good time to start!)

lib/hanami/slice_registrar.rb
```diff
@@ -85,6 +82,16 @@ def with_nested
...
+      list = []
+      ancestor = parent
+      while ancestor
+        list.unshift(ancestor)
```
timriley, line 89: Can we please use `prepend` here instead of `unshift`? I don't know about you, but `unshift` always breaks my brain. And for a very abstract method like this, which is already encroaching into brain-breaking territory, the kinder we can be to the readers' brains, the better 😄

lib/hanami/slice_registrar.rb
timriley: I think this is fine as a first cut. I'm sure it could possibly be done via some kind of `reduce` fanciness, but I think this procedural approach is simple to read, and we can always come back and do a targeted refactoring later. My biggest concerns here are about making the _purpose_ of this method super clear, which I think we can address via doc comments.

lib/hanami/slice_registrar.rb
```diff
@@ -118,6 +122,15 @@ def load_slice(slice_name)
       register(slice_name, slice_class)
     end
 
+    def find_slice_require_path(slice_name)
```
timriley, line 125: A doc comment about this would be good too. Explain the algorithm and the order of preference for slice file lookup.

lib/hanami/slice_registrar.rb
```diff
@@ -118,6 +122,15 @@ def load_slice(slice_name)
...
+      ancestors
+        .map { _1.root.join(CONFIG_DIR, SLICES_DIR, slice_name) }
+        .push(root.join(SLICES_DIR, slice_name, CONFIG_DIR, "slice"))
+        .uniq
+        .find { _1.sub_ext(RB_EXT).file? }
+        &.to_s
```
timriley, line 131: The implementation here is nice and straightforward to follow! 👏🏼

spec/integration/slices_spec.rb
```diff
@@ -81,6 +81,188 @@ class Slice < Hanami::Slice
...
+        end
+      RUBY
+
+      write "slices/main/config/slice.rb", <<~RUBY
+        module Main
+          class Slice < Hanami::Slice
+            config.actions.format :json
+          end
+        end
+      RUBY
```
timriley, line 137: My understanding of the implementation is that, in this scenario, this file here should never get loaded at all. Is that right?

To make that even clearer for a test like this, could we actually just have the file look something like this?

```ruby
write "slices/main/config/slice.rb", <<~RUBY
  raise "I should never be loaded"
RUBY
```

I tested this locally and everything still passes. What do you think?

If we did make this change, we should probably change the one remaining slice class to use  `.format :json` just for overall consistency.

spec/integration/slices_spec.rb
```diff
@@ -81,6 +81,188 @@ class Slice < Hanami::Slice
...
+
+      require "hanami/prepare"
+
+      expect(Hanami.app.slices[:main]).to be Main::Slice
+      expect(Hanami.app.slices[:main].slices[:nested].config.actions.format).to eq([:json])
+    end
+  end
+
+  it "Loading a nested slice with its own defined slice class prefers the parent slice's defined slice class" do
```
timriley, line 174: Getting real _“How many slices would a Hanami app slice if a Hanami app could slice slices?”_ vibes from this description 😆

spec/integration/slices_spec.rb
```diff
@@ -81,6 +81,188 @@ class Slice < Hanami::Slice
...
+
+      write "slices/main/slices/nested/config/slice.rb", <<~RUBY
+        module Main
+          module Nested
+            class Slice < Hanami::Slice
+              config.actions.format :json
+            end
+          end
+        end
+      RUBY
```
timriley, line 205: Likewise, let's replace this with a simple `raise` if you agree with my thoughts on the earlier example. After seeing this a second time (in this much more complex nested structure), I think this approach can really improve understanding of the test.

timriley: (Same for the test below, but I won't bother you with another inline comment about that.)

spec/integration/slices_spec.rb
```diff
@@ -81,6 +81,188 @@ class Slice < Hanami::Slice
...
+        require "hanami"
+
+        module Main
+          module Nested
+            class Slice < Hanami::Slice
+              config.actions.format :csv
+            end
+          end
+        end
+      RUBY
```
timriley, line 235: This one actually surprised me! I'm not sure we want this. Because `nested` is not this slice's actual “fully qualified” name, IMO; I'd expected it to be `main/nested`.

Nesting of slices is already complex enough. If possible, I think for now we should eliminate this particular method of falling back to the "grandparent" (or "great parent" or so on for even deeper nesting) for loading a slice's class file.

So, given a main->nested slice class, here's where I'd expect us to allow its slice class to be defined:

1. `slices/main/config/slices/nested.rb`
2. `slices/main/slices/nested/config/slice.rb`

There is no step three. Same applies for any arbitrary level of nesting: load the slice class by it's parent's `config/slices/[slice_name].rb`, or by its own internal `config/slice.rb` only.

What do you think? We could always extend this later, but given how niche and potentially complex nested slices are, I'd prefer to keep the rules simpler for starters.

I think this might have the beneficial side effect of allowing you to simplify or even remove the need for the `SliceRegistrar#ancestors` method?

What do you think?

timriley (changes requested):
Thanks for taking care of this, @parndt. Overall, I'm very happy to see this change go in, and your implementation looks solid to me.

I've left some feedback for you via inline comments. Nothing major: a few nips and tucks, and a bit of extra doco via comments.

Just one big thing did leap out: per my comment on the very last test, I think we should restrict how far up the chain of ancestor slices we go in terms of looking up the slice class, at least for now. Keen for your thoughts.

spec/integration/slices_spec.rb
parndt: Ah yeah, good call

spec/integration/slices_spec.rb
parndt: ALL OF THEM

spec/integration/slices_spec.rb
parndt: Yeah, I'm totally fine with adding this limitation, I would want this order of precedence
- <app>/config/slices/main/nested.rb
- slices/main/config/slices/nested.rb
- slices/main/slices/nested/config/slice.rb

so the app's file loads ahead of the enclosing slice's file which loads ahead of the actual slice's file.

spec/integration/slices_spec.rb
parndt: works great!

lib/hanami/slice_registrar.rb
```diff
@@ -122,10 +112,27 @@ def load_slice(slice_name)
...
+    # 3. `slices/[slice_name]/config/slice.rb`
+    #
+    # If the slice is nested under another slice then it will look in the following order:
+    #
+    # 1. `config/slices/[parent_slice_name]/[slice_name].rb`
+    # 2. `slices/[parent_slice_name]/config/[slice_name].rb`
+    # 3. `slices/[parent_slice_name]/[slice_name]/config/slice.rb`
     def find_slice_require_path(slice_name)
+      app_slice_file_path = [slice_name]
+      app_slice_file_path.prepend(parent.slice_name) unless parent.eql?(parent.app)
```
timriley, line 131: I think this won't work for deeply nested slices, right? So in the case of a slice structure like:

```
slices/top/slices/middle/slices/bottom/
```

Then this code would look for `config/slices/middle/bottom.rb`, when it should _actually_ be looking for `config/slices/top/middle/bottom.rb`, right?

timriley (commented):
One last comment here, and I'll see if I can take a look at adjusting this myself, since I know you're at Ruby Retreat, @parndt :)

I've re-merged main to make sure the tests pass locally for me.

timriley:
Update: this is actually looking fine. Added new tests to exercise/demonstrate the behavior for deeply nested slices.

lib/hanami/slice_registrar.rb
timriley: I was wrong. For a deeply nested slice (e.g. `Top::Middle::Bottom::Slice`), its parent will have a `slice_name.to_s` of `top/middle`, which means we look for the correct slice path inside `/config/slices/`. The extra specs I've added in dc71e89 demonstrate this.

timriley:
Thanks @parndt! Just did a bit of spec expansion/cleanup and merged this in! Just in time for 2.1.0.rc1 😄

parndt:
Awesome!!!!!!!!!! Thank you @timriley !

PR #1352, opened 2023-10-13
Ensure compatibility with Ruby `Logger` from stdlib

### LINKED ISSUE #1295: Hanami 2 expects dry logger and generic ruby Logger class doesn't work
# Overview

When creating a new Hanami 2.0.3 application, and registering a logger provider, I am unable to use Ruby's Logger class -- Hanami seems to expect a dry-logger Dispatcher given it attempts to call a `#tagged` method.

# Steps to Reproduce

1. `gem install hanami`
2. `hanami new bookshelf && cd bookshelf`
3. Create a logger provider, using the standard ruby logger:
```
# frozen_string_literal: true

Hanami.app.register_provider :logger do
  start do
    require "logger"
    logger = Logger.new($stdout)

    register "logger", logger
  end
end
```
4. `hanami server`
5. Browse to http://localhost:2300/

The following stack trace now shows:
```
Puma caught this error: undefined method `tagged' for #<Logger:0x000000010328b7f8 @level=0, @progname=nil, @default_formatter=#<Logger::Formatter:0x000000010328b5f0 @datetime_format=nil>, @formatter=nil, @logdev=#<Logger::LogDevice:0x000000010328b3e8 @shift_period_suffix=nil, @shift_size=nil, @shift_age=nil, @filename=nil, @dev=#<IO:<STDOUT>>, @binmode=false, @mon_data=#<Monitor:0x000000010328b348>, @mon_data_owner_object_id=4360>>

        logger.tagged(:rack) do
              ^^^^^^^ (NoMethodError)
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/hanami-2.0.3/lib/hanami/web/rack_logger.rb:79:in `log_request'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/hanami-2.0.3/lib/hanami/web/rack_logger.rb:67:in `block in attach'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-monitor-1.0.1/lib/dry/monitor/notifications.rb:35:in `block in instrument'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:38:in `block in process'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:34:in `each'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-events-1.0.1/lib/dry/events/bus.rb:34:in `process'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-events-1.0.1/lib/dry/events/publisher.rb:269:in `process'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-monitor-1.0.1/lib/dry/monitor/notifications.rb:31:in `instrument'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-monitor-1.0.1/lib/dry/monitor/notifications.rb:25:in `stop'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/dry-monitor-1.0.1/lib/dry/monitor/rack/middleware.rb:37:in `call'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/hanami-router-2.0.2/lib/hanami/middleware/app.rb:40:in `call'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/hanami-2.0.3/lib/hanami/slice.rb:758:in `call'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/puma-6.1.0/lib/puma/configuration.rb:269:in `call'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/puma-6.1.0/lib/puma/request.rb:98:in `block in handle_request'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/puma-6.1.0/lib/puma/thread_pool.rb:340:in `with_force_shutdown'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/puma-6.1.0/lib/puma/request.rb:97:in `handle_request'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/puma-6.1.0/lib/puma/server.rb:431:in `process_client'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/puma-6.1.0/lib/puma/server.rb:233:in `block in run'
/Users/em.dwyer/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/gems/puma-6.1.0/lib/puma/thread_pool.rb:147:in `block in spawn_thread'
```

jodosha:
Closes #1295

### DIFF
--- a/lib/hanami/web/rack_logger.rb
+++ b/lib/hanami/web/rack_logger.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "delegate"
+
 module Hanami
   # @api private
   module Web
@@ -53,10 +55,33 @@ module Hanami
         end
       end
 
+      # @since 2.1.0
+      # @api private
+      class TaggedLogger < SimpleDelegator
+        # @since 2.1.0
+        # @api private
+        def self.call(logger)
+          return logger if logger.respond_to?(:tagged)
+          new(logger)
+        end
+
+        class << self
+          # @since 2.1.0
+          # @api private
+          alias [] call
+        end
+
+        # @since 2.1.0
+        # @api private
+        def tagged(*, &blk)
+          blk.call
+        end
+      end
+
       # @api private
       # @since 2.0.0
       def initialize(logger, env: :development)
-        @logger = logger
+        @logger = TaggedLogger[logger]
         extend(Development) if %i[development test].include?(env)
       end
 
--- a/lib/hanami/web/rack_logger.rb
+++ b/lib/hanami/web/rack_logger.rb
@@ -79,7 +79,7 @@ module Hanami
           end
 
           def accepts_entry_payload?(logger)
-            logger.method(:add).parameters.last.then { |type, _| type == :keyrest }
+            logger.method(:info).parameters.last.then { |type, _| type == :keyrest }
           end
         end
 
@@ -106,9 +106,18 @@ module Hanami
         #
         # @since 2.1.0
         # @api private
-        def add(severity, message = nil, **payload)
+        def info(message = nil, **payload)
           payload[:message] = message if message
-          logger.add(severity, JSON.fast_generate(payload))
+          logger.info(JSON.fast_generate(payload))
+        end
+
+        # @see info
+        #
+        # @since 2.1.0
+        # @api private
+        def error(message = nil, **payload)
+          payload[:message] = message if message
+          logger.info(JSON.fast_generate(payload))
         end
       end
 
@@ -136,11 +145,7 @@ module Hanami
       # @since 2.0.0
       def log_request(env, status, elapsed)
         logger.tagged(:rack) do
-          logger.add(
-            ::Logger::Severity::INFO,
-            nil,
-            **data(env, status: status, elapsed: elapsed),
-          )
+          logger.info(**data(env, status: status, elapsed: elapsed))
         end
       end
 
@@ -148,11 +153,7 @@ module Hanami
       # @since 2.0.0
       def log_exception(env, exception, status, elapsed)
         logger.tagged(:rack) do
-          logger.add(
-            ::Logger::Severity::ERROR,
-            exception,
-            **data(env, status: status, elapsed: elapsed)
-          )
+          logger.error(exception, **data(env, status: status, elapsed: elapsed))
         end
       end
 
--- a/lib/hanami/web/rack_logger.rb
+++ b/lib/hanami/web/rack_logger.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "delegate"
+require "json"
 
 module Hanami
   # @api private
@@ -57,18 +58,39 @@ module Hanami
 
       # @since 2.1.0
       # @api private
-      class TaggedLogger < SimpleDelegator
-        # @since 2.1.0
-        # @api private
-        def self.call(logger)
-          return logger if logger.respond_to?(:tagged)
-          new(logger)
-        end
-
+      class UniversalLogger
         class << self
           # @since 2.1.0
           # @api private
-          alias [] call
+          def call(logger)
+            return logger if compatible_logger?(logger)
+
+            new(logger)
+          end
+
+          # @since 2.1.0
+          # @api private
+          alias_method :[], :call
+
+          private
+
+          def compatible_logger?(logger)
+            logger.respond_to?(:tagged) && accepts_entry_payload?(logger)
+          end
+
+          def accepts_entry_payload?(logger)
+            logger.method(:add).parameters.last.then { |type, _| type == :keyrest }
+          end
+        end
+
+        # @since 2.1.0
+        # @api private
+        attr_reader :logger
+
+        # @since 2.1.0
+        # @api private
+        def initialize(logger)
+          @logger = logger
         end
 
         # @since 2.1.0
@@ -76,12 +98,24 @@ module Hanami
         def tagged(*, &blk)
           blk.call
         end
+
+        # Logs the entry as JSON.
+        #
+        # This ensures a reasonable (and parseable) representation of our log payload structures for
+        # loggers that are configured to wholly replace Hanami's default logger.
+        #
+        # @since 2.1.0
+        # @api private
+        def add(severity, message = nil, **payload)
+          payload[:message] = message if message
+          logger.add(severity, JSON.fast_generate(payload))
+        end
       end
 
       # @api private
       # @since 2.0.0
       def initialize(logger, env: :development)
-        @logger = TaggedLogger[logger]
+        @logger = UniversalLogger[logger]
         extend(Development) if %i[development test].include?(env)
       end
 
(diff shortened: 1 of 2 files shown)


solnic (approved):

timriley (changes requested):
At surface level, this looks like a good tactical fix to me 👍🏼 

One thing to note, however: supplying a logger without tagging capabilities does mean that the logger will output differently formatted log lines compared to our default configuration.

Default log lines for rack requests:

```
[test_app] [INFO] [2023-10-14 12:06:41 +1100] GET 200 43µs 127.0.0.1 / -
```

Log lines for rack requests when supplying a stdlib logger:

```
I, [2023-10-14T12:06:41.714334 #87825]  INFO -- : {:verb=>"GET", :status=>200, :ip=>"127.0.0.1", :path=>"/", :length=>"-", :params=>{}, :elapsed=>"187µs"}
```

This is because we supply a hash of data to the logger when logging from rack requests (see `Hanami::Web::RackLogger#log_request` and `#data`). Then we set up our default development logger to use a different formatter when log entries are tagged as `:rack` (see `Hanami::Config::Logger#development_logger`).

So in the case of a non-tagged logger, we're just `to_s`-ing the output, which is why we get the default Ruby string-formatted Hash appearing in the log lines.

This also impacts production-mode logs.

Hanami's default logger:

```
{"progname":"test_app","severity":"INFO","time":"2023-10-14T01:17:01Z","verb":"GET","status":200,"ip":"127.0.0.1","path":"/","length":"-","params":{},"elapsed":10,"elapsed_unit":"µs"}
```

When using stdlib logger:

```
I, [2023-10-14T12:17:01.122225 #88072]  INFO -- : {:verb=>"GET", :status=>200, :ip=>"127.0.0.1", :path=>"/", :length=>"-", :params=>{}, :elapsed=>22, :elapsed_unit=>"µs"}
```

With our default setup, we get nice JSON logs in production, but when using the stdlib logger, we have a string-formatted Hash (for the same reasons as outlined above).

It's also worth noting that none of our logger filtering config will apply when using a directly assigned logger like we have here in this new tests.

---

So what do we do with this finding?

My initial thought here was that we could add some kind of "compatibility-mode" logging capability to `Hanami::Web::RackLogger`. For example, if the logger didn't `respond_to?(:tagged)` or even if it wasn't a `kind_of?(Dry::Logger`), we could have separate methods that appropriately formats the string before passing it to the logger, e.g. `#log_request_compat` and `#log_exception_compat` to match the existing methods.

But that feels like the wrong place to do all of that stuff. It's a lot of specific logger setup handling in a place that should really just be logging the request details and getting on with its business.  I do have some working code here if you wanted to see it.

My next thought was that we could update `app.config.logger=` to wrap the given logger instance in a new `Dry::Logger::Dispatcher` so we can apply consistent rules to it as our default logger. I also have some working code here if you wanted to see it; it's nicer than the code for the above, but I also realise that this strips away some of the power of being able to provide your own logger.

And _then_ I realised that the new tests we set up here (based on the example in #1295) didn't even use `app.config.logger=`, instead they set up a provider that registers the logger in the container. At that point, the framework really has no control whatsoever over what kind of logger object ends up being registered.

So, after all of that, here's what I think we should do:

- Firstly, update our documentation to make it **very clear** that anyone providing a full replacement logger object is _opting out entirely_ from all framework-provided logging niceties, such as the development vs production formatting, and _especially_ the filtering of log lines. Anyone providing their own logger **MUST** make sure they provide their own filtering, at minimum.
- Then, start with the changes you've already made in this PR
- And update `Web::RackLogger` to check if the logger is a "legacy" logger, and if so, serialize the log entry as JSON before passing it to the logger. This at least ensures we have a standard format that we're passing to these legacy loggers.

I took the liberty of seeing what this would look like. Check it out here: https://github.com/hanami/hanami/compare/fix/ruby-logger-stdlib-compat...fix/ruby-logger-stdlib-compat--parseable-format

This feels like the most reasonable compromise solution of all the ideas I mooted above. I think we should very much try and avoid accepting an incidental `Hash#to_s` as part of our log output: it feels accidental and broken.

And one benefit of using JSON here instead is that it allows people using replacement loggers to re-parse it and provide their own filtering implementation based on that, which is the thing we should be emphasising in our docs.

Later on, if we ended up wanting to log some elsewhere inside framework code, we could lift this `UniversalLogger` wrapper and put it into a single reusable place, too, so we have consistent handling everywhere.

After getting to the bottom of this essay (thank you for reading this far!), what do you think?

I'm making this as "request changes" only because I think the `Hash#to_s` output is something we should be avoiding here. If you like the code in my branch, we could bring the commit into this one, and then I think we're good to go. Let me know! ❤️

timriley:
I realised my branch had failing tests. I've fixed those now.

---

We do have one other option here, and it's think it's worth considering: we make it so that any custom logger (whether it is set via `app.config.logger=` or registered directly via a provider) _must_ be wrapped by a Dry::Logger::Dispatcher, and provide some additional conveniences to make that easier.

dry-logger does play nicely with others — you can add your own logger instances as backends for the dispatcher, and the default instance is based on `::Logger` too, which serves as an example of how it could work with other systems. It does enforce a few opinions, though (like accepting log entries as structured data), and folks might find that's a step too far for them.

On the other hand, we *already take advantage of those opinions* inside our standard request logging code, given that we're doing tagged logging _and_ passing structured data, so at a framework level, we're already kind of tied to these dry-logger opinions.

Ultimately, I think the branch I pushed at least provides a reasonable fallback. So my opinion at this point is that we should incorporate that, or something like it.

And then in the future (post-2.2), we could possibly look at ways to make integrating custom loggers a but nicer.

jodosha:
@timriley 
> must be wrapped by a Dry::Logger::Dispatcher, and provide some additional conveniences to make that easier.

👎 We're getting in the way of our users.
One Hanami principle is: here's your default, here's your extension point.
In case advanced users opt in for the latter, we're just accepting it and moving on.

timriley:
@jodosha Hooray, I'm glad my deep dive here was helpful to you! 😅 

And yes, I agree with you that we should just accept any old regular logger and then do our best to make it work.

In terms of configuring custom loggers, we already support this via config! This works: `app.config.logger = my_logger_instance`. I believe this is the approach we should document first; providing a logger via a provider is advanced usage and we can hold that in reserve. 

Based on my background knowledge of #1295, I believe that the provider example was taken from code that likely existed before we added support for replacing the logger via config.

timriley (approved):

jodosha:
@timriley OK, if that is the primary approach, I'll change the integration tests accordingly.
We want to ensure that the method that we advertise is always working.

PR #1353, opened 2023-10-14
Render a welcome page when no routes are defined

### LINKED ISSUE #1350: Render new welcome view as the starter route

timriley:
The page includes the current Hanami and Ruby versions. Design and content for this page will be added in a later change.

The page is rendered by a new `Hanami::Web::Welcome` middleware that is added as the first in the stack when (a) in development mode, and (b) when no routes are defined.

To support this introspection into the routes (which are passed to the router as an opaque block), wrap the routes in a `RoutesProc` class that adds an `#empty?` method, returning true when no routes are defined.

Resolves #1350.

### DIFF
--- a/lib/hanami/routes.rb
+++ b/lib/hanami/routes.rb
@@ -60,6 +60,35 @@ module Hanami
       end
     end
 
+    # Wrapper class for the (otherwise opqaque) proc returned from {.routes}, adding an `#empty?`
+    # method that returns true if no routes were defined.
+    #
+    # This is useful when needing to determine behaviour based on the presence of user-defined
+    # routes, such as when determining to show the Hanami welcome page in {Slice#load_router}.
+    #
+    # @api private
+    # @since 2.1.0
+    class RoutesProc < DelegateClass(Proc)
+      # @api private
+      # @since 2.1.0
+      def self.empty
+        new(proc {}, empty: true)
+      end
+
+      # @api private
+      # @since 2.1.0
+      def initialize(proc, empty: false)
+        @empty = empty
+        super(proc)
+      end
+
+      # @api private
+      # @since 2.1.0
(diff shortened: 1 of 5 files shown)


lib/hanami/routes.rb
```diff
@@ -60,6 +60,35 @@ def initialize(endpoint)
       end
     end
 
+    # Wrapper class for the (otherwise opqaque) proc returned from {.routes}, adding an `#empty?`
```
parndt, line 63:
```suggestion
    # Wrapper class for the (otherwise opaque) proc returned from {.routes}, adding an `#empty?`
```

lib/hanami/routes.rb
```diff
@@ -60,6 +60,35 @@ def initialize(endpoint)
...
+    # method that returns true if no routes were defined.
+    #
+    # This is useful when needing to determine behaviour based on the presence of user-defined
+    # routes, such as when determining to show the Hanami welcome page in {Slice#load_router}.
```
parndt, line 67:
```suggestion
    # routes, such as determining whether to show the Hanami welcome page in {Slice#load_router}.
```

lib/hanami/web/welcome.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "erb"
```
parndt, line 3: do we now need a runtime dependency for erb or are we assuming it'll continue to ship with the standard library (this is true for upcoming version 3.3 so probably fine for another year at least!)

timriley: Good question! Personally, I'm comfortable relying on erb as distributed with Ruby for as long as we can.

jodosha (approved):

PR #1354, opened 2023-10-15
Use an intermediary `helpers` object on view parts

### LINKED ISSUE #1348: Isolate assets in view parts to a `helpers` object

timriley:
Instead of mixing the standard helpers directly into part classes (which runs a high risk of naming collisions, given parts wrap all kinds of objects), instead create an intermediary `helpers` object that contains all the helper methods.

This relies on a small change to the standard helper module interfaces (to make their methods remain public when included) in https://github.com/hanami/view/pull/242.

Resolves #1348.

### DIFF
--- a/lib/hanami/extensions/view/part.rb
+++ b/lib/hanami/extensions/view/part.rb
@@ -57,7 +57,7 @@ module Hanami
         attr_reader :_context
 
         # @api public
-        # @since 2.1.10
+        # @since 2.1.0
         alias_method :context, :_context
 
         # @api private
--- a/lib/hanami/extensions/view/part.rb
+++ b/lib/hanami/extensions/view/part.rb
@@ -4,19 +4,66 @@ module Hanami
   module Extensions
     module View
       # @api private
+      # @since 2.1.0
       module Part
         def self.included(part_class)
           super
 
           part_class.extend(Hanami::SliceConfigurable)
-          part_class.include(StandardHelpers)
           part_class.extend(ClassMethods)
         end
 
         module ClassMethods
           def configure_for_slice(slice)
-            extend SliceConfiguredHelpers.new(slice)
+            const_set :PartHelpers, Class.new(PartHelpers) { |klass|
+              klass.configure_for_slice(slice)
+            }
           end
+        end
+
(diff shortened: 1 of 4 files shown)


lib/hanami/extensions/view/part.rb
```diff
@@ -4,20 +4,67 @@ module Hanami
...
+        def self.configure_for_slice(slice)
+          extend SliceConfiguredHelpers.new(slice)
+        end
+
+        # @api public
+        # @since 2.1.0
+        attr_reader :_context
+
+        # @api public
+        # @since 2.1.10
```
jodosha, line 60:
```suggestion
        # @since 2.1.0
```

lib/hanami/extensions/view/part.rb
```diff
@@ -4,20 +4,67 @@ module Hanami
...
+
+        include StandardHelpers
+
+        def self.configure_for_slice(slice)
+          extend SliceConfiguredHelpers.new(slice)
+        end
+
+        # @api public
+        # @since 2.1.0
+        attr_reader :_context
```
jodosha, line 57: @timriley Why is this Public API?
As per Ruby conventions, a name starting with an underscore is a warning sign for the reader not to use that method because it's private ("for real").

I understand that we want to expose the context, but can't we just make `context` as Public API? So it's cleaner and it follows the convention described above.

timriley: @jodosha Inside hanami-view (previously dry-view), access to the context object is indeed public API, but we use this underscored version because the context is always located in objects where the user is supplying their own values, such as scopes/templates and parts.

Here's how we explain it in the parts documentation: https://dry-rb.org/gems/dry-view/0.7/parts/#accessing-the-context

In both the scope (templates) _and_ parts we actually provide `#context` as a convenience alias. But as soon as there is (a) a local in the scope or (b) wrapped value in a part that uses the `context` name, we preference that instead of our internal context object. This is why any library- or framework-provided code _must_ use `_context` instead. The non-prefixed `context` alias is there only as a nicety for user-land code.

Does that help explain things for you?

PR #1356, opened 2023-10-16
Remove short names for assets helpers

### LINKED ISSUE #1349: Roll back all "short names" for asset helpers (image, video, favicon, js, javascript, css, stylesheet)
See [this thread](https://hanamiruby.slack.com/archives/C04CD2W0A4D/p1696730569084679).

Having these short names is breaking:

1. Using those names as exposures passed to templates
2. _Also_ using those names as **_methods_** on those exposure values (because these helpers are mixed into part classes)

This breaks this:

```ruby
# in the view class
expose :image

# in the template
<%= image.whatever %>
```

It also breaks this:

```erb
<%= user_profile.image %>
```

Because of how likely it is these helper short names will conflict with:

- Exposure names
- Method names on values (wrapped as Parts) passed to views

I think we need to remove them. The view system in Hanami 2 is more powerful due to the presence of parts and the fact that our helpers are available there too, and because of this, it does mean we need to be more mindful around part naming.

I think we should also remove the `js`/`javascript`, `css`/`stylesheet` helper names too, because of this same risk naming collisions.

timriley:
We discovered these short names are likely to conflict with exposures passed to templates/scopes. To leave as many names open for the user’s values, remove the short names and keep the longer, more descriptive names for these helpers.

Specifically, that means keeping:

- `javascript_tag` (remove `javascript` and `js` aliases)
- `stylesheet_link_tag` (remove `stylesheet` and `css` aliases)
- `favicon_link_tag` (remove `favicon` alias)
- `image_tag` (remove `image` alias)
- `video_tag` (remove `video` alias)
- `audio_tag` (remove `audio` alias)

Resolves #1349.

### DIFF
--- a/lib/hanami/helpers/assets_helper.rb
+++ b/lib/hanami/helpers/assets_helper.rb
@@ -154,7 +154,7 @@ module Hanami
       #
       #   # <script src="https://assets.bookshelf.org/assets/application-28a6b886de2372ee3922fcaf3f78f2d8.js"
       #   #         type="text/javascript"></script>
-      def javascript(*source_paths, **options)
+      def javascript_tag(*source_paths, **options)
         options = options.reject { |k, _| k.to_sym == :src }
 
         _safe_tags(*source_paths) do |source|
@@ -172,14 +172,6 @@ module Hanami
           tag.script(**attributes).to_s
         end
       end
-
-      # @api public
-      # @since 2.1.0
-      alias_method :js, :javascript
-
-      # @api public
-      # @since 2.1.0
-      alias_method :javascript_tag, :javascript
 
       # Generate `link` tag for given source(s)
       #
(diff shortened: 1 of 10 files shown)


timriley:
@jodosha if there's any short names that you think are critical we keep, let's figure it out here. We'll just need to strike an appropriate balance given the risk of conflicts with exposure names.

jodosha (approved):
@timriley TL;DR: We're (almost) good to go.

PROS:
* These names are reducing the likelihood of name clashing
* We're exposing a small public API (no method aliases). If we find these methods inconvenient, we may consider later reintroducing aliases or any user-convenient method.

CONS:
* These are hard to type and remember

---

Can we have `stylesheet_tag` instead of `stylesheet_link_tag`?

timriley:
@jodosha Yeah, I'm happy to go with `stylesheet_tag` instead. We can make clear in the API docs that it generates a `<link>` tag, and folks would expect it to do the right thing anyway.

I might make the same change for `favicon_link_tag` too, changing it to `favicon_tag`. As one resulting benefit, the names becomes consistent.

timriley:
@jodosha pushed these changes. I know you've approved this already, but I'll wait for confirmation from you before merging.

jodosha (approved):

Issue #1355, opened 2023-07-02
Fix use of helpers within parts


bkuhlmann:
**Describe the bug**

Hello. :wave: I've run into an issue with using helpers within parts and wanted to detail the cause and temporary workaround.

**To Reproduce**

The problem is with using *any* helper within a part. Here's a simple implementation:

``` ruby
class Demo < Hanami::View::Part
  include Hanami::View::Helpers::TagHelper

  def label = tag.span value.label
end
```

<details>
  <summary>Spec</summary>

The following spec is being provided in case it helps:

``` ruby
require "spec_helper"

RSpec.describe Demo do
  subject(:part) { described_class.new value: commit, rendering: view.new.rendering }

  let :view do
    Class.new Hanami::View do
      config.paths = [Bundler.root.join("lib/demo/templates")]
      config.template = "n/a"
    end
  end

  let(:commit) { Struct.new(:label).new "Hi" }

  describe "#label" do
    it "answers label" do
      expect(part.label).to eq("<span>Hi</span>")
    end
  end
end
```
</details>

The error you'll get when running the above is:

``` 
 NoMethodError: undefined method `inflector' for #<Demo::Views::Context>...
```

**Expected behavior**

I would not expect to see this error as I think helpers are meant to be used within views and parts?

**Workaround**

There is a workaround to this issue and it has to do with overwriting the private `#tag_builder_inflector` method. Here's a modification of the above impelemtation which makes this work:

``` ruby
class Demo < Hanami::View::Part
  include Hanami::View::Helpers::TagHelper

  def label = tag.span value.label

  private

  def tag_builder_inflector = Dry::Inflector.new
end
```

I'm not sure if this breaks other aspects of Hanami but seems to work. The reason this works is because of these tag helper [lines](https://github.com/hanami/view/blob/main/lib/hanami/view/helpers/tag_helper.rb#L198C1-L200) which check for a context but the problem is that you're getting a `Hanami::View::Rendering` object back which has no `#inflector` message.

**Your environment**

- Affects my production application: **YES**
- Ruby version: ruby 3.2.2 (2023-03-30 revision e51014f9c0) +YJIT [arm64-darwin22.4.0]
- OS: macOS 13.4.1 (22F82)
- **Hanami View**: I'm using the latest SHA on the `main` branch.

timriley:
And just moved this issue into hanami/hanami, since that's where our view/default helpers integration resides.

bkuhlmann:
[...]

PR #1357, opened 2023-10-17
Provide default renderings to app/slice view parts

timriley:
In response to the feedback in #1355, make it possible to directly initialize a part (whose class is inside an app/slice namespace) with a value only, and have a default rendering provided from the respective app/slice, ensuring the slice has access to all the integrated view facilities from the app/slice, such as helpers.

This helps reduce boilerplate with unit testing parts. Now, this is possible:

```ruby
RSpec.describe(TestApp::Views::Parts::Post) do
  subject(:part) { described_class.new(value:)

  let(:value) { Struct.new(:title).new("Hello world") }

  it "works" do
    # This title_tag method depends on helpers being present inside the part,
    # which only comes courtesy of the app's integrated views
    expect(part.title_tag).to eq "<h1>Hello world</h1>"
  end
end
```

So streamlined!

### DIFF
--- a/lib/hanami/extensions/view/slice_configured_part.rb
+++ b/lib/hanami/extensions/view/slice_configured_part.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+module Hanami
+  module Extensions
+    module View
+      # Provides slice-specific configuration and behavior for any view part class defined within a
+      # slice's module namespace.
+      #
+      # @api private
+      # @since 2.1.0
+      class SliceConfiguredPart < Module
+        attr_reader :slice
+
+        # @api private
+        # @since 2.1.0
+        def initialize(slice)
+          super()
+          @slice = slice
+        end
+
+        # @api private
+        # @since 2.1.0
+        def extended(klass)
+          define_new
+        end
+
+        # @api private
+        # @since 2.1.0
+        def inspect
+          "#<#{self.class.name}[#{slice.name}]>"
+        end
+
+        private
+
+        # Defines a `.new` method on the part class that provides a default `rendering:` argument of
+        # a rendering coming from a view configured for the slice. This means that any part can be
+        # initialized standalone (with a `value:` only) and still have access to all the integrated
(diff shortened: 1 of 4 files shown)


timriley:
@bkuhlmann After our discussion earlier today, I couldn't help myself. Take a look at this and let me know if it'll solve your issues :)

lib/hanami/extensions/view/slice_configured_part.rb
```diff
@@ -0,0 +1,72 @@
...
+        def define_new
+          slice = self.slice
+
+          define_method(:new) do |**args|
+            return super(**args) if args.key?(:rendering)
+
+            slice_rendering = Class.new(Hanami::View)
+              .configure_for_slice(slice)
+              .new
+              .rendering
```
timriley, line 64: For the sake of efficiency, I originally had this rendering initialised _outside_ the `defined_method` block and then accessed within it via standard closure behaviour, but this turned out to cause some infinite loops since `Hanami::View.configure_for_slice` creates a part class, which would bring us here, which would. then call `Hanami::View.configure_for_slice` again, and so on.

Creating the rendering inside the method still feels reasonable enough given this code will only ever be exercised in testing. Parts created as part of ordinary view rendering will always have an explicit `rendering:` argument provided.

bkuhlmann:
Thanks! Yes, this works nicely and speeds up the convenience of testing and working with parts in general. I was able to confirm this by requiring the `provide-default-rendering-to-view-parts` branch locally and using:

``` ruby
# Implementation
class Task < Hanami::View::Part
  def assignee = helpers.tag.span user.name
end

# Specification
RSpec.describe Tasks::Views::Parts::Task do
  subject(:part) { described_class.new value: task }

  let(:task) { Test::Factory.structs[:task, id: 1, user:] }

  describe "#assignee" do
    it "answers user" do
      expect(part.assignee).to eq("<span>Jane Doe</span>")
    end
  end
 end
```

jodosha (approved):
@timriley Thanks for making Parts ergonomics so simple. 🙏

timriley:
Thanks both! 😄

PR #1359, opened 2023-10-24
Update app-level view context extensions to work in the same spirit as `Hanami::View::Context` itself

timriley:
In [earlier changes to hanami-view](https://github.com/hanami/view/pull/223) that will soon be released as 2.1.0, we removed the `Hanami::View::Context#with` method, instead expecting that a view context need only be initialized once, and then kept for the entire rendering.

However, at the time of that change, we did not make commensurate changes here in the app-level view context extension, instead just keeping a local `#with` method, and continuing to (needlessly) initialize a view contexet to pass to the action, only to _re-initialize it_ inside the action (via `#with`) with the request before rendering the view.

With this change, make the view app-level view context work in a way that matches the standard `Hanami::View::Context`: it should only be initialized once. To achieve this, inject a view context *class* into actions (instead of an instance) and then call `.new` on that context class when preparing the options for view rendering.

This simplifies the view context extension internals and allows us to remove a few tests, too. It will also make view contexts easier for users to modify, since they no longer have to think about the case of `#with` being called somewhere deep inside the framework, out of their sight. Instead, their context classes will be initialized just once, like any ordinary Ruby object. And the only expectation we have is that their `#initialize` accepts `**kwargs`.

### DIFF
--- a/lib/hanami/extensions/action.rb
+++ b/lib/hanami/extensions/action.rb
@@ -40,7 +40,7 @@ module Hanami
         attr_reader :view
 
         # @api private
-        attr_reader :view_context
+        attr_reader :view_context_class
 
         # Returns the app or slice's {Hanami::Slice::RoutesHelper RoutesHelper} for use within
         # action instance methods.
@@ -70,9 +70,9 @@ module Hanami
         #
         #   @api public
         #   @since 2.0.0
-        def initialize(view: nil, view_context: nil, rack_monitor: nil, routes: nil, **kwargs)
+        def initialize(view: nil, view_context_class: nil, rack_monitor: nil, routes: nil, **kwargs)
           @view = view
-          @view_context = view_context
+          @view_context_class = view_context_class
           @routes = routes
           @rack_monitor = rack_monitor
 
@@ -104,7 +104,7 @@ module Hanami
 
         # @api private
         def view_options(request, response)
-          {context: view_context&.with(**view_context_options(request, response))}.compact
+          {context: view_context_class&.new(**view_context_options(request, response))}.compact
         end
 
         # @api private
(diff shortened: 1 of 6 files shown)


waiting-for-dev (commented):
Thanks for taking care of all the details, @timriley!

Could you help me understand something? It looks like the action extension [is always providing the `context:` option when rendering the view](https://github.com/hanami/hanami/blob/314a069acf4ded43e70604624c70636160e52176/lib/hanami/extensions/action.rb#L107), so it looks like the [configuration of a default context](https://github.com/hanami/hanami/blob/314a069acf4ded43e70604624c70636160e52176/lib/hanami/extensions/view/slice_configured_view.rb#L117) is useless. Am I missing something? If I'm correct, should it make sense to modify `hanami-view` upstream so it's responsible for expecting a context class instead of an instance as a configuration?

jodosha (approved):

timriley:
@waiting-for-dev Thanks for checking this out :)

> It looks like the action extension [is always providing the context: option when rendering the view](https://github.com/hanami/hanami/blob/314a069acf4ded43e70604624c70636160e52176/lib/hanami/extensions/action.rb#L107), so it looks like the [configuration of a default context](https://github.com/hanami/hanami/blob/314a069acf4ded43e70604624c70636160e52176/lib/hanami/extensions/view/slice_configured_view.rb#L117) is useless.

In full stack Hanami app, for ordinary view rendering within actions, you're right, the default context is never used.

However, the default context is still useful if you're doing standalone view rendering outside of the context of an action, or if you're doing direct unit testing of views (where those views don't depend on anything from a `request` object), so it's still worth having properly configured.

> should it make sense to modify hanami-view upstream so it's responsible for expecting a context class instead of an instance as a configuration?

I don't think so. The `default_context` is used as-is when `#call`-ing a view and not providing a `context:` argument, so we really do already want that to be an instance. In addition, expecting the user to initialize it means that in cases where a view context _does_ needs arguments for it to be initialized properly, we give the user full control of that (remember: hanami-view is also a standalone view system that could be used in any number of ways!).

PR #1360, opened 2023-10-25
Defer Dry::System::Container freeze

alassek:
This small change defers the freeze of the container until after hooks have been run. This allows you to make alterations to the container in a `after_finalize` hook such as decorating keys by enumeration.

[Use-case: APM monitoring](https://discourse.hanamirb.org/t/newrelic-apm-on-hanami-2/872)

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -307,7 +307,7 @@ module Hanami
 
         prepare
 
-        container.finalize!
+        container.finalize!(freeze: false).freeze
         slices.each(&:boot)
 
         @booted = true


timriley:
Hi @alassek! Thanks for submitting this change :) 

I'd be happy to see something like this go in, but I'd love for this PR to include a test that exercises some functionality now achievable with this change that was otherwise not achievable before. This will be important to make sure we don't regress in future (which I could totally imagine happening with this code), and would also serve as a useful demonstration of what is now possible. 

Would you mind adding a small test like this?

Thanks!

alassek:
@timriley certainly. I wanted to get feedback on whether this approach was acceptable first, but I agree some regression testing is called for.

timriley:
Thanks @alassek :) 

I just checked out that forum post again, and it had more replies than I saw the last time. 

Based on that, it does feel like the addition of a new hook method to dry-system (or dry-container? I'm on my phone with very bad reception this weekend, so I trust you'll know which is right) feels like the better approach to take here. 

If we want finalize to continue to freeze by default (which feels right to me), and if we want to allow a user to be able to add their own behavior after all keys are populated (which, again, feels reasonable!) then it feels like a before_freeze might be a good hook to add?

This would mean that no change is needed in hanami, and we'd instead have a more directed change inside dry-system that will be helpful to everyone who uses it, not just Hanami users. 

What do you think?

timriley:
Another option here could be to update dry-system so that [its `:after_finalize` hook](https://github.com/dry-rb/dry-system/blob/5e49fcb93d0daca266946d98eaf3ede0a2edfe6c/lib/dry/system/container.rb#L333) simply runs one line earlier, before the `freeze`.

If we think about it, this feels like a far more useful place for the hook, since—as you've discovered—the container already being frozen means there are far fewer possibilities for helpful additional behaviour to be injected.

alassek:
@timriley Well, let's talk about something you brought up on the discourse

> From my perspective I think that running a prepared-only Hanami app (i.e. not running Hanami.boot) is a completely valid way of operating Hanami for certain circumstances (imagine e.g. a serverless function that needs to run just one narrow slice of a fuller app).

None of the solutions so far cover this use-case. What if we went the direction of an `after_register` hook that gets fired when a new key is registered? This would work in all cases, and it would allow things like matching keys by regexp.

I still think that `after_finalize` should probably happen before freeze, since it is far less useful otherwise. But I think `after_register` might be more in the spirit of not requiring finalization to decorate your keys.

timriley:
> What if we went the direction of an `after_register` hook that gets fired when a new key is registered? This would work in all cases, and it would allow things like matching keys by regexp.

I love it! Would definitely support something like this being added to dry-system.

> I still think that `after_finalize` should probably happen before freeze, since it is far less useful otherwise. But I think `after_register` might be more in the spirit of not requiring finalization to decorate your keys.

Yep, I'd love to see that to `after_finalize` in dry-system too.

@alassek, Would you be up for making these changes?

timriley:
Since I think these changes are now clearly targeting dry-system, I'm going to close this PR for the sake of tidiness. I'm happy for further conversation to still happen here, of course :)

alassek:
@timriley Yes, now that we have consensus on this I'll get started on a PR for dry-system. Thanks for the feedback!

Issue #1367, opened 2023-11-08
2.1 RC1 - bundler: command not found: hanami on Heroku deploy


olivierlacan:
Not sure whether this lies in hanami/cli or elsewhere but I'm opening this here for visibility.

No matter how I try, I can't get Hanami 2.1 RC1 to skip or ignore the `bundle exec hanami assets compile` task. 

## Context

I initially did not set any environment variables: no `HANAMI_ENV` or `SERVE_STATIC_ASSETS`. But I later tried these to no avail.

I didn't set the buildpack order either, Heroku is running with this:
> Detected buildpacks: Ruby,Node.js

My `Gemfile` does include `gem "hanami-cli", "2.1.0rc1"`, not just in the :development group.

## Error 

Here's the error I keep getting on deploy, for reference: 

```
remote: -----> Precompiling assets
remote:        Running: rake assets:precompile
remote:        rake aborted!
remote:        Hanami::CLI::HanamiExecError: `bundle exec hanami assets compile' failed
remote:
remote:
remote:        "bundler: command not found: hanami\nInstall missing gem executables with `bundle install`"
remote:        /tmp/build_f036228f/vendor/bundle/ruby/3.1.0/gems/hanami-cli-2.1.0.rc1/lib/hanami/cli/bundler.rb:94:in `block in hanami_exec'
remote:        <internal:kernel>:90:in `tap'
remote:        /tmp/build_f036228f/vendor/bundle/ruby/3.1.0/gems/hanami-cli-2.1.0.rc1/lib/hanami/cli/bundler.rb:93:in `hanami_exec'
remote:        /tmp/build_f036228f/vendor/bundle/ruby/3.1.0/gems/hanami-2.1.0.rc1/lib/hanami/rake_tasks.rb:57:in `run_hanami_command'
remote:        /tmp/build_f036228f/vendor/bundle/ruby/3.1.0/gems/hanami-2.1.0.rc1/lib/hanami/rake_tasks.rb:47:in `block (3 levels) in <top (required)>'
remote:        /tmp/build_f036228f/vendor/bundle/ruby/3.1.0/gems/rake-13.1.0/exe/rake:27:in `<top (required)>'
remote:        Tasks: TOP => assets:precompile
remote:        (See full trace by running task with --trace)
```

Here are my successive attempts: 
- Pushed a brand new Hanami 2.1 RC1 app to Heroku
- Repeat with `SERVE_STATIC_ASSETS` set to true, same error.
[...]

PR #1368, opened 2023-11-10
Avoid calling bundler exec in assets:precompile

timriley:
Instead of invoking `bundle exec hanami assets compile` as a shell command, instead call the `Hanami::CLI::Commands::App::Assets::Compile` CLI conmand object directly.

This still achieves the same goal, and it works around an issue we were seeing on Heroku where `bundle exec` could not find the `hanami` command when this rake task is called during the app's build phase.

Later on, once we've had time to properly debug the issue, we may want to return to the previous approach of shelling out to `bundle exec hanami`, this change gives us working Heroku deploys and asset compilation for 2.1.0 and buys us the time to do that further investigation without the pressure of us blocking new users wanting to deploy their apps.

Fixes https://github.com/hanami/hanami/issues/1367

### DIFF
--- a/lib/hanami/rake_tasks.rb
+++ b/lib/hanami/rake_tasks.rb
@@ -44,17 +44,9 @@ Hanami::CLI::RakeTasks.register_tasks do
   if Hanami.bundled?("hanami-assets")
     namespace :assets do
       task :precompile do
-        run_hanami_command("assets compile")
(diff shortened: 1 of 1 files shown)


jodosha (approved):
@timriley thanks for the investigation and the fix. 🙏

PR #1370, opened 2023-11-26
Pass through kwargs to middleware

pat:
Most middleware expect the old-school approach of hash options, but some are now using keyword arguments instead (such as Rack::Timeout). So let's presume that any keyword arguments that aren't known by Hanami should be passed through.

There is the edge-case of both the middleware and Hanami expecting arguments with the same name - this does not address that. 😓

The CI for this won't be green until https://github.com/hanami/router/pull/267 is merged in (but I've tested it locally with that branch, and is happy).

### DIFF
--- a/lib/hanami/slice/routing/middleware/stack.rb
+++ b/lib/hanami/slice/routing/middleware/stack.rb
@@ -91,16 +91,16 @@ module Hanami
           #
           # @api public
           # @since 2.0.0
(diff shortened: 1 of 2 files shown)


timriley:
Thanks for this, @pat! I think GitHub lost the initial test run for this, and a rebase was necessary, so I've just done this and force pushed, along with an expansion of your test so that it can ensure the keyword arg passing also works for `use` directly inside routes files.

PR #1399, opened 2024-05-25
Add #app? predicate to Hanami::Slice

### LINKED ISSUE #1397: Add `Slice.source_dir` method
This would help us clean up code like this:

```ruby
      def assets_dir?
        assets_path = app.eql?(self) ? root.join("app", "assets") : root.join("assets")
        assets_path.directory?
      end
```

We could then change it to:

```ruby
assets_path = root.join(source_dir, "assets")
```

The `source_dir` would be `nil` for slices, and `"app"` for assets.

Maybe this should actually just be a `source_path` that already joins onto the root:

```ruby
def source_path
  app? ? root.join("app") : root
end
```

(See #1398 for adding an `app?` method to slices).

Would would make the `assets_path` determination even simpler:

```ruby
assets_path = source_path.join("assets")
```

### LINKED ISSUE #1398: Add `Slice.app?` method
We have a bunch of checks like this throughout the hanami code:

```ruby
if slice.app.eql?(slice)
```

This is ungainly and not very intention revealing.

It would be better if this was just:

```
if slice.app?
```

alassek:
Closes #1398
Closes #1397

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -213,6 +213,16 @@ module Hanami
         config.root || app.root.join(SLICES_DIR, slice_name.to_s)
       end
 
+      # Return's the slice's root component directory, accounting for App as a special case.
+      #
+      # @return [Pathname]
+      #
+      # @api public
+      # @since 2.2.0
+      def source_path
+        app? ? root.join("app") : root
+      end
+
       # Returns the slice's configured inflector.
       #
       # Unless explicitly re-configured for the slice, this will be the app's inflector.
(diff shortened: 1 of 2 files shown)


lib/hanami/slice.rb
```diff
@@ -1057,18 +1077,15 @@ def render_detailed_errors?
...
       def db_dir?
-        db_path = app.eql?(self) ? root.join("app", "db") : root.join("db")
-        db_path.directory?
+        source_path.join("db").directory?
       end
 
       def relations_dir?
-        relations_path = app.eql?(self) ? root.join("app", "relations") : root.join("relations")
-        relations_path.directory?
+        source_path.join("relations").directory?
```
timriley, line 1088: These are so much clearer now 😍

timriley (approved):

timriley:
Thanks @alassek! FYI, I made a couple of commits to fix a typo and use the APP_DIR constant. Will merge this in now!

PR #1403, opened 2024-05-27
Share parent DB via `config.db.import_from_parent`

### LINKED ISSUE #1394: Multiple slices sharing app-level ROM setup
See https://discourse.hanamirb.org/t/integrating-rom-into-hanami-2-2/971#slices-may-alternatively-share-the-app-level-rom-setup-and-relations-etc-13

timriley:
This new setting defaults to false. When set to true, instead of initializing an independent ROM setup per slice, the app’s ROM setup will be imported directly into each slice, allowing the app’s relations (and other ROM components) to be shared across all slices.

Resolves #1394.

### DIFF
--- a/lib/hanami/config.rb
+++ b/lib/hanami/config.rb
@@ -216,6 +216,20 @@ module Hanami
     # @since 2.0.0
     attr_reader :actions
 
+    # Returns the app's db config, or a null config if hanami-db is not bundled.
+    #
+    # @example When hanami-db is bundled
+    #   config.db.import_from_parent # => false
+    #
+    # @example When hanami-db is not bundle
+    #   config.db.import_from_parent # => NoMethodError
+    #
+    # @return [Hanami::Config::DB, Hanami::Config::NullConfig]
+    #
+    # @api public
+    # @since 2.2.0
+    attr_reader :db
+
     # Returns the app's middleware stack, or nil if hanami-router is not bundled.
     #
     # Use this to configure middleware that should apply to all routes.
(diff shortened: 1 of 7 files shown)


lib/hanami/providers/db.rb
```diff
@@ -19,6 +19,8 @@ class DB < Dry::System::Provider::Source
 
       # @api private
       def prepare
+        prepare_and_import_parent_db and return if import_from_parent?
```
timriley, line 22: With a few steps to go, it's clear the code this DB provider is going to get somewhat complex.

Since this PR effectively introduces the "second mode" of DB setup (sharing with the app, instead of slice independence), I felt this guard clause + helper method + early return was the most reasonable way to do it.

This keeps the "import"-related functionality together lower in the file, and allows the rest of the DB setup (which is important, because it still has to run at least once) to remain inside the main body of each of the provider lifecycle methods.

lib/hanami/config.rb
```diff
@@ -288,16 +302,27 @@ def initialize(app_name:, env:)
...
+        Hanami::Config::Assets.new
+      }
+
+      @db = load_dependent_config("hanami-db") { DB.new }
+
+      @logger = Config::Logger.new(env: env, app_name: app_name)
+
+      @middleware = load_dependent_config("hanami-router") {
+        Slice::Routing::Middleware::Stack.new
+      }
```
timriley, line 322: Some extra diff around here since I ordered these config objects consistently at the same time.

spec/unit/hanami/config/router_spec.rb
```diff
@@ -29,7 +29,7 @@
   context "hanami-router is not bundled" do
     before do
       allow(Hanami).to receive(:bundled?).and_call_original
-      expect(Hanami).to receive(:bundled?).with("hanami-router").and_return(false)
+      allow(Hanami).to receive(:bundled?).with("hanami-router").and_return(false)
```
timriley, line 32: This change was needed because the `expect` led to the `false` only being returned for the first call, but now I've tidied up the initialization in `Config` (so that `@middleware` and `@router` are assigned separately), we needed this `bundled?` check to be called twice.

PR #1410, opened 2024-06-01
Integrate repos

timriley:
Repos automatically use the ROM container that exists for their slice.

They also infer a root relation from their name. For example, PostRepo will infer a root of `:posts`.

### DIFF
--- a/lib/hanami/extensions/db/repo.rb
+++ b/lib/hanami/extensions/db/repo.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: true
+
+require "hanami/db"
+
+module Hanami
+  module Extensions
+    # @api private
+    # @since 2.2.0
+    module DB
+      # @api private
+      # @since 2.2.0
+      module Repo
+        def self.included(repo_class)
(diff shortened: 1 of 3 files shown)


lib/hanami/extensions/db/repo.rb
```diff
@@ -0,0 +1,89 @@
...
+require "hanami/db"
+
+module Hanami
+  module Extensions
+    # @api private
+    # @since 2.2.0
+    module DB
+      # @api private
+      # @since 2.2.0
+      module Repo
```
timriley, line 12: All of this _could_ live in the hanami-db gem, since unlike our other gems, hanami-db is not intended to provide any kind of standalone experience for users.

However, I opted to provide an extensions module for the app integration aspects of `Hanami::DB::Repo` because it means we can keep _all_ such code together in the same place. I think there's value in that locality for us as maintainers.

Later on, once we provide a proper first-party API for framework extensions, we can move this code (along with the code for extending views, actions, etc.) back into their respective gems.

Issue #1395, opened 2024-05-24
Automatically adjust database URL in test mode


timriley:
See details at top of https://discourse.hanamirb.org/t/integrating-rom-into-hanami-2-2/971/2

alassek:
@timriley I am running into a lot of trouble figuring out how to test this.

The unit tests were easy enough, but integration testing this is going to require adding all the supported databases to the test matrix. Is this something that you were planning on doing?

Running `#prepare` on the DB provider with a connection URL is going to require actually opening the connection when `ROM.setup` is called.

timriley:
@alassek I don't think we should worry about testing all databases in CI. The unit tests over in hanami-db give us confidence that we cover the different URL schemes for the different databases we care about.

To test this in hanami/hanami, I think we just want one integration-style test that makes sure the framework is calling the database naming logic over in hanami-db. I think a test for a single SQLite-using app would be enough for now, just like we're doing for our other tests.

I do think we'll eventually want to add real working database servers into a test matrix (Before or after 2.2? I think this is open for debate!). In the meantime, if we can defer this for just a little while longer, it'll allow us to more quickly move through the rest of our work on establishing the database layer in the first place.

timriley:
So while I think we can get away with just SQLite for testing this feature in hanami/hanami, I think I spoke too soon for our efforts more generally: I've just hit this same issue looking to test `db structure dump` in hanami/cli. And in that case I think it's important our tests actually exercise a real Postgres database. So I'm going to investigate getting a proper test setup over there.

(In the future I'd really like to merge hanami/cli and hanami/hanami somehow. It would help us avoid a lot of repetition in our testing, not to mention make it easier for the CLI and the core framework to evolve more gracefully together).

PR #1415, opened 2024-06-05
Apply Hanami::DB::Testing.database_url

alassek:
This integrates test database renaming when HANAMI_ENV=test

Closes hanami/hanami#1395

As per [our discussion](https://github.com/orgs/hanami/projects/6/views/1?pane=issue&itemId=64371170), I am relying on the unit testing in hanami-db. Integration testing this won't be possible until there is a testing database running in CI.

### DIFF
--- a/lib/hanami/providers/db.rb
+++ b/lib/hanami/providers/db.rb
@@ -92,8 +92,10 @@ module Hanami
 
         # For "main" slice, expect MAIN__DATABASE_URL
         slice_url_var = "#{target.slice_name.name.gsub("/", "__").upcase}__DATABASE_URL"
+        chosen_url = config.database_url || ENV[slice_url_var] || ENV["DATABASE_URL"]
+        chosen_url &&= Hanami::DB::Testing.database_url(chosen_url) if Hanami.env?(:test)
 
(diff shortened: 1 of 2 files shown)


timriley (approved):
The change here looks good, thanks @alassek!

To clarify one thing from my earlier comment: I think it would be good if we could add one full integration test here for the database URL rewriting feature, just using SQLite, since we're already able to test using that.

Would you mind adding that before merging? Happy for it to go into a new test file that describes this one specific feature.

Once you've done that, I'm happy to see this merge :)

alassek:
> To clarify one thing from my earlier comment: I think it would be good if we could add one full integration test here for the database URL rewriting feature, just using SQLite, since we're already able to test using that.
>
> Would you mind adding that before merging? Happy for it to go into a new test file that describes this one specific feature.

I am unfamiliar with how GIthub Actions work. If the runtime environment is writeable I should be able to test SQLite. If not it should be possible to mount a scratch volume or something.

timriley (approved):
Nice work figuring out the testing side, @alassek 😄 Happy to see this go in! It's a great feature, thanks for proposing it.

PR #1416, opened 2024-06-05
Replace ROM::Inflector in DB Provider

### LINKED ISSUE #1400: Set ROM inflector be the same as the Hanami app inflector
In the persistence provider for my example app decafsucks, [we need to do the following](https://github.com/decafsucks/decafsucks/blob/5923add8c29f23e4259e7d43d69de3b57b12c737/config/providers/persistence.rb#L9-L10):

```ruby
    # TODO(Hanami): As part of built-in rom setup, configure ROM with app inflector
    silence_warnings { ROM::Inflector = Hanami.app["inflector"] }
```

ROM 5.x doesn't support a configurable inflector.

We're either going to need to hardcode a hack like this, or update ROM to allow this to be configurable, and then properly pass in the inflector.

alassek:
Inconsistent inflections between ROM and Hanami will lead to constant loading errors. Since ROM 5.3 does not include a configuration setting for inflection, we have to swap out the constant manually.

Closes #1400

I chose the `remove_const` approach for this. The other common solution is to set `$VERBOSE` to false momentarily, but this felt like a more surgical approach.

### DIFF
--- a/lib/hanami/providers/db.rb
+++ b/lib/hanami/providers/db.rb
@@ -24,6 +24,7 @@ module Hanami
         prepare_and_import_parent_db and return if import_from_parent?
 
         apply_parent_provider_config
+        override_rom_inflector
 
         require "hanami-db"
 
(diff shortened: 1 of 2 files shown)


timriley (approved):
Looks great! And I agree that the remove_const approach feels nicer than the approach I hinted at earlier ❤️

PR #1430, opened 2024-06-18
Add DB provider adapter config & sensible defaults

### LINKED ISSUE #1389: Expand configurability of db provider
Can we move rom plugins into config, so the user doesn't have to do this?

```ruby
          after(:prepare) do
            @rom_config.plugin(:sql, relations: :auto_restrictions)
          end
        end
```

Are there other aspects of ROM config/setup that could be moved into config?

The more that goes into config, the more power we get in sharing this across slices, thanks to #1408.

### LINKED ISSUE #1409: Expand default set of Sequel extensions

timriley:
_Note to reviewers: I'm 99% confident this is a good approach and should work well for us. I'm interested in your perspectives on the config API here. Have I missed anything?_

Add adapter config and sensible adapter defaults to the DB provider.

Now you can configure your adapters like so:

```ruby
# config/providers/db.rb

  config.adapter :sql do |a|
    # Add a Sequel extension (available in :sql adapter only)
    a.extension :some_extension

    # Add a ROM plugin
    a.plugin relations: :some_plugin
  end
end
```

The following defaults are provided for the :sql adapter:

```ruby
plugin relations: :instrumentation do |plugin|
  plugin.notifications = target["notifications"]
end

plugin relations: :auto_restrictions

extension(
  :caller_logging,
  :error_sql,
  :sql_comments
)
```

And in addition for Postgres databases:

```ruby
extension(
  :pg_array,
  :pg_enum,
  :pg_json,
  :pg_range
)
```

These defaults are set as part of the db provider's `prepare` lifecycle step. This is necessary (instead of doing it earlier, i.e. at the stage of establishing the setting defaults) to allow the user to configure and set their `database_url`, which is required to determine the set of defaults to provide.

When configuring the db provider, the defaults may be skipped by the user:

```ruby
  config.adapter :sql do |a|
    # Skip all defaults
    a.skip_defaults

    # Or plugins only
	a.skip_defaults :plugins

    # Or extensions only
	a.skip_defaults :extensions
  end
end
```

If you want to configure the db provider to use a non-sql adapter, then you can configure it independently:

```ruby
  config.adapter = :yaml

  config.adapter :yaml do |a|
    # configure plugins here
    # (extensions not available since these are sql-only)
  end
end
```

You can also configure ROM plugins for _any_ adapter via `any_adapter`:

```ruby
  config.any_adapter do |a|
    # configure plugins here
  end
end
```

This may be useful in an advanced setup where your app uses different ROM adapters, but you want certain plugins to be used consistently across all of them.

For db providers in slices, when using the default `app.config.db.configure_from_parent` arrangement (`true`), adapter config is applied from the parent slice just like any other config.

However, if the configures their own `:sql` adapter in a slice provider (either plugins or extensions), then the respective plugins or extensions from the parent will _not_ be applied. We might be able to provide more sophisticated handling here, but this felt like a reasonable first approach. In a Hanami app with multiple slices but a single shared database, I think it would be unlikely that different ROM plugins or Sequel extensions would be needed on a slice-by-slice basis anyway.

This change depends on https://github.com/dry-rb/dry-configurable/pull/164 to allow for the custom config class we use for the db provider. I'm going to make sure that PR merges first before this one, but I wanted to open this one up for review first.

Resolves #1389, resolves #1409

### DIFF
--- a/lib/hanami/providers/db.rb
+++ b/lib/hanami/providers/db.rb
@@ -132,7 +132,6 @@ module Hanami
 
         parent_db_provider.source.config.adapters.each do |adapter_name, parent_adapter|
           adapter = config.adapters[adapter_name]
-          # binding.irb if target == Main::Slice
 
           adapter.class.settings.keys.each do |key|
             next if adapter.config.configured?(key)
--- a/lib/hanami/providers/db.rb
+++ b/lib/hanami/providers/db.rb
@@ -17,15 +17,24 @@ module Hanami
       setting :adapters, mutable: true, default: Adapters.new
       setting :relations_path, default: "relations"
 
-      # @api private
+      def initialize(...)
+        super(...)
+
+        @configured_for_database = false
+      end
+
+      def finalize_config
+        apply_parent_config and return if apply_parent_config?
+
+        configure_for_database
+      end
+
       def prepare
         prepare_and_import_parent_db and return if import_from_parent?
 
         override_rom_inflector
 
-        apply_parent_config
-        configure_for_database
-        # mark as configured?, i.e. freeze config? might be a good idea
+        finalize_config
 
         require "hanami-db"
 
@@ -46,7 +55,9 @@ module Hanami
 
         config.each_plugin do |plugin_spec, config_block|
           if config_block
-            @rom_config.plugin(config.adapter, plugin_spec, &config_block)
+            @rom_config.plugin(config.adapter, plugin_spec) do |plugin_config|
+              instance_exec(plugin_config, &config_block)
+            end
           else
             @rom_config.plugin(config.adapter, plugin_spec)
           end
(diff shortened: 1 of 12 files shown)


lib/hanami/providers/db.rb
```diff
@@ -9,23 +10,31 @@ module Providers
...
-      # TODO: Switch extensions based on configured adapter
-      setting :extensions, default: [:error_sql]
+      def initialize(...)
+        super(...)
 
-      setting :relations_path, default: "relations"
+        @configured_for_database = false
+      end
+
+      def finalize_config
```
timriley, line 26: This is pulled out into a public method on the provider source so we can call it on the parent provider in `#apply_parent_config`. Doing it this way is important to avoid unnecessary connections to the database that would otherwise occurred if we called a straight-up `#prepare`.

alassek (approved):
This looks great ❤️ 

I think this covers any customizations I would need to do.

PR #1104, opened 2021-05-04
Remove phantomjs

ekremkaraca:
Hi there,

Since #1090 was merged, no need to have phantomjs binary placed here.

PR #1443, opened 2024-07-25
Remove PhantomJS binary

cllns:
Like #1104 for but for our current (v2) codebase.


timriley:
Nice find! Thank you!

PR #1444, opened 2024-07-26
Don't disable prompt for select

cllns:
Disabling the prompt makes it not show as the selected option when it's displayed in the DOM. Instead, the first (non-prompt) option is shown as selected. 

One _could_ mark the prompt as both selected and disabled but then if people pick one of the options, there's no way to go back and make it empty, which is a bad UX anti-pattern.


timriley:
Yep, this makes sense, thanks @cllns!

PR #1445, opened 2024-07-26
MissingActionError relative path instead of absolute path

kyleplump:
Fixes issue [#1438](https://github.com/hanami/hanami/issues/1438)

### DIFF
--- a/lib/hanami/routes.rb
+++ b/lib/hanami/routes.rb
@@ -38,7 +38,7 @@ module Hanami
         action_constant = slice.inflector.camelize(
           "#{slice.inflector.underscore(slice.namespace.to_s)}#{PATH_DELIMITER}#{action_path}"
         )
-        action_file = slice.root.join("#{action_path}#{RB_EXT}")
+        action_file = "#{action_path}#{RB_EXT}"
 
         super(<<~MSG)
           Could not find action with key #{action_key.inspect} in #{slice}


cllns:
Thanks for this! Do you mind adding specs for this, for both `app/` and within a slice? 

Right now, from within a slice, this change will output:
 `To fix this, define the action class Foo::Actions::Bar::Index in actions/bar/index.rb` 

e.g.:

```ruby
slice :foo, at: "/foo" do
  get "/bar", to: "bar.index"
end
```

When it should say:
 `To fix this, define the action class Foo::Actions::Bar::Index in slices/foo/actions/bar/index.rb`.

In other places we use `Pathname#relative_path_from` with `slice.root` as the arg. That won't as-is here because we're working with a string, but it should be do-able. Let me know if you need any more pointers!

kyleplump:
Thanks for the pointers @cllns! (still new to Hanami, and learning 😅 ).  I made the change you suggested using `Pathname`. 

On the spec front: I noticed there was already an integration test that covered this?  Although it was using regex to wildcard match any character before the relative path, I changed the test to more explicitly check  that the error message contains _only_ the relative path.  Let me know if this works

cllns:
Oh nice! Great work. 

LGTM, I'll let @timriley weigh in too.

I wonder if we should add a `Hanami::Slice#relative_root` that does this for us, it seems like a useful utility method.

kyleplump:
Thanks! Lmk I'd be happy to add that utility method in if you guys think it would be useful

PR #1446, opened 2024-07-29
Introduce Hanami::Provider::Source

### LINKED ISSUE #1417: Inside providers, add a `slice` alias for `target`

alassek:
`Hanami::ProviderRegistrar` uses a custom superclass, `Hanami::Provider::Source`, for user-defined sources. This provides a `slice` method within the source.

Closes #1417

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem "hanami-webconsole", github: "hanami/webconsole", branch: "main"
 
 gem "hanami-devtools", github: "hanami/devtools", branch: "main"
 
-gem "dry-system", github: "alassek/dry-system", branch: 'provider-class-override'
+gem "dry-system", github: "dry-rb/dry-system", branch: "main"
 gem "dry-inflector", github: "dry-rb/dry-inflector", branch: "main"
 
 # This is needed for settings specs to pass
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,6 @@ gem "hanami-webconsole", github: "hanami/webconsole", branch: "main"
 gem "hanami-devtools", github: "hanami/devtools", branch: "main"
 
 gem "dry-system", github: "dry-rb/dry-system", branch: "main"
-gem "dry-inflector", github: "dry-rb/dry-inflector", branch: "main"
 
 # This is needed for settings specs to pass
 gem "dry-types"
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,9 @@ gem "hanami-webconsole", github: "hanami/webconsole", branch: "main"
 
 gem "hanami-devtools", github: "hanami/devtools", branch: "main"
 
+gem "dry-system", github: "dry-rb/dry-system", branch: "provider-sources-2"
+gem "dry-inflector", github: "dry-rb/dry-inflector", branch: "main"
+
 # This is needed for settings specs to pass
 gem "dry-types"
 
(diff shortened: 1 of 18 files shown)


Gemfile
```diff
@@ -20,6 +20,9 @@ gem "hanami-webconsole", github: "hanami/webconsole", branch: "main"
 
 gem "hanami-devtools", github: "hanami/devtools", branch: "main"
 
+gem "dry-system", github: "alassek/dry-system", branch: 'provider-class-override'
```
alassek, line 23: change this back once dry-rb/dry-system#275 is merged

timriley: 👍🏼 — let's get https://github.com/dry-rb/dry-system/pull/275 merged first, so we can switch this to `branch: "main"` before merging.

Gemfile
```diff
@@ -20,6 +20,9 @@ gem "hanami-webconsole", github: "hanami/webconsole", branch: "main"
...
+gem "dry-inflector", github: "dry-rb/dry-inflector", branch: "main"
```
timriley, line 24: Is this dry-inflector line strictly needed for this PR?

timriley (approved):
@alassek Thanks for getting this ready to go!

I hope you don't mind, but I just pushed one commit onto this branch: aside from a bit of tidying up, what it does is centralize all our provider customization to a single place: this newly introduced `Hanami::Provider::Source` class.

Previously, this was split across two places: the `target_container`->`slice` aliasing in the registrar, and the extra `slice` reader in the source class. Having these both live together feels much nicer, and would help people more quickly understand our custom provider experience by just looking in one place.

If you're happy with this, then I think this is ready to go in as soon as https://github.com/dry-rb/dry-system/pull/275 is merged :)

lib/hanami/provider/source.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Hanami
+  module Provider
+    class Source < Dry::System::Provider::Source
+      attr_reader :slice
+
+      alias_method :target_container, :slice
+      alias_method :target, :slice
```
timriley, line 9: I don't love that I had to add both of these lines here.

But I think this can be temporary.

If we update `Dry::System::Provider::Source`, and replace `alias_method :target, :target_container` with `def target = target_container`, then I think we can reduce these lines here to just:

```ruby
def target_container = slice
```

Which I think will be a good final state :)

alassek: Well this appears to have totally blown up the suite so I will need to investigate further

timriley: @alassek Where have things blown up? This PR looks green to me.

timriley: @alassek after my extra adjustment to https://github.com/dry-rb/dry-system/pull/275, I've now pushed another commit to this branch with just the single `def target_container = slice` override, and now everything is looking good 👍🏼

timriley:
@alassek Are you happy with this PR (and https://github.com/dry-rb/dry-system/pull/275) in their current states?

alassek:
@timriley Yes, I think this is ready to go once dry-rb/dry-system#275 merges

Gemfile
timriley: Have adjusted this back to dry-rb/dry-system main 👍🏼

Gemfile
timriley: I removed the dry-inflector line here and everything still passes, so I've pushed that change up too.

timriley:
@alassek Thanks also for your work on this one! 😄

Issue #1442, opened 2024-07-16
Logger stream in development defaults to $stdout, when it should be log/development.log


timriley:
This is especially noticeable now we can do SQL queries from the console, e.g.

```
prerelease22[development]> repo = app["repos.post_repo"]
=> #<Prerelease22::Repos::PostRepo struct_namespace=Prerelease22::Structs auto_struct=true>
prerelease22[development]> repo.create(title: "hello world")
[prerelease22] [INFO] [2024-07-16 20:30:37 +1000]   Loaded :sqlite in 1ms INSERT INTO `posts` (`title`) VALUES ('hello world')
[prerelease22] [INFO] [2024-07-16 20:30:37 +1000]   Loaded :sqlite in 0ms SELECT `posts`.`id`, `posts`.`title` FROM `posts` WHERE (`id` IN (1)) ORDER BY `posts`.`id`
=> #<Prerelease22::Structs::Post id=1 title="hello world">
prerelease22[development]> repo.all
[prerelease22] [INFO] [2024-07-16 20:42:47 +1000]   Loaded :sqlite in 0ms SELECT `posts`.`id`, `posts`.`title` FROM `posts` ORDER BY `posts`.`id`
=> [#<Prerelease22::Structs::Post id=1 title="hello world">]
```

timriley:
I realised that logging to stdout is needed for our `hanami console` and `hanami dev` outputs to look right, so we do indeed want to keep stdout logging in development mode.

Maybe in the future we can think about adjusting the formatting of the logger when we know its initiated from _a console in development mode_, but for now I think we can just leave things as they are.

PR #1447, opened 2024-07-30
default to development.log instead of stdout

kyleplump:
closes [#1442](https://github.com/hanami/hanami/issues/1442)

### DIFF
--- a/lib/hanami/config/logger.rb
+++ b/lib/hanami/config/logger.rb
@@ -124,7 +124,7 @@ module Hanami
         case env
         when :development, :test
           config.level = :debug
-          config.stream = File.join("log", "#{env}.log") if env == :test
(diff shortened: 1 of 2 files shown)


timriley:
Thanks for taking the inititative on this one, @kyleplump! However, I think I was a little too quick to file #1442. I realised that we needed the stdout logging for the hanami server to show its logs alongside the other tools when running `hanami dev`, and in addition, it turns out that logging to stdout from the console is also standard for other tools like Rails too, so I think we can just keep this for now, and maybe consider a console-only quietened mode in the future.

I really appreciate the help, though! Sorry that this particular PR won't go anywhere.

kyleplump:
no problem @timriley! you're reasoning makes sense - happy to help regardless

PR #1450, opened 2024-08-09
Add connection_options to Hanami::Providers::DB::Gateway

### LINKED ISSUE #1440: Allow custom connection options to be configured in :db provider

alassek:
Closes #1440

This adds a `connection_options` to gateway configuration that is passed through to `ROM::Gateway.setup`, so that various options specific to your datastore may be set without encoding them into the DATABASE_URL itself.

```ruby
  config.gateway :default do |gw|
    gw.connection_options timeout: 10_000, readonly: true
  end
end
```

You can see an overview of these settings in [the Sequel documentation](https://sequel.jeremyevans.net/rdoc/files/doc/opening_databases_rdoc.html)

### DIFF
--- a/lib/hanami/providers/db/gateway.rb
+++ b/lib/hanami/providers/db/gateway.rb
@@ -15,6 +15,7 @@ module Hanami
         setting :database_url
         setting :adapter_name, default: :sql
         setting :adapter, mutable: true
+        setting :connection_options, default: {}
 
         # @api public
         # @since 2.2.0
@@ -34,6 +35,22 @@ module Hanami
           end
         end
 
+        # @api public
+        # @since 2.2.0
+        def connection_options(**options)
+          if options.any?
+            config.connection_options.merge!(options)
+          end
+
+          config.connection_options
+        end
+
+        # @api public
+        # @since 2.2.0
+        def options
+          {**connection_options, **adapter.gateway_options}
+        end
+
         # @api private
         def configure_adapter(default_adapters)
           default_adapter = default_adapters[config.adapter_name]
@@ -48,7 +65,7 @@ module Hanami
 
         # @api private
         def cache_keys
-          [config.database_url, config.adapter.gateway_cache_keys]
+          [config.database_url, config.connection_options, config.adapter.gateway_cache_keys]
         end
 
         private
--- a/lib/hanami/providers/db/sql_adapter.rb
+++ b/lib/hanami/providers/db/sql_adapter.rb
@@ -85,7 +85,7 @@ module Hanami
 
         # @api private
         def gateway_options
-          {extensions: config.extensions}
+          {extensions: extensions}
         end
 
         # @api public
--- a/lib/hanami/providers/db/sql_adapter.rb
+++ b/lib/hanami/providers/db/sql_adapter.rb
@@ -73,7 +73,7 @@ module Hanami
           )
 
           # Extensions for specific databases
-          if database_url.to_s.start_with?(%r{postgres(ql)*://}) # FIXME: what is the canonical postgres URL?
+          if database_url.to_s.start_with?(%r{postgres(ql)*://})
             extension(
               :pg_array,
               :pg_enum,
--- a/lib/hanami/providers/db.rb
+++ b/lib/hanami/providers/db.rb
@@ -207,7 +207,7 @@ module Hanami
             ROM::Gateway.setup(
               gw_config.adapter_name,
               gw_config.database_url,
-              **gw_config.adapter.gateway_options
+              **gw_config.options
             )
           }
         }
--- a/spec/integration/db/gateways_spec.rb
+++ b/spec/integration/db/gateways_spec.rb
@@ -81,7 +81,48 @@ RSpec.describe "DB / Gateways", :app_integration do
     end
   end
 
+  it "configures connection options on their respective gateways" do
+    with_tmp_directory(@dir = Dir.mktmpdir) do
+      write "config/app.rb", <<~RUBY
+        require "hanami"
 
+        module TestApp
+          class App < Hanami::App
+          end
+        end
+      RUBY
+
+      write "db/.keep", ""
+
+      write "config/providers/db.rb", <<~RUBY
+        Hanami.app.configure_provider :db do
+          config.gateway :default do |gw|
+            gw.connection_options timeout: 10_000
+
+            gw.adapter :sql do |a|
+              a.skip_defaults :extensions
+              a.extension :error_sql
+            end
+          end
+
+          config.gateway :extra do |gw|
+            gw.connection_options readonly: true
+          end
+        end
+      RUBY
+
+      ENV["DATABASE_URL"] = "sqlite::memory"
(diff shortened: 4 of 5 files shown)


lib/hanami/providers/db/sql_adapter.rb
```diff
@@ -6,10 +6,27 @@ class DB < Hanami::Provider::Source
...
+        # @api public
+        # @since 2.2.0
+        def connection_options
+          config.connection_options ||= {}
+        end
+
+        # @api public
+        # @since 2.2.0
+        def options(**options)
+          options.delete(:extensions)
```
alassek, line 26: There are two ways to go here, and I believe it's better for long-term maintainability to not allow `options` to be a back-door for `extensions` since that has a definite public api.

timriley (commented):
Thanks for keeping things moving here, @alassek, I really appreciate it! 😍😍

I have a couple of suggestions here for you. Keen for your thoughts. Once these are addressed I think this will be good to merge.

lib/hanami/providers/db/sql_adapter.rb
```diff
@@ -6,10 +6,27 @@ class DB < Hanami::Provider::Source
...
 
+        # @api public
+        # @since 2.2.0
+        def connection_options
+          config.connection_options ||= {}
+        end
+
+        # @api public
+        # @since 2.2.0
+        def options(**options)
```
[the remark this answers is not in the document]
wuarmin: I always read along eagerly. Danke, Tim, dass du auch auf diese Details achtest. Ich glaube nämlich, dass einen Unterschied machen kann. 

> What do you think about the idea of having the #connection_options method also being the method that can merge in user-supplied options?
Makes a lot of sense to me in this case.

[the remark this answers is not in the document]
alassek: My motivation for naming it `options` stems from `connection_options` being kind of long, and inconvenient to type. But for the sake of clarity I agree that we should probably be consistent. This isn't an API that will be used a lot, so brevity is less important.

lib/hanami/providers/db/sql_adapter.rb
```diff
@@ -8,20 +8,34 @@ class DB < Hanami::Provider::Source
...
         # @api public
         # @since 2.2.0
-        setting :extensions, mutable: true
+        setting :connection_options, default: {}
 
         # @api public
         # @since 2.2.0
-        def extension(*extensions)
-          self.extensions.concat(extensions)
+        setting :extensions, default: []
```
alassek, line 15: @timriley Thinking about your feedback:

> Having `mutable: true` on its own with no default value felt kind of like a smell to me too, since the `mutable:` flag tells the reader that something special is happening with that setting, but the lack of the corresponding default value means that the reader has to go look elsewhere to discover the full story for that setting, which is not ideal.

This applies equally to how `extensions` works, and this made me discover that `{extensions: nil}` is a possible result if the helper method is not called.

I took the initiative to clean up this interface in the same way, so that now we always return an array for this.

timriley: @alassek Thanks for bringing this one up! We should look to find a way to make the `extensions` setting work the same, or if it needs to exist like this for a reason, leave a comment above it explaining why. Let's do this in a follow up PR.

In the meantime, I think the test case you pointed out is a good one to consider for this new `connection_options` setting too.

Is it possible to have a set of connection settings configured in an app provider and then have those settings changed (or cleared) in a slice that is inheriting the app config?

alassek: @timriley excellent point, I did not test configuration hierarchy. I'll add some tests around that.

wuarmin:
@alassek

lib/hanami/providers/db/sql_adapter.rb
timriley: @alassek Are you still happy to add these tests?

timriley:
Now that #1452 is merged, we'll need to move the `connection_options` setting into `Hanami::Providers::DB::Gateway` (where it can live alongside the respective `database_url` setting).

lib/hanami/providers/db/gateway.rb
```diff
@@ -34,6 +35,22 @@ def adapter(name = Undefined)
...
+          end
+
+          config.connection_options
+        end
+
+        # @api public
+        # @since 2.2.0
+        def options
+          {**connection_options, **adapter.gateway_options}
+        end
```
alassek, line 52: Since these concerns have now been split into two places, I opted to create an `option` method on Gateway to combine them, otherwise I would have to completely duplicate the setting in Adapter and that didn't feel right.

lib/hanami/providers/db/sql_adapter.rb
```diff
@@ -73,7 +73,7 @@ def configure_for_database(database_url)
           )
 
           # Extensions for specific databases
-          if database_url.to_s.start_with?(%r{postgres(ql)*://}) # FIXME: what is the canonical postgres URL?
+          if database_url.to_s.start_with?(%r{postgres(ql)*://})
```
alassek, line 76: Note that while the [PostgreSQL documentation](https://www.postgresql.org/docs/current/app-psql.html#id-1.9.4.21.8) uses `postgresql://`, [Heroku uses `postgres://`](https://devcenter.heroku.com/articles/connecting-heroku-postgres#external-connections-ingress) and in my testing `psql` supports both.

A codesearch through GitHub reveals that both are quite commonplace, so I recommend that we continue to support both ways. Annoying, but it is what it is.

lib/hanami/providers/db/sql_adapter.rb
```diff
@@ -85,7 +85,7 @@ def configure_for_database(database_url)
 
         # @api private
         def gateway_options
-          {extensions: config.extensions}
+          {extensions: extensions}
```
alassek, line 88: This eliminates an edge-case where you would end up with `{extensions: nil}` if you didn't set anything. Now the type signature of this value should always be consistent.

timriley:
This is looking great, thanks @alassek!

In relation to [the comment I left earlier](https://github.com/hanami/hanami/pull/1450#discussion_r1721118744) about parent config copying to child slices, now that the connection options are in the gateway, this has changed the overall behaviour: since [gateways are not copied to child slices](https://github.com/hanami/hanami/blob/a270767ef77d2d5aa382d84163de85367d042215/lib/hanami/providers/db.rb#L116-L117), the connection options will not be copied either.

On the one hand, this is simple to explain/document, but on the other hand, it would mean you have to re-configure the same connection options in every slice if you want multiple slices connecting to the same database. In fact, if you have multiple slices, then configuring connection options in one place but not another would lead to spurious DB connections being created since the `connection_options` are included as part of the gateway's cache keys.

I think what you've done here is good and I'm going to merge it as is, but I'll make another issue for us to handle copying gateway config from the parent if a slice happens to be using the same database URL.

timriley (approved):

timriley:
The other issue: https://github.com/hanami/hanami/issues/1457 - [Copy gateway config from parent slice DB provider](https://github.com/hanami/hanami/issues/1457)

Issue #1387, opened 2024-05-24
Raise helpful error from db provider if the necessary database driver gem is missing


timriley:
Based on the database_url, we'll know if the user is using a Postgres, MySQL or SQLite database. If the corresponding driver gem is not bundled (e.g. `"pg"` for Postgres), we should proactively raise a helpful error, telling the user which gem to install.

PR #1453, opened 2024-09-12
raises error if necessary db driver gem not installed

kyleplump:
closes [#1387](https://github.com/hanami/hanami/issues/1387)

this check runs when the db provider is spinning up

currently, it will return just a message with the gem name that needs to be installed.  it might be more helpful if we output a link to the rubygems page for the offending gem? also - does this error need a stack trace to come with it? thoughts?

thanks!


timriley:
Hey @kyleplump! Thanks so much for taking this on! I've just made a few of adjustments (you can check out my commits to see) and I'll merge this in now. I'm happy we'll have this in place for beta2 😄

kyleplump:
thanks for the finesse @timriley! i learn something new every time i read through the changes (like `Hanami#bundled?`!)

PR #1456, opened 2024-10-07
`Dry::Operation` extensions

### LINKED ISSUE #1437: Automatically include ROM extension for operations used inside Hanami apps
Eliminate this little line of boilerplate, which we (temporarily) introduce in https://github.com/hanami/cli/pull/180.

```ruby
include Dry::Operation::Extensions::ROM
```

alassek:
Resolves #1437

Automatically extend Dry::Operation to include ROM extensions when hanami-db is present. Removes a line of boilerplate from generated Operation files.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,9 @@ gem "dry-system", github: "dry-rb/dry-system", branch: "main"
 # This is needed for settings specs to pass
 gem "dry-types"
 
+# For testing operation integrations
+gem "dry-operation", github: "dry-rb/dry-operation", branch: "main"
+
 # For testing the DB layer
 gem "sqlite3"
 
--- a/lib/hanami/extensions.rb
+++ b/lib/hanami/extensions.rb
@@ -19,3 +19,7 @@ end
 if Hanami.bundled?("hanami-router")
   require_relative "extensions/router/errors"
 end
+
+if Hanami.bundled?("dry-operation")
+  require_relative "extensions/operation"
+end
--- a/lib/hanami/extensions/operation.rb
+++ b/lib/hanami/extensions/operation.rb
@@ -27,8 +27,7 @@ module Hanami
         # @api private
         # @since 2.2.0
         def configure_for_slice(slice)
-          namespace = slice.slice_name.namespace
-          include namespace::Deps["db.rom"]
+          include slice.namespace::Deps["db.rom"]
         end
 
         # @api private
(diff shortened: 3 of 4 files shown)


timriley:
Thanks for getting onto this, @alassek!

I have a couple of questions about ROM integration and gateway selection, though.

On ROM integration, based on the [extension’s own docs](https://github.com/dry-rb/dry-operation/blob/main/lib/dry/operation/extensions/rom.rb), we'll need to provide a `#rom` method to expose the ROM container to the transaction. Unless I've missed something in this PR, I don't think we're doing that yet?

To achieve this, I think we'd want to go down the route of using `SliceConfigurable` and injecting the `"db.rom"` object for the operation's specific slice. Would you be up for doing this? There are examples of us doing in the other Hanami extensions, but please let me know if it would be helpful for me to provide any extra detail/context here :)

Next, on gateway selection. Again, based on the [extension docs](https://github.com/dry-rb/dry-operation/blob/main/lib/dry/operation/extensions/rom.rb), we can specify a gateway name at `include`-time:

```ruby
include Dry::Operation::Extensions::ROM[gateway: :my_gateway]
```

or runtime:

```ruby
transaction(gateway: :my_gateway) do
```

If we don't specify a gateway name, then it uses `:default`.

I think if we make operations slice-aware (per my suggestion above), then I think we should be pretty safe to simply allow the default gateway of `:default`, since that's what we already specify as the name of the default gateway via Hanami's `:db` provider. And if a user needs a transaction for a different gateway, they can specify it themselves via the `#transaction` runtime API.

So I don't think this aspect requires us to do any extra work, but I just wanted to raise it here in case you had any other thoughts :)

Thank you again, and please let me know if I can help any further!

alassek:
@timriley Okay this has been updated to use `SliceConfigurable`. Let me know if this wasn't what you had in mind.

timriley:
@alassek Looks great, much more succinct than some of the other SliceConfigurable-based extensions! 😍 I don't know what, but I never thought to directly `include slice.namespace::Deps` like you did here.

I made a couple of very minor tidy-ups (see commits above) and will squash/merge this in now 😄

alassek:
@timriley aha! I didn't know `Slice#namespace` was a thing. That's better.

Many people don't know that the left-hand side of `::` can be _any expression_, that's one of those Ruby things that is either extremely useful or extremely cursed depending on who you're talking to :laughing:

PR #1459, opened 2024-10-23
Apply gateway config from parent slice (and simplify config)

### LINKED ISSUE #1457: Copy gateway config from parent slice DB provider

timriley:
Follow up from @alassek's work in #1450 and ensure that a gateway configured with `connection_options` in an app's DB provider can have those same `connection_options` used in child slices configured to use the same gateway.

To achieve this, when a slice's `:db` provider prepares itself, copy the config from matching gateways in the parent slice. This will apply `connection_options` configured from the parent.

This is an important change, because it ensures that the parent and its child slices can share database connections for the gateway (this is because `connection_options` is now included in the cache key for the gateways that we cache and reuse across slices).

Some additional particulars about the behavior:

- Config from parent gateways is only copied if the gateway has an identical (a) name, and (b) database URL.
- Aside from `database_url`, gateways only have three other settings. Of these, `adapter_name` and `connection_options` are copied directly.
- The `adapter` setting has some special handling: if there is a custom default adapter configured for the provider (i.e. `config.adapter(:sql) do ... end`), then the `.adapter` from the parent gateway will _not_ be copied. This is because we want the provider's custom default adapter to be applied to the gateway in order to honour the user's intention for that config.
- To support the above, change `config.adapters[...]` to return `nil` if an adapter has not been explicitly configured for the provider, rather than have it create a new adapter and automatically assign it to the given key. To offer the existing behaviour of "return or create+assign me an adapter, please", we add `config.adapters.adapter(...)`, which we use internally where required. (n.b. this is entirely private API, the public interface of `config.adapter(...)` is unchanged.
- If a user has set `config.db.configure_from_parent = false` in the slice, then we uphold the existing behaviour for this flag, and do not copy any gateway config from the parent.

---

While here, simplify the gateway config a little: removing the `config.any_adapter` from DB providers.

Keeping this meant another complicating factor in how we apply config from parent gateways, and it made me realise that it just complicates how we think about gateway/adapter config _in general._

A user can still achieve what they want without having this method. If they need a ROM plugin to be used in multiple adapter, they can just configure it explicitly in each adapter. Given that the number of adapters in use will tend to be low, this does not feel onerous at all. We can always choose to reintroduce something like this based on actual user feedback, which I think is the right way to go in this case.

Resolves #1457

### DIFF
--- a/lib/hanami/providers/db.rb
+++ b/lib/hanami/providers/db.rb
@@ -113,7 +113,7 @@ module Hanami
           # Preserve settings already configured locally
           next if config.configured?(key)
 
-          # Do not copy gateways, these are always configured per slice
+          # Skip gateway config, we handle this in #configure_gateways
           next if key == :gateways
 
           # Skip adapter config, we handle this below
@@ -123,12 +123,12 @@ module Hanami
         end
 
         parent_db_provider.source.config.adapters.each do |adapter_name, parent_adapter|
-          adapter = config.adapters[adapter_name]
+          adapter = config.adapter(adapter_name)
 
           adapter.class.settings.keys.each do |key|
             next if adapter.config.configured?(key)
 
-            adapter.config[key] = parent_adapter.config[key]
+            adapter.config[key] = parent_adapter.config[key].dup
           end
         end
       end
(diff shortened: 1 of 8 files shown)


alassek (approved):
Looks great! :shipit:

PR #1495, opened 2025-02-07
Add APP_ENV as a fallback of HANAMI ENV

svoop:
Deprecate `RACK_ENV` in favor of `APP_ENV` as the alternative and fallback to `HANAMI_ENV`.

Background: The [purpose of`RACK_ENV` in `Rack::Server`](https://github.com/rack/rack/issues/1546) is different from setting the environment of the application.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The web, with simplicity.
 
+## [Unreleased]
+
+### Deprecated
+
+- [Sven Schwyn] Add `APP_ENV` as a fallback of `HANAMI_ENV` (#1487)
+
 ## v2.2.1 - 2024-11-12
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@ The web, with simplicity.
 
 ## [Unreleased]
 
-### Deprecated
+### Added
 
-- [Sven Schwyn] Add `APP_ENV` as a fallback of `HANAMI_ENV` (#1487)
+- Check `ENV["APP_ENV"]` for the Hanami env if `ENV["HANAMI_ENV"]` is not set. The order of environment variable checks is now `HANAMI_ENV`->`APP_ENV`->`RACK_ENV` (@svoop in #1487).
 
 ## v2.2.1 - 2024-11-12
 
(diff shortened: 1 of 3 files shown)


runlevel5 (approved):

timriley:
Thanks for looking into this, @svoop.

Now that the Rack project has removed their use of the `RACK_ENV` var (as noted in that issue you linked), the ambiguity over its use has been cleared up, I think: it's no longer used by Rack.

Given this, I wonder whether it's really worth the disruption/deprecation cycle for us to drop our support for it. It's only there for a fallback, anyway. HANAMI_ENV is our documented and preferred var for this.

So if we did leave RACK_ENV support in place, what's left is the question of whether we should look for APP_ENV in addition to it. To be honest, I have been completely unaware of APP_ENV as any kind of emerging community standard. The situation seems kind of mixed. RACK_ENV is respected [by Roda](https://roda.jeremyevans.net/rdoc/classes/Roda/RodaPlugins/Environments.html), whereas APP_ENV is used [by Sinatra](https://sinatrarb.com/intro.html). I do note it's supported [by Sidekiq and Puma](https://github.com/puma/puma/issues/2692), though.

This tips me over the edge: I think we should support APP_ENV, but I don't think it's worth deprecating RACK_ENV. It can remain as an undocumented compatibility affordance.

Would you be happy to reorient this PR so it checks HANAMI_ENV -> APP_ENV -> RACK_ENV? Thank you!

svoop:
> Now that the Rack project has removed their use of the RACK_ENV var [...]
> Would you be happy to reorient this PR so it checks HANAMI_ENV -> APP_ENV -> RACK_ENV? Thank you!

Fair point! I'd rather get rid of RACK_ENV, but then again, this is not one of my baby-gems with a a few k downloads. 😄 (Also, I'm a bit drawn by underdog positions, why else would I replace RSpec with Minitest in my Hanami pet project and try whether `gems.rb` instead of `Gemfile` breaks things.)

That said, I'll change things the way you suggested this afternoon.

svoop:
@timriley There you go, `APP_ENV` is now an alternative to `HANAMI_ENV` in the order you suggested.

timriley:
Thanks @svoop! Massaged the CHANGELOG entry and docs, and now this is good to go.

svoop:
@timriley Thanks for rewording things, I've had a bit of a hassle hesitating between brevity and clarity. Lesson learned: Clarity beats brevity. 👍

PR #1496, opened 2025-02-07
Use version agnostic URL to API doc in README

### LINKED ISSUE #1491: [website] link to latest API documentation is broken
On the webpage for the latest API (https://docs.hanamirb.org/ which redirects to https://docs.hanamirb.org/2.1.0/) there is a line under the "Contact" section:

**API Doc:** http://rdoc.info/gems/hanami

Clicking on that link ends up in error. Chrome complains of too many redirects.
On the other hand https://rubydoc.info/gems/hanami can be reached... 

Also, all the non-latest versions pages (https://docs.hanamirb.org/2.1.0/, https://docs.hanamirb.org/2.0.0/, etc...) contain the same link pointing to the latest version. I suggest to point instead to the version specific documentation: https://rubydoc.info/gems/hanami/2.1.0, https://rubydoc.info/gems/hanami/2.0.0 etc...

svoop:
fixes #1491

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ You can give back to Open Source, by supporting Hanami development via [GitHub S
 * Guides: https://guides.hanamirb.org
 * Snippets: https://snippets.hanamirb.org
 * Mailing List: http://hanamirb.org/mailing-list
-* API Doc: https://gemdocs.org/gems/hanami/latest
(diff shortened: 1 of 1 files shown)


cllns:
Great, thanks @svoop!

PR #1503, opened 2025-02-26
Support optional nonce in Rack requests, CSP header rules and view helpers

svoop:
If the `Hanami::Middleware::ContentSecurityPolicyNonce` middleware is in use:

* A random unique 8 character nonce is generated on every request and written to the Rack env `hanami.content_security_policy_nonce`key.
* All occurrences of the placeholder `'nonce'` in the generated CSP header are replaced with its current specific incarnation like `'nonce-A12OggyZ'`.
* The new `content_security_policy_nonce` view helper returns the current nonce.
* The `javascript_tag` and `stylesheet_tag` view helpers are decorated with an nonce attribute like `nonce=A12OggyZ`.

A few things to check when this PR is reviewed:

* Is the use of `Hanami.app.config.actions.content_security_policy` in the Rack middleware permissible? (Plan B: Don't assume a CSP header to be present, but in this case, the Rack env nonce key is always generated whether it is needed or not.)
* The helper is named `content_security_policy_nonce` which appears a bit verbose since most apps don't use other nonces. However, you never know and as the corresponding helper in Rails has this verbose name as well, it might ease the transition to Hanami just a tiny little bit.
* The tests are not where other helpers like `javascript_tag` are tested because it takes a real request for this to work.

### DIFF
--- a/lib/hanami/config/actions/content_security_policy.rb
+++ b/lib/hanami/config/actions/content_security_policy.rb
@@ -96,6 +96,29 @@ module Hanami
           @policy.delete(key)
         end
 
+        # Returns true if 'nonce' is used in any of the policies.
+        #
+        # @return [Boolean]
+        #
+        # @api public
+        # @since x.x.x
+        def nonce?
+          @policy.any? { _2.match?(/'nonce'/) }
+        end
+
+        # Returns an array of middleware name to support 'nonce' in
+        # policies, or an empty array if 'nonce' is not used.
+        #
+        # @return [Array<(Symbol, Array)>]
+        #
+        # @api public
+        # @since x.x.x
+        def middleware
+          return [] unless nonce?
+
+          [Hanami::Middleware::ContentSecurityPolicyNonce]
+        end
+
         # @since 2.0.0
         # @api private
         def to_s
--- a/lib/hanami/helpers/assets_helper.rb
+++ b/lib/hanami/helpers/assets_helper.rb
@@ -2,6 +2,7 @@
 
 require "uri"
 require "hanami/view"
+require "../constants"
 
 # rubocop:disable Metrics/ModuleLength
 
@@ -713,7 +714,7 @@ module Hanami
       def content_security_policy_nonce
         return unless _context.request?
 
-        _context.request.env["hanami.content_security_policy_nonce"]
+        _context.request.env[CONTENT_SECURITY_POLICY_NONCE_REQUEST_KEY]
       end
 
       private
--- a/lib/hanami/helpers/assets_helper.rb
+++ b/lib/hanami/helpers/assets_helper.rb
@@ -711,7 +711,7 @@ module Hanami
       #
       #   <script nonce="<%= content_security_policy_nonce %>">
       def content_security_policy_nonce
-        _context.request.env['hanami.content_security_policy_nonce']
+        _context.request.env["hanami.content_security_policy_nonce"]
       rescue Hanami::ComponentLoadError
         nil
       end
--- a/lib/hanami/helpers/assets_helper.rb
+++ b/lib/hanami/helpers/assets_helper.rb
@@ -88,6 +88,11 @@ module Hanami
       # name of the algorithm, then a hyphen, then the hash value of the file.
       # If more than one algorithm is used, they"ll be separated by a space.
       #
+      # If the `Hanami::Middleware::ContentSecurityPolicyNonce` middleware is
+      # in use, the nonce value of the current request is automatically added
+      # as an attribute. See {#content_security_policy_nonce} for how to
+      # configure this correctly.
+      #
       # @param source_paths [Array<String, #url>] one or more assets by name or absolute URL
       #
       # @return [Hanami::View::HTML::SafeString] the markup
@@ -163,7 +168,8 @@ module Hanami
         _safe_tags(*source_paths) do |source|
           attributes = {
             src: _typed_path(source, JAVASCRIPT_EXT),
-            type: JAVASCRIPT_MIME_TYPE
+            type: JAVASCRIPT_MIME_TYPE,
+            nonce: content_security_policy_nonce
           }
           attributes.merge!(options)
 
@@ -191,6 +197,11 @@ module Hanami
       # If the "subresource integrity mode" is on, `integrity` is the
       # name of the algorithm, then a hyphen, then the hashed value of the file.
       # If more than one algorithm is used, they"ll be separated by a space.
+      #
+      # If the `Hanami::Middleware::ContentSecurityPolicyNonce` middleware is
+      # in use, the nonce value of the current request is automatically added
+      # as an attribute. See {#content_security_policy_nonce} for how to
+      # configure this correctly.
       #
       # @param source_paths [Array<String, #url>] one or more assets by name or absolute URL
       #
@@ -257,7 +268,8 @@ module Hanami
           attributes = {
             href: _typed_path(source_path, STYLESHEET_EXT),
             type: STYLESHEET_MIME_TYPE,
-            rel: STYLESHEET_REL
+            rel: STYLESHEET_REL,
+            nonce: content_security_policy_nonce
           }
           attributes.merge!(options)
 
@@ -673,6 +685,35 @@ module Hanami
         return source_path if _absolute_url?(source_path)
 
         _context.assets[source_path].url
+      end
+
+      # Random per request nonce value for Content Security Policy (CSP) rules.
+      #
+      # If the `Hanami::Middleware::ContentSecurityPolicyNonce` middleware is
+      # in use, this helper returns the nonce value for the current request
+      # or `nil` otherwise.
+      #
+      # For this policy to work in the browser, you have to add the `'nonce'`
+      # placeholder to the script and/or style source policy rule. It will be
+      # substituted by the current nonce value like `'nonce-A12OggyZ'.
+      #
+      # @return [String, nil] nonce value of the current request
+      #
+      # @since x.x.x
+      #
+      # @example App configuration
+      #
+      #   config.middleware.use Hanami::Middleware::ContentSecurityPolicyNonce
+      #   config.actions.content_security_policy[:script_src] = "'self' 'nonce'"
+      #   config.actions.content_security_policy[:style_src] = "'self' 'nonce'"
+      #
+      # @example View helper
+      #
+      #   <script nonce="<%= content_security_policy_nonce %>">
+      def content_security_policy_nonce
+        _context.request.env['hanami.content_security_policy_nonce']
+      rescue Hanami::ComponentLoadError
+        nil
       end
 
       private
--- a/lib/hanami/helpers/assets_helper.rb
+++ b/lib/hanami/helpers/assets_helper.rb
@@ -2,7 +2,7 @@
 
 require "uri"
 require "hanami/view"
-require "../constants"
+require_relative "../constants"
 
 # rubocop:disable Metrics/ModuleLength
 
--- a/lib/hanami/helpers/assets_helper.rb
+++ b/lib/hanami/helpers/assets_helper.rb
@@ -89,10 +89,10 @@ module Hanami
       # name of the algorithm, then a hyphen, then the hash value of the file.
       # If more than one algorithm is used, they"ll be separated by a space.
       #
-      # If the `Hanami::Middleware::ContentSecurityPolicyNonce` middleware is
-      # in use, the nonce value of the current request is automatically added
-      # as an attribute. See {#content_security_policy_nonce} for how to
-      # configure this correctly.
+      # If the Content Security Policy uses 'nonce' and the source is not
+      # absolute, the nonce value of the current request is automatically added
+      # as an attribute. You can override this with the `nonce: false` option.
+      # See {#content_security_policy_nonce} for more.
       #
       # @param sources [Array<String, #url>] one or more assets by name or absolute URL
       #
@@ -145,6 +145,10 @@ module Hanami
       #
       #   # <script src="/assets/application.js" type="text/javascript" defer="defer"></script>
       #
+      # @example Disable nonce
+      #
+      #   <%= javascript_tag "application", nonce: false %>
+      #
       # @example Absolute URL
       #
       #   <%= javascript_tag "https://code.jquery.com/jquery-2.1.4.min.js" %>
@@ -165,12 +169,13 @@ module Hanami
       #   #         type="text/javascript"></script>
       def javascript_tag(*sources, **options)
         options = options.reject { |k, _| k.to_sym == :src }
+        nonce_option = options.delete(:nonce)
 
         _safe_tags(*sources) do |source|
           attributes = {
             src: _typed_url(source, JAVASCRIPT_EXT),
             type: JAVASCRIPT_MIME_TYPE,
-            nonce: content_security_policy_nonce
+            nonce: _nonce(source, nonce_option)
           }
           attributes.merge!(options)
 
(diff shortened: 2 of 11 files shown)


runlevel5 (approved):
LGTM 👍

svoop:
[521d610](https://github.com/hanami/hanami/pull/1503/commits/521d610d9b609c45ca1452e7ab080480ad3e6961): Rebased to bring it back in-sync with main.

timriley:
Thanks @svoop, this PR is next for me to review :)

spec/integration/web/content_security_policy_nonce_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+  describe "HTML request" do
+    context "CSP enabled" do
+      def before_prepare
+        write "config/app.rb", <<~RUBY
+          require "hanami"
+
+          module TestApp
+            class App < Hanami::App
+              config.middleware.use Hanami::Middleware::ContentSecurityPolicyNonce
+              config.actions.content_security_policy[:script_src] = "'self' 'nonce'"
```
timriley, line 85: Could we explore a way of making this nonce support work _without_ the user having to write this `config.middleware.use` line?

We already have some patterns for this. For example, check out `#middleware` [inside lib/hanami/config/actions/sessions.rb](https://github.com/hanami/hanami/blob/f3f0149db658db4293cb084b537a9e7557eb563f/lib/hanami/config/actions/sessions.rb#L62-L66). This is then invoked as part of the framework [setting up the router](https://github.com/hanami/hanami/blob/f3f0149db658db4293cb084b537a9e7557eb563f/lib/hanami/slice.rb#L1050-L1052) for the app or slice.

Maybe we could do the same thing inside `Hanami::Config::Actions::ContentSecurityPolicy`? It could then automatically return the middleware if the `script-src` or `style-src` values include `'nonce'`?

What I like about an approach like this:

1. It's streamlined for the user. Less boilerplate config. Instead, they reveal their intention (“I want a nonce”), and then _everything just works_.
2. It gains us more flexibility for future refactors. If our users are not explicitly configuring the middleware, then we can more easily change its naming, structure or purpose as we continue to improve our content security policy support.

lib/hanami/middleware/content_security_policy_nonce.rb
```diff
@@ -0,0 +1,50 @@
...
+      def initialize(app)
+        @app = app
+        @nonce = random_nonce
+      end
+
+      # @api private
+      # @since x.x.x
+      def call(env)
+        if Hanami.app.config.actions.content_security_policy == false
+          @app.call(env)
```
timriley, line 29:
```suggestion
          return @app.call(env)
```

Little code thought: we could turn this into a simple guard clause with an early return here, then flatten the main logic below.

timriley (commented):
This is looking good, @svoop! Thanks for taking this on :) 

One main bit of design feedback from me, via inline comment: let's see if we can evolve this so the user doesn't need to explicitly include the middleware.

Let me know if that makes sense. Happy to discuss!

spec/integration/web/content_security_policy_nonce_spec.rb
svoop: Sure does, I'll look into it!

lib/hanami/helpers/assets_helper.rb
```diff
@@ -672,6 +684,35 @@ def asset_url(source_path)
...
+      #   config.middleware.use Hanami::Middleware::ContentSecurityPolicyNonce
+      #   config.actions.content_security_policy[:script_src] = "'self' 'nonce'"
+      #   config.actions.content_security_policy[:style_src] = "'self' 'nonce'"
+      #
+      # @example View helper
+      #
+      #   <script nonce="<%= content_security_policy_nonce %>">
+      def content_security_policy_nonce
+        _context.request.env['hanami.content_security_policy_nonce']
+      rescue Hanami::ComponentLoadError
```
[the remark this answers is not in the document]
svoop: @timriley Couple a dozen tests fail with "Request not available. Only views rendered from Hanami::Action instances have a request" unless I rescue here. Maybe you see a better way to work around this?

spec/integration/web/content_security_policy_nonce_spec.rb
svoop: @timriley See https://github.com/hanami/hanami/pull/1503/commits/70dd139e7135d7e2abeabc376a700b4749b020b3 for a version which auto-discovers the use of nonce.

timriley (commented):
Thanks for continuing to work on this, @svoop!

I've replied to all the inline comments now, and I've also pushed up a few changes myself. They're all in small commits so hopefully they're easy to follow.

Let me know if you have any questions about them.

The biggest change was changing the middleware so a new nonce would be generated _for each request_ (which is the whole point of nonces) rather than just once when the middleware is attached (the `@nonce` ivar that was there previously).

It would be good to add a test that exercise two requests in succession and verifies that the nonce is different for each. Would you be OK to give that a go?

My last question for you is around configurability. Security is one of those areas where it's valuable to pay attention to existing, well-honed implementations. So I looked at Rails' [CSP nonce support](https://guides.rubyonrails.org/security.html#adding-a-nonce). 

Rails' nonce support is more configurable than this one here, which is to be expected. I don't want us to copy every bell and whistle as a starting point, but as the person who's introducing CSP nonces to Hanami, I'm interested in your take on these things in particular:

**Should the nonce generator be configurable?**

Right now we're not supporting this, but the Rails guides do make note of needing difference nonce generation strategies to support HTTP Conditional GET strategies.

My feeling is that it wouldn't hurt to allow the user to configure their own nonce generator proc in the same way.

**Should nonces in asset tag helpers be opt-in (or opt-out)?**

Rails will only put the nonce on a tag when you explicitly opt in, e.g. `javascript_tag(nonce: true)`.

Given our `javascript_tag` and `stylesheet_tag` helpers can both reference external URLs, I'm thinking that a similar approach in our case might be good too. Only include the nonce when opting in.

What do you think?

lib/hanami/config/actions/content_security_policy.rb
```diff
@@ -96,6 +96,29 @@ def delete(key)
...
         end
 
+        # Returns true if 'nonce' is used in any of the policies.
+        #
+        # @return [Boolean]
+        #
+        # @api public
+        # @since x.x.x
+        def nonce?
+          @policy.any? { _2.match?(/'nonce'/) }
```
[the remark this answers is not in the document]
svoop: I thought about that but decided against because CSP evolves and new policies might be introduced which allow nonces as well. By scanning all policies, the implementation is upwards compatible in that respect, and the overhead shouldn't matter since this is only done once. But it's certainly a matter of taste, so since you're the Majordomo at this party, I suggest you call it!

svoop:
> The biggest change was changing the middleware so a new nonce would be generated for each request (which is the whole point of nonces) rather than just once when the middleware is attached (the @nonce ivar that was there previously).

@timriley Good catch, not sure what I thought here. It's a while since I last touched middleware, but still, should have known.

lib/hanami/config/actions/content_security_policy.rb
```diff
@@ -96,6 +96,29 @@ def delete(key)
...
+        #
+        # @return [Array<(Symbol, Array)>]
+        #
+        # @api public
+        # @since x.x.x
+        def middleware
+          return [] unless nonce?
+
+          [Hanami::Middleware::ContentSecurityPolicyNonce]
+        end
```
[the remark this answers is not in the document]
svoop: ???

svoop:
@timriley 

> It would be good to add a test that exercise two requests in succession and verifies that the nonce is different for each. Would you be OK to give that a go?

Sure thing!

> Should the nonce generator be configurable? Right now we're not supporting this, but the Rails guides do make note of needing difference nonce generation strategies to support HTTP Conditional GET strategies.

The Rails guides are right about this one IMO, a session bound nonce is better in terms of caching... provided a safe session identifier is available. And since the "safe session" bit is out of the hands of the stack, a custom nonce generator would delegate this responsibility to the developer. Seems like the best approach indeed.

> Should nonces in asset tag helpers be opt-in (or opt-out)? Rails will only put the nonce on a tag when you explicitly opt in, e.g. javascript_tag(nonce: true). Given our javascript_tag and stylesheet_tag helpers can both reference external URLs, I'm thinking that a similar approach in our case might be good too. Only include the nonce when opting in.

Or add the nonce for paths (local request) and omit it for URLs (remote request), both overrideable with a `nonce: true|false` argument. This should cover more cases out of the box, but it introduces a tad more complexity. What do you think?

lib/hanami/middleware/content_security_policy_nonce.rb
```diff
@@ -38,7 +38,7 @@ def call(env)
       private
 
       def random_nonce
-        SecureRandom.alphanumeric(8)
+        SecureRandom.urlsafe_base64(16)
```
svoop, line 43: @timriley 👍 I peeked at another source than the Rails guides for inspiration – and an entrophy of 3.5 trillion seemed more than enough for an nonce. But since MDN also suggests 16bit, better safe than sorry. 😄

lib/hanami/middleware/content_security_policy_nonce.rb
```diff
@@ -29,10 +29,12 @@ def call(env)
 
         request_nonce = generate_nonce
         env[CONTENT_SECURITY_POLICY_NONCE_REQUEST_KEY] = request_nonce
-        @app.call(env).tap do |response|
-          headers = response[1]
-          headers["Content-Security-Policy"] = sub_nonce(headers["Content-Security-Policy"], request_nonce)
-        end
+
+        _, headers, _ = response = @app.call(env)
```
svoop, line 33: Not such a fan of `_`, but I ack that I like `tap` and `then` maybe a tad too much. 😅

spec/integration/web/content_security_policy_nonce_spec.rb
svoop: Yes, we can.

svoop:
TODO:

* [ ] Remove hardcoded middleware use in tests
* [ ] Customizable nonce generator
* [ ] Test for different nonce across two requests
* [ ] `nonce: true|false` argument for asset helpers

spec/integration/web/content_security_policy_nonce_spec.rb
svoop: 🏁

lib/hanami/config/actions/content_security_policy.rb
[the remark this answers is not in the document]
timriley: That was me just saying I loved this outcome :) 

Unknown

lib/hanami/config/actions/content_security_policy.rb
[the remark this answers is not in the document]
timriley: Yeah, this reasoning is fair, let's leave it.

timriley:
> Or add the nonce for paths (local request) and omit it for URLs (remote request), both overrideable with a nonce: true|false argument. This should cover more cases out of the box, but it introduces a tad more complexity. What do you think?

@svoop This sounds good to me!

spec/integration/web/content_security_policy_nonce_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+
+    context "CSP disabled" do
+      def before_prepare
+        write "config/app.rb", <<~RUBY
+          require "hanami"
+
+          module TestApp
+            class App < Hanami::App
+              config.middleware.use Hanami::Middleware::ContentSecurityPolicyNonce
+              config.actions.content_security_policy = false
```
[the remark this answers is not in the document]
svoop: I've added such a test.

svoop:
@timriley I've implemented [all your suggestions](#issuecomment-2804830589), however, there's one nut I can't quite crack myself:

The `nonce` generator is not configurable as per 0c5cfd3305da2f086409844e138aef2f201b6211. This works fine for random values, however, building session-bound values based on e.g. `request.session[:uuid]` is trickier. I pass the Rack request object to the `content_security_policy_nonce_generator` proc, but at that point, the session data is not yet ready:

```ruby
request.session
=> #<Rack::Session::Abstract::PersistedSecure::SecureSessionHash:0x2d78 not yet loaded>
```

The loading should be triggered when accessing the data e.g. with `request.session["_whatever"]`, but that's not happening here. Maybe a loading order issue, so I switched from Redis session to Cookie session and configured it before the CSP stuff in `config/app.rb` – same result. 

I know, it's not likely to nail the cause by just reading this comment, then again, you're so familiar with the stack and load order, maybe you have an idea where I should look next. (Worth a shot before I create a working example app.)

---

Other than that, the following LOC might merit your special attention for review:

* The `nonce` generator proc mentioned above takes the Rack request object as an optional argument. I check this with `arity` to omit having to build the request object if it's not needed in the proc. This adds a bit of cognitive complexity to the code though, would you prefer always passing it instead? https://github.com/hanami/hanami/blob/0c5cfd3305da2f086409844e138aef2f201b6211/lib/hanami/middleware/content_security_policy_nonce.rb#L30
* Commit b9fa3b646071c9b30dcc3eaa45a7895254dd9224 consolidates the wording of `source_path` and `source`. IMO, both terms were used inconsistently but interchangeably even between `javascript_tag` and `stylesheet_tag`which bit me in the butt. The tests all pass, so it's okay, but I'm not 100% sure whether there was a reasoning for this divergent wording I miss to see.
* The way the `nonce` argument for `javascript_tag` and `stylesheet_tag` works now requires a conditional which probably won't win a beauty contest. However, besides the behavior we discussed above `nonce: (nil|true|false)`, it also allows for an external value to be explicitly set with `nonce: "foobar"`. Not sure whether this is a real use case, but since `nonce` is an attribute of both `<script>` and `<link>`, I figured it should be possible to set it this way similar to other arguments like `defer`. https://github.com/hanami/hanami/blob/c6fd30b0297238cf1dee405613501b55cbefcd85/lib/hanami/helpers/assets_helper.rb#L771

Happy Easter 🐇 !

svoop:
@timriley With a fresh head after a few days off, I revisited the session problem (described in my now hidden comment just above this one) and... there is no problem. I simply made a mistake when I tried it with `curl`.

So all your suggestions are implemented and it would be cool if you could take a look eventually. Here are a few explanatory remarks you might want to read beforehand:

* The `nonce` generator proc mentioned above takes the Rack request object as an optional argument. I check this with `arity` to omit having to build the request object if it's not needed by the proc. This adds a bit of cognitive complexity to the code though, would you prefer always passing it instead? https://github.com/hanami/hanami/blob/0c5cfd3305da2f086409844e138aef2f201b6211/lib/hanami/middleware/content_security_policy_nonce.rb#L30
* Commit b9fa3b646071c9b30dcc3eaa45a7895254dd9224 consolidates the wording of `source_path` and `source`. IMO, both terms were used inconsistently but interchangeably – even between `javascript_tag` and `stylesheet_tag`which bit me in the butt. The tests all pass, but I'm not 100% sure whether there was a reasoning for this divergent wording I miss to see.
* The way the `nonce` argument for `javascript_tag` and `stylesheet_tag` works now requires a conditional which probably won't win a beauty contest. However, besides the behavior we discussed above `nonce: (nil|true|false)`, it also allows for an external value to be explicitly set with `nonce: "foobar"`. Not sure whether this is a real use case, but since `nonce` is an attribute of both `<script>` and `<link>`, I figured it should be possible to set it this way similar to other arguments like `defer`. https://github.com/hanami/hanami/blob/c6fd30b0297238cf1dee405613501b55cbefcd85/lib/hanami/helpers/assets_helper.rb#L771

Thanks a bunch!

svoop:
@timriley Ping, thanks for taking another look!

timriley:
@svoop hopefully later this week. I've been in talk preparation mode for the full live rehearsal I'm giving tomorrow :)

timriley (commented):
@svoop Fantastic work, thank you! This all looks good to me now. Happy with all your recent changes :)

Could you please do a squash and merge (along with an appropriate commit message) whenever you're ready?

PR #1507, opened 2025-03-04
Fix CI: Upgrade actions to latest versions

cllns:
CI is now failing due to Github no longer supporting old version of their actions. Example: https://github.com/hanami/hanami/actions/runs/13646312642 and the link it sends us to: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

I looked up the latest versions and changed it to them, without investigating if anything needs to change. CI still passes though so I think we're in the clear.

I'm happy to do this for the other repos too. With your approval @timriley, I'll open PR's, make sure they're green, then merge without seeking code review for each one. That work for you?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
           - "3.1"
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v4
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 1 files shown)


runlevel5 (approved):

timriley:
Thanks for doing this, @cllns! 😄 

I just handled the other repos now (see the linked PRs above), and I also pushed [f57155f](https://github.com/hanami/hanami/commit/f57155f63aca44a1393d2a6445e98f9bdccd91ca) just for a little bit of extra tidy up/consistency.

PR #1508, opened 2025-03-06
Remove unused shared examples

cllns:
These are old, from Hanami 1, so we can get rid of them. 

This is based of #1507 so CI will run and pass.


timriley (approved):
Amazing! Thank you @cllns! Let's get this in (after rebasing over main again). 🙇🏼

PR #1509, opened 2025-03-07
Fix docs for config.no_auto_register_paths

cllns:


lib/hanami/config.rb
```diff
@@ -104,7 +104,11 @@ class Config
     # @!attribute [rw] no_auto_register_paths
     #   Sets the paths to skip from container auto-registration.
     #
-    #   Defaults to `["entities"]`.
+    #   Paths are relative to the app root or the slice root.
```
timriley, line 107:
```suggestion
    #   Paths are relative to `/app` or a slice root.
```

I think this would be clearer and more correct?

"App root" I would consider to be the `Hanami.app.root`, which will be the root of the whole project, not the `/app` directory.

lib/hanami/config.rb
```diff
@@ -104,7 +104,11 @@ class Config
...
+    #
+    #   The `/lib` directory is always excluded from auto-registration.
```
timriley, line 109: If we clarify the `/app` detail above, I wonder whether this line is needed anymore? I think this kind of information is better suited to our guides, vs API-level docs for an individual setting?

timriley (commented):
A couple of little thoughts!

timriley (approved):
Thanks!

PR #1513, opened 2025-03-17
Update changelog about router redirect

inouire:
Is it a change or a fix ? It's fixing a regression compared to Hanami 1, but adds more power with custom URL schemes !
For now I made it a change ;)


timriley:
@inouire I think "Changed" is the right section here. But since the change was in hanami/router (https://github.com/hanami/router/pull/282), you'll want to put it in router's changelog. We keep changelog notes in the gem where the relevant change was made. Are you OK to move it over?

katafrakt:
This is already present in `hanami-router`'s changelog (https://github.com/hanami/hanami-router/blob/main/CHANGELOG.md#changed-4), so I'm closing this one.

Issue #1511, opened 2025-03-13
Fix sqlite warnings in tests


timriley:
As of some relatively recent sqlite3 gem changes, our test suite now emits a bunch of warnings like this:

```
/Users/tim/.local/share/mise/installs/ruby/3.4.2/lib/ruby/gems/3.4.0/gems/sqlite3-2.5.0-arm64-darwin/lib/sqlite3/fork_safety.rb:43: warning: Writable sqlite database connection(s) were inherited from a forked process. This is unsafe and the connections are being closed to prevent possible data corruption. Please close writable sqlite database connections before forking.
```

It would be really great to get rid of these :)

It's something to do with how we manage/setup/teardown sqlite databases between tests.

WToa:
Hey @timriley, I'm exploring the codebase so I tried addressing this in https://github.com/hanami/hanami/pull/1514, let me know your thoughts

timriley:
Foxed by #1514

PR #1514, opened 2025-03-25
Ensure DB connections are closed in app_integration tests

WToa:
Addresses https://github.com/hanami/hanami/issues/1511

More context can be found here: https://github.com/sparklemotion/sqlite3-ruby/pull/558, essentially open SQLite DB connections that exist prior to forking will trigger a warning.

This change ensures we disconnect all connections at the end of each `app_integration` test case so that we pre-emptively do so instead of letting the `sqlite3` gem do it for us and show a warning.


timriley:
@WToaThank you so much for doing this! 🥰

PR #1515, opened 2025-03-25
Update Contact section to point Chat URL to Discord

WToa:
Small change, noticed that it's now pointing to Discord on Contributing.md so wanted this change to be consistent


timriley (approved):
Great catch, thank you!

PR #1518, opened 2025-04-10
Fix deprecation warning about URI::DEFAULT_PARSER.

### LINKED ISSUE #1517: Fix URI RFC3986_PARSER warnings
## Why

Hello. :wave: I'm getting the following warnings in my console:

``` 
ruby/gems/3.4.0/gems/hanami-2.2.1/lib/hanami/helpers/assets_helper.rb:64: warning: URI::RFC3986_PARSER.make_regexp is obsolete. Use URI::RFC2396_PARSER.make_regexp explicitly.
```

## How

This would only require using `URI::RFC2396_PARSER.make_regexp` to fix the issue.

wuarmin:
Hey :wave:,

this should fix #1517. The fix is the same that was performed in the rack repo, see https://github.com/rack/rack/commit/3d0b9f377a4753d1fca40fca011c2c0e9282ab50.

Best regards


lib/hanami/helpers/assets_helper.rb
```diff
@@ -61,7 +61,9 @@ module AssetsHelper
 
       # @since 0.3.0
       # @api private
-      ABSOLUTE_URL_MATCHER = URI::DEFAULT_PARSER.make_regexp
+      ABSOLUTE_URL_MATCHER = (
+        defined?(URI::RFC2396_PARSER) ? URI::RFC2396_PARSER : URI::DEFAULT_PARSER
```
bkuhlmann, line 65: :thought_balloon: Does the older parser even need to be supported or does this break on older versions? Probably should be a spec for this especially if CI builds across multiple Ruby versions?

timriley: So we currently support [Ruby 3.1.0 and up](https://github.com/hanami/assets/blob/bf837a0539e4bb7c4b00ba41efb4f393968a4dec/hanami-assets.gemspec#L21).

On Ruby 3.1.0, `URI::RFC2396_PARSER` is not defined:

```
irb(main):001> RUBY_VERSION
=> "3.1.0"
irb(main):002> require "uri"; URI::RFC2396_PARSER
(irb):2:in `<main>': uninitialized constant URI::RFC2396_PARSER (NameError)

require "uri"; URI::RFC2396_PARSER
                  ^^^^^^^^^^^^^^^^
Did you mean?  URI::RFC2396_Parser
               URI::RFC3986_Parser
               URI::RFC3986_PARSER
```

Interestingly, _something_ like it is available in camel-cased form:

```
irb(main):004> RUBY_VERSION
=> "3.1.0"
irb(main):005> URI::RFC2396_Parser
=> URI::RFC2396_Parser
```

By the time we get to the last Ruby 3.1 patch release (3.1.7), `URI::RFC2396_PARSER` _is_ available:

```
irb(main):001> RUBY_VERSION
=> "3.1.7"
irb(main):002> require "uri"; URI::RFC2396_PARSER
=> #<URI::RFC2396_Parser:0x00000001080214b8>
```

An a surprising twist, it's gone again for the first release of 3.2 (3.2.0):

```
irb(main):001> RUBY_VERSION
=> "3.2.0"
irb(main):002> require "uri"; URI::RFC2396_PARSER
(irb):2:in `<main>': uninitialized constant URI::RFC2396_PARSER (NameError)

require "uri"; URI::RFC2396_PARSER
                  ^^^^^^^^^^^^^^^^
Did you mean?  URI::RFC2396_Parser
               URI::RFC3986_Parser
               URI::RFC3986_PARSER
```

But it's back for the last patch release of 3.2 (3.2.8):

```
irb(main):001> RUBY_VERSION
=> "3.2.8"
irb(main):002> require "uri"; URI::RFC2396_PARSER
=> #<URI::RFC2396_Parser:0x0000000105dd7948>
```

It's also present in 3.3.0.

Our stance is to support the latest 3 versions of Ruby, so when we make our next significant release, we can drop 3.1 support, but the variation from Ruby 3.2.0 to 3.2.8 means we should keep this `defined?` check in place.

Maybe we can add 3.2.0 explicitly to our test matrix just for the purposes of ensuring compatibility for this next year, until we can drop 3.2 support. @wuarmin — could you please do that as part of this PR?

It would also be good if you can add a comment above this new line saying we can drop the check and fallback once Ruby 3.3 becomes our minimum required version. This way someone will see it at the right time and remove it 😄

wuarmin: Thanks for the detailed feedback @timriley.
I resolved your suggestions.

timriley (approved):
Thanks @wuarmin!

PR #1519, opened 2025-04-16
Don't capture on absolute URL detection

svoop:
The URI parser regexp contain capture groups and are quite large to begin with. We can save a cycle or two by ignoring the capture groups here.

### DIFF
--- a/lib/hanami/helpers/assets_helper.rb
+++ b/lib/hanami/helpers/assets_helper.rb
@@ -703,7 +703,7 @@ module Hanami
       # @since 2.1.0
       # @api private
       def _absolute_url?(source)
-        ABSOLUTE_URL_MATCHER.match(source)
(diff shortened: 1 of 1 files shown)


svoop:
@kyleplump Not speed-wise, but since the monster produced by `URI::DEFAULT_PARSER.make_regexp` doesn't anchor the expression to the beginning of the string, `start_with?` would enforce that. Not sure though if that's desirable because it would falsely classify ` http://foobar.com` (leading whitespace) as non-absolute.

kyleplump:
@svoop good point - in my (albeit limited) bench marking `start_with?` can usually resolve faster, but in this case we'd need additional logic to cover any edge cases like the one you described.  so no real gain there in practice!

lgtm 🙂

timriley:
Thank you both 😄

PR #1528, opened 2025-07-03
Add support for single-character slice names

### LINKED ISSUE #1527: remove 2 character requirement for slice names
Currently the `SliceRegistrar` will throw an error for slices less than two characters long.  The only justification I can see for this behavior is simply to pass the regex `/^[a-z][a-z0-9_]+$/`.  A valid use case for single character slice names is given a parent slice `Auth` I would like to create two children slices called `N` and `Z` (`Auth::N` and `Auth::Z`).

aaronmallen:
resolves #1527


aaronmallen:
> Thanks! Looks good to me. I don't see why we shouldn't support this.

what are next steps? This is my first contribution. Do we just wait for Tim to merge or do I need to do anything more?

cllns:
Merged! Thanks @aaronmallen :) Will be in the next release, but of course you can get it now with: `gem "hanami", github: "hanami/hanami"`

PR #1534, opened 2025-09-12
Add memoize_component_namespaces config

### LINKED ISSUE #1533: Memoize actions and views in the container, measure performance impact
From @timriley in https://discourse.hanamirb.org/t/hanami-hack-day-at-baltic-ruby-oss-expo-2025/1228#p-3071-memoize-actions-and-views-in-the-container-measure-performance-impact-6

> **Memoize actions and views in the container, measure performance impact**
> 
> Right now our app and slice containers return new instances of these objects every time they are fetched. Instead, we could have the container set up these components using the memoize: true option, which would mean the same object is returned each time. We expect this may bring some good performance improvements! Let’s make this small adjustment, and then measure the impact.

A key benefit of our Actions and Views being immutable is that we can re-use objects, so let's actually try that out!

An idea for implementing this is to have `config.memoize_component_prefixes = ["actions.", "views."]` as the default value, which would let people add their own (with `<<`)  or disable it (with ` = []` or `= false`)as they choose.

And looking into it further, dry-system's [memoize happens via component directories rather than keys](https://dry-rb.org/gems/dry-system/1.0/component-dirs/#memoize), so perhaps the simplest way to do it is `config.memoize_component_dirs = ["actions/", "views/"]` which would be relative to app/ or a slice path

cllns:
Closes #1533.

I've included a simple benchmarking script. By default it runs with 0 instance variables, but you can go in and edit `minimal_view_code` and `minimal_action_code` an un-comment several lines to get 4 instance variables for each. We probably don't want to keep this in the final version, though creating boilerplate benchmarking scripts that others can use may be worthwhile.

Results: 8-9x speed up for 0 instance variables and 10-11x speed up for 4 instance variables.

This PR only _adds_ the ability to set `config.memoize_component_dirs`, it doesn't set a default. This is the safest option, to avoid breaking existing apps. Actions are already immutable so there's little risk of anything breaking there. In practice, Views don't hold state either but they aren't frozen on creation so there's slightly more risk. We could modify the app generator so that Hanami 2.3 apps `config.memoize_component_dirs = ["actions/", "views/"]`

We should also consider if there are other dirs we want to memoize as well.

I'd also like to add specs to ensure that only the dirs we specify are memoized. We can add a smoke test to ensure they, in fact, return the same objects, though dry-system's tests obviously handle this.

## Benchmark results
**No instance variables: 8+ times speed up**
```
Simple Hanami application loaded!

Configuration verification:
Memoized slice memoize_component_dirs: ["actions/", "views/"]
Normal slice memoize_component_dirs: []

Testing memoization behavior:
Memoized Actions - same instance: true
Normal Actions - same instance: false
Memoized Views - same instance: true
Normal Views - same instance: false

=== SIMPLE ACTION MICROBENCHMARK ===
Measuring minimal component resolution overhead:
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin23]
Warming up --------------------------------------
    Memoized Actions    72.552k i/100ms
      Normal Actions     8.596k i/100ms
Calculating -------------------------------------
    Memoized Actions    726.681k (± 0.4%) i/s    (1.38 μs/i) -      2.249M in   3.095088s
      Normal Actions     82.152k (± 4.2%) i/s   (12.17 μs/i) -    249.284k in   3.039788s

Comparison:
    Memoized Actions:   726680.9 i/s
      Normal Actions:    82152.3 i/s - 8.85x  slower


============================================================
=== SIMPLE VIEW MICROBENCHMARK ===
Measuring minimal component resolution overhead:
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin23]
Warming up --------------------------------------
      Memoized Views    72.760k i/100ms
        Normal Views     8.510k i/100ms
Calculating -------------------------------------
      Memoized Views    725.935k (± 0.7%) i/s    (1.38 μs/i) -      2.183M in   3.007053s
        Normal Views     84.915k (± 1.7%) i/s   (11.78 μs/i) -    255.300k in   3.007486s

Comparison:
      Memoized Views:   725934.8 i/s
        Normal Views:    84915.2 i/s - 8.55x  slower
```

4 instance variables (10+ times speed up):
```
Simple Hanami application loaded!

Configuration verification:
Memoized slice memoize_component_dirs: ["actions/", "views/"]
Normal slice memoize_component_dirs: []

Testing memoization behavior:
Memoized Actions - same instance: true
Normal Actions - same instance: false
Memoized Views - same instance: true
Normal Views - same instance: false

=== SIMPLE ACTION MICROBENCHMARK ===
Measuring minimal component resolution overhead:
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin23]
Warming up --------------------------------------
    Memoized Actions    71.374k i/100ms
      Normal Actions     6.811k i/100ms
Calculating -------------------------------------
    Memoized Actions    721.595k (± 1.1%) i/s    (1.39 μs/i) -      2.213M in   3.066641s
      Normal Actions     66.267k (± 4.6%) i/s   (15.09 μs/i) -    204.330k in   3.089977s

Comparison:
    Memoized Actions:   721595.4 i/s
      Normal Actions:    66266.7 i/s - 10.89x  slower


============================================================
=== SIMPLE VIEW MICROBENCHMARK ===
Measuring minimal component resolution overhead:
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [arm64-darwin23]
Warming up --------------------------------------
      Memoized Views    72.258k i/100ms
        Normal Views     6.899k i/100ms
Calculating -------------------------------------
      Memoized Views    718.231k (± 0.9%) i/s    (1.39 μs/i) -      2.168M in   3.018438s
        Normal Views     67.050k (± 2.4%) i/s   (14.91 μs/i) -    206.970k in   3.088560s

Comparison:
      Memoized Views:   718230.9 i/s
        Normal Views:    67049.7 i/s - 10.71x  slower

```


lib/hanami/slice.rb
```diff
@@ -913,6 +914,20 @@ def prepare_container_component_dirs
...
+
+        container.config.component_dirs.each do |component_dir|
+          component_dir.memoize = -> component {
+            relative_path = component.file_path.relative_path_from(root).to_s
+
+            config.memoize_component_dirs.any? do |dir_pattern|
+              relative_path.start_with?(dir_pattern)
+            end
+          }
+        end
```
timriley, line 928: I see an `.any?` nested inside an `.each` and I wonder if there's a way we could flatten these out somehow and improve the startup performance when fully booting an app...

timriley (commented):
This is fantastic, thank you for doing this, @cllns! 🥰

Looks like this will indeed give us the performance gains we thought it might. I'd also be interested in comparing (a) the difference in _startup_ performance when memoizing vs not, as well as (b) memory usage over time.

In terms of the config interface, I wonder if we want to orient this around _component keys_ rather than dirs. Even though _most_ opponents will come from files in a directory, not everything will, so I think it could be more powerful to orient this around keys, e.g. maybe a key prefix? or `File.fnmatch`-style pattern if we wanted to be really fancy?

cllns:
Agree we should implement using keys instead of dirs. That's more consistent. I wanted to get an early spike up in the quickest way possible and saw dry-system already had support for memoizing based on dirs.

I also think we may want to freeze Views as part of these changes? To ensure data is leaked between instances. @timriley Do you know why they're not frozen on creation?

lib/hanami/slice.rb
cllns: Found a different approach, one that uses a splat, which is quite fast

cllns:
Closing in favor of #1573

PR #1537, opened 2025-09-26
Remove injected `settings` from view context

### LINKED ISSUE #1536: `settings` in default view context should not be exposed to templates
See https://github.com/hanami/hanami/blob/d7bce6d7d6a01957a5bb8f5964495b9e7afd7fe3/lib/hanami/extensions/view/context.rb#L93-L99. The `settings` are not used by our standard view context, we we should either (a) remove it, or (b) make it private.

The current situation is confusing, because you can call `settings` in a template and get the settings object, even though you've done nothing to put it there.

This is not something we've documented so I think it's safe to change. Easy to communicate in a CHANGELOG, too.

baweaver:
Fixes #1536

### Summary

The settings attribute in view contexts was unexpectedly exposed to templates, causing confusion since users could call settings without explicitly making it available.

### Changes

• Move settings `attr_reader` from public to private section in view context
• Update tests to verify settings is not publicly accessible
• Add `CHANGELOG` entry

### Breaking Change

Templates can no longer access settings directly. This was undocumented behavior that should not have been relied upon.

### DIFF
--- a/lib/hanami/extensions/view/slice_configured_context.rb
+++ b/lib/hanami/extensions/view/slice_configured_context.rb
@@ -37,27 +37,20 @@ module Hanami
         #
         # This includes the following app components:
         #   - the configured inflector as `inflector`
-        #   - "settings" from the app container as `settings`
         #   - "routes" from the app container as `routes`
         #   - "assets" from the app container as `assets`
         def define_new
           inflector = slice.inflector
-          resolve_settings = method(:resolve_settings)
           resolve_routes = method(:resolve_routes)
           resolve_assets = method(:resolve_assets)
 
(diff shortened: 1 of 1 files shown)


timriley:
Thanks for picking this up, @baweaver! :D

After some reflection, I think it's an overreach to put settings in here in the first place. The settings object tends to contain very low-level info, and generally much more than should be accessible in the view layer.

If as user wants to access something from their settings in their view context, they can:

1. Create their own context class and inject settings as a dep
2. Or inject a higher-level dep that _uses_ the settings

So would you mind just removing it here, instead of making it private? I appreciate your help!

Issue #1538, opened 2025-09-27
Support user-provided methods in `hanami console`


alassek:
@jaredcwhite Suggested we provide an API for users to customize the helper methods in their REPL.

[Discourse conversion](https://discourse.hanamirb.org/t/add-helpers-to-irb-console/1304)

timriley:
@minaslater this is the one I mentioned to you in Slack, now conveniently a GitHub issue. Thanks @alassek 😄 

For anyone following here, I reckon an app/slice-level setting is the way to go:

config.console.include MyModule

Or something like that. We can finesse the exact experience as we go.

This change will need to span two repos: Hanami, for the setting itself, and CLI for the `console` command to use that config.

alassek:
@timriley I wrote this up because I have an implementation of it already! I figured we may as well make `engine` configurable in addition to accepting a cli flag.

```ruby
config.console.engine = :pry
config.console.include Mymodule
config.console.extensions # => [Mymodule]
```

PR #1540, opened 2025-09-28
Add `console` to configuration

alassek:
Relates to #1538

This creates a configuration object for users to define custom modules to include in the console context. Also supports a permanent change in engine setting.

### DIFF
--- a/lib/hanami/config.rb
+++ b/lib/hanami/config.rb
@@ -6,6 +6,7 @@ require "dry/configurable"
 require "dry/inflector"
 
 require_relative "constants"
+require_relative "config/console"
 
 module Hanami
   # Hanami app config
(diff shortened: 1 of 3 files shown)


timriley (approved):
Great stuff, thanks @alassek!

Left you a note about removing the conditionality, but otherwise let's get this in.

I presume you'll put together a hanami-cli PR to go along with this?

To answer some of the things you mentioned in https://github.com/hanami/hanami/issues/1538, I think it's completely fine for this to be simple and top-level right now. We can decide later on if we want to add more slice-wise sophistication to it, but my feeling is that probably won't be too necessary.

alassek:
@timriley 
> I presume you'll put together a hanami-cli PR to go along with this?

Yes, I have a corresponding PR ready to go, I wasn't sure if it was possible to open both at once since that depends on this.

PR #1542, opened 2025-10-01
Router: Implement resources / resource methods

afomera:
👋 Saw this on the roadmap (https://github.com/hanami/roadmap/issues/20) and wanted it, dare say I _needed_ it to be built. So here I am!

Bear with me, this will be a bit of a long one. 

This introduces two new DSL methods to the Router classes:

In RESTful applications there are a few primary actions we can count on. `index`, `show`, `create`, `new`, `edit`, `update`, `destroy`. For `resources` keys, we include all of them, for `resource` there's no `index` action so it's omitted. 

## `resources`

| Argument | Description |
|--------|--------|
| name | name of the route, string or symbols allowed Ex: "users", ":user" |
| options | allows specific routing options to be defined |
| block | allows nesting resources / resource blocks inside  |

### Options
| Name | Description |
|--------|--------|
| path | url path for a given resource - `resources :users` #=>  /users/* |
| only | Allows defining only specific actions `only: [:new, :create]` for example. Exclusive, cannot be paired with `except` |
| except | Opt-out of specific named actions. Exclusive, cannot be paired with `only` option.  |
| to | Override action path if you need specific namespaces `to: 'users.profiles'`  |
| as | Override route name `resources :avatar, as: :super_avatar  #=> path(:super_avatar)`  |

## `resource`

Generates the singular version of a given resource. Without the `index` action, and routes for editing will not have a specific id parameter passed to the routes `resource :profile` i.e generates

```
/profile
/profile/edit
/profile/new
```

| Argument | Description |
|--------|--------|
| name | name of the route, string or symbols allowed Ex: "users", ":user" |
| options | allows specific routing options to be defined |
| block | allows nesting resources / resource blocks inside  |

### Options
| Name | Description |
|--------|--------|
| path | url path for a given resource - `resources :users` #=>  /users/* |
| only | Allows defining only specific actions `only: [:new, :create]` for example. Exclusive, cannot be paired with `except` |
| except | Opt-out of specific named actions. Exclusive, cannot be paired with `only` option.  |
| to | Override action path if you need specific namespaces `to: 'users.profiles'`  |
| as | Override route name `resources :avatar, as: :super_avatar  #=> path(:super_avatar)`  |

## Example routes:

```rb
    resources :comments, path: "skeets"
    resources :users, except: [:new, :create] do
      resource :profile, only: [:show, :edit, :update], to: "users/profiles"
      resource :avatar, only: [:edit, :update], to: "users.avatars", as: :super_avatar
    end

    slice :storage, at: "/storage" do
      get "/:key", to: "serve", as: :serve_file
      resource :upload, only: [:new, :create], to: "uploads"
      resources :files, only: [:index, :show], to: "files"
    end
```

Generates this from `bundle exec hanami routes`

```
± be hanami routes
GET     /skeets                       comments.index                as :comments
GET     /skeets/new                   comments.new                  as :new_comment
POST    /skeets                       comments.create               as :comment
GET     /skeets/:id                   comments.show                 as :comment
GET     /skeets/:id/edit              comments.edit                 as :edit_comment
PATCH   /skeets/:id                   comments.update               as :comment
DELETE  /skeets/:id                   comments.destroy              as :comment
GET     /users                        users.index                   as :users
GET     /users/:id                    users.show                    as :user
GET     /users/:id/edit               users.edit                    as :edit_user
PATCH   /users/:id                    users.update                  as :user
DELETE  /users/:id                    users.destroy                 as :user
GET     /users/:user_id/profile       users.profiles.show           as :user_profile
GET     /users/:user_id/profile/edit  users.profiles.edit           as :edit_user_profile
PATCH   /users/:user_id/profile       users.profiles.update         as :user_profile
GET     /users/:user_id/avatar/edit   users.avatars.edit            as :edit_user_super_avatar
PATCH   /users/:user_id/avatar        users.avatars.update          as :user_super_avatar
GET     /storage/:key                 serve                         as :storage_serve_file
GET     /storage/upload/new           uploads.new                   as :storage_new_upload
POST    /storage/upload               uploads.create                as :storage_upload
GET     /storage/files                files.index                   as :storage_files
GET     /storage/files/:id            files.show                    as :storage_file
```

# TODO:

- [x] Specs for classes
- [x] Better code organization
- [x] Use inflector in Dry / Hanami already (needs passing through)

### DIFF
--- a/lib/hanami/slice/router.rb
+++ b/lib/hanami/slice/router.rb
@@ -146,13 +146,15 @@ module Hanami
 
         resource_builder.add_routes(self)
 
-        resource_scope(name, resource_builder.nested_scope_path, &block) if block
+        resource_scope(resource_builder, &block) if block
       end
 
       # @api private
-      def resource_scope(resource_name, path, &block)
-        @resource_scope.push(resource_name)
-        scope(path) { instance_eval(&block) }
+      def resource_scope(resource_builder, &block)
+        @resource_scope.push(resource_builder.name)
+        scope(resource_builder.scope_path, as: inflector.singularize(resource_builder.name)) do
+          instance_eval(&block)
+        end
       ensure
         @resource_scope.pop
       end
@@ -163,16 +165,18 @@ module Hanami
       class ResourceBuilder
         ROUTE_CONFIGURATIONS = {
           index: {method: :get},
-          new: {method: :get, path_suffix: "/new", name_prefix: "new_"},
+          new: {method: :get, path_suffix: "/new", name_prefix: "new"},
           create: {method: :post},
           show: {method: :get, path_suffix: "/:id"},
-          edit: {method: :get, path_suffix: "/:id/edit", name_prefix: "edit_"},
+          edit: {method: :get, path_suffix: "/:id/edit", name_prefix: "edit"},
           update: {method: :patch, path_suffix: "/:id"},
           destroy: {method: :delete, path_suffix: "/:id"}
         }.freeze
 
         PLURAL_ACTIONS = %i[index new create show edit update destroy].freeze
         SINGULAR_ACTIONS = %i[new create show edit update destroy].freeze
+
+        attr_reader :name
 
         def initialize(name:, type:, resource_scope:, options:, inflector:)
           @name = name
@@ -190,7 +194,7 @@ module Hanami
           end
         end
 
-        def nested_scope_path
+        def scope_path
           if plural?
             "#{@path}/:#{@inflector.singularize(@path.to_s)}_id"
           else
@@ -220,9 +224,11 @@ module Hanami
         end
 
         def add_route(router, action, route_config)
+          path = route_path(route_config[:path_suffix])
+
           router.public_send(
             route_config[:method],
-            route_path(route_config[:path_suffix]),
+            path,
             to: "#{key_path_base}#{CONTAINER_KEY_DELIMITER}#{action}",
             as: route_name(action, route_config[:name_prefix])
           )
@@ -234,9 +240,13 @@ module Hanami
           suffix.empty? ? base_path : "#{base_path}#{suffix}"
         end
 
-        def route_name(action, prefix)
-          base_name = action == :index ? @inflector.pluralize(route_name_base) : route_name_base
-          prefix.to_s.empty? ? base_name : "#{prefix}#{base_name}"
+        def normalize_suffix(suffix)
+          return "" if suffix.nil? || suffix.empty?
+
+          # For singular resources, remove :id from paths
+          return suffix.gsub("/:id", "") if singular?
+
+          suffix
         end
 
         def key_path_base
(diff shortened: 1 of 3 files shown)


lib/hanami/slice/router.rb
```diff
@@ -76,11 +76,278 @@ def slice(slice_name, at:, &blk)
...
+      # @example
+      #   resources :users
+      #   # Generates:
+      #   # GET    /users          users.index
+      #   # GET    /users/new      users.new
+      #   # POST   /users          users.create
+      #   # GET    /users/:id      users.show
+      #   # GET    /users/:id/edit users.edit
+      #   # PATCH  /users/:id      users.update
+      #   # PUT    /users/:id      users.update
```
alassek, line 97: `PATCH` and `PUT` should not be treated as equivalents. They are [entirely different concepts](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PATCH).

`PATCH` means to mutate an existing resource. It is not idempotent.

`PUT` is an upsert. You could be creating the resource new, or replacing what already exists. Importantly, it is an idempotent operation—the only write operation that is.

Personally, I usually go with something like `users.upsert` for this.

afomera: thanks @alassek, does the framework need to support that additional action by default? i was also tempted to remove the default PUT handling and just use patch like other frameworks but wanted to get feedback on it. 

what do you think?

alassek: I think `PUT` is special enough that it would make sense to me to drop it by default and use `PATCH` alone.

I'm curious to hear if @timriley has any vision for what a resource helper should do. I'm also not that familiar with other frameworks here—is everyone just copying what Rails did? I don't think we should copy their legacy baggage like the `PUT`/`PATCH` confusion, but I wonder how other frameworks differ.

afomera: i'm good with dropping PUT from here by default. i've not really used it in 12 years of rails, or sinatra. i'll get it prepped and updated when im back at a computer. 

https://laravel.com/docs/12.x/controllers#shallow-nesting laravel tends to do both to the update action, but i'm not pressed about it either way. 

https://hexdocs.pm/phoenix/routing.html#resources phoenix in elixir also does this as well. 

now i'm torn but ill error on the side of removal and we can always add it back if we change our mind

cllns: I personally think PATCH only is good for update actions. People can always manually define their own PUT route if they need it.

afomera: Removed the PUT route 👍

pixeltrix: Just to put (no pun intended) this in a historical context, the legacy framework used to use PUT as PATCH didn't exist at the time RESTful routing was added. Support was added back in 2011 for PATCH and for backwards compatibility reasons the two are effectively identical. I would go with not having the PUT route as there's no need for this backwards compatibility here.

timriley:
@afomera I'm done with my structural changes now :) All that's left is some testing for the _names_ that we generate for the routes. After that I think this is good to go, pending any further feedback from your side!

PR #439, opened 2016-01-01
Fix ghostly failing for middleware session test

davydovanton:
I often see that one test in middleware_test file ghostly failed with this error ([for example](https://travis-ci.org/lotus/lotus/jobs/99671905#L233)):
```
  1) Error:
Lotus::Middleware::when it's configured with sessions::with other middleware#test_0001_prepends sessions:
NoMethodError: undefined method `<' for nil:NilClass
    /home/travis/build/lotus/lotus/test/middleware_test.rb:92:in `block (4 levels) in <top (required)>'
```
I hope that this change will fix this ghostly bug.

What do you think?

PR #1549, opened 2025-11-01
Enable body parser middleware by default

timriley:
Make it easier for users to process multipart form posts and JSON request bodies by enabling the `Hanami::Middleware::BodyParser` middleware by default. This will give them the `request.params` that they expect, without requiring them to do any extra work.

Having this middleware enabled is safe because it will only attempt to parse bodies for the content types that we specify, which are these:

```ruby
DEFAULT_MIDDLEWARE_PARSERS = {
  form: ["multipart/form-data"],
  json: ["application/json", "application/vnd.api+json"]
}.freeze
```

The body parser middleware will skip handling all other requests.

For a user who wants to add e.g. JSON body parsing for different content types, they can configure another body parser themselves in their app config, e.g.

```ruby
# in the app
config.middleware.use(:body_parser, [json: ["application/scim+json"]])
```

I'll make sure this is documented for 2.3.

With this middleware being enabled by default, we need to make sure it behaves more respectfully with regard to rewindable bodies, which I've done here: https://github.com/hanami/hanami-router/pull/287.

Overall, this is a good improvement for the most common use cases, and IMO a good thing to ship with 2.3.

For 2.4, I want to move body parsing into Hanami Actions, and properly connect that with the action format config. See https://github.com/hanami/hanami-controller/issues/484 for more. This will still provide a reasonable upgrade path because: (a) for the common cases like JSON and multipart forms, it will require no user config, just like we have here, and (b) for the cases where a user has customised body parsing in 2.3, it will be a straightforward config adjustment for 2.4, nothing major. 

Fixes #439.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ end
 
 gem "hanami-utils", github: "hanami/hanami-utils", branch: "main"
 gem "hanami-db", github: "hanami/hanami-db", branch: "main"
-gem "hanami-router", github: "hanami/hanami-router", branch: "rewindable-input-in-body-parser"
+gem "hanami-router", github: "hanami/hanami-router", branch: "main"
 gem "hanami-controller", github: "hanami/hanami-controller", branch: "main"
 gem "hanami-cli", github: "hanami/hanami-cli", branch: "main"
 gem "hanami-view", github: "hanami/hanami-view", branch: "main"


katafrakt (approved):

alassek (approved):
This will be a nice quality-of-life improvement 👍

parndt:
So this can be 100% overridden in apps if needed?

parndt:
@timriley perfect!

timriley:
With this, I think I've covered everything, so I'll merge these in now :)

Issue #1535, opened 2025-09-16
Use blocks for logging to improve performance


p8:
While looking at the Vernier results of the TechEmpower json test I noticed a lot of time is spend in the logger, even though the log level is set to `:error`.

<img width="869" height="754" alt="Image" src="https://github.com/user-attachments/assets/0e913f10-5385-411f-8463-2cb2990ef272" />

https://gist.github.com/p8/5bafee12df11d572880ec35649d4fc46

For performance Rails (and Ruby's Logger) [use blocks](https://github.com/rails/rails/blob/0b5af58d2f0b8916f8211fa620c48da71d6da657/actionpack/lib/action_controller/log_subscriber.rb#L26-L44) to only execute code in the block if the severity is at least the log level.

I've created a PR to add this to DryLogger as well: https://github.com/dry-rb/dry-logger/pull/33
Would it be possible for Hanami to use blocks for logging as well?

PR #1550, opened 2025-11-01
Don't generate logging payload if severity is less than logger level

p8:
When logging a request only generate the JSON payload if the severity is equal to or higher than the logger level.

Fixes: #1535

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,6 @@ gem "dry-types"
 
 # For testing operation integrations
 gem "dry-operation", github: "dry-rb/dry-operation", branch: "main"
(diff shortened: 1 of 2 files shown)


timriley:
I've just rebased over main and force pushed to get the tests working again (main's Gemfile had a GitHub branch reference for a day or two that no longer exists now).

p8:
That sounds great!

timriley:
Thanks again for your help here, @p8! I've just released dry-logger v1.2.0 and updated this branch to require that version via Hanami's gemspec. With that done, this is ready to merge! 🎉

PR #1564, opened 2026-02-21
Add `views.default_template_engine` setting

katafrakt:
This setting is used by Hanami CLI to choose the template engine to use (unless another one is specified directly via a CLI option) - see https://github.com/hanami/hanami-cli/pull/389


timriley (approved):
Looks good! Please add a CHANGELOG entry before you merge :)

PR #1581, opened 2026-04-19
Use hanami-asset-js#44 to prepare assets

cllns:
Fix needed due to https://github.com/hanami/hanami-assets-js/pull/43

Testing CI here ~then I will put in repo-sync once we confirm it's correct~

Changed approach here to test the fix on a branch for https://github.com/hanami/hanami-assets-js/pull/44. This appears to validate that fix is correct.

This can be closed (and not merged!) once that is good.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,8 +59,6 @@ jobs:
           node-version: 20.x
       - name: Install dependencies
         run: npm install
-      - name: Build assets
-        run: npm run build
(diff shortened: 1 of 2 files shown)


timriley:
I re-ran the main branch build after merging https://github.com/hanami/hanami-assets-js/pull/44, and it looks like we're good. Thanks for figuring this out, @cllns! And I'm sorry for the disruption to your work 🙇🏼

PR #1582, opened 2026-04-21
Update references to hanami-action

cllns:
Renamed from hanami-controller in https://github.com/hanami/hanami-action/pull/507


timriley (approved):
Thanks @cllns! We'll need to do cli next. I'll get onto it when I can, but also very happy for you to do if you have time :)

PR #1585, opened 2026-04-29
Prevent Hanami::View integration from breaking DI

alassek:
This is a weird case that requires multiple things to line up. Because Hanami::Extensions::Action::InstanceMethods is _prepended_ into Hanami::Action, this makes the implementation of `initialize` interfere with `define_initialize` from Dry::AutoInject.

If any injected dependency happens to share the name of an argument in the prepended method, _and_ a dependency is injected in a parent action that does **not** define a dependency with that name, then it will cause the dependency to get nilled out.

This is particularly problematic for `:view` here because this is public interface. Our guides explicitly show injecting a custom view as a dependency named `view`, which will trigger this bug.

### DIFF
--- a/lib/hanami/extensions/action.rb
+++ b/lib/hanami/extensions/action.rb
@@ -67,15 +67,33 @@ module Hanami
         #   These dependencies are injected automatically so that a call to `.new` (with no
         #   arguments) returns a fully integrated action.
         #
+        #   WARNING: This is prepended into a class intended for dependency injection, so the
+        #   implementation of this method using `**kwargs` instead of named keyword arguments
+        #   is intentional. Adding named kwargs will break Dry::AutoInject.
+        #
+        #   @param view [Hanami::View]
+        #   @param view_context_class [Hanami::View::Context]
+        #   @param rack_monitor [Dry::Monitor::Rack::Middleware]
         #   @param routes [Hanami::Slice::RoutesHelper]
         #
         #   @api public
         #   @since 2.0.0
-        def initialize(view: nil, view_context_class: nil, rack_monitor: nil, routes: nil, **kwargs)
-          @view = view
-          @view_context_class = view_context_class
-          @routes = routes
-          @rack_monitor = rack_monitor
+        def initialize(**kwargs)
+          if kwargs.key?(:view)
+            @view = kwargs.delete(:view)
+          end
+
+          if kwargs.key?(:view_context_class)
+            @view_context_class = kwargs.delete(:view_context_class)
+          end
+
+          if kwargs.key?(:rack_monitor)
+            @rack_monitor = kwargs.delete(:rack_monitor)
+          end
+
+          if kwargs.key?(:routes)
+            @routes = kwargs.delete(:routes)
+          end
 
           super(**kwargs)
         end
--- a/spec/integration/action/dependency_injection_spec.rb
+++ b/spec/integration/action/dependency_injection_spec.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+RSpec.describe "App action / Dependency injection", :app_integration do
+  specify "Custom view injection works as intended" do
+    with_tmp_directory(Dir.mktmpdir) do
+      write "config/app.rb", <<~RUBY
+        require "hanami"
+
+        module TestApp
+          class App < Hanami::App
+          end
+        end
+      RUBY
+
+      write "app/logger", <<~RUBY
+        require "logger"
+
+        module TestApp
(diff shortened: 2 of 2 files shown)


spec/integration/action/dependency_injection_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      write "config/app.rb", <<~RUBY
+        require "hanami"
+
+        module TestApp
+          class App < Hanami::App
+          end
+        end
+      RUBY
+
+      write "app/logger", <<~RUBY
```
timriley, line 15:
```suggestion
      write "app/logger.rb", <<~RUBY
```

This file won't be picked up by auto-registration without the `.rb` extension. Maybe it isn't actually needed for this test? Our default logger gets registered as `"logger"` anyway.

spec/integration/action/dependency_injection_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+          end
+        end
+      RUBY
+
+      write "app/actions/users/show.rb", <<~RUBY
+        module TestApp
+          module Actions
+            module Users
+              class Show < TestApp::Action
+                include Deps[view: "views.users.show"]
```
timriley, line 53: Would it be more valuable to show this injecting a view that is _not_ the same as the view that would be automatically injected anyway as part of our action/view integration? Give it a slightly different name or something, and probably leave a comment above explaining why we're doing it for the purposes of this test?

lib/hanami/extensions/action.rb
```diff
@@ -67,15 +67,33 @@ module InstanceMethods
         #   These dependencies are injected automatically so that a call to `.new` (with no
         #   arguments) returns a fully integrated action.
         #
+        #   WARNING: This is prepended into a class intended for dependency injection, so the
+        #   implementation of this method using `**kwargs` instead of named keyword arguments
+        #   is intentional. Adding named kwargs will break Dry::AutoInject.
```
timriley, line 72: This is very helpful, thank you for adding it!

timriley (approved):
Thanks for doing this, @alassek!

I definitely approve of getting this fixed in whatever way possible, and this spot seems like the most sensible, given its the integration of these things _within Hanami_ that's affected.

I left a couple of notes for you, which I'd be happy for you to address in whatever way you see fit before merging :)

Lastly, a question — with all the time you've spent investigating this (thank you!), do you think there's some possible way in which we could (eventually) fix this in dry-auto_inject such that it renders these within hanami unnecessary? I really do want to make AutoInject as unobtrusive as possible, and while there might be practical limits to that, I always want to at least ask the question. Thank you again!

timriley:
Oh, could you please add a CHANGELOG entry while you're at this?

spec/integration/action/dependency_injection_spec.rb
alassek: The key that is being resolved is irrelevant, what triggers the bug is the fact that we're naming the dep `:view`. I can do that if you want, but I think it would obscure what we're actually testing.

alassek:
> @timriley do you think there's some possible way in which we could (eventually) fix this in dry-auto_inject such that it renders these within hanami unnecessary? I really do want to make AutoInject as unobtrusive as possible, and while there might be practical limits to that, I always want to at least ask the question.

I'm still investigating dry-auto_inject, but yes potentially there is a more permanent fix there. But my suspicion is that the way it detects pass-through implementations is actually wrong, and we would need to change the behavior. That has a substantially greater blast-radius, so I want to proceed with greater caution.

spec/integration/action/dependency_injection_spec.rb
timriley: Right, gotcha. I just want to make sure that what we're testing here is made clear to future readers (future-us) when we're assessing why this test exists.

So maybe just a comment above this `include Deps` explaining a bit more? What we're testing here is essentially "manual injection of a deep that's otherwise supplied automatically?"

alassek: Sure, that makes sense—if dry-auto_inject is patched in the future then this will always pass and thus be unclear what we're testing in the first place.

alassek:
I have a minimal reproduction case for dry-auto_inject and I'm much more confident I understand whats happening. We might just close this and fix it there. Prepend was actually a red-herring, it's the multiple injections combined with the fact that an injected dep matches an optional key further up the ancestor chain.

alassek:
Closing this in favor of dry-rb/dry-auto_inject#97

PR #1586, opened 2026-04-30
Raise informative error message on Rack #call method when no router is available

### LINKED ISSUE #1557: Improve error message when attempting to rackup a Hanami app without router in the bundle
See https://github.com/the-benchmarker/web-frameworks/pull/8873 for an example of the current confusing error.

sandbergja:
Closes #1557


katafrakt (approved):
Looks good, thanks!

PR #1587, opened 2026-05-20
Allow setting log level for DB queries

katafrakt:
This came up on Discord once, when someone asked how to make SQL logs log on `:debug` level, so they can be excluded in production. This adds `:log_level` setting for `:db`, allowing to set that. I don't imagine anyone would set it to something else than `:debug`, but it's theoretically possible now.

### DIFF
--- a/lib/hanami/config/db.rb
+++ b/lib/hanami/config/db.rb
@@ -15,7 +15,7 @@ module Hanami
 
       setting :import_from_parent, default: false
 
-      setting :sql_log_level, default: :info
+      setting :log_level, default: :info
 
       private
 
--- a/lib/hanami/config/db.rb
+++ b/lib/hanami/config/db.rb
@@ -15,6 +15,8 @@ module Hanami
 
       setting :import_from_parent, default: false
 
+      setting :sql_log_level, default: :info
+
       private
 
       def method_missing(name, *args, &block)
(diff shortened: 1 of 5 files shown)


lib/hanami/config/db.rb
```diff
@@ -15,6 +15,8 @@ class DB
 
       setting :import_from_parent, default: false
 
+      setting :sql_log_level, default: :info
```
timriley: I wonder if this should just be `:log_level`, and have it apply to the DB system logs _in general_?

I know that the majority of our users will use SQL, and we do make specific accommodations for that, but because Rom is a multi-adapter framework, I do prefer to keep things generic if we can.

What do you think, @katafrakt?

katafrakt: Good idea with renaming this to `:log_level`. But do we have any other logs than SQL queries coming from the DB? I didn't see any.

timriley (approved):
This looks good, thanks @katafrakt.

As part of this PR, or a follow-up, I wonder if we want to recalibrate our local vs production log levels in general.

For example, I think it would actually make sense to default the db.log_level to `:debug`, and then change the Hanami logger level in test/development to `:debug`, keeping it at `:info` in production.

This will mean we get SQL logging locally, but suppressed when the app is deployed, which I think is a better arrangement.

How does that sound to you? Would you be up for doing it?

timriley:
Opened https://github.com/hanami/hanami/issues/1588 as an issue/reminder for us about this. I think it'll be a good change as part of 3.0.

katafrakt:
I think setting the default to `:debug` makes a lot of sense, but I'm not sure we should swap that depending on the env. I think most people actually don't want SQL logs in production (?), then it will be naturally filtered out via minimal log level set to `:info` in production.

timriley:
@katafrakt yes, that’s what I meant, I think I just worded it clumsily 😊

- db log level stays at debug in all envs 
- the hanami app logger minimum level stays at info in production, but debug in test and development (this is the change)

katafrakt:
Good we are on the same page then 👍 
I will merge this one and follow up with the default log level change, as I think it is a separate change (especially changelog-wise, maybe even one that should be considered potentially breaking)

PR #1593, opened 2026-05-23
Add i18n helpers to Hanami::Action

timriley:
Make I18n helpers available within `Hanami::Action` when the i18n gem is bundled. The helpers delegate to the slice's own i18n backend via a new `#i18n` accessor on the action, automatically filled by our app-level action extension.

Expand relative translation keys (with a leading dot) against the action class's name, relative to its slice namespace. For example, `t(".title")` inside `Main::Actions::Posts::Show` resolves to `posts.show.title`.

Store this name in a new `config.i18n_key_base` setting on actions, set per-class by `SliceConfiguredAction`.

Since the i18n helper logic is now used by both the view and action layers, extract the common logic into a `Hanami::Helpers::I18nHelper::Methods`, which expects implementations to be provided for `#_i18n` (providing the i18n backend) and `#_resolve_i18n_key` (for relative key expansion). Implement each of these through dedicated helper modules for the action and view.

To ensure the action helper has no hard dependency on hanami-view, provide fallback HTML escaping via `CGI.escapeHTML` when hanami-view is not bundled.


katafrakt (approved):
> To ensure the action helper has no hard dependency on hanami-view, provide fallback HTML escaping via CGI.escapeHTML when hanami-view is not bundled.

nice! 👍

PR #1594, opened 2026-05-24
Keep base `Hanami::View::Context` unmodified by app integration

### LINKED ISSUE #1522: Fix loading of standalone view contexts
## Why

When loading a subclass of `Hanami::View::Context` from a gem (as a gem dependency to your own Hanami application), Hanami will layer on a bunch of Hanami application/slice specific behavior via the `Hanami::Extensions::View::Context` module. This all happens -- before any gem which uses the Hanami View gem in a standalone capacity -- is loaded within the Hanami application. This also means there isn't a great way for your own gem to detect and prevent this behavior.

## How

The reason this crops up is that the following code is applied before any standalone gem is loaded. Here's the file/code in question:

``` ruby
# hanami/extensions/view/context.rb
Hanami::View::Context.include(Hanami::Extensions::View::Context::ClassExtension)
```

## Notes

If a gem uses the Hanami View gem in a standalone capacity, I don't think Hanami should try to force application specific behavior upon the gem since it's surprising and unexpected behavior. I would also expect that you can render a standalone view within a Hanami Action too.

timriley:
Previously, `Hanami::Extensions::View::Context::ClassExtension::InstanceMethods` was prepended onto `Hanami::View::Context` itself at gem load time. This meant that any subclass, even one intending to serve in a standalone capacity, would receive all our app-integrated behavior: a specific `#initialize` and a whole bunch of instance methods. There was no way to opt out.

To address this, move the prepend into `ClassMethods#configure_for_slice` so it only applies to classes living legitimately within a Hanami app. This keeps the base class clean of any app-specific behavior.

A few implementation notes on how we support this change:

- Avoid a double-prepend in slice-subclass chains (e.g. `Main::Views::Context < TestApp::Views::Context`), by marking integrated classes via an `IntegratedContext` module. Skip the prepend on classes that already have that marker.
- Change `Hanami::SliceConfigurable` so it early returns rather than raises if there is no Hanami app defined. This is what allows a gem to `require "hanami"` (which adds `SliceConfigurable` to `Hanami::View::Context`) and still go on to subclass `Hanami::View::Context` even before an actual Hanami app is defined.
- Split out a small `Initializer` module from `InstanceMethods`, containing the `#initialize` method, which is the only method really requiring a prepend. `InstanceMethods` is now now included rather than prepended, so user subclasses can override its methods naturally.

**NOTE:** This change addresses `Hanami::View::Context` _only_. This has a small enough blast radius that I feel comfortable squeezing it into our impending release. I want to revisit our extensions system more broadly and make things nicer for third-party extensions and even more standalone usage of individual components, but that is a much larger change, and something that will need to wait until a later Hanami release.

Fixes #1522

### DIFF
--- a/lib/hanami/extensions/view/context.rb
+++ b/lib/hanami/extensions/view/context.rb
@@ -65,7 +65,7 @@ module Hanami
           # A marker indicating the context class has had Hanami app-specific behaviour mixed in.
           #
           # Checked by {ClassMethods#configure_for_slice} so a chain of slice-configured subclasses
-          # does not re-apply the integration. This matters for the prepended `#initialize`, which
+          # does not re-apply the integration. This matters for the prepended {Initializer}, which
           # would otherwise fire once per slice in the chain instead of once per instance.
           #
           # @api private
@@ -77,7 +77,8 @@ module Hanami
             # @api private
             def configure_for_slice(slice)
               unless is_a?(IntegratedContext)
-                prepend InstanceMethods
+                prepend Initializer
+                include InstanceMethods
                 extend IntegratedContext
               end
 
@@ -85,21 +86,14 @@ module Hanami
             end
           end
 
-          # @api public
-          # @since 2.1.0
-          module InstanceMethods
-            # Returns the app's inflector.
-            #
-            # @return [Dry::Inflector] the inflector
-            #
-            # @api public
-            # @since 2.1.0
-            attr_reader :inflector
-
+          # Captures the integration kwargs into ivars. Prepended so it runs before any
+          # user-defined initializer in the subclass.
+          #
+          # @api private
+          module Initializer
             # @see SliceConfiguredContext#define_new
             #
             # @api private
-            # @since 2.1.0
             def initialize(
               inflector: nil,
               routes: nil,
@@ -129,6 +123,21 @@ module Hanami
               # state across distinct view renderings.
               @content_for = source.instance_variable_get(:@content_for).dup
             end
+          end
+
+          # Readers exposing the ivars captured by {Initializer}. Included rather than prepended
+          # so user subclasses can override them naturally.
+          #
+          # @api public
+          # @since 2.1.0
+          module InstanceMethods
+            # Returns the app's inflector.
+            #
+            # @return [Dry::Inflector] the inflector
+            #
+            # @api public
+            # @since 2.1.0
+            attr_reader :inflector
 
             # Returns the app's assets.
             #
--- a/lib/hanami/extensions/view/context.rb
+++ b/lib/hanami/extensions/view/context.rb
@@ -16,7 +16,6 @@ module Hanami
           # a class named `Views::Context` within the slice's namespace.
           #
           # @api private
-          # @since 2.1.0
           def context_class(slice)
             views_namespace = views_namespace(slice)
 
@@ -31,8 +30,15 @@ module Hanami
 
           private
 
-          # @api private
-          # @since 2.1.0
+          def views_namespace(slice)
+            # TODO: this could be moved into the top-level Extensions::View
+            if slice.namespace.const_defined?(:Views)
+              slice.namespace.const_get(:Views)
+            else
+              slice.namespace.const_set(:Views, Module.new)
+            end
+          end
+
           def context_superclass(slice)
             return Hanami::View::Context if Hanami.app.equal?(slice)
 
@@ -46,36 +52,35 @@ module Hanami
               Hanami::View::Context
             end
           end
-
-          # @api private
-          # @since 2.1.0
-          def views_namespace(slice)
-            # TODO: this could be moved into the top-level Extensions::View
-            if slice.namespace.const_defined?(:Views)
-              slice.namespace.const_get(:Views)
-            else
-              slice.namespace.const_set(:Views, Module.new)
-            end
-          end
         end
 
-        # @api private
-        # @since 2.1.0
         module ClassExtension
           def self.included(context_class)
             super
 
             context_class.extend(Hanami::SliceConfigurable)
             context_class.extend(ClassMethods)
-            context_class.prepend(InstanceMethods)
+          end
+
+          # A marker indicating the context class has had Hanami app-specific behaviour mixed in.
+          #
+          # Checked by {ClassMethods#configure_for_slice} so a chain of slice-configured subclasses
+          # does not re-apply the integration. This matters for the prepended `#initialize`, which
+          # would otherwise fire once per slice in the chain instead of once per instance.
+          #
+          # @api private
+          module IntegratedContext
           end
 
           # @api private
-          # @since 2.1.0
           module ClassMethods
             # @api private
-            # @since 2.1.0
             def configure_for_slice(slice)
+              unless is_a?(IntegratedContext)
+                prepend InstanceMethods
+                extend IntegratedContext
+              end
+
               extend SliceConfiguredContext.new(slice)
             end
           end
@@ -115,7 +120,6 @@ module Hanami
             end
 
             # @api private
-            # @since 2.1.0
             def initialize_copy(source)
               # The standard implementation of initialize_copy will make shallow copies of all
               # instance variables from the source. This is fine for most of our ivars.
--- a/lib/hanami/slice_configurable.rb
+++ b/lib/hanami/slice_configurable.rb
@@ -31,11 +31,9 @@ module Hanami
 
         inherited_mod = Module.new do
           define_method(:inherited) do |subclass|
-            unless Hanami.app?
-              raise ComponentLoadError, "Class #{klass} must be defined within an Hanami app"
-            end
-
             super(subclass)
+
+            return unless Hanami.app?
 
             subclass.instance_variable_set(:@configured_for_slices, configured_for_slices.dup)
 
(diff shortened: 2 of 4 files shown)


timriley:
Hi @bkuhlmann, would you mind giving this branch a test and seeing if it unblocks the standalone `Hanami::View::Context` usage you were looking to do?

lib/hanami/extensions/view/context.rb
```diff
@@ -31,8 +30,15 @@ def context_class(slice)
 
           private
 
-          # @api private
-          # @since 2.1.0
+          def views_namespace(slice)
```
timriley, line 33: I just moved this method higher up in the file. No change to its actual code.

bkuhlmann:
Yeah, I gave this a test by doing the following:

``` bash
git clone https://github.com/hanami/hanami
git switch extend-in-app-view-contexts-only
cd hanami
gemsmith --install
```

I then added the following actions as a quick test:

``` ruby
require "milestoner"

module Terminus
  module Actions
    module Versions
      class Index < Action
        include Initable[view: Milestoner::Views::Milestones::Index.new]

        def handle(*, response) = response.render view, tags: []
      end
    end
  end
end
```

Upon rebooting the app, I first started seeing Dry Schema and Dry Validation missing constant errors from my Hanami app.rb. Upon fixing those, this caused the following error:

``` 
undefined method 'content_security_policy' for an instance of Hanami::Config::NullConfig (NoMethodError)
```

None of this happens when reverting back to the current version of Hanami (i.e. `gem pristine hanami`). It seems your patch, by moving up higher, causes a slew of dependencies to not be available.

timriley:
@bkuhlmann thanks for testing this out. I think you may need to try using the main branches of all the Hanami gems. They've all moved on since the last 2.x release. Along with this you'll need to rename the "hanami-controller" entry in the Gemfile to "hanami-action". And add dry-validation directly to your Gamfile (which I think you already figured out).

If that's not possible for you to do, I can put together a demo project later today.

bkuhlmann:
Yeah, I ended up cloning and updating my `Gemfile` as such (this included renaming the Hanami Controller project as Hanami Action to match the Gem specification):

``` ruby
gem "hanami", path: "~/Scratch/hanami"
gem "hanami-action", path: "~/Scratch/hanami-action"
gem "hanami-assets", path: "~/Scratch/hanami-assets"
gem "hanami-db", path: "~/Scratch/hanami-db"
gem "hanami-router", path: "~/Scratch/hanami-router"
gem "hanami-validations", path: "~/Scratch/hanami-validations"
gem "hanami-view", path: "~/Scratch/hanami-view"
```

I use [Cogger](https://alchemists.io/projects/cogger) as my logger for the entire application but the new Hanami logging capabilities is causing several issues and I haven't fully unpacked all the changes in the new release as of yet.

Putting a pause in this for today and will resume tomorrow. Instead of testing all of this against my existing stack, I'll spin up a new Hanami project which should reduce the amount of swirl trying to get this hybrid Hanami app up and running.

timriley:
Thanks for your persistence, @bkuhlmann! I can also share you a simple cogger setup to help unblock you. Should be able to share that within the next 6-12 hours :)

timriley:
I've just pushed up an app to demonstrate/test the changes in this PR: https://github.com/timriley/demo-hanami-pr-1594

You can review the changes commit-by-commit:

- [Use head gems](https://github.com/timriley/demo-hanami-pr-1594/commit/aeb6e3f68ef670d1fb6ccd5bd5a021717fabd63f) uses the main branch and adjusts other `Gemfile` entries as required
- [Demonstrate integrated vs standalone contexts](https://github.com/timriley/demo-hanami-pr-1594/commit/aca1ab50f1b69d520dc474bc699223da999c1078) shows me setting up a pretend third-party gem inheriting from `Hanami::View::Context`, and demonstrates that class being left unchanged even after the app loads
- [Log with cogger](https://github.com/timriley/demo-hanami-pr-1594/commit/8a380aa5b3fe11b32ccfcd6df26391e784b96712) introduces a working Cogger logger, which is much simpler with our improved logger support
- [Demonstrate SQL logging with Cogger](https://github.com/timriley/demo-hanami-pr-1594/commit/db2fd4d703f1bdc013cad5848233aa9ebc63d5f9) adds a table and some SQL calls to demonstrate SQL logging working with Cogger with zero patches/config required

### Setup

To try this, `bundle`, `bin/hanami db prepare` (this is using SQLite only), and then run `bin/hanami dev`.

### Testing standalone view contexts

Once the dev server is up, you can hit http://localhost:2300/view_context and see the difference in view context methods:

CleanShot 2026-05-25 at 23 01 05@2x

Check out `app/views/view_context/show.rb` for the two different classes we're pulling those methods from.

### Testing Hanami's new logging capabilities with Cogger

While you're there, you can also hit http://localhost:2300/sql_logging a few times and see Cogger working beautifully with our new logging setup:

```
🟢 [test_app_1594] [sql] [query=INSERT INTO `posts` (`title`) VALUES ('Test post 1')] [db=sqlite] [elapsed=1] [elapsed_unit=ms]
🟢 [test_app_1594] [sql] [query=SELECT `posts`.`id`, `posts`.`title` FROM `posts` ORDER BY `posts`.`id`] [db=sqlite] [elapsed=0] [elapsed_unit=ms]
🟢 [test_app_1594] [rack] [verb=GET] [status=200] [ip=127.0.0.1] [path=/sql_logging] [length=-] [params={}] [elapsed=25ms]
🟢 [test_app_1594] [rack] [verb=GET] [status=304] [ip=127.0.0.1] [path=/assets/app.js] [length=-] [params={}] [elapsed=611µs]
🟢 [test_app_1594] [rack] [verb=GET] [status=304] [ip=127.0.0.1] [path=/assets/app.css] [length=-] [params={}] [elapsed=528µs]
🟢 [test_app_1594] [sql] [query=INSERT INTO `posts` (`title`) VALUES ('Test post 2')] [db=sqlite] [elapsed=2] [elapsed_unit=ms]
🟢 [test_app_1594] [sql] [query=SELECT `posts`.`id`, `posts`.`title` FROM `posts` ORDER BY `posts`.`id`] [db=sqlite] [elapsed=0] [elapsed_unit=ms]
🟢 [test_app_1594] [rack] [verb=GET] [status=200] [ip=127.0.0.1] [path=/sql_logging] [length=-] [params={}] [elapsed=6ms]
🟢 [test_app_1594] [sql] [query=INSERT INTO `posts` (`title`) VALUES ('Test post 3')] [db=sqlite] [elapsed=1] [elapsed_unit=ms]
🟢 [test_app_1594] [sql] [query=SELECT `posts`.`id`, `posts`.`title` FROM `posts` ORDER BY `posts`.`id`] [db=sqlite] [elapsed=0] [elapsed_unit=ms]
🟢 [test_app_1594] [rack] [verb=GET] [status=200] [ip=127.0.0.1] [path=/sql_logging] [length=-] [params={}] [elapsed=5ms]
```

^ I don't know why github is wrapping those, but they're each just one line in the actual logs.

To make Cogger look close to how I've seen you configure it, I did have to create a custom formatter, but I'm sure you can think of better approaches to this, as the maintainer of that gem.

The upshot to Hanami's logging changes is that I would hope most of the words in https://alchemists.io/articles/hanami_logging can go away after we make our next release :)

bkuhlmann:
OK, I tried your demo and verified the fix works. However, it doesn't work when using [Milestoner](https://alchemists.io/projects/milestoner) because Milestoner leverages the *entire* view layer (i.e. custom views, contexts, parts, and scopes). I've created a [Demo](https://github.com/bkuhlmann/demo) app to illustrate. You only need to run `bin/setup` and then `hanami server`. Each commit explains *what* the changes are and *why* they were made.

When you launch the server and visit `http://localhost:2300`, you'll get the following error in your browser:

> undefined local variable or method 'project_uri_home' for an instance of Hanami::View::Scope

The reason this is happening is:

- Milestoner is using a custom scope as defined via the [Index](https://github.com/bkuhlmann/milestoner/blob/main/lib/milestoner/views/milestones/index.rb#L13) view. This includes a custom context and parts. This is also a slight red herring because the issue remains with the context object.
- The Context [delegates](https://github.com/bkuhlmann/milestoner/blob/main/lib/milestoner/views/context.rb#L18-L35) to Milestoner's settings.
- The corresponding [index.html.erb](https://github.com/bkuhlmann/milestoner/blob/main/lib/milestoner/templates/milestones/index.html.erb#L3) template then fails to make use of the *delegated* `project_uri_home ` method (and other delegated methods) because the context isn't providing them.

If I directly modify the source of the currently installed Milestoner gem version and remove the messaging of the delegated methods from the `index.html.erb` template, then I can get the app to render properly.

---

Yeah, I'll update my Hanami Logging article when the new Hanami version is released as all of my articles are *living* documents as I constantly update them as I evolve in order to refine my learnings. I reference my articles constantly via [Pennyworth](https://alchemists.io/projects/pennyworth) macros.

timriley:
Thanks for trying this and sharing all that detail, @bkuhlmann. Given this will need a change to how we integrate every aspect of the view layer, this is unfortunately too much to squeeze into a release that I'm hoping will go out in the next couple of weeks.

This has been a really helpful experiment, though, and I appreciate the feedback you've shared along the way. This is going to give us a head start for working on teasing things apart and improving standalone usage in a follow-up release. I'll make sure I start with the view layer then.

PR #1595, opened 2026-05-26
Change default log level for database to debug

### LINKED ISSUE #1588: Switch default development/test log level to debug
Make it so we don't log SQL queries by default in production. See discussion at https://github.com/hanami/hanami/pull/1587#pullrequestreview-4341286528.

katafrakt:
closes #1588

### DIFF
--- a/spec/integration/db/logging_spec.rb
+++ b/spec/integration/db/logging_spec.rb
@@ -74,7 +74,7 @@ RSpec.describe "DB / Logging", :app_integration do
       log_lines = logger_stream.read.split("\n")
 
       expect(log_lines.length).to eq 1
-      expect(strip_ansi(log_lines.first)).to match(/\[test_app\] \[INFO\] \[.*\] SQL sqlite \d+(\.\d+)?ms SELECT `posts`.`title` FROM `posts` ORDER BY `posts`.`id`/)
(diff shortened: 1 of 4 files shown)


spec/integration/db/logging_spec.rb
```diff
@@ -293,26 +293,87 @@ class Posts < Hanami::DB::Relation
...
+            create_table :authors do
+              primary_key :id
+            end
+          end
+        end
+        migration.apply(gateway, :up)
+        gateway.connection.execute("INSERT INTO posts (title) VALUES ('Together breakfast')")
+
+        relation = Hanami.app["relations.posts"]
+        expect(relation.select(:title).to_a).to eq [{title: "Together breakfast"}]
```
katafrakt, line 360: How do we feel about these large chunks of setup repeated in most examples? I would personally extract and parameterize them, but that might be my Elixir habits speaking.

timriley: I feel two things about such setup:

- OT1H, it _is_ nice for it to be local and let the test "tell its own story" in a very clear and immediate way
- OTOH, it can definitely get repetitive, especially when there are a lot of similar-ish tests within one file.

I'm open to extracting and parameterizing the basic repeated setup across several tests, so long as the test themselves continue to read well, i.e. you don't have to go jumping too much around the file to understand them.

It's all a balance, and I'd be happy to see you give it a go :)

timriley (approved):
Thanks @katafrakt!

spec/integration/db/logging_spec.rb
katafrakt: I'll think about a follow-up. Let's not block the release with that.

PR #1604, opened 2026-06-19
Add `Slice.with_slices` method

timriley:
This returns the slice itself and all nested slices. For the case of calling this on `Hanami.app`, it returns every slice in the app, including the app itself.

This encapsulates manual slice iteration code that had been spreading across all of our generated test setup files, and will allow those files to be simplified to a single call to `Hanami.app.with_slices`.


timriley:
Merging this now so I can use it from hanami-rspec and hanami-minitest.

PR #1609, opened 2026-06-30
Fix template resolution for mailers inheriting from slice-level base class

timriley:
A concrete mailer inheriting from an app-level base class (e.g. `MyApp::Mailers::Welcome < MyApp::Mailer`) would be configured with the wrong template (`"mailer"`) and raise a TemplateNotFoundError when rendering.

This was due to the `SliceConfiguredMailer` only inferring a template name once per slice, rather than once per class. Fix this by adding a `define_inherited` hook to `SliceConfiguredMailer` that configures the template name for every subclass, mirroring how we do it in `SliceConfiguredView`.

Also update the `write_welcome_mailer` spec helper to mirror a generated app (concrete mailers inheriting from a base `Mailer`), so we have a range of tests executing this realistic arrangement.

### DIFF
--- a/lib/hanami/extensions/mailer/slice_configured_mailer.rb
+++ b/lib/hanami/extensions/mailer/slice_configured_mailer.rb
@@ -25,6 +25,7 @@ module Hanami
         def extended(mailer_class)
           configure_mailer(mailer_class)
           define_new
+          define_inherited
         end
 
         def inspect
@@ -41,6 +42,19 @@ module Hanami
               delivery_method: kwargs.fetch(:delivery_method) { resolve_delivery_method.() },
               **kwargs
             )
+          end
+        end
+
+        # Reconfigures the template name for each subclass.
+        def define_inherited
+          template_name = method(:template_name)
+
+          define_method(:inherited) do |subclass|
+            super(subclass)
+
+            if (template = template_name.(subclass))
+              subclass.config.template = template
+            end
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/hanami/extensions/mailer/slice_configured_mailer.rb
```diff
@@ -44,6 +45,19 @@ def define_new
...
+        # Reconfigures the template name for each subclass.
+        def define_inherited
+          template_name = method(:template_name)
+
+          define_method(:inherited) do |subclass|
+            super(subclass)
+
+            if (template = template_name.(subclass))
+              subclass.config.template = template
+            end
```
katafrakt, line 57: @timriley should this be wrapped in a check if `hanami-view` is present? I have a test failure here in Palaver (which infamously lacks Hanami View):

```
     NoMethodError:
       undefined method 'template=' for an instance of Dry::Configurable::Config
     # ./vendor/bundle/ruby/4.0.0/gems/dry-configurable-1.4.0/lib/dry/configurable/config.rb:237:in 'Dry::Configurable::Config#method_missing'
     # ./vendor/bundle/ruby/4.0.0/gems/hanami-3.0.0/lib/hanami/extensions/mailer/slice_configured_mailer.rb:56:in 'block in define_inherited'
     # ./vendor/bundle/ruby/4.0.0/gems/dry-core-1.2.0/lib/dry/core/class_attributes.rb:95:in 'block (2 levels) in defines'
```

timriley: Yes, it should! I can patch it up today and make a release.

katafrakt: i created https://github.com/hanami/hanami/pull/1611, but TBH not sure if we need to release right away. It's really an edge case, so maybe let's wait few days, in case we have more fixes to release?

PR #1613, opened 2026-07-05
Resolve slice-owned config via the slice, not the global app

parndt:
`Providers::Inflector` registered `Hanami.app.inflector`, and `Providers::Logger` built its logger from `Hanami.app.config`, even though every provider source already knows the slice it is registered on.

When the app registers these providers (as it does automatically during prepare), the slice and the app are one and the same, so this changes nothing there. But provider sources can be registered on any slice, and a slice registering its own logger or inflector provider would get components built from the app's config, with its own config silently ignored:

```ruby
# slices/admin/config/slice.rb
module Admin
  class Slice < Hanami::Slice
    config.shared_app_component_keys -= ["logger"]
    config.logger.stream = "log/admin.log" # was ignored
  end
end

# slices/admin/config/providers/logger.rb
Admin::Slice.configure_provider(:logger)
```

Resolves via the provider's own slice instead, so these providers honour the slice they belong to. The behaviour is unchanged for app-registered providers.


timriley:
Thanks @parndt!

timriley (approved):

PR #1614, opened 2026-07-05
Support preparing, booting, and serving a slice without an app

parndt:
An `Hanami::Slice` can now be defined, prepared, booted, and served as a Rack app with no `Hanami::App` defined in the process. Behaviour is unchanged when an app is present.

This is built on the concept of a slice's host: the app, when one is defined in the process, or the slice itself, when running standalone. The host provides shared facilities to the slices it hosts: config, base providers, shared components, and the routes helper. A slice that is its own host provisions for itself exactly what an app would otherwise provide, following the principle that the app is just the outermost slice:

- config: built fresh via `Hanami::Config.new`, as `App.inherited` does
- root: falls back to `Dir.pwd`, the same default an app gets
- providers: registers notifications, inflector, logger (with `UniversalLogger` decoration), and rack (when bundled) itself, mirroring `App#prepare_app_providers`, instead of importing them from an app
- `i18n/mailers` providers: always registered standalone (nothing to share from)

Slice-configurable classes (actions, views) previously raised `ComponentLoadError` when subclassed with no app defined. When no app is present, the owning slice is now resolved by scanning `Hanami::Slice` subclasses (most specific namespace first); only classes still reachable via their constant are considered, and subclassing outside any slice namespace is a no-op. Actions resolve routes and rack.monitor from their own slice, view contexts build on `Hanami::View::Context` directly, and the assets middleware receives the slice's own config rather than defaulting to the app's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ A complete Hanami app is composed of multiple gems. For a complete overview of c
 
 ### Added
 
+- Slices can be defined, prepared, booted, and served as Rack apps without an `Hanami::App` defined in the process. (@parndt in #1614)
+
 ### Changed
 
 ### Deprecated
--- a/lib/hanami/app.rb
+++ b/lib/hanami/app.rb
@@ -117,6 +117,9 @@ module Hanami
       def prepare_container_component_dirs; end
       def prepare_container_imports; end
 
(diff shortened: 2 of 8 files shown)


parndt:
@timriley maybe we can think of a better way to do `prepare_host_providers`?

PR #1617, opened 2026-07-11
Support JRuby in CI + code

parndt:
This was started in other pull requests e.g. #1524 but we lost traction.
I'm opening this to restart the conversation and maybe have a solution.

This goes with hanami/hanami-cli#404

Other relevant work: hanami/hanami-cli#434


lib/hanami/providers/db.rb
```diff
@@ -230,13 +230,41 @@ def prepare_gateways
           fetch_or_store(gw_config.cache_keys) {
             ROM::Gateway.setup(
               gw_config.adapter_name,
-              gw_config.database_url,
+              database_url_for_jdbc(gw_config.database_url),
```
katafrakt: Do we need this translation? I think JRuby users should use JDBC URLs already, but perhaps worth discussing.

spec/support/app_integration.rb
```diff
@@ -90,8 +90,12 @@ def generate_assets_config
   let(:app_modules) { %i[TestApp Admin Main Search] }
 end
 
+def slice_autoloaders(klass)
+  klass.subclasses.flat_map { |subclass| [subclass.autoloader] + slice_autoloaders(subclass) }
+end
```
katafrakt, line 95: Is this an equivalent of `ObjectSpace.each_object`? I feel dumb, because we have been discussing it few months ago and did not come up with this. Usually we enabled `ObjectSpace` in JRuby tests to get them pass.

parndt: Yeah, it's an attempt to work around that

lib/hanami/providers/db.rb
parndt: This is a good point! Less code would be better if we can expect it to be specified

spec/support/app_integration.rb
katafrakt: Ah, I misread it initially. I thought this iterates on subclasses of `Zeitwerk::Loader`, but we are iterating slices here. Yeah, it's pretty smart. Although enabling `ObjectSpace` is not a problem IMO too. I'm okay with both approaches.

katafrakt:
I think with https://github.com/hanami/hanami-cli/pull/434 merged it would be great to pick this up again. 

The only additional comment I get is that maybe the database-gem-loading code could be extracted to `hanami-db`, as it's kind of similar here and in Hanami CLI. Do you think it would be a good idea @parndt @timriley?

PR #1620, opened 2026-08-14
Use exposures over params for view auto-rendering args

### LINKED ISSUE #1619: Request parameters override exposures
## Describe the bug

In our app we do lots of things like:
```ruby
before def set_user_id(request, response)
  response.exposures[:user_id] = request.session[:user_id]
end
```

And then in our views:
```ruby
expose :user, decorate: false do |user_id:|
  user_repo.by_pk(user_id).one
end
```

So imagine my surprise when Codex alerted me that I can just do `/profile?user_id=1234` to see any user's profile!

## To Reproduce

```ruby
require "hanami"
require "hanami/view"

module Test
  class App < Hanami::App
  end

  module Views
    class Rig < Hanami::View
      expose :foo, decorate: false, default: nil

      def call(format: config.default_format, context: config.default_context, layout: config.layout, **input)
        input.inspect
      end
    end
  end

  module Actions
    class Rig < Hanami::Action
      before def set_foo_exposure(request, response) = response.exposures[:foo] = 1

      def view
        Test::Views::Rig.new
      end
    end
  end
end

instance = Test::Actions::Rig.new

puts instance.call({}).body.join
# => {foo: 1}
puts instance.call({foo: 1234}).body.join
# => {foo: 1234}
```

## Expected behavior

I would expect inputs to views defined in exposures to take precedence over client-supplied data. Trust the server over the user! I understand that is probably a pretty severe change in behaviour, but it's also not documented anywhere as far as I can see?

The only workarounds I see are:
1. Monkey-patch `Hanami::Action::Response` to override `#render`
2. Require **every** action to have a contract so that only permitted request parameters are sent to views.

## My environment

- Hanami version: 2.3.2
- Ruby version: 4.0.5
- OS: MacOS

timriley:
When an action auto-renders its view, prefer response exposures over request params for the view's input.

When a response exposure is set, this is an explicit intention set in code. It should not be overridden by client-provided params, which can lead to unexpected or unwanted behavior.

Fixes #1619.

I'll release this soon as a 3.0 patch release.

For 3.1, I'll propose that we remove the params passthrough to the view altogether. It's better app authors to be explicit about this, I think. To smooth the transition we'll keep a module that preserves the 3.0 behaviour but adds a warning when a param is passed through, so it's easy to track down and fix any issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ A complete Hanami app is composed of multiple gems. For a complete overview of c
 
 ### Fixed
 
+- Prefer response exposures over request params when preparing input for the view as part of an action auto-rendering. (@timriley in #1620)
 - Don't attempt to inherit mailer template when `hanami-view` if not bundled. (@katafrakt in #1611)
 - Return `false` from `Hanami::Slice.app?`, rather than raising `Hanami::AppLoadError`, when no app is defined. (@parndt in #1612)
 - Build `logger` and `inflector` providers registered on a slice from the slice's own config, rather than the app's. (@parndt in #1613)
--- a/lib/hanami/extensions/action.rb
+++ b/lib/hanami/extensions/action.rb
@@ -37,6 +37,12 @@ module Hanami
       #
       # @since 2.0.0
       module InstanceMethods
+        # Options accepted by `Hanami::View#call` itself, which must never be given from request
+        # params.
+        #
+        # @api private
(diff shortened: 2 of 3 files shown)


timriley:
@madleech Thanks for reporting #1619! Would this fix work for you?

madleech:
Yes that appears to work. Thanks.

timriley:
Thanks for checking, @madleech. 🙇🏼 I'll get this into a patch release sometime soon.

PR #1621, opened 2026-08-14
Build the database URL from component ENV vars

aaronmallen:
This change makes Hanami build a database URL from its component environment variables. If `DATABASE_URL` is set, behavior stays the same. If it is not set but the component variables are, Hanami builds the URL from them.

Valid components are:

- `DATABASE_ADAPTER` (such as postgres, mysql2, sqlite)(**required**)
- `DATABASE_HOST` (**defaults to "localhost"**)
- `DATABASE_NAME`
- `DATABASE_PASSWORD`
- `DATABASE_PORT` (**defaults to the appropriate port for the given adapter**)
- `DATABASE_USER`

## Why

Many environments, such as docker, expect their own connection variables (`POSTGRES_PASSWORD`, `POSTGRES_USER`, and so on). Docker compose handles this well through its variable syntax:

```yaml
x-database-password: &database-password ${DATABASE_PASSWORD:-password}
x-database-user: &database-user ${DATABASE_USER:-postgres}

services:
  postgres:
    # ...
    environment:
        POSTGRES_PASSWORD: *database-password
        POSTGRES_USER: *database-user
```

But Hanami has always expected a `DATABASE_URL`. This makes environments like docker conflict with Hanami, because you must keep both sets of variables in sync. To change `DATABASE_PASSWORD`, you would have to update both `DATABASE_PASSWORD` and `DATABASE_URL`.

Hanami does offer a workaround today:

```ruby
  database = case Hanami.env
             when :development then "blog_development"
             when :test then "blog_test"
             else "blog"

  settings = {
    database:,
    host: ENV.fetch("DATABASE_HOST", "localhost"),
    password: ENV.fetch("DATABASE_PASSWORD", "password"),
    port: ENV.fetch("DATABASE_PORT", 5432),
    user: ENV.fetch("DATABASE_USER", "postgres"),
  }

  config.gateway :default do |gateway|
    gateway.database_url = format("postgres://%<user>s:%<password>s@%<host>s:%<port>s/%<database>s", settings)
  end
end
```

But writing this config by hand is a poor experience for our users.

### DIFF
--- a/lib/hanami/providers/db.rb
+++ b/lib/hanami/providers/db.rb
@@ -26,7 +26,7 @@ module Hanami
       end
 
       def finalize_config
-        return if @config_finalized
+        return self if @config_finalized
 
         apply_parent_config if apply_parent_config?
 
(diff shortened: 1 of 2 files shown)


timriley:
Thanks @aaronmallen, this is looking good, and I'm glad it'll help get rid of your Docker-related boilerplate!

My one suggestion at this point is that we should try and make this work for all forms of our ENV-supplied database URLs, e.g.

`{SLICE_NAME__}DATABASE_URL{__GATEWAY_NAME}`

Right now I this change only supports the SLICE_NAME prefix, but not the GATEWAY_NAME suffix?

aaronmallen:
@timriley done!

lib/hanami/providers/db.rb
```diff
@@ -261,13 +261,132 @@ def detect_database_urls_from_env
...
+        gateway_suffix = "__#{gateway_name.to_s.upcase}" unless gateway_name == :default
+
+        components = DATABASE_URL_COMPONENT_VARS.to_h { |var|
+          [var, database_url_component_from_env(var, env_var_prefix, gateway_suffix)]
+        }
+
+        return if components.values.all?(&:nil?)
+
+        # Without a host or a port, we have nothing to connect to
+        return unless components["DATABASE_HOST"] || components["DATABASE_PORT"]
```
katafrakt: What about the default values? Can we just swap the host and keep the default port and have it work? Also, not sure if this is true for SQLite.

lib/hanami/providers/db.rb
```diff
@@ -261,13 +261,132 @@ def detect_database_urls_from_env
...
+        DATABASE_NAME
+      ].freeze
+      private_constant :DATABASE_URL_COMPONENT_VARS
+
+      # Database URL schemes to consider when no DATABASE_URL_SCHEME is given, in order of
+      # preference. The first scheme whose gem is bundled is used.
+      #
+      # @api private
+      DATABASE_URL_SCHEMES = %w[postgres mysql2 sqlite].freeze
+      private_constant :DATABASE_URL_SCHEMES
```
katafrakt: Probably does not support JRuby, but this can be fixed later

lib/hanami/providers/db.rb
```diff
@@ -261,13 +261,132 @@ def detect_database_urls_from_env
...
+        end
+      end
+
+      def escape_database_url_component(value)
+        value.gsub(/[^A-Za-z0-9\-._~]/) { |char| char.bytes.map { format("%%%02X", _1) }.join }
+      end
+
+      def detect_database_url_scheme
+        DATABASE_URL_SCHEMES.find { |scheme| Hanami.bundled?(DATABASE_GEMS.fetch(scheme)) }
+      end
```
katafrakt: Not sure about this logic TBH. If a dependency adds `pg` as a transitive dependency, this function would detect PostgreSQL here. Do we need to attempt to detect that?

katafrakt (commented):
Some questions, probably non-blocking

lib/hanami/providers/db.rb
timriley: Yeah, I think we're better off requiring this be set, I think it's too much chance for something to go amiss.

aaronmallen: I'm thinking the better var would be `DATABASE_ADAPTER` and we could derive from there? I think `DATABASE_URL_SCHEME` is possibly too confusing.

lib/hanami/providers/db.rb
aaronmallen: Jruby doesn't support private_const?

lib/hanami/providers/db.rb
aaronmallen: 🤔 yeah this needs some rework

aaronmallen:
@timriley @katafrakt this is good for another review

lib/hanami/providers/db.rb
katafrakt: No, sorry I wasn't clear. I mean that JRuby uses `jdbc:postgresql` scheme etc. But I stand by my comment, that it does not need to be addressed right now. I will be working on bringin JRuby support in next weeks, so I can adjust it then.

lib/hanami/providers/db.rb
```diff
@@ -261,13 +261,151 @@ def detect_database_urls_from_env
...
+          [var, database_url_component_from_env(var, env_var_prefix, gateway_suffix)]
+        }
+
+        return if components.values.all?(&:nil?)
+
+        adapter = components["DATABASE_ADAPTER"]
+        return unless adapter
+
+        scheme = DATABASE_ADAPTER_SCHEMES.fetch(adapter, adapter)
+        components["DATABASE_HOST"] ||= DEFAULT_DATABASE_HOST
```
katafrakt, line 351: not sure if we should use default host for SQLite

lib/hanami/providers/db.rb
```diff
@@ -261,13 +261,151 @@ def detect_database_urls_from_env
...
+        components["DATABASE_PORT"] ||= DEFAULT_DATABASE_PORTS[scheme]
+
+        name = components["DATABASE_NAME"]
+
+        "#{scheme}://#{database_url_authority(components)}#{"/#{name}" if name}"
```
katafrakt, line 356: username and password are escaped via `escape_database_url_component` but name is just put directly without escaping - is this intentional or an overlook?

katafrakt (commented):
Looking good. Just two more things to make sure we're doing the correct thing.

lib/hanami/providers/db.rb
aaronmallen: Definitely an overlook

PR #1625, opened 2026-08-20
Add in-place code reloading via `Slice#reload!`

afomera:
`Hanami::Slice.reload!` rebuilds a slice and its nested slices from scratch: it stops providers, unloads the constants Zeitwerk defined, discards the container, routes and settings, and prepares again. Slices are rediscovered from disk, so ones added, removed or redefined are picked up too.

The app class object is preserved, so `run Hanami.app` in config.ru stays valid across a reload. Slice classes are not, so anything holding one across a reload holds a stale one. `config/app.rb` is never reloaded, since it defines the app class a reload keeps.

Gated on `config.code_reloading`, which defaults to true in development and false elsewhere. With it off, Zeitwerk does no reloading bookkeeping and `reload!` raises rather than quietly doing nothing.

### DIFF
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -986,11 +986,6 @@ module Hanami
         instance_exec(container, &@prepare_container_block) if @prepare_container_block
         container.configured!
 
-        # Zeitwerk only tracks the constants it defines (and so can only unload them later) when
-        # reloading is enabled before `setup`. Both this class and `Hanami::App` call `setup` from
-        # `prepare_autoloader`, so this must happen here, ahead of either. See {ClassMethods#reload!}.
-        autoloader.enable_reloading if code_reloading?
-
         prepare_autoloader
         @autoloader_setup = true
 
@@ -1030,15 +1025,6 @@ module Hanami
         return config.code_reloading unless Hanami.app?
 
         app.config.code_reloading
-      end
-
-      # Returns the realpaths of the slice's provider files, which dry-system `require`s.
-      #
-      # TODO: Remove along with the `$LOADED_FEATURES` handling in `prepare_container_providers`.
-      def provider_file_paths
-        container.config.provider_dirs
-          .flat_map { |dir| Dir[File.join(root, dir, "**", "*#{RB_EXT}")] }
-          .filter_map { |path| File.realpath(path) if File.file?(path) }
       end
 
       def ensure_slice_name
@@ -1094,7 +1080,11 @@ module Hanami
           :zeitwerk,
           loader: autoloader,
           run_setup: false,
-          eager_load: false
+          eager_load: false,
+          # Zeitwerk only tracks the constants it defines (and so can only unload them later) when
+          # reloading is enabled ahead of `setup`. The plugin applies this from its
+          # `after(:configure)` hook, which runs before `prepare_autoloader` calls `setup`.
+          enable_reloading: code_reloading?
         )
       end
 
@@ -1166,18 +1156,10 @@ module Hanami
 
       # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       def prepare_container_providers
-        on_unload do
-          # Stop providers so they can release the resources they hold (database connections, and
-          # so on) before the components holding them are discarded.
-          container.shutdown!
-
-          # Provider files are `require`d by dry-system, so dropping them from `$LOADED_FEATURES`
-          # is what allows them to be evaluated again against the newly built container.
-          #
-          # TODO: Remove once dry-system's `ProviderRegistrar` uses `load` for provider files, as
-          # its `ManifestRegistrar` already does for registrations.
-          provider_file_paths.each { |path| $LOADED_FEATURES.delete(path) }
-        end
+        # Stop providers so they can release the resources they hold (database connections, and so
+        # on) before the components holding them are discarded. dry-system `load`s provider files,
+        # so the freshly built container evaluates them again by itself.
+        on_unload { container.shutdown! }
 
         # Check here for the `routes` definition only, not `router` itself, because the
         # `router` requires the slice to be prepared before it can be loaded, and at this
--- a/lib/hanami/slice.rb
+++ b/lib/hanami/slice.rb
@@ -40,10 +40,21 @@ module Hanami
       @_mutex.synchronize do
         subclass.class_eval do
           @_mutex = Mutex.new
-          @autoloader = Zeitwerk::Loader.new
-          @container = Class.new(Dry::System::Container)
+          build_container_and_autoloader
         end
       end
+    end
+
+    # Builds the slice's container and autoloader.
+    #
+    # Called when the slice is first defined, and again by {ClassMethods#reload!}, which discards
+    # the previous pair wholesale rather than attempting to invalidate them in place.
+    #
+    # @api private
+    # @since 3.1.0
+    def self.build_container_and_autoloader
+      @autoloader = Zeitwerk::Loader.new
+      @container = Class.new(Dry::System::Container)
     end
 
     module ClassMethods
@@ -355,6 +366,62 @@ module Hanami
       def shutdown
         slices.each(&:shutdown)
         container.shutdown!
+        self
+      end
+
+      # Reloads the slice in place, picking up changes to its source files.
+      #
+      # This tears down the slice and each of its nested slices - stopping providers, unloading
+      # autoloaded constants, and discarding the container, routes and settings - then prepares
+      # them again from scratch.
+      #
+      # Crucially, the slice class objects themselves are preserved, so an already-mounted Rack app
+      # (such as `run Hanami.app` in `config.ru`) remains valid across a reload.
+      #
+      # Slices are rediscovered from disk on each reload, so slices added, removed or redefined are
+      # picked up too. Note this means slice class objects are replaced, unlike the app class.
+      #
+      # `config/app.rb` is the one file that is never reloaded: it defines the app class that this
+      # method deliberately preserves, so changes to it still require a full restart.
+      #
+      # This is intended for development use only.
+      #
+      # @return [self]
+      #
+      # @see #prepare
+      #
+      # @api public
+      # @since 3.1.0
+      def reload!
+        # `@awaiting_reload_prepare` means an earlier reload tore this slice down but then raised
+        # while preparing it again (a syntax error in a reloaded file, most often). The slice is
+        # torn down but not prepared, so this reload must still run to give it another chance.
+        return self unless prepared? || @awaiting_reload_prepare
+
+        unless code_reloading?
+          raise SliceLoadError,
+            "#{self} cannot be reloaded because `config.code_reloading` was false when it was " \
+            "prepared. Set `config.code_reloading = true` on the app before preparing it."
+        end
+
+        # Skipped when a previous reload already tore the slice down but failed to prepare it
+        # again. Tearing down twice would raise, and there is nothing left to tear down anyway.
+        if prepared?
+          # Tear down nested slices first: their containers import from their parent's, so
+          # unwinding in the other order would leave them holding references into a dead container.
+          (slices.with_nested + [self]).each { |slice| slice.__send__(:teardown_for_reload) }
+
+          # Then discard the slice tree itself, so `prepare` rediscovers it from disk and picks up
+          # slices that have been added, removed or redefined.
+          discard_slices_for_reload
+
+          @awaiting_reload_prepare = true
+        end
+
+        prepare
+
+        @awaiting_reload_prepare = false
+
         self
       end
 
@@ -854,6 +921,11 @@ module Hanami
         instance_exec(container, &@prepare_container_block) if @prepare_container_block
         container.configured!
 
+        # Zeitwerk only tracks the constants it defines (and so can only unload them later) when
+        # reloading is enabled before `setup`. Both this class and `Hanami::App` call `setup` from
+        # `prepare_autoloader`, so this must happen here, ahead of either. See {ClassMethods#reload!}.
+        autoloader.enable_reloading if code_reloading?
+
         prepare_autoloader
 
         # Load child slices last, ensuring their parent is fully prepared beforehand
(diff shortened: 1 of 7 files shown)


lib/hanami/slice.rb
```diff
@@ -358,6 +369,62 @@ def shutdown
...
       end
 
+      # Reloads the slice in place, picking up changes to its source files.
+      #
+      # This tears down the slice and each of its nested slices - stopping providers, unloading
+      # autoloaded constants, and discarding the container, routes and settings - then prepares
+      # them again from scratch.
+      #
+      # Crucially, the slice class objects themselves are preserved, so an already-mounted Rack app
+      # (such as `run Hanami.app` in `config.ru`) remains valid across a reload.
```
timriley: Does this mean that reloading won't pick up added/changed routes?

lib/hanami/slice_registrar.rb
```diff
@@ -72,6 +72,30 @@ def to_a
...
+    # Drops each registered slice's definition file from $LOADED_FEATURES and removes its slice
+    # class and namespace module, leaving nothing behind that would shadow what is on disk. This is
+    # what allows slices to be added, removed or redefined without restarting.
+    #
+    # The slice classes themselves are discarded here (unlike the app class, which {Slice.reload!}
+    # preserves), so a reload replaces them with fresh objects.
+    #
+    # @api private
+    # @since 3.1.0
+    def unload_for_reload
```
timriley: Is there anything here that's specific to the eventual reload? If not, then maybe just `#unload` is a nicer name.

lib/hanami/slice.rb
```diff
@@ -866,17 +938,134 @@ def prepare_slice
         self
       end
 
+      # Unwinds everything {#prepare} built up, leaving the slice class itself intact and ready to
+      # be prepared again.
+      #
+      # @api private
+      # @since 3.1.0
+      def teardown_for_reload
```
timriley: Can I just say the comments throughout this method (and in general!) are really helpful 🙏🏼

lib/hanami/slice.rb
```diff
@@ -866,17 +938,134 @@ def prepare_slice
...
+
+        # Stop providers so they can release resources (DB connections, and so on) before the
+        # components holding them are discarded.
+        container.shutdown!
+
+        # Drop every constant Zeitwerk defined for this slice, then retire the loader itself. A
+        # fresh loader is built below rather than reusing this one via `reload`, which would
+        # require `enable_reloading` to have been set before `setup`.
+        autoloader.unload
+        autoloader.unregister
```
timriley: > A fresh loader is built below rather than reusing this one via `reload`, which would require `enable_reloading` to have been set before `setup`.

Don't we _already_ set `enable_reloading` on the autoloader before calling `setup`? Does this mean we can preserve the existing autoloader rather than creating another one?

lib/hanami/slice.rb
```diff
@@ -866,17 +938,134 @@ def prepare_slice
...
+
+        remove_consts_for_reload
+
+        build_container_and_autoloader
+
+        # These files live under `config/` and so are `require`d rather than autoloaded. Dropping
+        # them from $LOADED_FEATURES is what allows them to be evaluated again against the newly
+        # built container.
+        reloadable_paths.each { |path| $LOADED_FEATURES.delete(path) }
```
timriley: I wonder if we should also think about local files required _inside_ these `reloadable_paths`? Or maybe that's taking things a step too far.

timriley:
Thanks for taking the initiative with this, @afomera, I'm *very* excited to see it happening!

I'm going to give you a big long feedback here in one place, rather than sprinkle comments throughout the code, because I think there's some design adjustments we can make to get this into a shape we can all be happy with.

But first, let me start with the positives!

- It works as advertised!
- You clearly understand the ins and outs of the different bits of machinery we use (dry-system, etc.), with the way you handle the different aspects of unloading.
- You made the right choice to put as much logic as possible into `Hanami::Slice` - we want to limit anything that needs to become a special case in `App`.
- Dropping and recreating the autoloaders makes sense, especially given they are per-slice, and we're tearing down slices fully anyway.
- You made the right separation between the hanami and hanami-reloader gems (at least in terms of their current shape; after your improvements, there's an argument to be made for dropping hanami-reloader altogether, I think).
- You've actually gone and _done_ this, giving me something real to grapple with and to help crystalise my views on what we should do next. Thank you!

## My recommendations

My biggest feeling when viewing the diff was, "wow, this is a lot, and complicated." And it makes sense! This is attempting to unwind Hanami's own boot process, which is itself quite involved, working across several tools and layers.

Hanami's own boot process is fairly "artisinal" right now, relying on everything being in one place and sequenced just right. There's nothing generic about it—at least not yet! I want us to get there, but that's still for the future. So my first thought here was whether we might be able to achieve the _beginnings_ of something more generic for this reloading machinery, even within the confines of the current Hanami boot structure.

### Register unwinds where the work happens

I think we could achieve this by adding to an "undo stack" every time we do each step of the boot process. For example:

```ruby
def prepare_container_consts
  namespace.const_set CONTAINER_CONST_NAME, container
  namespace.const_set DEPS_CONST_NAME, container.injector

  on_unload do
    namespace.__send__(:remove_const, CONTAINER_CONST_NAME)
    namespace.__send__(:remove_const, DEPS_CONST_NAME)
  end
end
```

This way, we could collapse the reload approach to something like this, instead of haing to maintain bespoke steps like `teardown_for_reload`, `remove_consts_for_reload`, etc.:

```ruby
def reload! = unload!.prepare

def unload!
  @unload_steps.reverse_each(&:call)
  @unload_steps = []
  @prepared = @booted = false
  self
end
```

We get a few nice properties from this:

- Since the unload stack holds exactly the steps that have completed, unloading after a partially failed `prepare` still works without any special handling. So that `@awaiting_reload_prepare` gate goes away.
- We get ordering (last-in, first-out) for free. We no longer have to delicately arrange things by hand.
- Access to values is maintained by closure semantics, so we don't need to worry about the objects we plan to remove being taken from underneath us.
- It will help minimise bugs from drift in the load/unload sides, since the unloading steps stay local to the code they relate to.
- It's an extension point we'll want as part of a more generic, extensible boot process in the long term.

### The container/autoloader can have their own ordinary unload step

These are built in `inherited` and not during `prepare`, so they look special (and in many ways are), but their unloading treatment can still be ordinary, via something like this at the beginning of `prepare`:

```ruby
on_unload do
  autoloader.unload
  autoloader.unregister
  build_container_and_autoloader
end
```

This way they get rebuilt on each prepare, so repeated reloads work. The `.inherited` code is just there for setting them up for the very first time.

### Change Dry System for our benefit

A bunch of the work in the current branch is covering up for things that Dry System does. But we're also the maintainers of Dry System! And I want us to empower Dry System with capabilities (as makes sense) whenever we expand upon Hanami's code loading.

There's a possible world where we make each part of Dry System "reloadable", able to undo anything it did itself, but I think we could start with some simple adjustments that would make Hanami's life a lot easier.

#### Use `load` for provider files

For starters, we could update `ProviderRegistrar` to use `load` rather than `require`, so it can match `ManifestRegistrar` and effectively give us provider reloading for free, without any `$LOADED_FEATURES` surgery.

#### Update Zeitwerk plugin

We can add an `enable_reloading:` option to Dry System's zeitwerk plugin. In this PR we're having to reach around that plugin to ensure we set this option, when instead we could just be relying on it to do the right setup steps itself.

This plugin runs via an `after(:configure)` hook, which is still ahead of the `autoloader.setup` call that we make within Hanami, so it'll safely configure the reloading in time.

### Use `load` for routes/settings

If we switch from `require` to `load` for these files, combined with the Dry System change above, it means we can do away with the `$LOADED_FEATURES` manipulations entirely.

---

With this all done, I _think_ we should end up with a clean unload/reload structure that is a lot easier to understand and maintain. I was very tempted to do these steps myself, but I think it's a far better outcome for Hanami for a second person to gain a deeper understanding of our boot process, so I'd be delighted if you could explore the above and share how you go :)

Please note that most of the suggestions above are grounded in what I _think_ should help and are as yet not tested, so please do share if you come across any hitches!

## Recommendations for hanami-reloader

These for the change inside hanami-reloader, but I'll share them here, since it's all part of the end-to-end feature.

### The middleware's mutex should cover dispatch

The middleware looks like this:

```ruby
def call(env)
  @mutex.synchronize { check_for_changes }
  @app.call(env) # outside the lock
end
```

By default, our Pumas will run 5 threads in dev, so a second thread can be dispatching inside the app while the first starts unloading things underneath it.

This looks like it needs a read/write lock - shared for dispatch, exclusive for reload. Some example code below, which ended up being more complicated than I first anticipated, because it has to handle lazy Rack responses.

```ruby
# in initialize
@lock = Concurrent::ReadWriteLock.new

def call(env)
  reload_if_needed
  
  @lock.acquire_read_lock
  dispatched = false
  begin
    status, headers, body = @app.call(env)
    dispatched = true
    # Rack bodies can be lazy, so returning from #call doesn't mean dispatch is over. Hold the read
    # lock until the body is closed. Otherwise a reload can pull the app out from under it.
    [status, headers, Rack::BodyProxy.new(body) { @lock.release_read_lock }]
  ensure
    @lock.release_read_lock unless dispatched
  end
end

def reload_if_needed
  return unless @file_checker.updated?  # cheap check, no lock held
  @lock.with_write_lock { check_for_changes } # re-checks under the lock
end
```

concurrent-ruby is already a transitive dependency of ours, so relying on it here doesn't change our overall dependency picture; we'd just need to bring it in as a top-level dependency of hanami-reloader.

### Allow reload errors to be presented by our error rendering

In the current state, an error raised during reloading goes straight up to Puma, not caught and presented nicely by our own webconsole middleware.

I think one way to solve this would be to have hanami-reloader add the webconsole middleware _before_ its own reloading middleware, but then we'd need to be aware of this inside Hanami::Slice so we don't try and mount a _second_ instance of the webconsole middlware. There might be a way through this, but I could see it either getting messy or requiring a substantial overhaul of some kind.

Definitely happy for you to investigate, but probably fine for us to accept this as a gap for now.

---

Anyway, that's my wall of feedback for you :) Definitely keen to hear your thoughts and how you'd like to proceed from here. Thanks again! ❤️

afomera:
Sounds good @timriley thanks for the feedback!! It's going to result in better code at the end of this.

A hitch as I start working through this.

1) if `prepare` raises before reaching `prepare_autoloader` the stack holds an `autoloader.unload` step for a loader that wasn't setup, and Zeitwerk will raise a SetupRequired on the next unload. Zeitwerk doesn't have a public predicate for the `@setup` flag so we have to track that ourselves and then reset it in `build_container_and_autoloader` in an improved setup.

2) const removal is still a bit of a tricky case we might need to handle delicately ie in the case of the Settings class, since re `loading the config/settings.rb with Settings defined would reopen the class and redeclare the dry-configurable settings. 

But I'm working through this now and will report back, nothing I don't think we won't be able to eventually fix in a proper way

lib/hanami/slice.rb
```diff
@@ -358,6 +370,73 @@ def shutdown
...
+      # @return [self]
+      #
+      # @see #unload!
+      # @see #prepare
+      #
+      # @api public
+      # @since 3.1.0
+      def reload!
+        unload!.prepare
+      end
```
timriley, line 390: I think we can just call these `reload` and `unload`, without the bang at the end. Nothing else on `Slice` uses the bang, and many of the other methods mutate the slice's state too, like `prepare`, `boot`, etc., so I think it makes sense to keep everything consistent.

lib/hanami/slice.rb
```diff
@@ -358,6 +370,73 @@ def shutdown
...
+      # @see #on_unload
+      #
+      # @api private
+      # @since 3.1.0
+      def unload!
+        # Zeitwerk can only unload constants it was told to track ahead of `setup`.
+        if @unload_steps.any? && !code_reloading?
+          raise SliceLoadError,
+            "#{self} cannot be unloaded because `config.code_reloading` was false when it was " \
+            "prepared. Set `config.code_reloading = true` on the app before preparing it."
```
timriley, line 408: Would something like this be clearer as the message?

```
#{self} cannot be unloaded because `config.code_reloading` is false (it
defaults to true only in development). Set `config.code_reloading = true`
on the app before preparing it.
```

Mostly the same as yours, just without the "false when it was prepared" bit, and the extra note about the default values.

The reason I don't think we need "when it was prepared" is because it's not actually possible to change `code_reloading` _after_ running `prepare` — the config is frozen at that point. So by taking away that bit, we remove the dimension of time from the error reasoning, and just focus on "was it on or off", which I think should be simpler to understand.

What do you think?

lib/hanami/slice.rb
```diff
@@ -358,6 +370,73 @@ def shutdown
...
+        end
+
+        # Popped rather than iterated, so a step registered while unwinding (a provider's `stop`
+        # memoizing `routes`, say) is unwound too instead of being dropped with the stack.
```
timriley, line 412: This felt a little dense to me, and it wasn't immediately clear how this could happen. I've had a go at expanding this with a concrete example

```suggestion
        # Pop here rather than iterating, because an unload step may register another unload step
        # while it runs.
        #
        # For example, in the unload step for `.prepare_container_providers`, we run
        # `container.shutdown!`. This calls `stop` on every provider, executing user code that may
        # interact with slice methods (e.g. `.routes`) that add further unload steps.
```

How does this feel?

lib/hanami/slice.rb
```diff
@@ -358,6 +370,73 @@ def shutdown
...
+          step.call
+        end
+
+        @prepared = false
+        @booted = false
+
+        self
+      end
+
+      # Registers a step for {#unload!} to run, undoing work {#prepare} just did.
```
timriley, line 423:
```suggestion
      # Registers a step for {#unload!} to run, to undo work done during {#prepare}.
```

Just a minor wording tweak.

lib/hanami/slice.rb
```diff
@@ -849,34 +951,66 @@ def prepare_slice
...
+          @_router = nil
+          @rack_app = nil
+        end
+
         @prepared = true
 
         self
       end
 
+      # Whether code reloading is enabled. Always an app-wide decision: slices import from their
```
timriley, line 990:
```suggestion
      # Returns true if code reloading is enabled. Always an app-wide decision: slices import from their
```

Just keeping the wording consistent with the docs for our other bool-returning methods.

lib/hanami/slice.rb
```diff
@@ -849,34 +951,66 @@ def prepare_slice
...
         prepare_autoloader
+        @autoloader_setup = true
 
         # Load child slices last, ensuring their parent is fully prepared beforehand
         # (useful e.g. for slices that may wish to access constants defined in the
         # parent's autoloaded directories)
         prepare_slices
 
+        # Registered here rather than where these are memoized: `router` memoizes inside a
+        # `@_mutex` block, and `on_unload` must not take a lock Ruby cannot re-enter.
```
timriley, line 979:
```suggestion
        # Register these here rather than where they're memoized. `@_router ||=` re-runs on every
        # call while the router is nil (a slice with no routes), and would register an unload step
        # each time. `@rack_app` is built from the router, so clear it alongside.
```

I think this might be the more accurate description of why we're doing this here?

(I think we could expand on the design of our loading/unloading approach to allow these to one day move alongside their respective lines, but I think the approach you've taken here makes a lot of sense for just getting this shipped in the first place.)

lib/hanami/slice.rb
```diff
@@ -913,7 +1052,10 @@ def prepare_container_plugins
           :zeitwerk,
           loader: autoloader,
           run_setup: false,
-          eager_load: false
+          eager_load: false,
+          # Applied from the plugin's `after(:configure)` hook, so it lands before
+          # `prepare_autoloader` calls `setup`, which is Zeitwerk's deadline for it.
```
timriley, line 1057:
```suggestion
```

This reads as very LLM-y, and I don't think we need it anyway — Zeitwerk will raise if we enable reloading _after_ running setup, so the fact that everything runs error-free is enough to tell us that things are setup right.

lib/hanami/slice_registrar.rb
```diff
@@ -72,6 +72,21 @@ def to_a
       slices.values
     end
 
+    # Unloads every registered slice, then removes the slice classes and namespace modules
+    # themselves. The caller discards this registrar afterwards, so the next prepare rediscovers
+    # slices from disk.
+    #
+    # @api private
+    # @since 3.1.0
+    def unload!
```
timriley, line 81: I think this can be just a regular (non-bang) `unload` name too.

I recognise that we have a bit of inconsistency here between Hanami and Dry System, but I think _internal consistency_ as at least the first thing we should reach for.

lib/hanami/slice_registrar.rb
```diff
@@ -98,16 +113,19 @@ def parent_slice_namespace
...
     #
     # If a slice class is not found, registering the slice will generate the slice class.
     def load_slice(slice_name)
-      slice_require_path = find_slice_require_path(slice_name)
-      require slice_require_path if slice_require_path
+      slice_path = find_slice_require_path(slice_name)
+
+      # `load`, so the file is evaluated again on each prepare; `unload!` above removed the
+      # constants it defines. `find_slice_require_path` omits the extension, which `load` needs.
+      load "#{slice_path}#{RB_EXT}" if slice_path
```
timriley, line 124: `#load_slice` is the only thing that calls `#find_slice_require_path`, and both methods are private, so let's just change the latter (we could even tweak its name too) so that we don't have to worry about adding this comment and adding `#{RB_EXT}` here?

timriley (approved):
This is looking great, thank you so much @afomera!

I've left you a bunch of small and (hopefully) directly actionable comments on this PR. I'm more than happy for you to merge this once they're addressed.

Thank you again, this is going to be such a great new feature ❤️ 

I'll go review the dry-system PR next, so this one isn't blocked.

timriley:
@afomera Oh, and can you please add a CHANGELOG entry before you merge this? Thanks!

PR #1627, opened 2026-08-31
Load settings from YAML files

aaronmallen:
Settings can now be loaded from YAML files at `config/settings/default.yml` and `config/settings/#{Hanami.env}.yml`, in addition to `ENV`.

## Why

Settings files are useful where `ENV` alone is awkward:

- They are evaluated as ERB, so a setting can be computed rather than just read.
- They keep non-secret, environment-varying config in the repo, reviewable in a diff, instead of spread across deployment environments.
- They give a natural home for shared defaults, with per-environment overrides layered on top.

```yaml
# config/settings/default.yml
database_url: <%= ENV.fetch("DATABASE_URL") %>
secret: <%= MySecretProvider.get("app_secret") %>
site_name: <%= Hanami.env?(:test) ? "test_site" : "site" %>
```

## How it works

The default settings store is now a `Hanami::Settings::CompositeStore` that resolves each setting from, in order:

1. `ENV` (via `Hanami::Settings::EnvStore`, as before)
2. `config/settings/#{Hanami.env}.yml`
3. `config/settings/default.yml`

Both files are optional. Each is evaluated as ERB, then parsed as YAML, with its top-level keys becoming setting names. Files are read lazily on the first fetch and then memoized, so an app that defines no settings never touches the filesystem.

Like `ENV` and `.env`, the files are read from the app root only and are shared by the app and all of its slices.

### `Hanami::Settings::YamlFileStore`

The new store is public API and can be used on its own:

```ruby
# config/app.rb
config.settings_store = Hanami::Settings::YamlFileStore.new(
  config.root.join("my_settings.yml")
)
```

It also accepts a block, for a path that depends on config not yet known when the store is built:

```ruby
config.settings_store = Hanami::Settings::YamlFileStore.new {
  config.root.join("my_settings.yml")
}
```

Details worth knowing:

- YAML aliases are enabled, and `Date`, `DateTime`, `Time` and `Symbol` values are permitted under Psych's safe load.
- A missing or empty file behaves as an empty store.
- A file whose top level is not a mapping raises a `Hanami::Error` naming the file, rather than failing obscurely later.

## Backwards compatibility

Fully backwards compatible. `ENV` keeps precedence over both files, so an existing app that configures its settings through the environment behaves exactly as it does today, whether or not the files are present.

Assigning `config.settings_store` still replaces the whole chain, `ENV` included. To add a store while keeping the defaults, compose it:

```ruby
config.settings_store = Hanami::Settings::CompositeStore.new(
  MyStore.new,
  config.settings_store
)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ A complete Hanami app is composed of multiple gems. For a complete overview of c
 
 ### Added
 
+- Load settings from `config/settings/default.yml` and `config/settings/[HANAMI_ENV].yml`, via the new`Hanami::Settings::YamlFileStore`. Values in `ENV` continue to take precedence. (@aaronmallen)
+
 ### Changed
 
 ### Deprecated
--- a/lib/hanami/config.rb
+++ b/lib/hanami/config.rb
@@ -48,16 +48,22 @@ module Hanami
     # @!attribute [rw] settings_store
     #   Sets the store used to retrieve {Hanami::Settings} values.
     #
-    #   Defaults to an instance of {Hanami::Settings::EnvStore}.
+    #   Defaults to a {Hanami::Settings::CompositeStore} that resolves each setting from, in order:
+    #   `ENV`, `config/settings/[HANAMI_ENV].yml`, then `config/settings/default.yml`.
+    #
+    #   The YAML files are read from the app root only, and are shared by the app and all its
+    #   slices, in the same way as `ENV`.
     #
     #   @return [#fetch]
     #
     #   @see Hanami::Settings
+    #   @see Hanami::Settings::CompositeStore#fetch
     #   @see Hanami::Settings::EnvStore#fetch
+    #   @see Hanami::Settings::YamlFileStore#fetch
     #
     #   @api public
     #   @since 2.0.0
-    setting :settings_store, default: Hanami::Settings::EnvStore.new
+    setting :settings_store
 
     # @!attribute [rw] slices
     #   Sets the slices to load when the app is preared or booted.
@@ -362,6 +368,7 @@ module Hanami
 
       # Apply default values that are only knowable at initialize-time (vs require-time)
       self.root = Dir.pwd
+      self.settings_store = default_settings_store
       self.render_errors = (env == :production)
       self.render_detailed_errors = (env == :development)
       load_from_env
@@ -535,6 +542,18 @@ module Hanami
     end
 
     private
+
+    # Returns the store used to retrieve settings values when none is configured.
+    #
+    # The YAML file paths are resolved lazily, since the root may be configured after this config is
+    # initialized.
+    def default_settings_store
+      Settings::CompositeStore.new(
+        Settings::EnvStore.new,
+        Settings::YamlFileStore.new { root.join(SETTINGS_PATH, "#{env}.yml") },
+        Settings::YamlFileStore.new { root.join(SETTINGS_PATH, "default.yml") }
+      )
+    end
 
     def load_from_env
       self.slices = ENV["HANAMI_SLICES"]&.split(",")&.map(&:strip)
--- a/lib/hanami/settings.rb
+++ b/lib/hanami/settings.rb
@@ -43,15 +43,25 @@ module Hanami
   # {Hanami::Slice::ClassMethods#prepare prepare} step, to ensure that the app boots only when valid
   # settings are present.
   #
-  # Setting values are loaded from a configurable store, which defaults to
-  # {Hanami::Settings::EnvStore}, which fetches the values from equivalent upper-cased keys in
-  # `ENV`. You can configure an alternative store via {Hanami::Config#settings_store}. Setting stores
-  # must implement a `#fetch` method with the same signature as `Hash#fetch`.
+  # Setting values are loaded from a configurable store, which defaults to a
+  # {Hanami::Settings::CompositeStore} resolving each value from, in order:
+  #
+  # 1. {Hanami::Settings::EnvStore}, which fetches the value from the equivalent upper-cased key in
+  #    `ENV`.
+  # 2. {Hanami::Settings::YamlFileStore}, for `config/settings/[HANAMI_ENV].yml`.
+  # 3. {Hanami::Settings::YamlFileStore}, for `config/settings/default.yml`.
+  #
+  # The YAML files are optional, and are read from the app root only: like `ENV`, they are shared by
+  # the app and all its slices.
+  #
+  # You can configure an alternative store via {Hanami::Config#settings_store}. Setting stores must
+  # implement a `#fetch` method with the same signature as `Hash#fetch`.
   #
   # [dry-c]: https://dry-rb.org/gems/dry-configurable/
   # [dry-t]: https://dry-rb.org/gems/dry-types/
   #
-  # @see Hanami::Settings::DotenvStore
+  # @see Hanami::Settings::EnvStore
+  # @see Hanami::Settings::YamlFileStore
   #
   # @api public
   # @since 2.0.0
--- a/lib/hanami/settings/yaml_file_store.rb
+++ b/lib/hanami/settings/yaml_file_store.rb
@@ -1,0 +1,104 @@
+# frozen_string_literal: true
+
+require "date"
+require "dry/core/constants"
+require "erb"
+require "yaml"
+
+require_relative "../errors"
+
+module Hanami
+  class Settings
+    # A settings store that loads settings from a YAML file.
+    #
+    # The file is evaluated as ERB, then parsed as YAML, with its top-level keys becoming the
+    # setting names. It is read lazily, on the first fetch, and its contents are then memoized. A
+    # file that does not exist, or that is empty, results in an empty store.
+    #
+    # @example
+    #   # config/app.rb
+    #   config.settings_store = Hanami::Settings::YamlFileStore.new(config.root.join("my_settings.yml"))
+    #
+    # @example Resolving the path lazily
+    #   # Give a block to resolve the path when the settings are first fetched, rather than when the
+    #   # store is created.
+    #   config.settings_store = Hanami::Settings::YamlFileStore.new { config.root.join("my_settings.yml") }
+    #
+    # @api public
+    # @since 3.1.0
+    class YamlFileStore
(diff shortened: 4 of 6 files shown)


katafrakt:
This is quite a big conceptual change. I wonder if majority of things mentioned as the justification here cannot be done using existing machinery?

> They keep non-secret, environment-varying config in the repo, reviewable in a diff, instead of spread across deployment environments

This can be done with per-env env files now, right? `.env.development` and `.env.test` should be checked into version control as much as the YAML config files from this PR.

> They give a natural home for shared defaults, with per-environment overrides layered on top

A shared default could perhaps be [set when defining a setting](https://hanakai.org/learn/hanami/v3.0/app/settings#default-values) instead?

> They are evaluated as ERB, so a setting can be computed rather than just read

This is true difference from the ENV approach, but I have some reservations about ERB-in-YAML approach, to be honest. The example shows `MySecretProvider.get("app_secret")` and it hints that maybe it should be an actual provider with it's lazy loading? If it's an expensive operation, like fetching something from an API, it could break the contract that Hanami app (re)loads very fast, so we can afford to do that on every code change in development.

I think I would like to understand the need behind such a change before discussing the implementation.

aaronmallen:
@katafrakt 

> This can be done with per-env env files now, right? .env.development and .env.test should be checked into version control as much as the YAML config files from this PR.

for the `.env` files committed to source sure, but what about `.env.local`

> A shared default could perhaps be [set when defining a setting](https://hanakai.org/learn/hanami/v3.0/app/settings#default-values) instead?

Yes but a setting can only have one default. Here's an example of where using the `default:` of a setting might fall sort:

```yaml
# config/settings/development.yml
redis_db: 1
```

```yaml
# config/settings/test.yml
redis_db: 2
```

You can technically do this with env as well. But relying purely on settings would look like this:

```ruby
module MyApp
  class Settings < Hanami::Settings
    default_redis_db = case Hanami.env
                       when :test then 2
                       when :development then 1
                       else 0
                       end
    setting :redis_db, default: default_redis_db 
  end
end
```

As you mentioned the **major** selling point is being able to eval ruby code, which you cannot do in env. There are countless examples of where being able to eval ruby code would be beneficial. The ones outlined in the PR are just what came top of mind. Manipulating env alone with `ENV.fetch` is more powerful than simply relying on env IMO.

Lastly env declarations continue to work here, so you're not forced to use the yaml files. In the inverse since env works for those who want to use env, I think it would be a bit of a disservice not to offer this to our users who would rather write configuration to configuration files.

aaronmallen:
I'd also like to point out that this opens the doorway to complex settings that currently aren't easy to achieve with env alone:

```yaml
redis:
  db: 1
  host: <%= ENV.fetch("REDIS_HOST", "localhost") %>
  password: <%= `op read "REDIS_PASSWORD"` %>
  port: <%= ENV.fetch("REDIS_PORT", 6379) %>
  user: <%= ENV["REDIS_USER"] %>
  ```
  
  ```ruby
  module MyApp
    class Settings < Hanami::Settings
      setting :redis, default: {}, constructor: Types::Hash.schema(
        db?: Types::Integer.default(0),
        host?: Types::String.default("localhost"),
        password?: Types::String,
        port?: Types::Integer.default(6379),
        username?: Types::String
      )
    end
  end
``` 

```ruby
Sidekiq.configure_server do |config|
  config.redis = Hanami.app.settings.redis.compact
end
```

afomera:
Yeah, I like this idea a lot personally.

I do not enjoy needing to reach into a .env file to tweak things that perhaps are environmental based or whatnot. 

One of my use cases would be git worktrees, where i might normally need a complicated script to manipulate a .env file for that worktree to setup and use separate databases, or override the default in the app.rb or whatever.

For example to use a separate pg db name with each worktree, or redis DB etc etc. I can use the yaml file to define erb that checks if those ENV vars are set, and load em from my worktree tool vs the default of needing to override the default the config setting way.

+1 from me on this PR

katafrakt:
@aaronmallen 

> for the .env files committed to source sure, but what about .env.local

I'm not sure I understand the question. You should not commit `.env.local`, as this is by design a local override. But how does YAML improve the situation with it?

> Yes but a setting can only have one default. Here's an example of where using the default: of a setting might fall sort:

Again, I think I'm missing something. In this case you would likely want to have:

```
# .env.development
REDIS_DB=1

# .env.test
REDIS_DB=2
```

... not play with the `Hanami.env` in `settings.rb`.

> I'd also like to point out that this opens the doorway to complex settings that currently aren't easy to achieve with env alone:
> `host: <%= ENV.fetch("REDIS_HOST", "localhost") %>`

This is actually how Dotenv works already. I think I'm, again, missing something here about these complex settings required. Not saying it's not true. Just that I don't see it.

aaronmallen:
@katafrakt 

> I'm not sure I understand the question. You should not commit .env.local, as this is by design a local override. But how does YAML improve the situation with it?

By using configuration files you can commit the setting, without committing its value:

```env
# .env
DATABASE_HOST=localhost
DATABASE_PORT=5432
DATABASE_USER=me
```
```env
# .env.local (not committed)
DATABASE_PASSWORD=secret
```
```yaml
database:
  host: <%= ENV.fetch("DATABASE_PASSWORD", "localhost" %>
  port: <%= ENV.fetch("DATABASE_PORT", 5432) %>
  password: <%= ENV["DATABASE_PASSWORD"] %> 
  user: <%= ENV["DATABASE_USER"] %>
```
here we can clearly see what the setting schema is suppose to look like without dividing things across env files that aren't committed.

> Again, I think I'm missing something. In this case you would likely want to have

This was in response to your suggestion to use a setting default... I'm sure there are better examples, and I stated you could achieve the same thing with dot env, but this was a case where a setting default was not sufficient. 

> This is actually how Dotenv works already.

Please provide an example of constructing a hash from env, as far as I'm aware that is not possible.

At the end of the day there are 3 strong arguments for this change:

1. Being able to eval ruby code in your settings
2. Env continues to work so why not also offer file based configuration
3. Configuration **belongs** in configuration files, Env was always meant to configure global state **not** application state.

afomera:
@katafrakt imo it's also plausible that we as Hanami should be extensible enough for different setup configurations as well, and I've worked on many applications that may have used dotenv or something like it but have also wanted yaml configuration files, and to do that currently you'd have to write something for each application or patch hanami itself. 

One job I worked used [Chamber](https://github.com/thekompanee/chamber) to load yaml configurations and then allow overriding with ENV vars.

katafrakt:
> Please provide an example of constructing a hash from env, as far as I'm aware that is not possible.

That's not what I was saying. I was referring to the example of taking the value from the env and falling back to the default.

But given the tone of this discussion, I'm not going to engage in it any further.

radar:
The example uses yaml to say that we can evaluate this with ERB. But we can already do this with the settings as they stand at the minute:

https://github.com/radar/twist-v3/blob/develop/config/settings.rb

This is a Ruby file, where I can evaluate whatever Ruby code I want already. 

Taking your example of:

```ruby
module MyApp
  class Settings < Hanami::Settings
    default_redis_db = case Hanami.env
                       when :test then 2
                       when :development then 1
                       else 0
                       end
    setting :redis_db, default: default_redis_db 
  end
end
```

You can do this already with a `.env.development` and a `.env.test` file. These will be loaded over `.env`, and depend on what `HANAMI_ENV` is ultimately set to, which judging by [this code](https://github.com/hanami/hanami/blob/2c785981724efbf25f13acbc9d5db287011e949a/lib/hanami.rb#L175) falls back to whatever RACK_ENV is set to.

I'm not yet seeing a "winning case" for using yaml here. There are also [convincing reasons](https://noyaml.com/) to not support the format at all.

I agree that configuration belongs in configuration. `config/settings.rb` _is_ that file, or `.env.*` are those files as well.

timriley:
Thanks for all the discussion here, everyone! Let me share a few thoughts of my own.

Firstly, @aaronmallen did check in with me about my openness to considering such a feature, and I said it was worth a look.

I do think something like this is worthy of consideration as one of those peripheral areas of the framework where we could take away a papercut without detracting in any way from our core framework experience.

The main interface to Hanami settings is the `Settings` class in `config/settings.rb`. This isn't changing.

Our choice of dotenv to populate the settings store was mostly pragmatic: it was what I was already using in Icelab's pre-Hanami proto-framework (see [here](https://github.com/icelab/snowpack/blob/fe44a990241c3a9e8b37f8cf60ff5e8380c3c959/lib/snowpack/generators/application/templates/system/boot/settings.rb.tt) and [here](https://github.com/dry-rb/dry-system/blob/09e77a53589b07d99975a08702e827e229177701/lib/dry/system/provider_sources/settings/loader.rb)), it was relatively low-touch, and it fit well with the Heroku-style deployments we were mostly doing at the time.

I do think dotenv still works well for modern Hanami apps, but I definitely don't see it as perfection. It's still awkward to spread things across multiple files, and it can only supply primitive values.

As Aaron shared, YAML files support nested structures and ERB interpolation that make certain kinds of config easier to manage. If someone comes to Hanami who is more used that way of managing settings, then this could make their life slightly easier and help them feel more comfortable while learning the rest of the new concepts in our framework.

I'd only support a change like this if it was non-disruptive and completely non-obtrusive to people who're running existing Hanami apps and who are happy with the dotenv-based loading, which will remain our default. Since the YAML support doesn't require any new dependencies, and is mostly just a new file in the codebase, I see the carrying cost being fairly low. I'd see us adding just one small section to our existing settings guide.

So to answer the question of “should we accept this,” I think we should focus more in on:

1. Does this allow for certain kinds of settings to be provided that could not be done before? (In this case, I think this boils down to hash structures and ERB eval; all the discussion about per-env defaults etc. seems like it's the basically the same for both dotenv and YAML files)
2. Might this make the Hanami experience more comfortable for a wider range of humans? (This one is more about squishy human preferences and emotions rather than code capabilities)
3. Is the maintenance/carrying cost acceptable?

I am genuinely interested in your thoughts on these.

While I tend to be pretty cautious about us adopting multiple ways of doing something, this proposal feels pretty acceptable to me. I'm also conscious that I only have my own personal experience when comes to app development, so I try to be open to other people and their experiences too, especially when we could support those while still fulfilling our core Hanami framework promise.

(While I'm here, one thing that I think would meaningfully improve our settings capability is per-slice settings definitions and loading. I'd love for someone to work on that!)

PR #1628, opened 2026-09-02
Enable ObjectSpace for JRuby tests

katafrakt:
Currently we have 575 tests are failing on JRuby. But most of them are related to `ObjectSpace` not being available. This really obfuscates how far we are from supporting JRuby. After enabling `ObjectSpace` it's only 75 failures (and will be even fewer after merging https://github.com/hanami/hanami-cli/pull/434) – so IMO it's a nice first step.


katafrakt:
Actually it's even 44 on CI, so I probably had some dependency not updated locally.

timriley:
Great thinking @katafrakt! Bringing this in now!

